Issue #3, opened 2019-06-06
Random points and vectors.


kellertuer:
In order to construct random points and vectors we have to find a way to specify the kind of randomness. For example tangent vectors of SPDs can be generated using Gaussian or Rician noise. For me that is independent of a manifold, i.e. should not be a field of `manifold` but maybe an optional parameter to the random functions? Maybe something like

`random_vector(M,x,:Gaussian,sigma)`

and

`random_vector(M,x,:Rician,tau)`

where I am also not so sure, whether other kinds of noise might even require more parameters.

the default `random_point(M,x)` should then be set to a default noise with default parameters for example `:Gaussian` with `sigma=1`. What do you think?
dependent o

sdaxen:
Many different distributions could be used to generate noise. A few off the top of my head would be uniform (for compact manifolds), tangent-X (for any X that can be defined on the tangent space), X (similar to tangent-X but with a Jacobian adjustment, i.e. it is not pulled back to the manifold), and Brownian diffusion.

I'm okay with just assuming tangent-X and passing the corresponding distribution (e.g. from Distributions.jl) to a function as syntactic sugar (i.e.. something like `random_vector(M,x,d) = exp(M,x,rand(d))`).

A separate package really should support more manifold statistics.

kellertuer:
Passing down the distribution a second (`random_point`) or third (`random_vector`) is a very neat idea.

sdaxen:
Oh sorry, that totally should have been `random_vector(M,x,d) = rand(d)`, `random_point(M,x,d) = exp(M,x,rand(d))`, potentially with the vectors wrapped in a special type if we do that.

RE your point in https://github.com/JuliaNLSolvers/ManifoldMuseum/issues/2#issuecomment-499385511, if we do have such custom types, those could be passed to these functions. But then it may make more sense to extend `Base.rand`.

mateuszbaran:
I've quickly sketched a Distributions.jl-style interface for random points:
```julia
using Distributions

abstract type Manifold
end

struct ManifoldPointVariate <: VariateForm
end

struct ManifoldGaussian{TM<:Manifold, TMean, TSigma<:Real} <: Distribution{ManifoldPointVariate, Continuous} where TMean
    M::TM
    μ::TMean
    σ::TSigma
end

function rand(d::ManifoldGaussian)
end
```
[...]

PR #5, opened 2019-06-06
Tests, isapprox and some work on Sphere

mateuszbaran:
Major points:

* I've introduced a new function, `isapprox`.
* Sphere is in a bit better shape.
* I've added some generic tests for manifolds.
* `dot`, `exp`, `log`, `isapprox`, `norm` extend standard functions.

### DIFF
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -6,7 +6,7 @@ import LinearAlgebra: dot, norm
 
 abstract type Manifold end
 
-isapprox(m::Manifold, x, y; kwargs...) = error("Not implemented")
+isapprox(m::Manifold, x, y; kwargs...) = isapprox(x, y; kwargs...)
 
 retract!(M::Manifold, y, x, v) = exp!(M, y, x, v)
 retract(M::Manifold, x, v) = retract!(M, copy(x), x, v)
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -1,30 +1,39 @@
 module ManifoldMuseum
 
-export Manifold
+import Base: isapprox, exp, log
+import LinearAlgebra: dot, norm
+
 
 abstract type Manifold end
 
-retract!(M::Manifold, y, x, v) = exp!(M,y,x,v)
-retract(M::Manifold, x, v) = retract!(M,copy(x),x,v)
+isapprox(m::Manifold, x, y; kwargs...) = error("Not implemented")
+
+retract!(M::Manifold, y, x, v) = exp!(M, y, x, v)
+retract(M::Manifold, x, v) = retract!(M, copy(x), x, v)
 
 project_tangent!(M::Manifold, w, x, v) = error("Not implemented")
-project_tangent(M::Manifold, x, v) = project_tangent!(M,copy(x),x,v)
+project_tangent(M::Manifold, x, v) = project_tangent!(M, copy(x), x, v)
 
 distance(M::Manifold, x, y) = error("Not implemented")
 
-dot(M::Manifold, x, v, w) = dot(v,w)
+dot(M::Manifold, x, v, w) = dot(v, w)
 norm(M::Manifold, x, v) = sqrt(dot(M, x, v, v))
 
 exp!(M::Manifold, y, x, v) = error("Not implemented")
-exp(M::Manifold, x, v) = exp!(M,copy(x),x,v)
+exp(M::Manifold, x, v) = exp!(M,copy(x), x, v)
 
 log!(M::Manifold, v, x, y) = error("Not implemented")
-log(M::Manifold, x, y) = error("Not implemented")
+
+function log(M::Manifold, x, y)
+    v = zero_tangent_vector(M, x)
+    log!(M, v, x, y)
+    return v
+end
 
 dimension(M::Manifold) = error("Not implemented")
 
-vector_transport!(M::Manifold, vto, x, v, y) = project_tangent!(M,w,x,v)
-vector_transport(M::Manifold, x, v, y) = vector_transport!(M,copy(v),x,y,v)
+vector_transport!(M::Manifold, vto, x, v, y) = project_tangent!(M, vto, x, v)
+vector_transport(M::Manifold, x, v, y) = vector_transport!(M, copy(v), x, y, v)
 
 random_point(M::Manifold) = error("Not implemented")
 random_tangent_vector(M::Manifold, x) = error("Not implemented")
@@ -36,5 +45,21 @@ zero_tangent_vector!(M::Manifold, v, x) = log!(M, v, x, x)
 geodesic(M::Manifold, x, y, t) = exp(M, x, log(M, x, y), t)
 
 include("Sphere.jl")
+
+export Manifold
+export dimension,
+    distance,
+    dot,
+    exp,
+    exp!,
+    geodesic,
+    isapprox,
+    log,
+    log!,
+    norm,
+    typical_distance,
+    zero_tangent_vector,
+    zero_tangent_vector!
+
 
 end # module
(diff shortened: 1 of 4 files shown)


src/ManifoldMuseum.jl
```diff
@@ -1,30 +1,39 @@
...
-export Manifold
+import Base: isapprox, exp, log
+import LinearAlgebra: dot, norm
+
 
 abstract type Manifold end
 
-retract!(M::Manifold, y, x, v) = exp!(M,y,x,v)
-retract(M::Manifold, x, v) = retract!(M,copy(x),x,v)
+isapprox(m::Manifold, x, y; kwargs...) = error("Not implemented")
```
antoine-levitt: Fallback to `isapprox(x,y)`?

mateuszbaran: OK, it might be a useful default.

src/Sphere.jl
```diff
@@ -2,10 +2,37 @@ struct Sphere{T} <: Manifold where {T}
     shape::T
 end
 
-project_tangent!(S::Sphere,w,x,v) = (w .= w - dot(x,v)*x)
-distance(S::Sphere,x,y) = norm(x-y) # FIXME
-exp!(S::Sphere,y,x,v) = retract!(S,y,x,v) # FIXME
-log!(S::Sphere,v,x,y) = project_tangent!(S,v,x,y-x) # FIXME
+isapprox(S::Sphere, x, y; kwargs...) = isapprox(x, y; kwargs...)
+project_tangent!(S::Sphere, w, x, v) = (w .= w .- dot(x, v).*x)
```
antoine-levitt: Typo here (mine): `v` instead of w

src/Sphere.jl
```diff
@@ -2,10 +2,37 @@ struct Sphere{T} <: Manifold where {T}
...
+    end
+    return v
+end
+
 dimension(S::Sphere) = sum(S.shape)
 random_point(S::Sphere) = (x = randn(S.shape); x / norm(x))
-random_tangent_vector(S::Sphere, x) = (v = randn(S.shape); project_tangent!(v,x,v))
+
+function random_tangent_vector(S::Sphere, x)
+    v = randn(S.shape)
```
antoine-levitt: Hm, we might have to parametrize on a float type. How about we parametrize `Sphere` on Tfloat, and add convenience constructors like `Sphere(x) = Sphere{eltype(x)}(size(x))`?

mateuszbaran: To be honest, I'd prefer to have different types for point and tangent vector distributions and extend `rand` (see #3 for a discussion)

antoine-levitt: OK; then we also extend `zero` and have no need for storing types or shapes in the manifold structure?

mateuszbaran: Do you mean renaming `zero_tangent_vector` to `zero`? `Sphere` will probably only have to store dimension (to implement `dimension`). I don't know if we should do this by a type parameter or a field.

antoine-levitt: Yes, it'd be natural to extend `zero`. But I don't know what to do about point vs tangent vector. Same goes for random.

Seems overkill to store dimension just to be able to print it back. If we really want dimension we can ask the user to give us a point: `dimension(S::Sphere,x) = length(x)-1`.

mateuszbaran: I'm not sure if it's clear enough that `zero(M, x)` returns a zero tangent vector. Why not zero element of a different vector bundle at `x`? Random points and tangent vectors are a separate issue, I'll have to think about it a bit more.

`dimension(M, x)` looks weird, all points on a manifold have the same dimension. A good use case for `dimension` would be nice to determine which option is better.

sdaxen: Is there a particular reason that we store dimension as a field instead of as a parameter? The latter enables dispatch on dimension. Of course, this may be moot if we use specialized vector types.

Do we want to also provide spheres of non-unit radius?

src/Sphere.jl
```diff
@@ -2,10 +2,36 @@ struct Sphere{T} <: Manifold where {T}
...
+    else
+        y .= cos(nv).*x .+ (sin(nv)/nv).*v
+    end
+    return y
+end
+
+function log!(S::Sphere, v, x, y)
+    θ = acos(dot(x, y))
+    if θ ≈ 0.0
+        zero_tangent_vector!(S, v, x)
```
sdaxen: Not certain this is great for AD. Would be nice to use the first order Taylor expansion around `θ = 0.0` instead of `zero_tangent_vector!`.

src/ManifoldMuseum.jl
```diff
@@ -1,30 +1,39 @@
...
+import LinearAlgebra: dot, norm
+
 
 abstract type Manifold end
 
-retract!(M::Manifold, y, x, v) = exp!(M,y,x,v)
-retract(M::Manifold, x, v) = retract!(M,copy(x),x,v)
+isapprox(m::Manifold, x, y; kwargs...) = isapprox(x, y; kwargs...)
+
+retract!(M::Manifold, y, x, v) = exp!(M, y, x, v)
```
sdaxen: Perhaps `y` should be `xto` for clarity, similar to `vto` below?

src/ManifoldMuseum.jl
```diff
@@ -1,30 +1,39 @@
...
 
 exp!(M::Manifold, y, x, v) = error("Not implemented")
-exp(M::Manifold, x, v) = exp!(M,copy(x),x,v)
+exp(M::Manifold, x, v) = exp!(M,copy(x), x, v)
 
 log!(M::Manifold, v, x, y) = error("Not implemented")
-log(M::Manifold, x, y) = error("Not implemented")
+
+function log(M::Manifold, x, y)
+    v = zero_tangent_vector(M, x)
```
sdaxen: Perhaps better to declare as undefined instead of filling with zeros that are then replaced.

src/ManifoldMuseum.jl
```diff
@@ -37,4 +46,20 @@ geodesic(M::Manifold, x, y, t) = exp(M, x, log(M, x, y), t)
 
 include("Sphere.jl")
 
+export Manifold
+export dimension,
+    distance,
+    dot,
+    exp,
```
sdaxen: Sorry, I'm still new to Julia, but since we extended `Base.exp` here, do we need to export `exp`?

src/ManifoldMuseum.jl
antoine-levitt: But then we'd need an `undefined_tangent_vector`? That seems overkill

src/Sphere.jl
antoine-levitt: Actually `log` is not very well conditionned for `x` close to `y`, so that `approx` can be replaced with `==` and have the same behavior. There's no way to get an accurate result here without doing something like `log1p`.

sdaxen: > Actually `log` is not very well conditionned for `x` close to `y`

That's interesting, can you explain a bit more?

src/ManifoldMuseum.jl
sdaxen: True, was thinking something like `TVector(M, x, undef)`, but since not all vectors can be instantiated with `undef`, I can see why we wouldn't do that.

src/Sphere.jl
antoine-levitt:
```
julia> rel_prec(x,y) = (x-y)/y
rel_prec (generic function with 1 method)

julia> rel_prec(log(1+1e-12),1e-12)
8.890058184100751e-5

julia> rel_prec(log1p(1e-12),1e-12)
-5.000242179397697e-13
```
Essentially, whenever `x` is close to `y`, you can't compute `x-y` to good relative accuracy.

antoine-levitt: In particular here, isapprox(1-y,0.0) is zero if and only if `y==1`.

src/ManifoldMuseum.jl
mateuszbaran: Right, functions from `Base` don't have to be exported. I'll remove these exports.

PR #7, opened 2019-06-06
Scaling argument for exp and retract, docs, tangent vector tests

mateuszbaran:
I've made a few new changes:

* `isapprox` for tangent vectors,
* scaling argument for `exp` and `retract`,
* some docs,
* new tests for tangent vectors,
* `log`, `exp` and `isapprox` are not explicitly exported (they don't have to be).

### DIFF
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -113,7 +113,13 @@ vector_transport(M::Manifold, x, v, y) = vector_transport!(M, copy(v), x, y, v)
 random_point(M::Manifold) = error("Not implemented")
 random_tangent_vector(M::Manifold, x) = error("Not implemented")
 
+"""
+    injectivity_radius(M::Manifold, x)
+
+Distance such that `log(M, x, y)` is defined for all points within this radius.
+"""
 injectivity_radius(M::Manifold, x) = 1.0
+
 zero_tangent_vector(M::Manifold, x) = log(M, x, x)
 zero_tangent_vector!(M::Manifold, v, x) = log!(M, v, x, x)
 
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -13,7 +13,7 @@ export dimension,
     norm,
     retract,
     retract!,
-    typical_distance,
+    injectivity_radius,
     zero_tangent_vector,
     zero_tangent_vector!
 
@@ -113,7 +113,7 @@ vector_transport(M::Manifold, x, v, y) = vector_transport!(M, copy(v), x, y, v)
 random_point(M::Manifold) = error("Not implemented")
 random_tangent_vector(M::Manifold, x) = error("Not implemented")
 
-typical_distance(M::Manifold) = 1.0
+injectivity_radius(M::Manifold, x) = 1.0
 zero_tangent_vector(M::Manifold, x) = log(M, x, x)
 zero_tangent_vector!(M::Manifold, v, x) = log!(M, v, x, x)
 
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -45,7 +45,9 @@ Retraction (cheaper, approximate version of exponential map) of tangent
 vector `t*v` at point `x` from manifold `M`.
 Result is saved to `y`.
 """
-retract!(M::Manifold, y, x, v, t=1) = exp!(M, y, x, v, t)
+retract!(M::Manifold, y, x, v) = exp!(M, y, x, v)
+
+retract!(M::Manifold, y, x, v, t) = retract!(M, y, x, t*v)
 
 """
     retract(M::Manifold, x, v, t=1)
@@ -53,19 +55,21 @@ retract!(M::Manifold, y, x, v, t=1) = exp!(M, y, x, v, t)
 Retraction (cheaper, approximate version of exponential map) of tangent
 vector `t*v` at point `x` from manifold `M`.
 """
-retract(M::Manifold, x, v, t=1) = retract!(M, copy(x), x, v, t)
+retract(M::Manifold, x, v) = retract!(M, similar(x), x, v)
+
+retract(M::Manifold, x, v, t) = retract(M, x, t*v)
 
 project_tangent!(M::Manifold, w, x, v) = error("Not implemented")
 project_tangent(M::Manifold, x, v) = project_tangent!(M, copy(x), x, v)
 
-distance(M::Manifold, x, y) = error("Not implemented")
+distance(M::Manifold, x, y) = norm(M, x, log(M, x, y))
 
 """
     dot(M::Manifold, x, v, w)
 
 Inner product of tangent vectors `v` and `w` at point `x` from manifold `M`.
 """
-dot(M::Manifold, x, v, w) = dot(v, w)
+dot(M::Manifold, x, v, w) = error("Not implemented")
 
 """
     norm(M::Manifold, x, v)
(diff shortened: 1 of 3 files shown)


antoine-levitt:
Thanks, sorry for merging the previous one too fast!

mateuszbaran:
No problem, opening new PRs is easy 😃 .

src/ManifoldMuseum.jl
```diff
@@ -3,24 +3,91 @@ module ManifoldMuseum
...
 
-exp!(M::Manifold, y, x, v) = error("Not implemented")
-exp(M::Manifold, x, v) = exp!(M,copy(x), x, v)
+"""
+    exp!(M::Manifold, y, x, v, t=1)
+
+Exponential map of tangent vector `t*v` at point `x` from manifold `M`.
+Result is saved to `y`.
+"""
+exp!(M::Manifold, y, x, v, t=1) = error("Not implemented")
```
sdaxen: I don't like always having the multiplication by a scalar, even if we don't need it. What about something like

```julia
exp!(M::Manifold, y, x, v) = error("Not implemented")
exp!(M::Manifold, y, x, v; t = 1) = exp!(M::Manifold, y, x, t .* v)
retract!(M::Manifold, y, x, v; kwargs...) = exp!(M, y, x, v; kwargs...)
retract(M::Manifold, x, v; kwargs...) = retract!(M, copy(x), x, v; kwargs...)
```

src/Sphere.jl
```diff
@@ -5,12 +5,13 @@ end
 project_tangent!(S::Sphere, w, x, v) = (w .= v .- dot(x, v).*x)
 distance(S::Sphere, x, y) = acos(dot(x, y))
 
-function exp!(S::Sphere, y, x, v)
-    nv = norm(S, x, v)
+function exp!(S::Sphere, y, x, v, t=1)
+    scaled_v = t*v
```
sdaxen: This doesn't seem necessary, since we then just compute the norm of `v`. What about

```julia
nv = norm(S, c, v)
tnv = t * nv
    if tnv ≈ 0.0
        y .= x
    else
        y .= cos(tnv).*x .+ (sin(tnv)/nv).*v
    end
    return y
end
```

src/ManifoldMuseum.jl
```diff
@@ -3,24 +3,91 @@ module ManifoldMuseum
...
 project_tangent(M::Manifold, x, v) = project_tangent!(M, copy(x), x, v)
 
 distance(M::Manifold, x, y) = error("Not implemented")
 
+"""
+    dot(M::Manifold, x, v, w)
+
+Inner product of tangent vectors `v` and `w` at point `x` from manifold `M`.
+"""
 dot(M::Manifold, x, v, w) = dot(v, w)
```
sdaxen: It seems odd that we leave `exp!` and `log!`, which are defined by the metric, as unimplemented but default to the inner product defined by the Euclidean metric here. Shouldn't we also leave unimplemented?

src/ManifoldMuseum.jl
```diff
@@ -3,24 +3,91 @@ module ManifoldMuseum
...
-retract(M::Manifold, x, v) = retract!(M, copy(x), x, v)
+Retraction (cheaper, approximate version of exponential map) of tangent
+vector `t*v` at point `x` from manifold `M`.
+"""
+retract(M::Manifold, x, v, t=1) = retract!(M, copy(x), x, v, t)
 
 project_tangent!(M::Manifold, w, x, v) = error("Not implemented")
 project_tangent(M::Manifold, x, v) = project_tangent!(M, copy(x), x, v)
 
 distance(M::Manifold, x, y) = error("Not implemented")
```
sdaxen: Since distance is defined by the inner product, couldn't we have a default implementation of
```julia
distance(M::Manifold, x, y) = norm(M, x, log(M, x, y))
```

Of course it could be specialized for certain manifolds.

kellertuer:
isn't there an infinite loop when calling `log(M,x,x)`? it calls `zero_tangent(M,x)`which calls `log(M,x,x)`again...?

src/ManifoldMuseum.jl
```diff
@@ -3,24 +3,91 @@ module ManifoldMuseum
...
+Result is saved to `y`.
+"""
+exp!(M::Manifold, y, x, v, t=1) = error("Not implemented")
+
+"""
+    exp(M::Manifold, x, v, t=1)
+
+Exponential map of tangent vector `t*v` at point `x` from manifold `M`.
+"""
+exp(M::Manifold, x, v, t=1) = exp!(M, copy(x), x, v, t)
```
sdaxen: `copy(x)` prevents the user from using static arrays (e.g. `SArray`). What about `similar(x)`, which creates a mutable uninitialized array instead?

kellertuer:
I do not yet see the need for `isapprox` – could you elaborate on that?

src/ManifoldMuseum.jl
```diff
@@ -3,24 +3,91 @@ module ManifoldMuseum
...
+export dimension,
+    distance,
+    dot,
+    exp!,
+    geodesic,
+    log!,
+    norm,
+    retract,
+    retract!,
+    typical_distance,
```
sdaxen: I know it was already defined, but what is the meaning and use case of `typical_distance`?

src/ManifoldMuseum.jl
mateuszbaran: Good point, I'll change it.

src/Sphere.jl
mateuszbaran: The other change will make this not necessary.

src/ManifoldMuseum.jl
mateuszbaran: Right, I'll leave it unimplemented by default.

src/ManifoldMuseum.jl
mateuszbaran: OK but this will have to be significantly different for proper support of AD anyway.

src/ManifoldMuseum.jl
kellertuer: Yes, I think that is a good default. If someone has a faster method for his/her manifold, you can really still specialise, but If you just want to get started, its a function less to implement.

src/ManifoldMuseum.jl
kellertuer: In Manopt (the Matlab one) it's used as an initial guess for trust region based methods and we could also rename it to refer to injectivity radius or something. I am using that number as a distance we definetly log is still defined.

mateuszbaran: This, I think, is from Manopt/Manopt.jl: https://www.manopt.org/tutorial.html#manifoldslibrary .

mateuszbaran:
`isapprox` is very useful for correctness checks of various operations. Usually you will make specialised tests for specific manifolds but `isapprox` allows us to have `test_manifold(M::Manifold, pts::AbstractVector)` which is supposed to check that given manifold conforms to the API.

mateuszbaran:
And yes, there is by default an infinite loop, I wasn't sure whether one should have to implement `zero_tangent_vector` and `log!` or `log` and `log!`, this way both variants work. One or the other should probably be left unimplemented though.

src/ManifoldMuseum.jl
sdaxen: I see. Yeah, I think `injectivity_radius(M::Manifold, x)` makes more sense, since it's point-dependent (e.g. on the upper hemisphere, points near the boundary have a lower injectivity radius than points far from the boundary).

kellertuer:
> And yes, there is by default an infinite loop, [...]

Can we find a way to avoid that loop? I just don't have a good idea yet.

mateuszbaran:
> > And yes, there is by default an infinite loop, [...]
> 
> Can we find a way to avoid that loop? I just don't have a good idea yet.

Support for AD will require changing the default implementation of `log` anyway, I'd wait a bit with this until we have some tests for AD.

Issue #2, opened 2019-06-06
An idea for “lightweight” manifolds


kellertuer:
One idea is to have manifolds that come without the “overhead” of types points and vectors. I actually like those, but I can understand the urge for easiness of non typed ones, that just accept matrices.
Maybe we could also handle that as a `Trait`? Do we often need to distinguish between those? For example on the sphere the old difference is really whether to set a field value of `x` or just return the array. So one could also call these `ArrayManifolds` then, if we need to distinguish between them.

sdaxen:
One nice feature of types points/vectors is that we can make it explicit what coordinates are being used and dispatch on different coordinate types. But maybe that's not how we want to implement coordinates.

Perhaps for each manifold if a single preferred canonical set of coordinates can be specified, we also have versions of the functions that accept those as inputs and then wrap them in the corresponding point/vector type?

kellertuer:
You're right. For any `Manifold` you could introduce different points and vectors with different coordinate systems (and hence computations of the corresponding functions). Still, for example for `random_point` that would require that the manifold is aware, which coordinates you currently prefer (maybe as a `pointType` field) then.

PR #8, opened 2019-06-07
Introduces ArrayManifolds

kellertuer:
As we discussed, for a lot of simple manifolds we do not necessarily require own types for points and/or tangent vectors. Since I would like to have such types for later use (maybe it is efficient for some manifold to store the value of an MPoint in 3 values but if you need the value of the point they are combined in a `getValue` call or something)...

So this is a WIP/RFC for a decorator pattern (as we have in mind for Lie groups and such), which should suffice for `Sphere` and other simple manifolds to implicitly get decorated (itself and its points and vectors).

Looking forward to your comments.

I was thinking that one could also call this type `ArrayManifold` since its even more general than just MatrixManifold.


src/MatrixManifold.jl
```diff
@@ -0,0 +1,29 @@
+"""
+    MatrixManifold{M <: Manifold} <: Manifold
+
+A manifold to encapsulate manifolds working on array representations of
+`MPoints` and `TVectors` in a transparent way, such that for these manifolds its
+not necessary to introduce explicit types for the points and tangent vectors,
+but they are encapusalted/stripped automatically when needed.
+"""
+struct MatrixManifold{M <: Manifold}
```
sdaxen: It's a bit confusing calling it a manifold when it doesn't inherit `Manifold`.

src/MatrixManifold.jl
```diff
@@ -0,0 +1,29 @@
...
+    manifold::M
+end
+convert(::Type{M},m::MatrixManifold{M}) where M <: Manifold = m.manifold
+convert(::Type{MatrixManifold{M}},m::M) where M <: Manifold = MatrixManifold(M)
+
+struct MatrixMPoint{V <: AbstractVector{<:Real}} <: MPoint
```
sdaxen: Perhaps I'm misunderstanding what these are for, but why is the wrapped array of a matrix point forced to be a vector?

Also, do we want to only support real matrix points? What about complex matrix groups like SU(2)?

mateuszbaran:
Our interface does not guarantee that all valid points inherit from `MPoint` (or any other abstract type). I think you should clarify what is the purpose of `MPoint` and `TVector`.

kellertuer:
Well, in the end, when you want to write a solver and have vom `x0` or something and we have a manifold where the data is not stored in _one_ array, so we need some struct to keep that data, I do not want to write for all these points `x0::Union{<:AbstractArray,MPoint}` that's all. Its a container for a point (or tangent vector) on a manifold. Just because otherwise several of my planned manifolds will not fit into this setting because the data is not representable as _one_ array.

src/MatrixManifold.jl
kellertuer: Well it's start to think about. Yes Complex would be a good idea – Oh and I meant to always take AbstractArrays – I am sorry, I was maybe not careful enough in writing.

src/MatrixManifold.jl
kellertuer: Oh sorry, forgot that :(

mateuszbaran:
Points and tangent vectors don't have to be represented by a single array in this API, why do you think so? You can have `MPoint` and `TVector` defined in Manopt.jl and everything should work. Do you need to assume something about the type of points and tangent vectors to implement solvers?

kellertuer:
Well, why do we need that for the manifold then? Maybe I am thinking into wrong directions, but _i_ want to know whether you give me a point on a manifold or a vector – semantically. Also, which manifold a point might belong to. `[1., 3., 0.]` is something that might be in Rn or you computed something wrongly and meant to get an point on the sphere. I can not validate that ever, because I have no semantic meaning to this vector. It might also be a valid tangent vector to the North Pole. That's where I would like to have tools to verify that. But I can't without knowing what it is (besides being an array). That gets more complicated, if you have a value that not only has one data item field.

mateuszbaran:
`Manifold` abstract type is needed for correct dispatch. Validation is a good point, it should be written in the docstrings for `MPoint` and `TVector`. BTW, you can always have something like
```julia
function validate_point(M::Manifold, x::MPoint)
    # your validation procedure
end

function validate_point(M::Manifold, x)
    return true # point is probably just an array, can't validate
end
```
Generally I think validation should be optional, not required.

mateuszbaran:
I think we could even have
```julia
function validate_point(M::Manifold, x::MPoint)
    error("Not implemented")
end

function validate_point(M::Manifold, x)
    return true
end
```
in ManifoldMuseum.

kellertuer:
That's already where we have a difference, I would return `false` in the second case, but yes that and something like

```
validate_tangent_vector(M::Manifold, x::MPoint, v::TVector) = error("Validation not implemented for a $(typeof(v)) in the tangent space of a $(typeof(x)) on a manifold $(typeof(M))")
```
is exactly what I would like to do.

The Matrix manifold then serves as a lazy way to switch between the sphere as it is now and a setting where you can use `MPoint` and `TVector` – without having to implement anything (despite the manifold functions acting on arrays as we already have).

I add the docstrings in a minute.

kellertuer:
Oh and I forgot to reference #2

src/ManifoldMuseum.jl
```diff
@@ -4,8 +4,48 @@ import Base: isapprox, exp, log
...
+"""
+    CoTVector
+
+Type for a cotangent vector of a manifold. While a [`Manifold`](@ref) not
+necessarily requires this type, for example when it is implemented for `Vector`s
+or `Matrix` type elements, this type can be used for more complicated
+representations, semantic verification or even dispatch for different
+representations of cotangent vectors and their types on a manifold.
+"""
+abstract type TVector end
```
mateuszbaran: This should be `CoTVector`, right?

mateuszbaran:
It looks better now :+1: . I don't have strong feelings about what `validate_point` or `validate_tangent_vector` should return if the argument is correct, it can be `false`. What other functions (apart from `convert` and `validate_*`) would you specialise in this interface for `MatrixManifold`,  `MPoint`, etc.?

src/ArrayManifold.jl
```diff
@@ -0,0 +1,29 @@
...
+    value::V
+end
+convert(::Type{V},x::ArrayMPoint{V}) where V <: AbstractArray{<:Number} = x.value
+convert(::Type{ArrayMPoint{V}},x::V) where V <: AbstractArray{<:Number} = MatrixPoint{V}(x)
+
+struct MatrixTVector{V <: AbstractArray{<:Number}} <: TVector
+    value::V
+end
+convert(::Type{V},x::MatrixTVector{V}) where V <: AbstractArray{<:Number} = x.value
+convert(::Type{MatrixTVector{V}},x::V) where V <: AbstractArray{<:Number} = MatrixPoint{V}(x)
```
mateuszbaran: This should be implemented as `= MatrixTVector{V}(x)`, right? (or `ArrayTVector`)

kellertuer:
Since the overhead compared to the manifolds that are not typed in `MPoint` and `TVector` is really mainly meant to do validation, one could even specialise something like

```
function exp(M::ArrayManifold,x::P,v::T) where {P <: MPoint, T <: TVector}
validate_tangent_vector(M,x,v)
y = exp(M,x,v)
validate_point(M,x)
return y
```

where we would expect that validate throws errors if it fails with further details _why_ it failed (for example a matrix is not symmetric or not positive definite or something).

Ah and probably one would do that on exp!`, right?

src/ArrayManifold.jl
kellertuer: Damn, I am really a little off today in concentration, I am sorry, too many copy paste errors...

mateuszbaran: No problem :slightly_smiling_face:

mateuszbaran:
I think `y = exp(M,x,v)` should be replaced by `y = exp(M.manifold, x.value, v.value)`, right? This is quite clever and could be done for all (or most) functions.

kellertuer:
Yes, one should dispatch on the values of course. That way you can easily enhance your array non-semantic calculation to a verified one by just changing M (to its `ArrayManifold`)

kellertuer:
I decided to not write `x.value` or `y.value` in the current commit, since we dispatch on manifolds – the `stripping`. from `ArrayMPoint` to the array is done by the convert anyways, since `M.manifold` dispatches already to the original function.

src/ManifoldMuseum.jl
```diff
@@ -44,16 +84,44 @@ zero_tangent_vector!(M::Manifold, v, x) = log!(M, v, x, x)
...
+
+"""
+    is_tangent_vector(M,x,v)
+
+check, whether `v` is a valid tangnt vector in the tangent plane of `x` on the
+[`Manifold`](@ref) `M`. An implementation should first check
+[`is_manifold_point`](@ref)`(M,x)` and then validate `v`. If it is not a tangent
+vector an error should be thrown.
+The default is to return `true`, i.e. if no checks are implmented,
+the assumption is to be optiimistic.
```
mateuszbaran: "optiimistic" should be "optimistic"

src/ManifoldMuseum.jl
```diff
@@ -44,16 +84,44 @@ zero_tangent_vector!(M::Manifold, v, x) = log!(M, v, x, x)
 
 geodesic(M::Manifold, x, y, t) = exp(M, x, log(M, x, y), t)
 
+"""
+    is_manifold_point(M,x)
+
+check, whether `x` is a valid point on the [`Manifold`](@ref) `M`. If it is not,
+an error is thrown.
+The default is to return `true`, i.e. if no checks are implmented,
+the assumption is to be optiimistic.
```
mateuszbaran: "optiimistic" should be "optimistic"

mateuszbaran:
Looks very nice, I'd like to see a test for `Sphere` wrapped in `ArrayManifold` and it could be merged then.

kellertuer:
Well, I started testing and the first thing I ran into is the infinite loop I mentioned – so the `Sphere` defaults should avoid that somehow. Despite that we should find a way to unify what to export (Sphere is currently not exported, while ArrayManifold is).

kellertuer:
Somehow my implementation of just `log!` in  `ArrayManifold` runs into your infinite loop when calling any `log(A,x,y)`.

mateuszbaran:
> However, the formula for `log` on the sphere does not return vectors orthogonal to their base point. Where is that formula from?

I took it from [1] (page 47). They represent tangent vectors with respect to the zero element of the ambient space, not the point they are tangent at.

[1]A. Srivastava and E. P. Klassen, Functional and Shape Data Analysis. Springer New York, 2016.

src/ArrayManifold.jl
```diff
@@ -38,34 +38,37 @@ function project_tangent!(M::ArrayManifold, w, x, v)
 end
 
 function distance(M::ArrayManifold, x, y)
-    is_manifold_point(M,x)
-    is_manifold_point(M,y)
+    is_manifold_point(M.manifold,x)
```
mateuszbaran: Maybe it would make sense to define
```julia
is_manifold_point(M::ArrayManifold, x) = is_manifold_point(M.manifold, x)
```

test/runtests.jl
```diff
@@ -24,6 +24,18 @@ function test_manifold(M::Manifold, pts::AbstractVector)
     @test norm(M, pts[1], tv1) ≈ sqrt(dot(M, pts[1], tv1, tv1))
 end
 
+function text_arraymanifold()
```
mateuszbaran: `text` should be `test`

kellertuer: I did not even read that yet, since I was really confused by the tangent vector not being orthogonal.

kellertuer:
> I took it from [1] (page 47). They represent tangent vectors with respect to the zero element of the ambient space, not the point they are tangent at.

Yes one always writes them that way, still, that should be a vector orthogonal to the base (i.e. dot(x,v) should be zero) and that's not the case here.

mateuszbaran:
Could you give an example where it's not zero (up to numerical errors)?
dot(x,v) = dot(x, y - cos(acos(dot(x, y)))*x) = dot(x, y - dot(x, y)*x) = dot(x, y) - dot(x, dot(x, y)*x) = dot(x, y) - dot(x, y)*dot(x, x) = dot(x, y) - dot(x, y) = 0
Unless I'm too tired to spot an error here.

kellertuer:
Oh, I found my bug, still the formula of log looks strange to me, could we add the formulae in math in the docstoring? that would maybe be helpful to users.

Since I had a typo in ArrayManifold, I will really only continue tomorrow finishing the tests.

src/Sphere.jl
```diff
@@ -35,3 +35,26 @@ end
 
 zero_tangent_vector(S::Sphere, x) = zero(x)
 zero_tangent_vector!(S::Sphere, v, x) = (v .= zero(x))
+
+function is_manifold_point(S::Sphere,x)
+    if size(x) != S.shape
+        throw(DomainError(size(x),"The point $(x) does not lie on $S, since its size is not $(S.shape)."))
+    end
+    if abs(norm(x)-1) > 10^(-15)
```
sdaxen: Maybe we should provide tolerance keyword arguments like in `Base.isapprox`, which could just be forwarded to an `isapprox` call on this line.

kellertuer: I tried, but I am still failing then providing a larger value of tolerance for the dot in the tangent test.

kellertuer:
I am not that much in favour of the current state since the explicit conversion takes some lines for every default function.

Further, the documentation is missing, I will provide that as soon as the implementation is fixed in its style.

mateuszbaran:
I'll resolve that infinite loop soon, I think generic version of `log`  shouldn't call `zero_tangent_vector`. You can change the formula for `log`, I've just copied something from FunManifolds to get started.

These explicit conversions should not be necessary, you can do:
```julia
is_manifold_point(M::ArrayManifold, x) = is_manifold_point(M.manifold, x)
```
and then you just use `is_manifold_point(M, x)` in methods of `ArrayManifold`.

kellertuer:
Great. For now this branch has working tests. I will refactor, when the infinite loop is gone. Maybe a default for the `zero_tangent_vector` can be `zero(x)` assuming that (as a default) `x` and its tangent vectors have similar representations? Otherwise one could also default that to an error.

mateuszbaran:
#16 resolves the infinite loop, what do you think about it?

kellertuer:
> #16 resolves the infinite loop, what do you think about it?

Well, I do see that there is no infinite loop, but the code you wrote is too clever for me, I don't get anything, what `similar_result` and `similar_result_type` do. What's with the `args...` for example, what are they for and wha `x...` and then you only use `x[1]`? But maybe you can just explain them over there... For me that whole code looks to me like – I've not yet learned enough Julia the last 1.5 years for that.

mateuszbaran:
I'll try to explain these two functions in #16, they are not totally nice but necessary to have modifying and non-modifying operations without too much repetition.

src/Sphere.jl
sdaxen: It feels a bit clunky passing a whole approx function. I was thinking something more like.

```julia
function is_manifold_point(S::Sphere, x; kwargs...)
   ...
   if isapprox(norm(x), 1; kwargs...)
      ...
   end
end
```
This also makes it easier to change the tolerances. Instead of needing to call something like `is_manifold_point(m, x, x->isapprox(x; atol=1e-3))`, you'd call `is_manifold_point(m, x; atol=1e-3)`.

src/Sphere.jl
```diff
@@ -36,25 +36,25 @@ end
 zero_tangent_vector(S::Sphere, x) = zero(x)
 zero_tangent_vector!(S::Sphere, v, x) = (v .= zero(x))
 
-function is_manifold_point(S::Sphere,x)
+function is_manifold_point(S::Sphere,x, local_isapprox=Base.isapprox)
```
sdaxen: It feels a bit clunky passing a whole approx function. I was thinking something more like.

```julia
function is_manifold_point(S::Sphere, x; kwargs...)
   ...
   if isapprox(norm(x), 1; kwargs...)
      ...
   end
end
```
This also makes it easier to change the tolerances. Instead of needing to call something like `is_manifold_point(m, x, x->isapprox(x; atol=1e-3))`, you'd call `is_manifold_point(m, x; atol=1e-3)`.

kellertuer: Thanks for the good idea, will add that today.

kellertuer:
Now the optional exactness for the tests is passed down. As soon as the infinite loop is gone on `master`, i.e. #16 is merged, I'll check which of the explicit conversions one can spare again (hopefully: all) and then this is finally also completed I hope.

kellertuer:
I still don't understand enough of the `similar_result` and `similar_result_type`. While the code is still too clever for my level of Julia understanding, it also does not work with my current idea of the Array Manifolds. I can not get rid of an error

```
MethodError: no method matching one(::Type{Any})
```

which for me might indicate that the type returned is `Any`?

Also trying to reduce the number of function definitions to carefully enhance the input with types did also not work, then I am always running into non-implemented fallbacks. I took the chance and enhanced the error messages at least :)

mateuszbaran:
Your error looks like you are trying to use this code with `Array{Any}`, I will fix that tomorrow.

> Also trying to reduce the number of function definitions to carefully enhance the input with types did also not work, then I am always running into non-implemented fallbacks.

I'll look at it.

> I took the chance and enhanced the error messages at least :)

Great 👍 .

mateuszbaran:
OK, I've fixed the code so that the tests pass now.

kellertuer:
I already thought about something like `array_value` (in Manopt I use `getValue` exactly for that), but `similar` and `eltype` – I still have to get used to that more :)

Thanks for the fix, the branch can not be merged.

PR #12, opened 2019-06-08
[WIP] Implementation of metrics

sdaxen:
In this conception of metrics, the metric wrapping the manifold must be passed to functions that rely on the metric. So functions like `exp`, `log` and `dot` will ultimately dispatch on the metric instead of the manifold. There is a default conversion from a manifold to one equipped with a(n ambient) Euclidean metric, but I'm not certain this is a good idea.

See discussion in https://github.com/JuliaNLSolvers/ManifoldMuseum/issues/9.


sdaxen:
Basic implementations of `dot`, `norm`, `distance`, `angle`, `log`, and `exp` are now implemented, with demonstration using ambient Euclidean metric on Euclidean space and on the sphere. I've left `retract`,`project_tangent`, `vector_transport`, `zero_tangent_vector`, `random_point`, and `random_tangent_vector` unchanged, since I'm not exactly certain what they mean in general.

The Euclidean metric induces the usual Euclidean inner product on the tangent vectors when expressed in the embedded Euclidean space, so we get concepts of norm for free, but `log!` and `exp!` must be implemented for each manifold with an induced Euclidean metric. At least, I think this is the case and would appreciate review.

sdaxen:
I've also changed the default behavior of `injectivity_radius`, which is dependent on the metric. By default, it returns `Inf`, since it will primarily be used for checking (the original value of 1.0 was arbitrary and meaningless, as far as I can tell). There's also a global `injectivity_radius` that returns the infimum of the injectivity radii of all points on the manifold.

kellertuer:
Oh, somehow my review got lost. Can you please think about  a nicer name than `local_matrix`?  Maybe `local_metric`?

sdaxen:
> Oh, somehow my review got lost. Can you please think about a nicer name than `local_matrix`? Maybe `local_metric`?

Hmm, I do think `local_matrix` captures what it returns well (the local matrix representation of the metric tensor), but I agree it's not a nice name. Maybe `local_metric` or `tensor_matrix` could work.

sdaxen:
Okay, so I greatly simplified the interface for metrics. When implementing a manifold, one may or may not implement `dot`, `exp!`, and `log!` (assuming some metric, but they don't need to specify one). If the user does, then they're done. When they equip the manifold with a metric via `MetricManifold`, most manifold functions should forward to the decorated manifold (when https://github.com/JuliaNLSolvers/ManifoldMuseum/pull/15 is merged), _except_ `dot`, `exp!`, and `log!`, to prevent accidental usage of the wrong metric. Once they implement `dot` or `local_metric`, `exp!`, and `log!` dispatched on their manifold and metric, then they're done. See the exampled of `EuclideanMetric` and `Sphere` for example. Here the default Euclidean metric was assumed for the manifolds, so when equipping the manifold with the Euclidean metric, it simply forwards to the wrapped manifold's version of the functions.

mateuszbaran:
I feel like there are so many manifolds with some sort of default metric that this approach of forcing all implementations to care about distinguishing manifolds and metrics on them is a bit too much hassle. How about making an additional abstract type for manifolds with metrics (for example `MetrizedManifold` or something), so you can still just do
```julia
struct Sphere{N} <: MetrizedManifold end
exp!(M::Sphere, ...) = ...
```
and when you *do* care about distinguishing manifolds and metric, you can do something like
```julia
struct SE{N} <: Manifold end
```
which wouldn't have all the defaults (as the defaults would only be for `MetrizedManifold`s and you could decorate that using `MetricManifold`?

sdaxen:
> I feel like there are so many manifolds with some sort of default metric that this approach of forcing all implementations to care about distinguishing manifolds and metrics on them is a bit too much hassle.

This implementation doesn't require implementations to consider a `Metric` at all. See for example the `Sphere` implementation. But one can implement a different metric and then apply it to the `Sphere` if one wishes.

> How about making an additional abstract type for manifolds with metrics (for example `MetrizedManifold` or something), so you can still just do

We _could_ do this. Currently, this is handled by simply implementing `dot`, `exp!`, and `log!`. I don't know that this warrants another type to the hierarchy.

mateuszbaran:
I think I'm finally starting to see what you are doing here. This would greatly benefit from some examples, such as implementation of a sphere by an euclidean manifold of equal dimension with a custom metric and tests that show that metric structure matches the existing implementation of sphere.

sdaxen:
Sorry for the delay, I was traveling last week and wasn't able to work on this much. I'm also not satisfied with this envisioning of metrics. `MetricManifold` _should_ make it easier to implement Riemannian manifolds.

I'll list some of my thoughts. The functionality I would like to provide includes

- The ability to equip an existing manifold with a new metric _without_ needing to reimplement all of the machinery for that manifold (`MPoint` and `TVector` types, checks, etc.).
- Out-of-the-box functionality for induced metrics on embedded manifolds.
- Default (inefficient) numerical methods for `exp!` and `log!` assuming the Levi-Civita connection that work as soon as you define `inner` or `local_metric`. [Absil et al](https://press.princeton.edu/sites/default/files/inline-files/Absil_Chap5.pdf) discuss this in Proposition 5.4.1.
- Conversion of Euclidean gradient to manifold gradient using metric.
- Adjustment of metric for coordinate change via `MPoint` and `TVector` (simply requires pre- and post- multiplication of metric matrix by Jacobian of coordinate change.
- Caching of local matrix and its inverse (some functions, such as `angle` or `geodesic`, use the local matrix multiple times). This is all the more important when a coordinate change is involved.

As an example of how this should work, take the sphere. One should be able to define the Sphere as an embedded submanifold of Euclidean space with the embedding constraint and _automatically_ get a fully-functioning manifold with geodesics, etc.

Alternatively, given an existing implementation of the sphere, one should be able to define an `MPoint` type for a new coordinate type (e.g. spherical coordinates), and equip the sphere with the Euclidean metric applied _to that coordinate system_  (note that this is not the same thing as the Euclidean metric applied to embedded coordinates). For free, one should get the induced metric on the embedded coordinates.

As the latter case makes clear, the metric is closely related to the coordinate chart, and that is currently not represented by this implementation.

kellertuer:
I never mind any delay. I am happy, we started this common approach :)

I am completely on your side with the idea of “decorating“ a manifold with a metric (where we usually assume some kind of default metric for now) and like your explanations and plans.

I would also see the “alternative” part, not as an alternative but as a second approach for the types points and tangent vectors.

But – how do you aim to change the current code to reflect your ideas? Maybe we can elaborate on this further by really starting with an example like the Sphere.

sdaxen:
> But – how do you aim to change the current code to reflect your ideas? Maybe we can elaborate on this further by really starting with an example like the Sphere.

I don't know yet. The thing that I'm currently thinking through is how to tie the metric to a set of coordinates. Maybe something like

```julia
using ForwardDiff
import LinearAlgebra.pinv

struct TransformedMetric{G<:Metric,P,PI,J,JI} <: Metric
    parent_metric::G
    map::P
    inv_map::PI
    jac::J
    inv_jac::JI
end

function TransformedMetric(g, f, f⁻¹)
    return TransformedMetric(g, f, f⁻¹,
                             x -> ForwardDiff.jacobian(f, x),
                             x -> ForwardDiff.jacobian(f⁻¹, x))
end

function local_metric(M::MetricManifold{MT,GT}, x) where {MT<:Manifold,GT<:TransformedMetric}
    g = metric(M)
    J = g.jac(x)
    return J' * local_metric(manifold(M), g.map(x)) * J
end

function inner(M::MetricManifold{MT,GT}, x, v, w) where {MT<:Manifold,GT<:TransformedMetric}
    J = metric(M).jac(x)
    return inner(manifold(M), x, J * v, J * w)
end

function exp!(M::MetricManifold{MT,GT}, y, x, v) where {MT<:Manifold,GT<:TransformedMetric}
    g = metric(M)
    J = g.jac(x)
    xprime = g.map(x)
    yprime = similar(xprime)
    vprime = J * v
    exp!(manifold(M), yprime, xprime, vprime)
    y .= g.inv_map(yprime)
end

function log!(M::MetricManifold{MT,GT}, v, x, y) where {MT<:Manifold,GT<:TransformedMetric}
    g = metric(M)
    J = g.jac(x)
    xprime = g.map(x)
    yprime = g.map(y)
    vprime = similar(xprime)
    log!(manifold(M), vprime, xprime, yprime)
    v .= pinv(J) * vprime
end
```

Because the exponential and logarithmic maps (and consequently geodesics) are coordinate invariant, this makes it easy to pull those back from the standard coordinates (usually embedded coordinates) to some other coordinate set without needing to re-derive the expressions in that coordinate set. Of course, they're not as fast, but the methods can always be specialized. Here's an example on the sphere:

```julia
julia> using LinearAlgebra, Plots
julia> import ManifoldMuseum.Sphere;
julia> sph_to_cart(x) = [cos(x[1]) * sin(x[2]), sin(x[1]) * sin(x[2]), cos(x[2])];
julia> cart_to_sph(x) = [atan(x[2], x[1]), acos(x[3])];
julia> S = MetricManifold(Sphere(2), EuclideanMetric());  # 2-sphere embedded in Euclidean space
julia> g = TransformedMetric(EuclideanMetric(), sph_to_cart, cart_to_sph);
julia> M = MetricManifold(S, g);
julia> x = cart_to_sph(normalize(randn(3)))
2-element Array{Float64,1}:
 -0.4924920930604519
  2.271021838720005 
julia> local_metric(M, x)
2×2 Array{Float64,2}:
 0.584761     1.33335e-17
 1.33335e-17  1.0        
julia> Diagonal([sin(x[2])^2, 1])  # round metric in spherical coordinates
2×2 Diagonal{Float64,Array{Float64,1}}:
 0.584761   ⋅ 
  ⋅        1.0
julia> y = cart_to_sph(normalize(randn(3)))
2-element Array{Float64,1}:
 0.1432196248965485
 1.628158500237012 
julia> points = hcat([sph_to_cart(geodesic(M, x, y, t)) for t=0:.1:10]...);
julia> plot(points[1, :], points[2, :], points[3, :])
``` 
newplot

sdaxen:
The most recent commit adds an early implementation of a numerically integrated exponential map. All it requires is a definition of `local_metric`, and it assumes the Levi-Civita connection. It computes the Christoffel symbols using automatic differentiation and then uses them to solve the ODE of the curve with zero acceleration. Here's an example with the Euclidean metric on spherical coordinates of the sphere.

```julia
julia> using LinearAlgebra, ManifoldMuseum, Plots;
julia> gr();
julia> import ManifoldMuseum: Sphere, local_metric;
julia> struct SphericalEuclideanMetric <: RiemannianMetric end;
julia> local_metric(::MetricManifold{<:Sphere,<:SphericalEuclideanMetric}, x) = Diagonal(ones(length(x)));
julia> sph_to_cart(x) = [cos(x[1]) * sin(x[2]), sin(x[1]) * sin(x[2]), cos(x[2])];
julia> cart_to_sph(x) = [atan(x[2], x[1]), acos(x[3])];
julia> S = MetricManifold(Sphere(2), SphericalEuclideanMetric());
julia> x = cart_to_sph(normalize(randn(3)))
2-element Array{Float64,1}:
 1.488795350111658
 1.103016376263013
julia> v = normalize(randn(2))
2-element Array{Float64,1}:
  0.8547018526419826
 -0.5191191993081771
julia> points = hcat([exp(S, x, t.*v) for t=0:.1:20]...);
julia> plot(points[1, :], points[2, :])
```
tmp2

```julia
julia> points = hcat([sph_to_cart(exp(S, x, t.*v)) for t=0:.1:20]...);
julia> plot(points[1, :], points[2, :], points[3, :])
```
tmp

As expected, the exponential map gives a straight line in spherical coordinates.

This implementation doesn't currently work for embedded coordinates, where the tangent space is a linear subspace of the tangent space of the ambient space. I'm working on figuring out the necessary changes. I still have no idea how to implement a default `log!`.

mateuszbaran:
Interesting. 

Default `log!` shouldn't be that hard, it can be calculated from geodesic between two given points and that's a pretty standard problem, up to all the numerical problems that will likely arise and the fact that sometimes geodesics don't exist or are not unique.

kellertuer:
I don't understand the middle image.

But concerning log! a default can also be to solve the problem os uniqueness, to implement the default only of the case that the distance of the two points is less than `injectivity_radius(M)`.

mateuszbaran:
> But concerning log! a default can also be to solve the problem os uniqueness, to implement the default only of the case that the distance of the two points is less than `injectivity_radius(M)`.

But by default you don't know the distance (you need a geodesic for that) and what you find may not necessarily be the shortest geodesic. So you really need all of them (or a decent sample) to decide whether you can return one or not. And that is assuming you know the `injectivity_radius`, which you sometimes may not, or it can be selected too conservatively and valid geodesics will be rejected.

kellertuer:
Well I assumed the distance might at least approximately be known, ok; but `infectivity_radius(M)` would even be the global minimum over all local infectivity radii, so yes it would be quite conservative. For me the default implementations are always conservative but reasonably working implementations (where one maybe can not do better). If for your special manifold you know better (i.e. a method that for that case accepts more geodesics) you have to reimplement for your case.

sdaxen:
> Default `log!` shouldn't be that hard, it can be calculated from geodesic between two given points and that's a pretty standard problem, up to all the numerical problems that will likely arise and the fact that sometimes geodesics don't exist or are not unique.

Could you point me to some references for how it's been done before?

> I don't understand the middle image, but despite that – very neat.

Middle image is the same geodesic as the first image but visualized in embedded coordinates instead of spherical coordinates.

The point (for me) of `MetricManifold` is to make it really easy to try different metrics and visualize the results without needing to do a lot of math. Having `exp!` is great, but we really need `log!` to compute distances between sets of points, which are what most people probably want to know. So a priori, the user doesn't know the injectivity radius, which is strictly defined as the minimum (of all geodesics) geodesic distance for which the exponential map is a diffeomorphism (equivalently, for which log is the inverse of exp), which is more conservative than we have defined it in the documentation. To even get the minimal geodesic distance, we already need the log map, so I don't think injectivity radius will help us here.

But suppose given 2 points, we have some factory that returns valid geodesics of maximum length d whose endpoints are the 2 points. The minimum value of d seen so far is then the maximum distance that we must consider along all future geodesics, so that defines a similar search radius.

mateuszbaran:
> > Default `log!` shouldn't be that hard, it can be calculated from geodesic between two given points and that's a pretty standard problem, up to all the numerical problems that will likely arise and the fact that sometimes geodesics don't exist or are not unique.
> 
> Could you point me to some references for how it's been done before?

This seems to be a good example: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2650001/ , or this: http://www.math.nus.edu.sg/~bao/teach/ma5233/so2.pdf . I doubt there are many examples of numerical calculation of geodesics on arbitrary manifolds but it looks to me like it should be possible using similar optimization methods. I assume you only consider geodesics between points covered by the same chart?

sdaxen:
> This seems to be a good example: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2650001/ , or this: http://www.math.nus.edu.sg/~bao/teach/ma5233/so2.pdf .

Thanks, I'll take a look.

> I assume you only consider geodesics between points covered by the same chart?

At the moment, yes, until we have the machinery for defining atlases. The Christoffel symbols/metric tensor are coordinate-dependent, so if we have a way to detect when a point has moved from an open subset covered by one chart to another, we can then change position/velocity coordinates and continue the geodesic.

@dkarrasch might be interested in this progress, as (I think) it touches on some of the features on their wish list.

src/Metric.jl
```diff
@@ -1,50 +1,202 @@
...
+function exp_diffeq_system!(du, u, p, t)
+    n = Int(size(u, 1) / 2)
+    M = p
+    x = view(u, 1:n)
+    dx = view(u, n+1:2n)
+    du[1:n] .= dx
+    Γ = christofell_symbols_second(M, x)
+    @tensor d²x[k] := -Γ[i,j,k] * dx[i] * dx[j]
+    du[n+1:end] .= d²x
+end
```
dkarrasch: In low-dimensional systems (dim<100), you may want to consider using the out-of-place form with signature `(u, p, t)`, returning `StaticVectors`, even though this may add another dependency. Not sure if performance is important here (like solving for many geodesics)?

mateuszbaran:
Very good progress :+1:

src/ManifoldMuseum.jl
```diff
@@ -87,6 +92,35 @@ then implements the feature itself.
...
 
+"""
+    manifold(M::Manifold; recurse::Bool=false)
+
+If `M` is a decorator manifold, return the underlying manifold. if `recurse` is
+`true`, continue un-decorating until a non-decorator manifold is encountered.
+"""
+function manifold end
+
+@traitfn function manifold(M::MT; recurse::Bool=false) where {MT<:Manifold;IsDecoratorManifold{MT}}
```
mateuszbaran: Are you sure the `recurse` keyword argument is actually necessary? I think normal dispatch should handle multiple layers of unwrapping well.

src/ManifoldMuseum.jl
```diff
@@ -290,15 +324,26 @@ export dimension,
...
+    MetricManifold,
+    IsRiemannianManifold,
+    isriemannianmanifold,
+    metric,
     local_metric,
-    inverse_local_metric
-export Euclidean, EuclideanMetric, TransformedEuclideanMetric
+    inverse_local_metric,
+    det_local_metric,
+    christofell_symbols_first,
```
mateuszbaran: Christoffel has two 'f' and one 'l'.

src/Metric.jl
```diff
@@ -1,50 +1,202 @@
...
+"""
 abstract type RiemannianMetric <: Metric end
 
-abstract type LorentzianMetric <: Metric end
+"""
+    LorentzMetric <: Metric
+
+Abstract type for Lorentz metrics, which have a single temporal dimension.
+"""
+abstract type LorentzMetric <: Metric end
```
mateuszbaran: Do you want `LorentzMetric`s to have the signature +++...+- or ---..-+?

src/Metric.jl
```diff
@@ -1,50 +1,202 @@
...
+    varying inner products on the tangent space. See [`inner`](@ref).
+"""
 abstract type Metric end
 
+@doc doc"""
+    RiemannianMetric <: Metric
+
+Abstract type for Riemannian metrics, a family of positive definite inner
+    products. The positive definite property means that for $v \in T_x M$,
+    the inner product $g(v,v) > 0$.
```
mateuszbaran: For `x` different from 0.

src/ManifoldMuseum.jl
mateuszbaran: By the way, I think the name of this function is not specific enough, maybe something like `unwrap_manifold` would be better?

src/Metric.jl
kellertuer: I would opt for `++...+-`, but that is (just) a design decision that should be well documented then.

src/Metric.jl
```diff
@@ -0,0 +1,202 @@
...
+    manifold::M
+    metric::G
+end
+
+@traitimpl IsDecoratorManifold{MetricManifold}
+
+@traitdef IsRiemannianManifold{M}
+@traitimpl IsRiemannianManifold{M} <- isriemannianmanifold(M)
+
+isriemannianmanifold(::Type{MT}) where {MT<:Manifold} = false
```
kellertuer: I know what you would like to achieve with this function, but it seems to be inconsistent, since this would lead to the case, that just `Sphere(3)` or `Euclidean(5)` are not riemannian manifolds, which is misleading. Not necessarily type-wise, but from the perception of the user, since both actually _are_ riemannian (just not _explicitly_ equipped with a metric in our toolbox at that point).

 could we do something like `hasMetric(::Type{MetricManifold{MT,GT},::GTc) where you could check whether `GT` and `GTc` are of same type? This would be more general than isriemannianmaifold – you could also check whether a `MetricManifold` has a `Lorentzian` metric – and it would just not be defined for other cases than `MetricManifolds`.

src/Euclidean.jl
```diff
@@ -0,0 +1,43 @@
...
+end
+
+function inverse_local_metric(M::MetricManifold{<:Manifold,EuclideanMetric}, x)
+    return local_metric(M, x)
+end
+
+det_local_metric(M::MetricManifold{<:Manifold,EuclideanMetric}, x) = one(eltype(x))
+
+inner(::Euclidean, x, v, w) = dot(v, w)
+inner(M::MetricManifold{<:Manifold,EuclideanMetric}, args...) = inner(manifold(M), args...)
```
kellertuer: Can you elaborate on the use case for this line? Because if I just would do `MetricManifold(SymmetricPositiveDefinite,EuclideanMetric)`(which is inconsistent in itself, yes!) this would return to the non-Euclidean metric the SymmetricPositiveDefinite matrices might have by default (the affine invariant). Yes that's not the only one, but `Hyperbolic(3)` could be used for the same inconsistency.

src/Metric.jl
mateuszbaran: I think we should care, at least to some degree, about performance. `StaticArrays` will likely be added as a dependency anyway as just using `SizedArray` can provide big benefits for operations small arrays.

src/ManifoldMuseum.jl
kellertuer: Or maybe `basic_manifold` in the sense of an undecorated one?

src/Metric.jl
kellertuer: Do you mean `v` different from the zero vector? I think so. It should be $g(v,v) \geq 0$ for $v \in T_x M$ and $g(v,v) = 0$ iff $v=0$,

mateuszbaran: Yes, of course, `v` is different from the zero vector.

src/ManifoldMuseum.jl
sdaxen: > Are you sure the recurse keyword argument is actually necessary?

Most of the time we only want to unwrap one layer of a multiply-decorated manifold. But sometimes we may want to get the underlying topological manifold by unwrapping all layers (e.g. `manifold_dimension`). Perhaps 2 different functions make sense: `base_manifold` for unwrapping one layer and `topological_manifold` for unwrapping all layers.

src/ManifoldMuseum.jl
sdaxen: Oh wow, thanks.

src/Metric.jl
sdaxen: I'm not certain which is best, (++...+-), (--...-+), (-+...++), (+-...--). All 4 conventions seem common. I wonder if it's too crazy to support multiple conventions via dispatch.

src/ManifoldMuseum.jl
kellertuer: but you never know the order of the decorators, so there is not much use in saying “one” layer. You might have a `DecoA{DecoB{M}}` or a `DecoB{DecoA{M}}` – so you're either looking for the manifold (itself) and want `M` or you are looking for a specific decorator (since you want to access the metric directly for example). `base_manifold` would return `M`. And maybe to access properties we better do `get`ers?

sdaxen: I think you're right, but it will be easier to tell when we get to multiply-decorated manifolds, like Lie groups equipped with metrics, which should have all of the functionality of the Lie group and of a metric manifold. And I think in a case like that, a specific ordering of decorators makes more sense. (ie I think it makes more sense to equip the Lie group with a metric than to equip a metric manifold with a group operation).

mateuszbaran: Multiple decorators is an important issue. One day, probably not that far away, we will start asking questions like "what metric should have a quotient manifold of a riemannian manifold with a custom metric by an action of a Lie group of rotations with a custom metric?" and even figuring out the correct order of decorators will be a problem.

Is there actually a reason to allow more than one (likely partial) order of decorators?

src/Metric.jl
mateuszbaran: I doubt I'll ever use it so all I care about is that the final decision should be documented.

src/ManifoldMuseum.jl
kellertuer: Of course there might be non-useful orderings – but should we be so strict to employ that if we can (in my opinion quite easily) ignore ordering by saying (with the trait) just say: Throw away deco until you find the on I need for my metric.

In short, my counter question is: Are we restricting too much by imposing a (eventually partial) order? I am just not sure ;)

src/Metric.jl
kellertuer: I might use it, but it's really a matter of taste (mine is `++...+-`), but as long as its documented, I am also fine with exactely one, because its easier to implement.

src/ManifoldMuseum.jl
mateuszbaran: It's hard to decide at this moment, there may be issues that are order-dependent. Something like "unwrap decorators to get X" functions should be good enough for now.

src/Euclidean.jl
sdaxen: Good point, this should forward to `dot` instead.

src/Metric.jl
sdaxen: Good point. Check out https://github.com/JuliaNLSolvers/ManifoldMuseum/pull/12/commits/5d0ed52d3edcebc531cec834e05023ac265199bb. It provides a simple interface for indicating the default metric for a manifold and getting automatic forwarding of key functions.

src/Metric.jl
sdaxen: I'm comfortable defaulting to this convention. Like anywhere else, a user is free to change the convention by implementing a special `MPoint` that changes it.

src/Euclidean.jl
```diff
@@ -34,10 +36,11 @@ end
 det_local_metric(M::MetricManifold{<:Manifold,EuclideanMetric}, x) = one(eltype(x))
 
 inner(::Euclidean, x, v, w) = dot(v, w)
-inner(M::MetricManifold{<:Manifold,EuclideanMetric}, args...) = inner(manifold(M), args...)
+inner(::MetricManifold{<:Manifold,EuclideanMetric}, x, v, w) = dot(v, w)
```
kellertuer: Wouldn't this require, that (for example for every matrix matrifold) `dot` does something resonable? I think it would either require the constructor for `MetricManifold{X,EuclideanMetric}` to avoid inconsistent constructions or this should default to an error.

I know what you are aiming at, I am just not yet sure how to define it without introducing too inconsistent ways of “misuse”, that might confuse the user then.

src/ManifoldMuseum.jl
```diff
@@ -332,14 +331,13 @@ export Metric,
     LorentzMetric,
     EuclideanMetric,
     MetricManifold,
-    IsRiemannianManifold,
-    isriemannianmanifold,
+    HasMetric,
```
kellertuer: One could also use `IsMetricManifold` which is a little more consistent with the `IsDecoratorManifold`, but I would be fine with either naming.

src/Metric.jl
```diff
@@ -78,41 +101,48 @@ Determinant of local matrix representation of the metric tensor $g$
 """
 det_local_metric(M::MetricManifold, x) = det(local_metric(M, x))
 
-inner(M::MetricManifold, x, v, w) = dot(v, local_metric(M, x) * w)
+@traitfn inner(M::MMT, x, v, w) where {MT<:Manifold,GT<:Metric,MMT<:MetricManifold{MT,GT};!HasMetric{MT,GT}} = dot(v, local_metric(M, x) * w)
+@traitfn inner(M::MMT, x, v, w) where {MT<:Manifold,GT<:Metric,MMT<:MetricManifold{MT,GT};HasMetric{MT,GT}} = inner(M.manifold, x, v, w)
+
```
kellertuer: That's a nice idea, because it enables the user to specify a performant default metric (with `MT`) that Is then just transparently passed through 👍

src/Euclidean.jl
sdaxen: Sure, but `dot` for matrix inputs is the Frobenius inner product, which to me seems the sensible choice when using a Euclidean metric. What would be a case where this default behavior wouldn't be sensible?

src/ManifoldMuseum.jl
sdaxen: I removed `IsRiemannianManifold` because I don't have a great use for it right now. It's easy to dispatch on whether a `MetricManifold` is using a Riemannian metric, but none of the functionality I'm implementing so far is unique to Riemannian manifolds, so this seems overkill.

src/Euclidean.jl
kellertuer: Maybe I'd prefer not to use this default at all but rather use your neat `HasMetric` trait approach. My fear is (I think I wrote that), that you can equip a non-Euclidean metricised manifold (SPDs for example) and then get strange results here. Maybe I'd like to more avoid that than really criticising the choice of `dot`. And maybe this user error can also not be avoided here; I mean _if_ you equip SPDs oh Hyperbolic space with Euclidean – it'S your fault not the code. So yeah, maybe dot is the right choice.

src/Metric.jl
sdaxen: Yeah this satisfies the two ideals we were discussing earlier. It keeps it simple to implement a new manifold without even thinking explicitly about metrics (but obviously assuming some sensible default), and if one then wants to implement several metrics, one can use `HasMetric` to designate the default that was assumed for the original manifold.

For now `local_metric` and its relatives take `MetricManifold`s only, since if you're implementing that, you're definitely thinking about metrics and should be working with a `Metric` and `MetricManifold`. And that gets you numerically integrated geodesics and the other tensors/curvature.

src/Euclidean.jl
sdaxen: Yeah, I'm perfectly fine with letting the user make that error if they like, so long as the documentation and code is unambiguous about what `EuclideanMetric` means. There's actually some inconsistency here now. I'm working on a prototype interface for maps, which will let us construct pullback metrics. `Sphere` would then default to a pullback metric from an embedding map in R^{n+1} instead of just defaulting to the misleading `EuclideanMetric`.

src/Metric.jl
sdaxen: I agree that we should support `SizedArray`s throughout the package. https://github.com/JuliaNLSolvers/ManifoldMuseum/pull/12/commits/6c3514bdb68a223bfc1cb49f99f91dca7e8f9607 and https://github.com/JuliaNLSolvers/ManifoldMuseum/pull/12/commits/5926eca563673be7d7596578a1cb3192d4438574 switches to the out-of-place form and supports sized arrays throughout, regardless of dimension. I'd appreciate any feedback.

We also should have a version of `exp` that takes a time vector and returns a vector of points corresponding to the curve. For numerically integrated geodesics, we're already computing the entire curve anyways, and this will be nice to have even for explicit geodesics.

test/runtests.jl
```diff
@@ -157,63 +160,104 @@ end
...
                     end
                 end
             end
+
+            R = riemann_tensor(M, x)
+            for i=1:n
+                for j=1:n
+                    for k=1:n
+                        for l=1:n
+                            if (i,j,k,l) == (1,1,2,2)
```
kellertuer: this quadruple for-loop can be shortened to (and I was quite happy when I learned that)
````
for i=1:n, j=1:n, j=1:n, l=1:n
    [...]
end
```
see at the end of  https://docs.julialang.org/en/v1/manual/control-flow/#man-loops-1

src/Metric.jl
kellertuer: You mean a function like `geodesic(M,x,y)` that basically returns the function `t -> exp(M,x, t*log(M,x,y))`? one could then also define `geodesic(M,x,y,t)` to evaluate the geodesic at t and the short hand `geodesic(M,x,y,T)` for a vector `T` of time points that simply broadcasts on t.

test/runtests.jl
sdaxen: Oh that is nice. Thanks for the tip!

src/Metric.jl
```diff
@@ -0,0 +1,245 @@
...
+    return G
+end
+
+function solve_exp_ode(M::MetricManifold, x, v, args...; kwargs...)
+    u0 = x
+    du0 = Base.convert(typeof(x), v)
+    tspan = (0.0, 1.0)
+
+    function exp_problem(du, u, p, t)
+        n = size(u, 1)
```
dkarrasch: What is `n` good for here? How would one treat matrix manifolds here? Would you `vec` them or compute on matrix level? Maybe it's more the `length(u)` that is relevant?

dkarrasch:
One more thing. I learned in the `Distances.jl` package that computing the geodesic distance on the 2-sphere via `acos(dot(v,w))` may be inaccurate, especially for points close to each other/close to antipodal. Do you think it's worth copying the `Haversine` distance from there for that special case? It requires transforming coordinates to "geographic" coordinates (longitude-latitude) though, if I'm not mistaken.

Project.toml
```diff
@@ -4,17 +4,19 @@ authors = ["Seth Axen <seth.axen@gmail.com>", "Ronny Bergmann <manopt@ronnybergm
 version = "0.1.0"
 
 [deps]
+DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Einsum = "b7d42ee7-0b51-5a75-98ca-779d3107e4c0"
```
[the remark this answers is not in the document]
mateuszbaran: Isn't `Tensors.jl` only good for some small tensors? https://kristofferc.github.io/Tensors.jl/stable/man/constructing_tensors/#Constructing-tensors-1

src/Metric.jl
sdaxen: Something like that, but I was thinking for `exp` too, since we may not always have a `log` map. On that note, `geodesic` maybe should be renamed to `minimal_geodesic`, since that's more descriptive of what it means (both return geodesics, but `geodesic` returns specifically the distance-minimizing geodesic between the provided points).

kellertuer: As long as we have a way to distinguish points and vectors, we could use `geodesic(M,x,y)` and `geodesic(M,x,v)`, otherwise we need two names. And yes, I am fine with adding the word `minimal`, though one could also use `shortest`.

sdaxen: How would `geodesic(M,x,v)` in this conception differ from `exp(M,x,v)`?

Project.toml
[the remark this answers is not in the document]
sdaxen: I would *love* to use `Tensors.jl`, but yes, it only supports up to dimension 3 right now, and I don't see any issues for supporting higher-dimensional tensors. If you know of any other good fits here, I'd be interested. Most of these tensors/tensor-like objects have symmetries/sparsities that it would be nice to explicitly handle.

Project.toml
```diff
@@ -4,17 +4,19 @@ authors = ["Seth Axen <seth.axen@gmail.com>", "Ronny Bergmann <manopt@ronnybergm
 version = "0.1.0"
 
 [deps]
+DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
```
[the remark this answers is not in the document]
sdaxen: Good advice, thanks!

src/Metric.jl
sdaxen: Oh, yeah that `n` is unused right now. Tbh, I've not put much thought into matrix manifolds yet, or even to what a metric matrix should mean in that space. I need to figure out the necessary changes for integrating geodesics on embedded manifolds in coordinates of the embedding first.

We've been using `Sphere` as our test case for vector manifolds. Perhaps we should pick a simple non-Euclidean non-group matrix manifold for testing as well.

src/Metric.jl
kellertuer: `exp(M,x,v)`returns a point `y`. `geodesic(M,x,v)` returns a function `t -> exp(M,x,t*v)`, so `geodesic(M,x,v)(1)` is the same as `exp` but one could again also implement `geodesic(M,x,v,T)` for a time vector `T` to get a point vector back.

sdaxen: Got it. Hmm, but in many cases, we may not have a way to distinguish points from vectors, so what about this?
`shortest_geodesic(M,x,y)` returns `geodesic(M,x,log(M,x,y))`
`shortest_geodesic(M,x,y,t)` returns `geodesic(M,x,log(M,x,y),t)`
`shortest_geodesic(M,x,y,T)` returns `geodesic(M,x,log(M,x,y),T)`
`geodesic(M,x,v)` returns `t->exp(M,x,v,t)`
`geodesic(M,x,v,t)` returns `exp(M,x,v,t)`
`geodesic(M,x,v,T)` returns `exp(M,x,v,T)`
`exp(M,x,v)` computes exponential map
`exp(M,x,v,t)` returns `exp(M,x,t*v)`
`exp(M,x,v,T)` returns `[exp(M,x, t*v) for t in T]`

This makes behavior/arguments more transparent and doesn't require any additional overloads than the usual `exp!` functions. In some cases, a more efficient form of `exp(M,x,v,T)` might be possible, such as with the numerically integrated geodesics, and the user is free to implement that.

kellertuer: At least in my head, any `geodesic(M,x,v)` would still be shortest (for example with length pi on the sphere you reach – not uniquely – the opposite point but all these are shortest geodesics). Thats why until now I called both (xy and xv geodesics) the same.

Having both is fine - I just think your default geodesics might be shortest, too.

sdaxen: Geodesics are only locally distance-minimizing, so they don't guarantee the shortest distance between any two points on the geodesic. A shortest geodesic would be a (not necessarily unique) geodesic that does guarantee the shortest distance and would require a `log` map.

I'll give an example. Suppose we have points `x=(0,0,1)`, `y=(-1,0,0)` with `v=(1,0,0)`. `geodesic(M,x,v)` will give a geodesic that includes `x` and `y`, but it will take `t=3pi/4` time units to get to `y` tracing out an arc of the same length, whereas  `shortest_geodesic(M,x,y)` is equivalent to `geodesic(M,x,(-pi/4,0,0))` and will get to `y` at `t=1` tracing out an arc length of `pi/4`. And indeed, `pi/4` is the distance between `x` and `y`.

sdaxen:
> One more thing. I learned in the `Distances.jl` package that computing the geodesic distance on the 2-sphere via `acos(dot(v,w))` may be inaccurate, especially for points close to each other/close to antipodal.

That's interesting. Could you point me to some more information on this?

src/Metric.jl
kellertuer: Yes I know that minimizing only holds locally :) but I wasn't aware of your idea of modelling here, so thanks for the explanation, with that idea the naming is consistent and very good.

sdaxen: Everything I said above, but replace `pi` with `tau=2pi`.

kellertuer: Yeah. I also got the idea on the smaller sphere ;)

sdaxen: Sorry, I commented that before I saw your comment :).

dkarrasch:
> That's interesting. Could you point me to some more information on this?

I guess I messed it up in my head. Haversine formula is when points are given in geographical coordinates in the first place. Nevermind.

src/Metric.jl
```diff
@@ -0,0 +1,245 @@
...
+inverse_local_metric(M::MetricManifold, x) = inv(local_metric(M, x))
+
+@doc doc"""
+    det_local_metric(M::MetricManifold, x)
+
+Determinant of local matrix representation of the metric tensor $g$
+"""
+det_local_metric(M::MetricManifold, x) = det(local_metric(M, x))
+
+@traitfn inner(M::MMT, x, v, w) where {MT<:Manifold,GT<:Metric,MMT<:MetricManifold{MT,GT};!HasMetric{MT,GT}} = dot(v, local_metric(M, x) * w)
```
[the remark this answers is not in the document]
sdaxen: See above discussion on `Tensors.jl` only supporting few dimensions. Will reconsider if that changes. Thanks for opening that issue, by the way.

sdaxen:
I'd like to revisit the status of this pull request, since it's getting quite involved. :) Here are the main things that are missing:

1. Numerically approximating exponential maps on embedded submanifolds. The techniques to do this generally don't use the metric and connection directly but rather use operations such as retraction and projection. Since we're defining those separately, I don't think this pull request needs to address them.
2. Default log maps. I'm looking into the papers @mateuszbaran shared earlier and geodesic shooting techniques, but I'm not certain any of these work in multiple charts or even a single chart that wrap around. I'm considering leaving these unimplemented, so we can revisit them later.
3. The currently implemented geodesic machinery works in a single coordinate chart. Right now it assumes that (most) of the manifold can be covered in a single unbounded chart. This won't even work correctly for the sphere, so we need an implementation of charts and atlases. I have some ideas for how to do this, but I don't know that this PR is the right place to develop that machinery.
4. I think we need a general framework for maps/morphisms. A minimal map should define its domain, its codomain, and wrap a function. If we implement various types of maps, then we can compose them in smart ways. (e.g. the inverse of a bijection is also a bijection, but composing a bijection with an injection produces an injection). It also enables us to handle pushforwards and pullbacks, such as pullback metrics. I also don't know that this PR is the right place for that.
5. The above components are necessary to define suitable, unambiguous defaults. e.g. our default `exp` and `log` maps on the `Sphere{n}` would be defined by an embedding map in `Euclidean{n}`. So something like `MetricManifold{Sphere{n},PullBackMetric{Embedding{Sphere{n},Euclidean{Tuple{n}}}}}` could forward to our default implementation.
5. More documentation

I'm not certain where a good stopping point is, whether I should just finish up a few of the things above and merge it or switch gears to work on charts and maps before continuing on this. Thoughts?

mateuszbaran:
Yes, this PR is getting quite large, it doesn't need to have everything.

1. Yes, this can be done in a separate PR.
2. AFAICT these methods were designed with a single chart in mind, without any wrapping around. Implementing default `log` based on the metric with support for multiple charts looks like an open problem. This can be in a separate PR as well.
3. This can also be done separately.
4. This looks like a functionality that could be in a separate package, there is something similar in `AbstractAlgebra.jl`: https://nemocas.github.io/AbstractAlgebra.jl/latest/map/
5. Looks reasonable.

I would advise merging this relatively soon, it changes quite a few things and merging will require more and more work.

kellertuer:
5 Looks reasonable, 6 should be done here; for the rest I'm +1 with @mateuszbaran .

src/Metric.jl
```diff
@@ -47,19 +47,25 @@ convert(::Type{MT},M::MetricManifold{MT,GT}) where {MT,GT} = M.manifold
...
+A `Trait` to mark a `Manifold` `M` as being shorthand for a
+`MetricManifold{M,G}` with metric `G`. This can be used to forward functions
+called on the `MetricManifold` to the already-imlemented functions for the
+`Manifold`.
 
 For example,
 
 ```
-@traitfn myFeature(M::MMT, k...) where {MT<:Manifold,GT<:Metric,MMT<:MetricManifold{MT,GT};HasMetric{MT,GT}} = myFeature(M.manifold, k...)
+@traitfn myFeature(M::MMT, k...) where {MT<:Manifold,
```
mateuszbaran: I'd change the function name to `my_feature` to be consistent.

sdaxen:
Okay, I think I've covered everything we discussed. Does it look alright to merge now?

src/ManifoldMuseum.jl
```diff
@@ -67,20 +79,48 @@ IsDecoratorManifold{M}`), a specific function should be implemented as a
...
 then implements the feature itself.
 """
 @traitdef IsDecoratorManifold{M}
 
+"""
+    base_manifold(M::Manifold)
+
+Strip all decorators on `M`, returning the underlying topological manifold.
+"""
+function manifold end
```
dkarrasch: There seems to be a mismatch between what is described in the docstring and the function name?

src/ManifoldMuseum.jl
```diff
@@ -166,13 +218,21 @@ exp!(M::Manifold, y, x, v) = error("Exponential map not implemented on a $(typeo
...
 
 exp(M::Manifold, x, v, t) = exp(M, x, t*v)
 
+"""
+    exp(M::Manifold, x, v, T::AbstractVector)
+
+Exponential map of tangent vector `t*v` at point `x` from manifold `M` for
+each `t` in `T`.
+"""
+exp(M::Manifold, x, v, T::AbstractVector) = [exp(M, x, v, t) for t in T]
```
dkarrasch: Do you need to define this? With broadcasting, wouldn't the user simply write `exp.(M, x, v, T)` for an `AbstractVector` of real numbers? By the way, maybe it makes sense to restrict the `t` (or `T`) arguments to `<:Real`?

src/ManifoldMuseum.jl
```diff
@@ -181,7 +241,59 @@ function log(M::Manifold, x, y)
...
+"""
+    geodesic(M::Manifold, x, y, t)
+
+Get the point at time `t` traveling from `x` along the geodesic with initial
+point `x` and velocity `v`.
+"""
+geodesic(M::Manifold, x, v, t) = exp(M, x, v, t)
+
+"""
+    geodesic(M::Manifold, x, y, T::AbstractVector)
```
dkarrasch: There's a typo: `y` should be `v`. Whatever you decide above, obviously, `geodesic` must be consistent regarding the `T`-vector.

src/ManifoldMuseum.jl
```diff
@@ -196,12 +308,22 @@ end
 
 Distance such that `log(M, x, y)` is defined for all points within this radius.
 """
-injectivity_radius(M::Manifold, x) = 1.0
+injectivity_radius(M::Manifold, x) = Inf
```
dkarrasch: This is a duplicate of line 318.

src/Sphere.jl
```diff
@@ -13,25 +13,30 @@ generates the $\mathbb S^{n}\subset \mathbb R^{n+1}$
 struct Sphere{N} <: Manifold end
 Sphere(n::Int) = Sphere{n}()
 
+@traitimpl HasMetric{Sphere,EuclideanMetric}
+
 @doc doc"""
-    dimension(S::Sphere)
+    manifold_dimension(S::Sphere)
 
 returns the dimension of the manifold $\mathbb S^n$, i.e. $n$.
```
dkarrasch: should be "Return"?

src/ManifoldMuseum.jl
sdaxen: > Do you need to define this? With broadcasting, wouldn't the user simply write `exp.(M, x, v, T)` for an `AbstractVector` of real numbers?

In some cases, it's significantly more efficient to compute the whole trajectory at once than to broadcast, the prime example being the numerically integrated exponential maps.

>By the way, maybe it makes sense to restrict the `t` (or `T`) arguments to `<:Real`?

Yeah, I think so. I'll change it.

dkarrasch: But this is not achieved by what is written here?
```julia
[exp(M, x, v, t) for t in T]
```
doesn't compute one geodesic and then evaluates at all `t in T`, but it re-computes the exponential map for every `t`.

src/ManifoldMuseum.jl
sdaxen: No, line 318 differs by an argument. :)

dkarrasch: Yes, I still seem to have trouble with reading... 😆

src/ManifoldMuseum.jl
sdaxen: Yes, this is just a suitable default that will be fine for most manifolds but which developers are free to specialize for their manifold if there's a more efficient approach. `MetricManifold` does just that when `HasMetric` is not defined (See https://github.com/JuliaNLSolvers/ManifoldMuseum/pull/12/files#diff-53108f2d38bac2c2779a8604bf2fc16fR307)

src/ManifoldMuseum.jl
sdaxen: No worries. :)

src/ManifoldMuseum.jl
mateuszbaran: Wouldn't `map(geodesic(M, x, v), T)` be a more reasonable default here? In case someone overloads `geodesic` with a more efficient implementation?

sdaxen: But now that I'm looking at it more closely, a better default is to use `map`, since it (tries to) respect the type of `T` and is more efficient.

sdaxen:
Thanks @dkarrasch for reviewing the PR. I've committed the recommended changes and a few fixes for broken tests. Unless there are any more issues, I think this PR is ready for merge.

src/ManifoldMuseum.jl
```diff
@@ -150,13 +188,27 @@ Norm of tangent vector `v` at point `x` from manifold `M`.
...
+angle(M::Manifold, x, v, w) = acos(inner(M, x, v, w) / norm(M, x, v) / norm(M, x, w))
+
 """
     exp!(M::Manifold, y, x, v, t=1)
 
 Exponential map of tangent vector `t*v` at point `x` from manifold `M`.
 Result is saved to `y`.
 """
-exp!(M::Manifold, y, x, v, t) = exp!(M::Manifold, y, x, t*v)
+exp!(M::Manifold, y, x, v, t::Real) = exp!(M::Manifold, y, x, t*v)
```
dkarrasch: `exp!(M::Manifold, y, x, v, t::Real) = exp!(M, y, x, t*v)`

sdaxen: Good catch, thanks.

PR #15, opened 2019-06-09
Decorator for Manifolds

kellertuer:
Since we use several manifolds as decorators (metric, array, Lie for example), this is a proposal to implement "pass through” fallsbacks for functions that a decorator provides, so for example the Lie group action or a metric or such.so that for a

```
ArrayManifold{LieGroup{MetricManifold{M}}}
```
not only the order is irrelevant but any operator “finds” its decorator manifolds automatically. This should solve #15.

### DIFF
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -2,11 +2,13 @@ module ManifoldMuseum
 
 import Base: isapprox, exp, log
 import LinearAlgebra: dot, norm
+import Markdown: @doc_str
+using SimpleTraits
 
 abstract type Manifold end
 
 """
-   IsDecoratorManifold
+    IsDecoratorManifold
 
 A `Trait` to mark a manifold as a decorator type. For any function that is only
 implemented for a decorator (i.e. a Manifold with `@traitimpl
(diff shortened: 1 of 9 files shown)


src/ManifoldMuseum.jl
```diff
@@ -3,10 +3,24 @@ module ManifoldMuseum
...
+   IsDecoratorManifold
+
+A `Trait` to mark a manifold as a decorator type. For any function that is only
+implemented for a decorator (i.e. a Manifold with `@traitimpl
+IsDecoratorManifold{M}`), a specific function should be implemented as a
+`@traitfn`, that transparently passes down through decorators, i.e.
+
+```
+@traitfn myFeature(M::Mt, k...) where {Mt; IsNice{Mt}} = myFeature(M.manifold, k...)
+myFeature(M::MyManifold, k...) = #... my explicit impementation
```
sdaxen: `IsNice{Mt}` should be `IsDecoratorManifold{Mt}`. Should also document that the pass through can be overriden if desired. e.g.
`myFeature(MyDecoratorManifold{MyManifold}, k...) =  #... my explicit override

sdaxen: We could also document the easier syntax `@traitfn myFeature(M::::IsDecoratorManifold, k...) = myFeature(M.manifold, k...)`

kellertuer:
While we do not have an explicit example for this trait yet, I would opt to also merge this one and maybe use the `MetricManifold` to illustrate a first example how to use this Trait.

PR #16, opened 2019-06-09
Initial pieces of support for AD

mateuszbaran:
`similar_result`should take us most of the way to a decent support of automatic differentiation. Additionally, a few tests for types other than `Array{Float64}`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,9 @@ version = "0.1.0"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [extras]
+DoubleFloats = "497a8b3b-efae-58df-a0af-a86822472b78"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
--- a/src/ManifoldMuseum.jl
+++ b/src/ManifoldMuseum.jl
@@ -55,12 +55,21 @@ retract!(M::Manifold, y, x, v, t) = retract!(M, y, x, t*v)
 Retraction (cheaper, approximate version of exponential map) of tangent
 vector `t*v` at point `x` from manifold `M`.
 """
-retract(M::Manifold, x, v) = retract!(M, similar(x), x, v)
+function retract(M::Manifold, x, v)
+    xr = similar_result(M, retract, x, v)
+    retract!(M, xr, x, v)
+    return xr
+end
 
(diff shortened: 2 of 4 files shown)


kellertuer:
I don't completely understand what `similar_result` does, can you elaborate on that?
Also, what's `MVector`?

mateuszbaran:
`similar_result` is like `similar` but it tries to guess the element type for array that will store the result based on types of arguments. Look at how it's used in `log`: it allocates the array that will be filled by `log!`. `similar` won't do because `x` and `y` might have different types and we need to determine the correct eltype for the new array. `similar_result` takes a few not-obviously-necessary arguments (manifold and function) that make it possible to override it in cases where the default guess is incorrect.

`MVector` is from [StaticArrays](https://github.com/JuliaArrays/StaticArrays.jl).

kellertuer:
Ah, I could not find `MVector` and thought `M` is for manifold, too narrow in terms ;) Ok with that I know _what_ it does, just not yet how. I'll try to understand that then. Thanks for also providing tests on different types.

mateuszbaran:
`similar_results` receives that manifold, function that is implemented (the non-modifying variant that will be implemented using the modifying variant) and arguments that were passed to that non-modifying variant. `similar_result_type` guesses the eltype for the matrix that will be the result of the operation. For example, `log` for sphere involves some arithmetic so that if we implemented the non-modifying variant, the eltype of result could be different from eltypes of arguments. `similar_result_type` takes eltypes of argument arrays, constructs values of them and adds them up and finally takes the type of the result. It won't always be the correct guess unfortunately but it should be close enough to be useful. I hope this helps 🙂

mateuszbaran:
About that `x[1]`, I just hope that this array has the correct size for the result -- this can be easily changed if it causes any problems.

kellertuer:
Hm, on my side several of the new tests from this branch fail for example with

```
ForwardDiff support: Test Failed at /Users/bergmann/Repositories/Software/ManifoldMuseum/test/runtests.jl:46
  Expression: d12 ≈ ForwardDiff.derivative(exp_f, t)
   Evaluated: 1.5707964f0 ≈ NaN
```
even on master. 

Can you check that?

mateuszbaran:
It fails for me too, it's a problem of `log` of Sphere being very AD-unfriendly for `t=0`. I'm not sure what we should do about it, this test will likely be problematic for many other manifolds. In FunManifolds I'm currently ignoring incorrect results of AD in such cases. BTW, it was discussed here: https://github.com/JuliaNLSolvers/ManifoldMuseum/pull/5#discussion_r291318806 .

kellertuer:
Oh, I must have missed that discussion. Thanks for the link. I will try to understand why it fails.

PR #18, opened 2019-06-21
Initial version of rotations manifold

mateuszbaran:
I've implemented most of the rotations manifold. Some things are not there yet, like support for Double64 (which requires https://github.com/JuliaMath/DoubleFloats.jl/pull/78 ) or support for AD.


kellertuer:
I will be travelling the next two weeks, so I will carefully look at your approach, especially towards AD and take another manifold then (maybe SPD) following the naming and implementation ideas. I will try during travels, to comment on/review your approach.

mateuszbaran:
This implementation is mostly based on yours from Manopt.jl (with the notable difference that `inner` is scaled by a factor of 1/2 to get this: `@test norm(M, pts[1], v) ≈ angles[2] - angles[1]`)

I'm going to implement a few more manifolds (probably power and product manifolds) in the near future. I'm also thinking about porting `FunManifolds.jl` to our API relatively soon but I'll probably wait until this package has CI and is registered.

kellertuer:
Ah, even better (will look at the code only now);
When the Interface is fixed and manifolds are covered as far as I need them, I will also migrate Manopt.jl (slowly), after this package has a CI and is registered.

docs/src/manifolds/rotations.md
```diff
@@ -0,0 +1,7 @@
+# Rotations
+
```
kellertuer: In the long run I would prefer to have a general explanation and maybe an illustration and applications in the doc (like explaining signs do...in a museum)

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+# Constructor
+
+    Rotations(n, retract! = rectract_qr!)
+
+generates the $\mathrm{SO}(n) \subset \mathbb R^{n\times n}$
+"""
+struct Rotations{N, TRetr<:Function} <: Manifold
+    retract!::TRetr
+end
+Rotations(n::Int, retract!::TRetr = retract_qr!) where TRetr = Rotations{n,TRetr}(retract!)
```
kellertuer: ah, also neat. I am currently putting the choice for the retraction into the problem (but we don't have that here of course) in the next version of Manopt.jl. This way introduces an implicit dispatch for the default retraction. Nice idea.

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+
+@doc doc"""
+    dimension(S::Rotations)
+
+returns the dimension of the manifold $\mathrm{SO}(n)$, i.e. $\frac{n(n-1), 2}$.
+"""
+dimension(S::Rotations{N}) where {N} = div(N*(N-1), 2)
```
kellertuer: I think we renamed it to manifold_dimension – but maybe we should just start a proper documentation with default names somewhen and then unify it globally again.

src/Rotations.jl
kellertuer: Oh and what about the default inverse retraction (first/second/higher order approximation of log)?

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+
+compute the inner product of the two tangent vectors `w,v` from the tangent
+plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
+restriction of the metric from the embedding, i.e. $ (v,w)_x = \frac{1,2}v^\mathrm{T}w $.
+"""
+inner(S::Rotations, x, w, v) = dot(w, v)/2
+
+project_tangent!(S::Rotations, w, x, v) = w .= (transpose(v).-v)./2
+
+function exp!(S::Rotations, y, x, v)
```
sdaxen: Here we are assuming the tangent `v` is a matrix, but above we assumed it was a vector?

sdaxen: Here we are assuming the tangent `v` is a matrix, but above we assumed it was a vector?

docs/src/manifolds/rotations.md
mateuszbaran: I totally agree, we need good docs here.

src/Rotations.jl
mateuszbaran: I skipped inverse retractions for now, I'll add them soon.

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+dimension(S::Rotations{N}) where {N} = div(N*(N-1), 2)
+
+@doc doc"""
+    inner(S,x,w,v)
+
+compute the inner product of the two tangent vectors `w,v` from the tangent
+plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
+restriction of the metric from the embedding, i.e. $ (v,w)_x = \frac{1,2}v^\mathrm{T}w $.
+"""
+inner(S::Rotations, x, w, v) = dot(w, v)/2
```
sdaxen: Would be good to have documentation on what `v` and `w` represent. I understand in the context of Lie algebras but not in the context of an embedded manifold. How is the factor of 1/2 a consequence of the embedding?

src/Rotations.jl
mateuszbaran: I'll make a test for `manifold_dimension`.

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+    is_tangent_vector(S,x,v; kwargs... )
+
+checks whether `v` is a tangent vector to `x` on the [`Rotations`](@ref)
+space `S`, i.e. atfer [`is_manifold_point`](@ref)`(S,x)`, `v` has to be of same
+dimension as `x` and orthogonal to `x`.
+The tolerance for the last test can be set using the ´kwargs...`.
+"""
+function is_tangent_vector(S::Rotations{N},x,v; kwargs...) where N
+    is_manifold_point(S,x)
+    if size(v) != (N, N)
```
sdaxen: Here `v` is called a vector but is expected to be a matrix.

src/Rotations.jl
sdaxen: Huh. I don't think I like this. `Rotations` will ultimately be decorated with a `SpecialOrthogonal` group structure and could then be decorated with a different metric and/or combined via a semidirect product to get a `SpecialEuclidean` element. It seems odd to have the base rotation manifold specialized on the retraction, which in many cases may not be used. I get that the desired behavior is to specify a default retraction for an optimization problem, but I wonder if there's a better way to do it that doesn't bake it into the manifold itself.

Since what differentiates these retractions is their order, I wonder if a good way to go would be to let `retract!` take an order parameter, which it would round up. So if a 2nd-order approximation is requested it would retract using the 2nd order approximation if available and perform the `exp!` map if not. Then you could do have something like `retraction(M,order=1)` that would return a suitable retraction of at least that order for your manifold.

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+"""
+struct Rotations{N, TRetr<:Function} <: Manifold
+    retract!::TRetr
+end
+Rotations(n::Int, retract!::TRetr = retract_qr!) where TRetr = Rotations{n,TRetr}(retract!)
+
+@doc doc"""
+    dimension(S::Rotations)
+
+returns the dimension of the manifold $\mathrm{SO}(n)$, i.e. $\frac{n(n-1), 2}$.
```
sdaxen: Should be `\frac{n(n-1)}{2}`

src/Rotations.jl
```diff
@@ -0,0 +1,148 @@
...
+"""
+dimension(S::Rotations{N}) where {N} = div(N*(N-1), 2)
+
+@doc doc"""
+    inner(S,x,w,v)
+
+compute the inner product of the two tangent vectors `w,v` from the tangent
+plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
+restriction of the metric from the embedding, i.e. $ (v,w)_x = \frac{1,2}v^\mathrm{T}w $.
```
sdaxen: Should be `\frac{1}{2}`

src/Rotations.jl
mateuszbaran: Specifying retraction order (or method) by an additional parameter is a good idea, but I'd prefer to have it resolved by dispatch instead. For example 
```julia
retract!(M::Manifold, y, x, v, method) = exp!(M, y, x, v)

retract!(M::Rotations, y, x, v, ::Val{:polar}) = ...
```

src/Rotations.jl
mateuszbaran: Where was it assumed that `v` is a vector?

sdaxen: The docstrings describe it as a vector.

src/Rotations.jl
kellertuer: I like that the mabnifold provides a default retration, it can still be overwritten by a decorator, why not?
but, hm, yeah a dispatch by value is fine, since in the end that also provides a proper default.

src/Rotations.jl
kellertuer: tangent vector ;) but that might still be represented by a matrix.

mateuszbaran:
@sdaxen This book: http://www.eeci-institute.eu/GSC2011/Photos-EECI/EECI-GSC-2011-M5/book_AMS.pdf looks like a good source about matrix manifolds like this one.

kellertuer:
You can get that book also officially open access https://press.princeton.edu/absil and yes I can recommend that book definetly.

src/Rotations.jl
mateuszbaran: Well, some spaces of matrices are also vector spaces. In particular here, tangent vector is represented by a matrix -- that's how matrix manifolds usually work.

src/Rotations.jl
mateuszbaran: I've improved the documentation, is it OK now?

src/Rotations.jl
kellertuer: I think (as said above) the misconception is, that the _representation_ of a _tangent vector_ can be a matrix (or even a higher dimensional tensor) but its then always meant as an element from the vectorspace (for matrices then the vectorspace R ^{n,n}.

sdaxen: Yeah, I'm just thinking that we need to be clear in the documentation that "tangent vector" may have different representations for different manifolds but always refers to an element from a "vector space". In cases like SO(3) where the tangent vectors are sometimes equivalently represented as vectors and sometimes as matrices, we could use dispatch to differentiate between the two.

kellertuer: Yes, we should carefully document that and always be exact, for example always really write “tangent vector” e.g. `the tangent vector v is here represented by a matrix` to keep that always clear.

src/Rotations.jl
sdaxen: It's more clear regarding tangent vectors now. It still seems that the  factor of 1/2 is chosen pragmatically, not as a consequence of the induced metric of the embedding, which is confusing. Or perhaps this is not an isometric embedding, where the Euclidean space is scaled by 1/sqrt(2) before the rotation matrices are embedded in it. I hadn't seen a discussion of this in Absil's book before, but I'll check again.

src/Rotations.jl
mateuszbaran: OK, done. I've decided to make an abstract type, it will be better once we get to product manifolds.

src/Rotations.jl
mateuszbaran: I've taken the factor 1/2 from this book actually: https://www.springer.com/gp/book/9781493940189 . Maybe it's a bad choice after all.

mateuszbaran:
@kellertuer , I'm finishing implementation of inverse retractions and I've found a problem. The polar inverse retraction does not seem to work for 2d rotations by 0 and pi/2. To be precise, an error is reported for Float32 but not for Float64 but AFAICT it's a problem with the method: `sylvester` function https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/index.html#LinearAlgebra.sylvester requires A and -B to have no eigenvalues with equal real parts but in my case both eigenvalues are purely imaginary. What should we do about it?

test/runtests.jl
```diff
@@ -186,16 +192,30 @@ end
...
             test_forward_diff = false,
             test_reverse_diff = false,
-            retraction_methods = retraction_methods)
+            retraction_methods = retraction_methods,
+            inverse_retraction_methods = inverse_retraction_methods)
 
         v = log(M, pts[1], pts[2])
         @test norm(M, pts[1], v) ≈ (angles[2] - angles[1])*sqrt(2)
+
+        if eltype(T) isa Float64
```
mateuszbaran: An error is thrown for `Float32`.

kellertuer:
> What should we do about it?

Like for the logarithmic map, also inverse retractions are only locally defined and the two points you mention are exactly that: Not within injectiviy radius. What did we agree on for the log (on the sphere)? We could provide a deterministic choice or throw an error, but it should be consistent throughout the Museum.

mateuszbaran:
This wasn't discussed for sphere yet. I think throwing a custom `OutsideOfInjectivityRadius` exception would be appropriate.

kellertuer:
I would maybe call it `OutOfInjectivityRadiusError` but in general I am fine with an error for that case, since `log` (or here `inverseRetraction` is not a function in that case (but multivalued).

src/Rotations.jl
```diff
@@ -0,0 +1,199 @@
...
+"""
+struct Rotations{N} <: Manifold end
+Rotations(n::Int) = Rotations{n}()
+
+@doc doc"""
+    manifold_dimension(S::Rotations)
+
+returns the dimension of the manifold $\mathrm{SO}(n)$, i.e. $\frac{n(n-1)}{2}$.
+"""
+manifold_dimension(S::Rotations{N}) where {N} = div(N*(N-1), 2)
```
sdaxen: I'd make this a generated function, since its value is known at compile time, and we may check dimension often.

src/Rotations.jl
```diff
@@ -0,0 +1,199 @@
...
+
+@doc doc"""
+    inner(S, x, w, v)
+
+compute the inner product of the two tangent vectors `w, v` from the tangent
+plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
+restriction of the metric from the embedding, i.e. $ (v,w)_x = \frac{1}{2}v^\mathrm{T}w $.
```
sdaxen: Docstring now outdated.

sdaxen:
A few general notes.

What about renaming `Rotations` to `RotationMatrices`? I mention this because there are other ways of representing rotations as abstract geometric objects (e.g. quaternions), but we're restriction ourself to the usual rotation matrix representation. I don't have a strong preference here, but I'm curious about your thoughts.

It seems that this implementation implicitly defines the tangent space at any point as the tangent space at the identity, the Lie algebra. This is different from how we have been discussing the tangent spaces in coordinates of the embedding so far, where they are a vector subspace of the ambient space that is tangent to the manifold at the point. I favor using the consistent definition of tangent space here (for `X \in SO(n), V \in T_e SO(n), W = X V \in T_X SO(n)`) (see for example Absil eq 3.26).  We could always have variants of the functions that drop the point of the manifold, assuming the identity element, but in my opinion, those make more sense when we wrap `Rotations{n}` with the Lie group structure (without that structure, why prefer the identity element over any other)? 

For at least `n=2` and `n=3`, the matrix exponential/logarithm can be computed analytically, so we could specialize those computations.

I believe the injectivity radius of SO(2) is pi, not pi/2.

src/Rotations.jl
mateuszbaran: In such simple cases the compiler can automatically detect that the value is constant, so the `@generated` annotation doesn't really help.

sdaxen: Oh, good to know!

mateuszbaran:
Renaming this to `RotationMatrices` is reasonable but I don't have strong preference either. Just using `Rotations` is shorter though and I doubt we will ever have an implementation that uses a different representation.

I'm also slightly in favor of changing the tangent space representation to the one that doesn't transform to `T_e SO(n)`, the only reason I've picked the current solution is that's how it's done in Manopt.jl and it was easier to do this.

Yes, we should have some better implementations for these two cases (`n=2`, `n=3`).

The injectivity radius for SO(2) for this embedding (for the logarithmic map) is pi*sqrt(2) I think? I've removed the scaling. It's smaller for the polar inverse retraction.

kellertuer:
Since for the Sphere we also have a default one (that does not say it's the embeddd one), I would like to keep `Rotations`. Another choice could be that as soon as we have several implementations, we use `Rotations` as a constructor with options that uses the “best” (most efficient) implementation for each dimension by default and can easily be used to switch to others. 

Concerning the usage of the Lie algebra for all tangent spaces – the Lie groups I worked with always had the advantages that you spare several transforms if you save tangent vectors as elements from the Lie algebra. But you're right that might be a good default when using exactly that machinery. As long as this way of working can be done somehow I am fine with another default (i.e. the full representation).

mateuszbaran:
@kellertuer Did you check your implementation of the QR retraction/inverse retraction in Manopt.jl? Code in this PR is almost identical and `inverse_retraction!` doesn't really give a vector that retracts to the original point. I've commented out the failing test for now.

sdaxen:
> The injectivity radius for SO(2) for this embedding (for the logarithmic map) is pi*sqrt(2) I think? I've removed the scaling. It's smaller for the polar inverse retraction.

Yes, `pi*sqrt(2)` looks right, though I worry now that this behavior might be too confusing for users. I'm starting to change my mind on the default embedding. I don't have a problem with embedding the rotation matrices in a Euclidean space with a metric of `I/2`, which would make distances angles and the injectivity radius length pi, as long as its clearly documented. 

> Concerning the usage of the Lie algebra for all tangent spaces – the Lie groups I worked with always had the advantages that you spare several transforms if you save tangent vectors as elements from the Lie algebra. But you're right that might be a good default when using exactly that machinery. As long as this way of working can be done somehow I am fine with another default (i.e. the full representation).

Yeah, this is the annoying part of it, but I think it's a common misconception that the Lie algebras is the tangent space at more than one point, and it would be nice if we were consistent for novice users to avoid that confusion.

mateuszbaran:
> Yes, `pi*sqrt(2)` looks right, though I worry now that this behavior might be too confusing for users. I'm starting to change my mind on the default embedding. I don't have a problem with embedding the rotation matrices in a Euclidean space with a metric of `I/2`, which would make distances angles and the injectivity radius length pi, as long as its clearly documented.

We can also change the embedding to have it isometric and get the geodesic distance be equal to the angle.

> > Concerning the usage of the Lie algebra for all tangent spaces – the Lie groups I worked with always had the advantages that you spare several transforms if you save tangent vectors as elements from the Lie algebra. But you're right that might be a good default when using exactly that machinery. As long as this way of working can be done somehow I am fine with another default (i.e. the full representation).
> 
> Yeah, this is the annoying part of it, but I think it's a common misconception that the Lie algebras is the tangent space at more than one point, and it would be nice if we were consistent for novice users to avoid that confusion.

Then I think I'll make a flag for this option, something like 
```julia
abstract type AbstractTangentRepresentation end
struct OrthogonalTangentRepresentation <: AbstractTangentRepresentation end
struct LieAlgebraTangentRepresentation <: AbstractTangentRepresentation end

struct Rotations{N, TSR<:AbstractTangentRepresentation} <: Manifold end
```
with orthogonal representation being the default.

sdaxen:
> We can also change the embedding to have it isometric and get the geodesic distance be equal to the angle.

Can you explain this a little more? I thought an isometric embedding was one where the pullback metric was the same as the ambient metric. So doesn't the current metric assume an isometric embedding? 

> Then I think I'll make a flag for this option, something like
> 
> ```julia
> abstract type AbstractTangentRepresentation end
> struct OrthogonalTangentRepresentation <: AbstractTangentRepresentation end
> struct LieAlgebraTangentRepresentation <: AbstractTangentRepresentation end
> 
> struct Rotations{N, TSR<:AbstractTangentRepresentation} <: Manifold end
> ```
> 
> with orthogonal representation being the default.

I think it's good we're discussing this, because we're providing both a base interface for manifolds and default manifold implementations. My understanding of our philosophy so far for manifolds is that we provide default implementations assuming one representation but taking any type for points and vectors, and users can change the representation by creating their own `MPoint`, `TVector`, and `CoTVector`. This is one approach, but another would be for manifolds to be created with flags like you have proposed here that enable the methods to do dispatch. The downside to this is one may want to call a method for `Rotations{N,OrthogonalTangentRepresentation}` that internally uses the functionality of `Rotations{N,LieAlgebraTangentRepresentation}`. As a design decision, which do we want to prefer?

mateuszbaran:
> Can you explain this a little more? I thought an isometric embedding was one where the pullback metric was the same as the ambient metric. So doesn't the current metric assume an isometric embedding?

My idea was to use a different representation for SO(n) than rotation matrices to have the pullback metric that results in nice angles for geodesic distance. After some thought, it would be even more confusing than these scaled distances.


> > Then I think I'll make a flag for this option, something like
> > ```julia
> > abstract type AbstractTangentRepresentation end
> > struct OrthogonalTangentRepresentation <: AbstractTangentRepresentation end
> > struct LieAlgebraTangentRepresentation <: AbstractTangentRepresentation end
> > 
> > struct Rotations{N, TSR<:AbstractTangentRepresentation} <: Manifold end
> > ```
> > 
> > 
> > with orthogonal representation being the default.
> 
> I think it's good we're discussing this, because we're providing both a base interface for manifolds and default manifold implementations. My understanding of our philosophy so far for manifolds is that we provide default implementations assuming one representation but taking any type for points and vectors, and users can change the representation by creating their own `MPoint`, `TVector`, and `CoTVector`. This is one approach, but another would be for manifolds to be created with flags like you have proposed here that enable the methods to do dispatch. The downside to this is one may want to call a method for `Rotations{N,OrthogonalTangentRepresentation}` that internally uses the functionality of `Rotations{N,LieAlgebraTangentRepresentation}`. As a design decision, which do we want to prefer?

That's a problem that somehow wasn't discussed yet, with the assumption that a single representation for each manifold would be good enough, at least inside `ManifoldMuseum`. It looks like it isn't. By the way, I thought `MPoint`, `TVector`, and `CoTVector` are primarily for decorating arrays (like in the case of `ArrayManifold`) or as a base type for non-array representations. A different representation would be implemented by making a separate type for the manifold. I'm not sure though.

What's the problem with a method for `Rotations{N,OrthogonalTangentRepresentation}` that internally uses the functionality of `Rotations{N,LieAlgebraTangentRepresentation}`? It should be definitely possible.

sdaxen:
> That's a problem that somehow wasn't discussed yet, with the assumption that a single representation for each manifold would be good enough, at least inside `ManifoldMuseum`. It looks like it isn't. By the way, I thought `MPoint`, `TVector`, and `CoTVector` are primarily for decorating arrays (like in the case of `ArrayManifold`) or as a base type for non-array representations. A different representation would be implemented by making a separate type for the manifold. I'm not sure though.

Hmm, you're right, it seems I've misunderstood what `MPoint` and such are for. But I do think they're general enough to use to wrap arrays for dispatch (e.g. for charts).

> What's the problem with a method for `Rotations{N,OrthogonalTangentRepresentation}` that internally uses the functionality of `Rotations{N,LieAlgebraTangentRepresentation}`? It should be definitely possible.

I have several issues with it. The first is that internally each call of `myfunc(M::Rotations{N,OrthogonalTangentRepresentation},...)` will need to create a new `Rotations{N,LieAlgebraTangentRepresentation}` manifold for dispatch. In this case, the manifold is a singleton, so there's no overhead, but what then do we do if we have non-singleton manifolds? (Or would the compiler optimize out any overhead?)

But for me the main issue is the conflation of the representation of a manifold and its (co)tangent spaces with the manifold itself by design. I'm just wondering what the downstream effects might be when we get to charts, etc. I'm not confident enough that this is a problem to oppose this approach though.

mateuszbaran:
> Hmm, you're right, it seems I've misunderstood what `MPoint` and such are for. But I do think they're general enough to use to wrap arrays for dispatch (e.g. for charts).

Yes, these types will also be useful for charts.

> I have several issues with it. The first is that internally each call of `myfunc(M::Rotations{N,OrthogonalTangentRepresentation},...)` will need to create a new `Rotations{N,LieAlgebraTangentRepresentation}` manifold for dispatch. In this case, the manifold is a singleton, so there's no overhead, but what then do we do if we have non-singleton manifolds? (Or would the compiler optimize out any overhead?)

In such cases we could just extract the part useful for both representations to a separate function.

> But for me the main issue is the conflation of the representation of a manifold and its (co)tangent spaces with the manifold itself by design. I'm just wondering what the downstream effects might be when we get to charts, etc. I'm not confident enough that this is a problem to oppose this approach though.

I'm worried about it too, it's yet another thing that will interact with other parts of the project and it's hard to predict these downstream effects. The easiest choice would be to merge this PR with one representation and come back to this issue at a later time, once we have support for multiple charts, custom metrics, Lie groups and more developed support for AD.

kellertuer:
For me the additional type seems a little overtyped, do you think that the tangent representation is really necessary for the basic manifold you are implementing here?

I mean we could try to approach the same model we did for the metric: There is an implicit default representation (well documented, maybe different for every manifold) and if you want to have more control over that representation you use a `TangentSpace` (without Manifold for the type) that acts as a decorator (lice `MetricManifold`)? This way you again to not conflate but can still specify that later.

mateuszbaran:
Yes, the decorator approach looks like a better idea. The only problem now is which tangent space representation should be the default one.

kellertuer:
Let's just agree on one.

* My favourite is the Lie algebra representation. It might be misleading for starters, since the elements you store are not the ones from the tangent space but the elements from the Lie algebra, but computationally that is the most effeícient way.
* One could store really the tangent vectors but then your manifold functions would be inefficient. This might help first time users, but it might annoy advanced once, because they always have to decorate to just get an efficient version

mateuszbaran:
I'm OK with this, documentation can just explain to first time users what representation is used and why.

sdaxen:
I'm also fine with this if well documented.

mateuszbaran:
I've added a distribution on the manifold of rotations and some documentation. The PR looks OK to me now. Is there something that should be done here before merging it?

src/DistributionsBase.jl
```diff
@@ -44,3 +44,13 @@ from manifold of type `TM`.
 """
 abstract type TVectorDistribution{TM<:Manifold, T} <: Distribution{TVectorvariate, TVectorSupport{TM, T}}
 end
+
+
+"""
+    get_support(d::TVectorDistribution)
+
+Get the object of type `TVectorSupport` for the distribution `d`.
+"""
```
kellertuer: Could. we maybe rename that to `domain`? `get_support`sounds a little bit like it starts a phone_call.

src/Rotations.jl
```diff
@@ -63,6 +63,11 @@ end
 
 injectivity_radius(M::Rotations, x) = π*sqrt(2.0)
 
+"""
+    PolarRetraction
+
+Retraction on the rotations manifold using the polar method.
+"""
```
kellertuer: In the long run, the formula would be nice, at least in the md in the docs, but I'd also prefer it here.

src/DistributionsBase.jl
mateuszbaran: OK, done

src/Rotations.jl
mateuszbaran: I've added some docs, similar to what manopt.jl has.

src/Rotations.jl
```diff
@@ -0,0 +1,318 @@
...
+retract!(M::Rotations, y, x, v) = retract!(M, y, x, v, QRRetraction())
+
+"""
+    PolarInverseRetraction
+
+Inverse retraction on the rotations manifold using the polar method.
+"""
+struct PolarInverseRetraction <: AbstractInverseRetractionMethod end
+
+injectivity_radius(::Rotations, x, ::PolarInverseRetraction) = π*sqrt(2.0)/2
```
sdaxen: I think we should update our documentation of `injectivity_radius` to clarify that we're not using the technical definition of "injectivity radius" (maximum radius of the geodesic ball over which the exponential map is diffeomorphic) and instead using "maximum radius over which the specified retraction is injective (has a left inverse)"

src/Rotations.jl
```diff
@@ -0,0 +1,318 @@
...
+
+"""
+    is_manifold_point(S,x; kwargs...)
+
+checks, whether `x` is a valid point on the [`Rotations`](@ref) `S`,
+i.e. is an array of size [`manifold_dimension`](@ref)`(S)` and represents a
+valid rotation.
+The tolerance for the last test can be set using the ´kwargs...`.
+"""
+function is_manifold_point(S::Rotations{N},x; kwargs...) where {N}
```
sdaxen: I would expect a function called `is_manifold_point` to return `true` if the point is on the manifold and `false` if not. Perhaps instead of throwing an error, this function should do that, and then a function named something like `check_manifold_point` would call this and throw an error if `false`.

docs/src/manifolds/rotations.md
```diff
@@ -0,0 +1,16 @@
...
+The manifold $\mathrm{SO}(n)$ of orthogonal matrices with determinant $+1$ in $\mathbb R^{n\times n}$, i.e.
+
+$\mathrm{SO}(n) = \bigl\{R \in \mathbb{R}^{n\times n} \big| RR^{\mathrm{T}} =
+R^{\mathrm{T}}R = \mathrm{I}_n, \operatorname{det}(R) = 1 \bigr\}$
+
+The $\mathrm{SO}(n)$ is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
+
+Tangent vectors are represented by elements of the corresponding Lie algebra, that is the tangent space at the identity element. This convention allows for more efficient operations on tangent vectors, although note that in such tangent vectors are not orthogonal to rotations the are tangent at in the ambient space of matrices. Tangent spaces at different points are different vector spaces.
```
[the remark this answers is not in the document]
sdaxen: Would be good to document how the usual definition of tangent space relates to the tangent space at the identity matrix.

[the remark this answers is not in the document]
mateuszbaran: I've removed that sentence, it's really quite confusing and a bit redundant now.

I've also added a paragraph that explains the relation of the tangent space at identity with other tangent spaces.

src/ManifoldMuseum.jl
```diff
@@ -102,29 +103,105 @@ Keyword arguments can be used to specify tolerances.
...
+
+Retraction method can be specified by the last argument. Please look at the
+documentation of respective manifolds for available methods.
 """
-retract!(M::Manifold, y, x, v) = exp!(M, y, x, v)
+retract!(M::Manifold, y, x, v, method::ExponentialRetraction) = exp!(M, y, x, v)
+
+retract!(M::Manifold, y, x, v) = retract!(M, y, x, v, ExponentialRetraction())
+
+retract!(M::Manifold, y, x, v, t::Number) = retract!(M, y, x, t*v)
```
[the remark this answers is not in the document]
mateuszbaran: Thanks, I didn't notice that Distributions has that function (it's not in the docs).

The problem with `t` being real is that `Dual` numbers (from e.g. ForwardDiff) derive from `Number` but not `Real`, so AD will fail here.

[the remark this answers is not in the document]
sdaxen: > The problem with `t` being real is that `Dual` numbers (from e.g. ForwardDiff) derive from `Number` but not `Real`, so AD will fail here.

`Dual` from `ForwardDiff` at least is derived from `Real`. https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/dual.jl#L14 (and [here](http://www.juliadiff.org/ForwardDiff.jl/stable/user/limitations.html) the documentation recommends `Real` as generic enough). Are there other packages this will cause a problem for?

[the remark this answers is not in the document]
mateuszbaran: Thanks, I must've mistaken it with something else. `Dual` from DualNumbers is not `Real` though: https://github.com/JuliaDiff/DualNumbers.jl/blob/677e2d52bfadc365f2b3f43b51c1dc66d284a9fa/src/dual.jl#L3 . There is likely a lot of other places where DualNumbers won't work for some manifolds so I'll make this `Real`.

docs/src/manifolds/rotations.md
```diff
@@ -0,0 +1,18 @@
+# Rotations
+
+The manifold $\mathrm{SO}(n)$ of orthogonal matrices with determinant $+1$ in $\mathbb R^{n\times n}$, i.e.
+
+$\mathrm{SO}(n) = \bigl\{R \in \mathbb{R}^{n\times n} \big| RR^{\mathrm{T}} =
+R^{\mathrm{T}}R = \mathrm{I}_n, \operatorname{det}(R) = 1 \bigr\}$
```
sdaxen: No need for `\ operatorname` here, since `\det` is defined

src/Rotations.jl
```diff
@@ -0,0 +1,318 @@
...
+returns the dimension of the manifold $\mathrm{SO}(n)$, i.e. $\frac{n(n-1)}{2}$.
+"""
+manifold_dimension(S::Rotations{N}) where {N} = div(N*(N-1), 2)
+
+@doc doc"""
+    inner(S, x, w, v)
+
+compute the inner product of the two tangent vectors `w, v` from the tangent
+plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
+restriction of the metric from the embedding, i.e. $ (v,w)_x = v^\mathrm{T}w $.
```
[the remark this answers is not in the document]
sdaxen: Actually, `\operatorname{tr}` is more fitting here.

docs/src/manifolds/rotations.md
```diff
@@ -0,0 +1,18 @@
...
+
+$\mathrm{SO}(n)$ is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
+
+Tangent vectors are represented by elements of the corresponding Lie algebra, which is the tangent space at the identity element. This convention allows for more efficient operations on tangent vectors. Tangent spaces at different points are different vector spaces.
+
+Let $L_R\colon \mathrm{SO}(n) \to \mathrm{SO}(n)$ where $R \in \mathrm{SO}(n)$ be the left-multiplication by $R$, that is $L_R(S) = RS$. The tangent space at rotation $R$, $T_R \mathrm{SO}(n)$, is related to the tangent space at the identity rotation $\mathrm{I}_n$ by the differential of $L_R$ at identity, $(\mathrm{d}L_R)_{\mathrm{I}_n} \colon T_{\mathrm{I}_n} \mathrm{SO}(n) \to T_R \mathrm{SO}(n)$. You can compare the functions [`log!(::ManifoldMuseum.Rotations, v, x, y)`](@ref) and [`exp!(::ManifoldMuseum.Rotations, y, x, v)`](@ref) to see how it works in practice.
```
sdaxen: I like the detail in the explanation, but it would be good to also explicitly say that  for `v \in T_I SO(n)`, `w=Rv \in T_R SO(n)`

src/Rotations.jl
```diff
@@ -0,0 +1,318 @@
...
+  \frac{1}{2} \bigl(\operatorname{Log}(x^{\mathrm{T}}y)
+  - (\operatorname{Log} x^{\mathrm{T}}y)^{\mathrm{T}}),
+```
+where $\operatorname{Log}$ denotes the matrix logarithm
+and save the result to `v`.
+"""
+function log!(S::Rotations, v::TV, x, y) where TV
+    U = transpose(x) * y
+    if (x != y) && (abs.(U) == one(U))
+        throw( ErrorException("The points $x and $y are antipodal, thus these input parameters are invalid.") )
```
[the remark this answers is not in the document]
mateuszbaran: OK, I've changed to option (2)

src/Rotations.jl
mateuszbaran: The function does the same thing for `Sphere`, I'll make a separate PR to change and discuss this.

src/Rotations.jl
mateuszbaran: I've edited the docstring for `injectivity_radius` that operates on inverse retraction objects (in `ManifoldMuseum.jl`)

mateuszbaran:
Thanks @sdaxen for the review :+1: . I think I've fixed everything except the issue with `is_manifold_point` but I'd prefer to do it in a separate PR.

src/Rotations.jl
```diff
@@ -25,7 +25,10 @@ manifold_dimension(S::Rotations{N}) where {N} = div(N*(N-1), 2)
 
 compute the inner product of the two tangent vectors `w, v` from the tangent
 plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
-restriction of the metric from the embedding, i.e. $ (v,w)_x = v^\mathrm{T}w $.
+restriction of the metric from the embedding, i.e.
+
+$ (v,w)_x = v^\mathrm{T}w = (v, w)_x = \mathrm{tr}(v^T w) $.
```
sdaxen: No need for the left half `(v,w)_x = v^\mathrm{T}w`

src/ManifoldMuseum.jl
```diff
@@ -278,8 +278,8 @@ injectivity_radius(M::Manifold, x) = 1.0
 """
     injectivity_radius(M::Manifold, x, R::AbstractInverseRetractionMethod)
 
-Distance such that `retract(M, x, y, R)` is defined for all points within
-this radius.
+Distance such that `inverse_retract(M, x, y, R)` is injective for all points
+within this radius (has a left inverse).
```
sdaxen: I think it's the other way around. For example, `exp(M, x, log(M, x, y))` always produces `y`, but `log(M, x, exp(M, x, v))` does not always equal `v`. So it's the radius in which `retract` is injective.

src/Rotations.jl
```diff
@@ -27,7 +27,7 @@ compute the inner product of the two tangent vectors `w, v` from the tangent
 plane at `x` on the special orthogonal space `S=`$\mathrm{SO}(n)$ using the
 restriction of the metric from the embedding, i.e.
 
-$(v,w)_x = v^\mathrm{T}w = (v, w)_x = \operatorname{tr}(v^T w)$.
+$v^\mathrm{T}w = (v, w)_x = \operatorname{tr}(v^T w)$.
```
sdaxen: Sorry to be pedantic, but I still think it's better without the left part,  so`$(v, w)_x = \operatorname{tr}(v^T w)$`

mateuszbaran: OK, I'll remove that part too.

src/ManifoldMuseum.jl
mateuszbaran: Makes sense, I'll change the docstring for the other method too.

sdaxen:
Nice! Okay, I think this PR is ready to merge now.

mateuszbaran:
Great! Do you want to merge yours first?

sdaxen:
Sure, I can do that.

PR #22, opened 2019-06-28
[WIP] Product manifold

mateuszbaran:
I've started working on the product manifold. Functionality is going quite nicely, remaining challenges:

1. Performance is far from ideal but I'm working on it.
2. I don't know how to do product metrics. @sdaxen , do you have any tips?
3. I haven't touched product distributions but this seems fairly easy.
4. There are likely some interactions with other manifolds that I haven't spotted yet. `zero_tangent_vector!` for sphere is an interesting example.


dkarrasch:
The canonical product metric is the block diagonal matrix with the individual metrics on the diagonal. I assume you will need to call `local_metric` on the individual metrics, evaluated at the corresponding subarrays of `x`. Or is the idea to actually keep the block structure in the manifold points/coordinates/tangent vectors etc.? That would make it perhaps even "easier".

kellertuer:
Quite a generic approach. How do you plan to be able to do a `Product(Manifold)(Euclidean(3),SymmetricPositiveDefinite(3))`? There, each element consists of a vector in R3 and a 3-by-3 matrix. I had your approach (roughly) in Matlab, and reshaped a lot to get things going and it never looked nice. Even less when you try to extract the SPD matrix for example.

mateuszbaran:
The main problem with the product metric is at the same time having a default metric structure for products of manifolds with such structure (e.g. `Euclidean`, `Sphere`) and being able to express products of custom metrics. This is the part I have a problem with.

I think renaming it to `ProductManifold` is a good idea.

> How do you plan to be able to do a `Product(Manifold)(Euclidean(3),SymmetricPositiveDefinite(3))`? There, each element consists of a vector in R3 and a 3-by-3 matrix. I had your approach (roughly) in Matlab, and reshaped a lot to get things going and it never looked nice. Even less when you try to extract the SPD matrix for example.

It's already implemented using reshapes of views (`uview_element`, `suview_element`). It's a tough problem though, it works, but too much is done at runtime at the moment. I'll work on it, it looks promising.

In FunManifolds, I use an `AbstractArray`-based wrapper of `Tuple`. https://github.com/mateuszbaran/FunManifolds.jl/blob/master/src/product_space.jl It's easier to avoid allocations and runtime operations there but immutability of tuples makes it problematic and it leads to a very fragmented memory which is bad for CPU caching.

src/Product.jl
```diff
@@ -0,0 +1,94 @@
...
+    return SizedAbstractArray{Tuple{S...}}(uview(x, range))
+end
+
+function suview_element(x::AbstractArray, range, shape::Size)
+    return reshape(uview(x, range), shape)
+end
+
+function det_local_metric(M::MetricManifold{Product{<:Manifold,TRanges,TSizes},ProductMetric}, x) where {TRanges, TSizes}
+    dets = map(ziptuples(M.manifolds, TRanges, TSizes)) do d
+        return det_local_metric(d[1], view_element(x, d[2], d[3]))
```
dkarrasch: If you `ziptuples` does what I suppose it's doing, you could take a recursive approach that I have seen in `Base` and elsewhere, also to guarantee type inference. Suppose `x` was a tuple of coordinate vectors. I guess you're accessing those "tuples" via `view_element`.
```julia
function det_local_metric(M::MetricManifold{Product{<:Manifold,TRanges,TSizes},ProductMetric}, x) where {TRanges, TSizes}
    xi = first(x)
    mani = first(M)
    return det_local_metric(Mi, xi) * det_local_metric(tail(M), tail(x))
end
```

src/Product.jl
```diff
@@ -0,0 +1,94 @@
...
+        push!(ranges, k:(k+len-1))
+        k += len
+    end
+    TRanges = tuple(ranges...)
+    TSizes = tuple(sizes...)
+    return Product{typeof(manifolds), TRanges, TSizes}(manifolds)
+end
+
+function representation_size(M::Product, ::Type{T}) where {T}
+    return (sum(map(m -> representation_size(m, T), M.manifolds)),)
```
dkarrasch: As below, this could be
```julia
function representation_size(M::Product, ::Type{T}) where {T}
    return representation_size(first(M), T) + representation_size(tail(M), T)
end
```

kellertuer:
> It's already implemented using reshapes of views (`uview_element`, `suview_element`). It's a tough problem though, it works, but too much is done at runtime at the moment. I'll work on it, it looks promising.
Oh, then I have not yet understood your implementation of such views (nor the abbreviation u and su), could you point to a documentation what that does? Otherwise I am looking forward to a documentation of that with an example :)

test/runtests.jl
```diff
@@ -259,4 +259,28 @@ end
...
+    types = [Vector{Float64},
+             SizedVector{5, Float64},
+             MVector{5, Float64},
+             Vector{Float32},
+             SizedVector{5, Float32},
+             MVector{5, Float32}]
+
+    for T in types
+        @testset "Type $T" begin
+            pts = [convert(T, [1.0, 0.0, 0.0, 0.0, 0.0]),
```
kellertuer: What I meant in my comment is that this construction of a point (just a 5-element array where the first 3 are on S2, the remaining 2 on Rn is quite complicated to read and hard to understand. I can understand that inernally that might be a good representation and that the views (I think I roughly understood them) are a good help, but would prefer a construction (at least additionally) of the type `([1.0, 0.0, 0.0], [0.0, 0.0])`, the elements on the product manifold are ordered anyways (hence the tuple idea).

src/ArrayManifold.jl
```diff
@@ -12,6 +12,10 @@ end
 convert(::Type{M},m::ArrayManifold{M}) where M <: Manifold = m.manifold
 convert(::Type{ArrayManifold{M}},m::M) where M <: Manifold = ArrayManifold(M)
 
+function representation_size(M::ArrayManifold, ::Type{T}) where {T}
```
sdaxen: Neither this nor the `manifold_dimension` dimension below should be necessary. Just implement the generic `IsDecoratorManifold` version in `Manifolds.jl`:

```julia
@traitfn function representation_size(M::MT, ::Type{T}) where {MT<:Manifold,T;!IsDecoratorManifold{MT}}
    error("representation_size not implemented for manifold $(typeof(M)) and type $(T).")
end

@traitfn function representation_size(M::MT, ::Type{T}) where {MT<:Manifold,T;IsDecoratorManifold{MT}}
    representation_size(base_manifold(M), T)
end
```

src/Euclidean.jl
```diff
@@ -19,6 +19,10 @@ struct Euclidean{T<:Tuple} <: Manifold where {T} end
 Euclidean(n::Int) = Euclidean{Tuple{n}}()
 Euclidean(m::Int, n::Int) = Euclidean{Tuple{m,n}}()
 
+function representation_size(::Euclidean{S}, ::Type{T}) where {S,T<:Union{MPoint, TVector, CoTVector}}
```
sdaxen: `T` should also permit `Array`s here.

src/Manifolds.jl
```diff
@@ -115,6 +116,16 @@ of each point of the manifold is homeomorphic.
 """
 function manifold_dimension end
 
+@doc doc"""
```
sdaxen: Replace with generic `IsDecoratorManifold` version. See above.

src/Metric.jl
```diff
@@ -40,6 +40,11 @@ struct MetricManifold{M<:Manifold,G<:Metric} <: Manifold
     metric::G
 end
 
+function representation_size(M::MetricManifold, ::Type{T}) where {T}
```
sdaxen: Not needed if implemented for `IsDecoratorManifold`

src/Product.jl
sdaxen: That would work if `M.manifolds` only contains two manifolds, but it could contain an arbitrary number.

sdaxen: Could do a `mapreduce`.

src/Product.jl
```diff
@@ -0,0 +1,94 @@
...
+
+generates the product manifold $M_1 \times M_2 \times \dots \times M_n$
+"""
+struct Product{TM<:Tuple, TRanges, TSizes} <: Manifold
+    manifolds::TM
+end
+
+function Product(manifolds...)
+    sizes = map(m -> representation_size(m, MPoint), manifolds)
+    lengths = map(s -> prod(s), sizes)
```
sdaxen: Could do `map(prod, sizes)`

src/Rotations.jl
```diff
@@ -13,6 +13,11 @@ generates the $\mathrm{SO}(n) \subset \mathbb R^{n\times n}$
 struct Rotations{N} <: Manifold end
 Rotations(n::Int) = Rotations{n}()
 
+function representation_size(::Rotations{N}, ::Type{T}) where {N,T<:Union{MPoint, TVector, CoTVector}}
```
sdaxen: Should return a `Size` for consistency.

src/Product.jl
kellertuer: > That would work if `M.manifolds` only contains two manifolds, but it could contain an arbitrary number.

Actually the second part recursively upwraps it :) Notice that `tail` in the second summand works on all remaining one (again doing the first plus the remaining ones and so on)... nice recursion!

test/runtests.jl
sdaxen: I favor this approach too. It also enables us to mix and match representations without hassle and doesn't force us to flatten matrices and such. I also think it's clearer in some cases. For example, `SE(n)`'s underlying topological manifold is `SO(n) \times R(n)`. Its elements _can_ be represented as `(n+1) \times (n+1)` matrices which act on elements of `R(n+1)`, but since `SE(n)` acts on elements of `R(n)`, it's cleaner and more efficient to represent its elements as a 2-tuple containing an element of `SO(n)` and one of `R(n)`.

sdaxen:
I was thinking a product of two `MetricManifolds` should become a `MetricManifold`. So `MetricManifold{MA,GA}() × MetricManifold{MB,GB}()` should become `MetricManifold{ProductManifold{MA,MB},ProductMetric{GA,GB}}()`. Whereas if any of the manifolds being multiplied are not a `MetricManifold`, we just get a standard `ProductManifold`. The idea is that if someone spent the time to equip their manifolds with a metric _before_ building a product manifold, they intend to use the standard (block-diagonal) product metric, which they can always override by re-equipping the product manifold. This would enable the machinery that depends on `local_metric`. A standard `ProductManifold` with non-`MetricManifold` members would still have access to `exp`, `log`, etc, just like manifolds with `HasMetric`, but to get the metric, you need to either equip the submanifolds with a metric or directly equip the `ProductManifold`.

test/runtests.jl
sdaxen: This could actually be implemented with a custom `MPoint` called `ProductMPoint`, which stores a list of points and enables dispatch on their types.

sdaxen:
To get the above behavior, we should implement `Base.prod` or `cartesian_product` for manifolds to create product manifolds, and this could be customized based on the types of the inputs. I also think we should have `LinearAlgebra.cross` for `Manifold` inputs forward to `Base.prod` so that we can use the × operator (which aliases to `LinearAlgebra.cross`) for manifold products. Unless there's another way to get that symbol.

test/runtests.jl
kellertuer: That would be one way, which might not be efficient (though from modelling side I like it). I like how easy @mateuszbaran can compute logs and so on with his internal single-dimension array representation. If a constructor like I proposed and accessorscan be realized (something like `p[1]`and `p[2]` accessing first and second manifold element of the Product manifold point `p`(and that might require a type to dispatch the array access on, maybe). But I would favour that just for ease of use. I am working a lot with product (and even more with power manifolds, even $ \mathcal M^{128\times 128} $ or something, so I really would like to see easy access to single manifolds values.

sdaxen: Yeah, I don't know how efficient this approach is. Still working through @mateuszbaran's array-based approach.

src/Product.jl
sdaxen: Oh! Cool, nice approach.

src/Product.jl
mateuszbaran: This won't work right now, `x` is not a tuple. `view_element` is supposed to take a statically-sized reshaped view of a part of the abstract vector `x`.

src/Product.jl
mateuszbaran: Result of `tail` would have to be re-wrapped into a new object of type `Product` which doesn't look ideal. `mapreduce` makes sense.

mateuszbaran: Or we would at least need a new function that operates on tuples of manifolds.

mateuszbaran:
> > It's already implemented using reshapes of views (`uview_element`, `suview_element`). It's a tough problem though, it works, but too much is done at runtime at the moment. I'll work on it, it looks promising.
> 
> Oh, then I have not yet understood your implementation of such views (nor the abbreviation u and su), could you point to a documentation what that does? Otherwise I am looking forward to a documentation of that with an example :)

I'll document it soon, it's still a WIP and things might change. Especially these two functions. They are supposed to take a statically-sized reshaped views of a part of an array representation of a point, tangent vector etc.

test/runtests.jl
mateuszbaran: I'll definitely add a more user-friendly method of creating points on product manifolds once I'm satisfied with the backend.

Tuples are not that great because they are:

1) Not `AbstractArray`s, which limits compatibility of the representation with some parts of the Julia ecosystem.
2) Immutable.

I'd love to have a mutable heterogeneous `AbstractArray` but it doesn't exist right now. I'm giving it a serious thought though.

I'll definitely add some convenient accessors.

src/ArrayManifold.jl
mateuszbaran: Thanks, good idea.

src/Euclidean.jl
mateuszbaran: Not really, `T` is supposed to refer what *kind* of object is represented (point on the manifold, some element of a vector space like tangent vector, cotangent vector, Riemann tensor etc.). I don't really like this solution, it's not clear what it is, but I thought some functionality of this sort is necessary and this was my first idea.

src/Product.jl
mateuszbaran: Why wouldn't this implementation of `representation_size` work when a product of more than two manifolds is represented?

Thanks for the tips about `mapreduce` and `prod`. :+1:

src/Rotations.jl
mateuszbaran: I think returning a tuple is quite consistent with what `size` returns.

test/runtests.jl
dkarrasch: I think you can mutate vectors in a tuple of Vectors/Arrays. Certainly, I'm missing context, but this works.
```julia
julia> x = (rand(2), rand(3,3))
([0.178285, 0.857401], [0.525022 0.700751 0.294022; 0.788598 0.253698 0.641965; 0.42019 0.56932 0.875257])

julia> x[1] .= 1
2-element Array{Float64,1}:
 1.0
 1.0

julia> x
([1.0, 1.0], [0.525022 0.700751 0.294022; 0.788598 0.253698 0.641965; 0.42019 0.56932 0.875257])
```

mateuszbaran: Hmm... it didn't work in FunManifolds because I wanted to be able to represent points as `SArray`s but since it's not supported here it might work. I'll try that and see what happens. Things like broadcasting, `representation_size` and  `ambient_space` will be challenging though.

src/Rotations.jl
sdaxen: That's fine, then it should be consistent everywhere (above and below returns `Size`)

mateuszbaran:
I've changed the internal representation of product manifolds quite a lot. Finally, it's a mix of a few approached discussed here: points, vectors etc. are represented by objects of type `ProductView` which are like `AbstractArray`s but have preallocated reshaped views to specific parts representing projections to manifolds from which the product manifold is composed. This way we don't lose the continuous representation, we can refer to parts relatively nice, performance is quite satisfactory (you can check with the brand new benchmark suite), broadcasting is supported and I'm quite happy with it. I still have to provide a nicer way to construct points or vectors on product manifolds but it should be easy.

What do you think about the new approach?

sdaxen:
I think the new approach is really cool (thanks for setting up the benchmark suite, by the way. Awesome!), and I'm kind of shocked that it performs so well. I'm still trying to understand how it works. 

How would it handle a mixture of vector and matrix manifolds? I'm not certain what to pass to `ProductView` to test this. Perhaps there should also be a convenience function for constructing a `ProductView` from a list of the desired representations for the corresponding manifolds. (e.g. for `M = ProductManifold(Rotations(3), Euclidean(3))`, you'd create a `ProductView` with `P = ProductView(M, [rot_mat, trans_vec])`, which would internally flatten it or whatever for `P.data` but create the necessary views for `P.views`. Maybe the elements of such a list could also be `MPoint`s, so that the user could then specify once what the representation is, and then `P.views` contains that structure wrapping the view, which I guess only works for points that wrap an `AbstractArray`.

Currently, `ProductView` seems to only work when the `ProductManifold` has 2 constituent manifolds. Example:

```julia
julia> R2 = Manifolds.Euclidean(2);

julia> M = Manifolds.ProductManifold(R2, R2);

julia> x = randn(4)
4-element Array{Float64,1}:
 0.7683675344912058  
 0.045762904704441366
 1.281251818793342   
 1.148563618371174   

julia> P = Manifolds.ProductView(M, x)
4-element Manifolds.ProductView{Manifolds.ProductManifold{Tuple{Euclidean{Tuple{2}},Euclidean{Tuple{2}}},(1:2, 3:4),Tuple{Tuple{2},Tuple{2}}},Float64,1,Array{Float64,1},Tuple{Manifolds.SizedAbstractArray{Tuple{2},Float64,1,1,SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}},Manifolds.SizedAbstractArray{Tuple{2},Float64,1,1,SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}}}}:
 0.7683675344912058  
 0.045762904704441366
 1.281251818793342   
 1.148563618371174   

julia> M = Manifolds.ProductManifold(R2, R2, R2);

julia> x = randn(6)
6-element Array{Float64,1}:
  0.6630836648853156 
  0.2331471356770406 
  0.2964616068062833 
 -0.14207468080177554
 -1.5358504686848118 
  0.3998359958877595 

julia> P = Manifolds.ProductView(M, x)
ERROR: MethodError: no method matching ziptuples(::Tuple{UnitRange{Int64},UnitRange{Int64},UnitRange{Int64}}, ::Type{Tuple{Tuple{2},Tuple{2},Tuple{2}}})
Closest candidates are:
  ziptuples(::Tuple{Vararg{Any,N}}, ::Tuple{Vararg{Any,M}}) where {N, M} at /Users/saxen/projects/Manifolds/src/utils.jl:9
  ziptuples(::Tuple{Vararg{Any,N}}, ::Tuple{Vararg{Any,M}}, ::Tuple{Vararg{Any,L}}) where {N, M, L} at /Users/saxen/projects/Manifolds/src/utils.jl:23
  ziptuples(::Tuple{Vararg{Any,N}}, ::Tuple{Vararg{Any,M}}, ::Tuple{Vararg{Any,L}}, ::Tuple{Vararg{Any,K}}) where {N, M, L, K} at /Users/saxen/projects/Manifolds/src/utils.jl:37
Stacktrace:
 [1] Manifolds.ProductView(::Manifolds.ProductManifold{Tuple{Euclidean{Tuple{2}},Euclidean{Tuple{2}},Euclidean{Tuple{2}}},(1:2, 3:4, 5:6),Tuple{Tuple{2},Tuple{2},Tuple{2}}}, ::Array{Float64,1}) at /Users/saxen/projects/Manifolds/src/ProductManifold.jl:50
 [2] top-level scope at none:0
```

mateuszbaran:
I'm glad you like the new approach :smiley:. I don't really know why it's so fast either, I was copying pieces of Julia's `Base` and `StaticArrays` until it was fast. Julia is particularly well suited for this kind of things.

I've added a product of three manifolds including a matrix one (SO(2)) to tests.

I've added `prod_point` and `proj_product_array` to easily construct and project points on a product manifold. I don't like these names but I didn't find better ones. I'll try converting `prod_point` to a new constructor for `ProductArray` (I think this names fits better than `ProductView`). `ProductArray` should also work for tangent vectors. I've also fixed products of more than two manifolds.

I didn't really think yet about how `ProductManifold` would support arbitrary `MPoint`s. Currently I assume that whatever is represented is an `AbstractArray` with some methods implemented. I'm sure I'll need a different type for points as flattening and un-flattening won't really be an option. I'll work on that later.

mateuszbaran:
I've restructured things a bit to allow products of arbitrary manifolds (not necessarily array-based). It has surprisingly good performance, generally on par with `ProductArray`. `similar` for `SizedArray` from `StaticArrays` is surprisingly slow though and that impacts some benchmarks.

mateuszbaran:
Product manifold is almost ready I think, I've added product distributions, some minor things and more tests. I think I'll leave product metrics for now. My next priorities are power manifold, tangent bundle, adding docs, setting up travis, appveyor and coveralls and finally porting FunManifolds.jl to this interface.

Do you have any suggestions?

kellertuer:
I checked your rework shortly and that looks quite nice; but since I am (again!) at a conference, I will only find time on Friday to look at that more precisely. I hope to then also start the manifold of symmetric positive definite matrices with affine invariant metric as default and the log euclidean as a first test how to dispatch on metrics (as well as a concise documentation including literature for both).

sdaxen:
I should have time to look at this closely between today and tomorrow. I'm currently working on Lie groups and their metrics with `Sp(1)` (derived from `Sphere(3)`) and `SO(n)` as test cases, then product groups (which will use `ProductManifold` somehow) and semidirect products. Then I'll tackle `SE(n)`.

src/Sphere.jl
```diff
@@ -62,11 +66,13 @@ end
 
 injectivity_radius(S::Sphere, args...) = π
 
-zero_tangent_vector(S::Sphere, x) = zero(x)
-zero_tangent_vector!(S::Sphere, v, x) = (v .= zero(x))
+function zero_tangent_vector!(S::Sphere, v, x)
```
sdaxen: I think we need one of these for `Euclidean`.

src/utils.jl
```diff
@@ -0,0 +1,160 @@
...
+
+@inline convert(::Type{AbstractArray}, sa::SizedAbstractArray) = sa.data
+@inline convert(::Type{AbstractArray{T}}, sa::SizedAbstractArray{S,T}) where {T,S} = sa.data
+@inline convert(::Type{AbstractArray{T,N}}, sa::SizedAbstractArray{S,T,N}) where {T,S,N} = sa.data
+
+Base.@propagate_inbounds getindex(a::SizedAbstractArray, i::Int) = getindex(a.data, i)
+Base.@propagate_inbounds setindex!(a::SizedAbstractArray, v, i::Int) = setindex!(a.data, v, i)
+
+SizedAbstractVector{S,T,M} = SizedAbstractArray{Tuple{S},T,1,M}
+@inline SizedAbstractVector{S}(a::AbstractArray{T,M}) where {S,T,M} = SizedAbstractArray{Tuple{S},T,1,M}(a)
```
sdaxen: This doesn't seem to work for me
```julia
julia> x = randn(3)
3-element Array{Float64,1}:
  0.10460674079990906
  0.02796008883007997
 -0.2567903778403676 

julia> x2 = StaticVector{3}(x)
ERROR: DimensionMismatch("No precise constructor for StaticArray{Tuple{3},T,1} where T found. Length of input was 3.")
Stacktrace:
 [1] StaticArray{Tuple{3},T,1} where T(::Tuple{Tuple{Tuple{Tuple{Float64,Float64,Float64}}}}) at /Users/saxen/projects/StaticArrays.jl/src/convert.jl:1
 [2] Type at /Users/saxen/projects/StaticArrays.jl/src/convert.jl:4 [inlined] (repeats 3 times)
 [3] _convert at /Users/saxen/projects/StaticArrays.jl/src/convert.jl:29 [inlined]
 [4] convert at /Users/saxen/projects/StaticArrays.jl/src/convert.jl:26 [inlined]
 [5] StaticArray{Tuple{3},T,1} where T(::Array{Float64,1}) at /Users/saxen/projects/StaticArrays.jl/src/convert.jl:7

```

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+    return ProductManifold(M1.manifolds..., M2.manifolds...)
+end
+
+"""
+    ShapeSpecification(manifolds::Manifold...)
+
+A structure for specifying array size and offset information for linear
+storage of points and tangent vectors on the product manifold of `manifolds`.
+"""
+struct ShapeSpecification{TRanges, TSizes} end
```
sdaxen: Would be nice to have documentation for `TRanges` and `TSizes`.

src/Euclidean.jl
```diff
@@ -19,6 +19,14 @@ struct Euclidean{T<:Tuple} <: Manifold where {T} end
 Euclidean(n::Int) = Euclidean{Tuple{n}}()
 Euclidean(m::Int, n::Int) = Euclidean{Tuple{m,n}}()
 
+function representation_size(::Euclidean{Tuple{n}}, ::Type{T}) where {n,T<:Union{MPoint, TVector, CoTVector}}
```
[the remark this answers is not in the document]
sdaxen: Oh, I think I get it now. If the array represents a point, you would use `representation_size (M, MPoint)` here to get the size. It would be nice to have that documented somewhere.

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+             SizedAbstractArray{Size3}(view(data, TRanges[3])))
+    return ProductArray{M, T, N, TData, typeof(views)}(data, views)
+end
+
+function ProductArray(M::ShapeSpecification{TRanges, TSizes}, data::TData) where {TM, TRanges, TSizes, T, N, TData<:AbstractArray{T,N}}
+    return ProductArray(typeof(M), data)
+end
+
+"""
+    prod_point(M::ProductManifold, pts...)
```
sdaxen: This docstring doesn't agree with the signature. It's not clear how to use this function.

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+
+function isapprox(M::ProductManifold, x, v, w; kwargs...)
+    return all(t -> isapprox(t...; kwargs...), ziptuples(M.manifolds, x.parts, v.parts, w.parts))
+end
+
+function representation_size(M::ProductManifold, ::Type{T}) where {T}
+    return (mapreduce(m -> sum(representation_size(m, T)), +, M.manifolds),)
+end
+
+manifold_dimension(M::ProductManifold) = sum(map(m -> manifold_dimension(m), M.manifolds))
```
sdaxen: Also can be a `mapreduce`

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+
+struct ProductMetric <: Metric end
+
+function det_local_metric(M::MetricManifold{ProductManifold, ProductMetric}, x::ProductArray)
```
sdaxen: Can use `mapreduce`

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+    dets = map(det_local_metric, M.manifolds, x.parts)
+    return prod(dets)
+end
+
+function inner(M::ProductManifold, x, v, w)
```
sdaxen: Can use `mapreduce`

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+    subproducts = map(inner, M.manifolds, x.parts, v.parts, w.parts)
+    return sum(subproducts)
+end
+
+function exp!(M::ProductManifold, y, x, v)
+    map(exp!, M.manifolds, y.parts, x.parts, v.parts)
+    return y
+end
+
+function exp(M::ProductManifold, x::ProductMPoint, v::ProductTVector)
```
sdaxen: I wonder how the efficiency of this approach compares to creating an empty `ProductMPoint` and then filling it using `exp!`, since (I think) the approach here creates a bunch of mutable points using `similar` and then copies those points to a nw array within the `ProductMPoint`.

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+    map(log!, M.manifolds, v.parts, x.parts, y.parts)
+    return v
+end
+
+function log(M::ProductManifold, x::ProductMPoint, y::ProductMPoint)
+    return ProductTVector(map(log, M.manifolds, x.parts, y.parts)...)
+end
+
+function distance(M::ProductManifold, x, y)
+    return sqrt(sum(map(distance, M.manifolds, x.parts, y.parts).^2))
```
sdaxen: Can use `mapreduce`. Between `map`, `sum`, and `.^2`, it looks like the manifolds are traversed 3 times, when it could be done in one.

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+function det_local_metric(M::MetricManifold{ProductManifold, ProductMetric}, x::ProductArray)
+    dets = map(det_local_metric, M.manifolds, x.parts)
+    return prod(dets)
+end
+
+function inner(M::ProductManifold, x, v, w)
+    subproducts = map(inner, M.manifolds, x.parts, v.parts, w.parts)
+    return sum(subproducts)
+end
+
```
sdaxen: Should probably also have an implementation of `norm`

src/ProductManifold.jl
sdaxen: Also, this should probably be further up with `inner`, since that's where we usually put it.

src/ProductManifold.jl
```diff
@@ -0,0 +1,416 @@
...
+Product retraction of `retractions`. Works on [`ProductManifold`](@ref).
+"""
+struct ProductRetraction{TR<:Tuple} <: AbstractRetractionMethod
+    retractions::TR
+end
+
+ProductRetraction(retractions::AbstractRetractionMethod...) = ProductRetraction{typeof(retractions)}(retractions)
+
+function retract!(M::ProductManifold, y, x, v, method::ProductRetraction)
```
sdaxen: Would be good to have a version that takes a usual `AbstractRetractionMethod` and applies it to each submanifold. I'm thinking this would be convenient for e.g. exponential retractions.

docs/src/manifolds/product.md
```diff
@@ -0,0 +1,9 @@
+# Product Manifold
+
+Product manifold $M = M_1 \times M_2 \times \dots M_n$ of manifolds $M_1, M_2, \dots, M_n$. Points on the product manifold can be constructed using [`Manifolds.prod_point`](@ref) with canonical projections $\pi_i \colon M \to M_i$ for $i \in 1, 2, \dots, n$ provided by [`Manifolds.proj_product`](@ref).
```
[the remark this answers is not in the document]
mateuszbaran: OK, I'll change that.

src/Euclidean.jl
```diff
@@ -37,12 +45,45 @@ det_local_metric(M::MetricManifold{<:Manifold,EuclideanMetric}, x) = one(eltype(
...
 
+distance(::Euclidean, x, y) = norm(x-y)
 norm(::Euclidean, x, v) = norm(v)
 norm(::MetricManifold{<:Manifold,EuclideanMetric}, x, v) = norm(v)
 
 exp!(M::Euclidean, y, x, v) = (y .= x + v)
 
 log!(M::Euclidean, v, x, y) = (v .= y - x)
+
+proj!(M::Euclidean, x) = x
```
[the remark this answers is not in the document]
mateuszbaran: Well, I think it wasn't exported because we weren't sure it's the right name for it. `project_tangent` does a similar thing and has a longer name, so perhaps `project_point` would be a better name?

src/Rotations.jl
```diff
@@ -33,13 +38,33 @@ Tangent vectors are represented by matrices.
...
 
-project_tangent!(S::Rotations, w, x, v) = w .= (transpose(v).-v)./2
+project_tangent!(S::Rotations, w, x, v) = w .= (v .- transpose(v))./2
 
 function exp!(S::Rotations, y, x, v)
     y .= x * exp(v)
     return y
 end
 
+function exp!(S::Rotations{2}, y, x, v)
```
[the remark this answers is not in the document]
mateuszbaran: Thanks, that's better :+1: . I only had to switch the minus signs in `y[2]` and `y[3]` to make the tests pass.

src/Rotations.jl
```diff
@@ -33,13 +38,33 @@ Tangent vectors are represented by matrices.
...
+    α = norm(S, x, v)/sqrt(2)
+    expv = [cos(α) sin(α); -sin(α) cos(α)]
+    y .= x * expv
+    return y
+end
+
+function exp!(S::Rotations{3}, y, x, v)
```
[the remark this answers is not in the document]
mateuszbaran: Thanks, I used your code instead.

src/Rotations.jl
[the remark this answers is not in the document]
dkarrasch: `y .= x * y` reads strange, it allocates the result on the rhs. This should be perhaps `copyto!(y, x*y)`, or, even better, simply write out the matrix product.
```julia
function exp!(S::Rotations{2}, y, x, v)
   α = norm(S, x, v) / sqrt(2)
   @assert size(y) == (2, 2)
   sinα, cosα = sincos(α)
   y[1] = x[1] * cosα + x[3] * sinα
   ... etc
   return y
end
```

src/Rotations.jl
```diff
@@ -59,11 +84,29 @@ point at `y`.
 function log!(S::Rotations, v::TV, x, y) where TV
```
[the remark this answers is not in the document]
mateuszbaran: I've just forgotten about removing it.

src/Rotations.jl
```diff
@@ -59,11 +84,29 @@ point at `y`.
...
     U = transpose(x) * y
     # MMatrix doesn't have `log` defined
-    U1 = TV(real(log(Array(U))))
+    U1 = real(log(Array(U)))
```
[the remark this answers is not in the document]
mateuszbaran: `log` is pretty expensive to calculate anyway, so removing a couple of allocations won't make any noticeable difference. I think switching to `_log` can be a part of #25 since it's already there.

src/Euclidean.jl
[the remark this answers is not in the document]
sdaxen: That's a good point. I like `project_point!` for consistency as well.

src/Rotations.jl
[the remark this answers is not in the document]
sdaxen: As a bonus, @dkarrasch's approach brings the runtime for non-sized matrices to equivalent to sized.

src/utils.jl
mateuszbaran: `StaticVector{3}` is an abstract type from `StaticArrays` and I don't think it's supposed to work. Did you mean `x2 = Manifolds.SizedAbstractVector{3}(x)`? I didn't work but it should, so I've fixed that.

src/ProductManifold.jl
mateuszbaran: Right, I've prepared documentation for it.

src/utils.jl
sdaxen: Yes, sorry, that's what I meant.

src/ProductManifold.jl
mateuszbaran: Right, I've improved the documentation.

src/ProductManifold.jl
mateuszbaran: Not really, `mapreduce` doesn't have a variant that accepts multiple collections.

src/ProductManifold.jl
mateuszbaran: Same as above.

src/ProductManifold.jl
mateuszbaran: The points are not copied to make `ProductMPoint`, there are just new references to the points allocated by `exp` on specific manifolds.

src/ProductManifold.jl
mateuszbaran: I moved it up but again, `mapreduce` doesn't iterate over multiple collections.

dkarrasch: Couldn't you `zip` them?
```julia
sqrt(mapreduce(u -> distance(u...)^2, +, zip(M.manifolds, x.parts, y.parts)))
```
Not sure it's worth all the effort, though...

src/ProductManifold.jl
mateuszbaran: If you need exponential retractions for each manifold then you can just use `exp`, right? I think it's hard to imagine a good use case for this.

kellertuer: It's quite easy: Any exponential map is a retraction. But retractions are more general (they only approximate exp). They can, however, be much more easy to be computed (i.e. (`x+y/||x+y||` is a retraction on the sphere). For me they are (btw) just called retractions (not neccessarily exponential).
However, I ghink the `ProductRetracion` should be fine, since one can build such a product retraction from any set (vector) of single retraction ones?

src/ProductManifold.jl
sdaxen: It seems you're right. Looks like it's coming in Julia 1.2: https://github.com/JuliaLang/julia/pull/31532.

We can still us `mapreduce` with `zip` for some speedups.

```julia
julia> manifolds = [MetricManifold(Euclidean(3), EuclideanMetric()) for i=1:10];

julia> xlist = [randn(3) for i=1:10];

julia> f(manifolds, xlist) = prod(map(det_local_metric, manifolds, xlist));

julia> g(manifolds, xlist) = mapreduce(x->det_local_metric(x...), *, zip(manifolds, xlist));

julia> @benchmark f(manifolds, xlist)
BenchmarkTools.Trial: 
  memory estimate:  240 bytes
  allocs estimate:  5
  --------------
  minimum time:     93.016 ns (0.00% GC)
  median time:      100.420 ns (0.00% GC)
  mean time:        135.009 ns (19.94% GC)
  maximum time:     4.056 μs (95.89% GC)
  --------------
  samples:          10000
  evals/sample:     951

julia> @benchmark g(manifolds, xlist)
BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  3
  --------------
  minimum time:     41.319 ns (0.00% GC)
  median time:      43.431 ns (0.00% GC)
  mean time:        56.510 ns (17.19% GC)
  maximum time:     6.070 μs (98.74% GC)
  --------------
  samples:          10000
```

src/Rotations.jl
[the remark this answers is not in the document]
mateuszbaran: OK, done.

src/ProductManifold.jl
mateuszbaran: Your benchmark is flawed, `manifolds` and `xlist` should be tuples. When that's the case, maps over multiple collections are optimized much better:
```julia
julia> using Manifolds

julia> using BenchmarkTools

julia> manifolds = ntuple( t-> MetricManifold(Euclidean(3), EuclideanMetric()), Val(10));

julia> xlist = ntuple(t -> randn(3), Val(10));

julia> f(manifolds, xlist) = prod(map(det_local_metric, manifolds, xlist));

julia> g(manifolds, xlist) = mapreduce(x->det_local_metric(x...), *, zip(manifolds, xlist));

julia> @benchmark f(manifolds, xlist)
BenchmarkTools.Trial: 
  memory estimate:  16 bytes
  allocs estimate:  1
  --------------
  minimum time:     22.136 ns (0.00% GC)
  median time:      22.786 ns (0.00% GC)
  mean time:        35.699 ns (28.56% GC)
  maximum time:     80.109 μs (99.97% GC)
  --------------
  samples:          10000
  evals/sample:     996

julia> @benchmark g(manifolds, xlist)
BenchmarkTools.Trial: 
  memory estimate:  48 bytes
  allocs estimate:  3
  --------------
  minimum time:     34.425 ns (0.00% GC)
  median time:      35.899 ns (0.00% GC)
  mean time:        50.878 ns (23.76% GC)
  maximum time:     86.106 μs (99.94% GC)
  --------------
  samples:          10000
  evals/sample:     993

```

src/ProductManifold.jl
mateuszbaran: I could but see here: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/22#discussion_r300322315

src/ProductManifold.jl
sdaxen: Wow, that's quite the speed-up. Cool!

src/ProductManifold.jl
mateuszbaran: Yes, `ProductRetraction` is about specifying retractions for each manifold, then they used respectively.

src/ProductManifold.jl
mateuszbaran: For the case of tuples, Julia can even totally constant-fold `f` (but not `g`):

```julia
julia> @code_native f(manifolds, xlist)
	.text
; ┌ @ REPL[13]:1 within `f'
	movabsq	$139763219890848, %rax  # imm = 0x7F1D291302A0
	vmovsd	(%rax), %xmm0           # xmm0 = mem[0],zero
	retq
	nop
; └
```

src/ProductManifold.jl
```diff
@@ -0,0 +1,458 @@
...
+end
+
+"""
+    proj_product(x::ProductArray, i::Integer)
+
+Project the product array `x` to its `i`th component. A new array is returned.
+"""
+function proj_product(x::ProductArray, i::Integer)
+    return copy(x.parts[i])
+end
```
kellertuer: I find the function name a little strange – but I think you discussed that already? For me its not a project but merely an access (but with copy). Btw. Why do you copy? Without that one could just overload `[I]` access for `ProductArray`?

src/ProductManifold.jl
```diff
@@ -0,0 +1,458 @@
...
+similar(x::ProductTVector, ::Type{T}) where T = ProductTVector(map(t -> similar(t, T), x.parts)...)
+
+(+)(v1::ProductTVector, v2::ProductTVector) = ProductTVector(map(+, v1.parts, v2.parts)...)
+(-)(v1::ProductTVector, v2::ProductTVector) = ProductTVector(map(-, v1.parts, v2.parts)...)
+(-)(v::ProductTVector) = ProductTVector(map(-, v.parts))
+(*)(a::Number, v::ProductTVector) = ProductTVector(map(t -> a*t, v.parts))
+
+function proj_product(x::ProductTVector, i::Integer)
+    return x.parts[i]
+end
```
kellertuer: similar as above – and why do you then _not_ copy here?

test/utils.jl
```diff
@@ -0,0 +1,178 @@
...
+
+Tests general properties of manifold `m`, given at least three different points
+that lie on it (contained in `pts`).
+
+# Arguments
+- `test_forward_diff = true`: if true, automatic differentiation using ForwardDiff is
+tested.
+- `retraction_methods = []`: retraction methods that will be tested.
+"""
+function test_manifold(M::Manifold, pts::AbstractVector;
```
kellertuer: It's great to have this in such generality, I think `manifold_dimension` and ìnjectivity_radius` are missing, but might also be done one by one per manifold. Similarly `is_manifold_point`(also testing the errors that might occur) is missing and `is_tangent_vector`.

kellertuer: and great that we have codecov running 👍

src/ProductManifold.jl
mateuszbaran: I think projections were discussed in the context of embedded manifolds. I've seen such functions from the Cartesian product to one of its factors also called projections but we can use a different name here. Unfortunately, `getindex`/`setindex!` (or `[I]`) is not an option, they are already defined in a different way, as required by broadcasting. I could probably write much more complex broadcasting rules that don't require `getindex` and `setindex!` defined this way but I don't think it's worth the trouble.

I'll remove that `copy`, I was experimenting with some things and didn't remove it.

src/ProductManifold.jl
mateuszbaran: Let's discuss it above and I'll change both methods to make them consistent.

src/ProductManifold.jl
kellertuer: Ah I did not see that difference and was just thinking about ease of use, since (just as a user view again) having a point `p` on a `Productmanifold` a just natural thing to access the `i`th entry would be `p[I]`; but yes I am also not yet so familiar with broadcast rules.

test/utils.jl
mateuszbaran: I'd like to have the whole interface for manifolds tested here and so I expand it slowly. `manifold_dimension` and `injectivity_radius` are there, `is_manifold_point` and `is_tangent_vector` are partially tested (only the positive case). I think the negative case (when an error is thrown) should be tested separately for each manifold.

kellertuer: Yes, I also think, that the negative cases can only be tested individually by carefully constructing invalid cases (if they even exist) and check for the right error messages.

src/ProductManifold.jl
mateuszbaran: So, how do we name this function then?

sdaxen: Well, it's a submanifold of the original `PowerManifold`, so maybe `submanifold` could return a specific manifold from a `ProductManifold` and `submanifold_point` could return the point? The downside to this name is that one would expect to be able to provide any number of indices to build a submanifold, not just a single one.

kellertuer: Even for `getIndex` I would expect that any sub manifold can be extracted with for example `p[1:2]`, so it would also be a nice idea to have that hear. I like the name `submanifold_point` for its generality.

mateuszbaran: OK, extracting using an arbitrary set of indices is not a problem, I can do that.

mateuszbaran: I just noticed that `submanifold_point` won't fit for extraction of parts of product tangent vectors, cotangent vectors etc. Can we use a name that works in all of these cases, like `submanifold_component` for example? Or something shorter, like just `component`? Or maybe `factor`?

mateuszbaran: Alternatively, we could overload `getproperty` to allow for something like `p.access[i]`.

sdaxen: I prefer if we're going to use `submanifold` to get a subset of manifolds that we use something like `submanifold_` here. I think `submanifold_point` _could_ still work, because no manifold is provided, and a `TVector` and `CoTVector` are points on manifolds in their own rights, but since we haven't been using "point" in this general way so far, I can see how it would be confusing. I think `_component` works. If this and power manifold is the only place where we'll be using these functions, it might make sense to use `factor_` instead of `submanifold_`, since that is the more descriptive term.

src/ProductManifold.jl
sdaxen: Awesome, thanks! It's very clear now. It would be convenient if the user didn't need to explicitly create a `ShapeSpecification`. Is there a reason to not have a constructor like `ProductArray(P::ProductManifold, data) = ProductArray(ShapeSpecification(P.manifolds...), data)`?

src/ProductManifold.jl
```diff
@@ -0,0 +1,476 @@
...
+corresponds to array representations expected by `Sphere(2)` and `Euclidean(2)`.
+
+    M1 = Sphere(2)
+    M2 = Euclidean(2)
+    Mshape = Manifolds.ShapeSpecification(M1, M2)
+
+Next, the desired point on the product manifold can be obtained by calling
+`Manifolds.prod_point(Mshape, [1.0, 0.0, 0.0], [-3.0, 2.0])`.
+"""
+function prod_point(M::ShapeSpecification, pts...)
```
sdaxen: Would be nice to bring this inline with `submanifold_component` or whatever we decide, since this isn't unique to points.

src/ProductManifold.jl
mateuszbaran: My first idea was to put the shape specification in the product manifold itself but since it is just for one point representation I thought it might be better to keep it separate. The main reason for not adding a constructor like that is because creating shape specification is relatively slow (I just benchmarked it and it takes about 8 microseconds for relatively simple cases).

sdaxen:
Building a `ProductArray` for a manifold with more than 3 factor manifolds fails:

```julia
julia> E = Euclidean(3)
Euclidean{Tuple{3}}()

julia> P = Manifolds.ProductManifold(E,E,E,E)
Manifolds.ProductManifold{NTuple{4,Euclidean{Tuple{3}}}}((Euclidean{Tuple{3}}(), Euclidean{Tuple{3}}(), Euclidean{Tuple{3}}(), Euclidean{Tuple{3}}()))

julia> shape = Manifolds.ShapeSpecification(P.manifolds...)
Manifolds.ShapeSpecification{(1:3, 4:6, 7:9, 10:12),NTuple{4,Tuple{3}}}()

julia> x = randn(12);

julia> a = Manifolds.ProductArray(shape, x)
ERROR: MethodError: no method matching Manifolds.ProductArray(::Type{Manifolds.ShapeSpecification{(1:3, 4:6, 7:9, 10:12),NTuple{4,Tuple{3}}}}, ::Array{Float64,1})
Closest candidates are:
  Manifolds.ProductArray(::Type{Manifolds.ShapeSpecification{TRanges,Tuple{Size1,Size2}}}, ::TData<:AbstractArray{T,N}) where {TRanges, Size1, Size2, T, N, TData<:AbstractArray{T,N}} at /Users/saxen/projects/Manifolds/src/ProductManifold.jl:118
  Manifolds.ProductArray(::Type{Manifolds.ShapeSpecification{TRanges,Tuple{Size1,Size2,Size3}}}, ::TData<:AbstractArray{T,N}) where {TRanges, Size1, Size2, Size3, T, N, TData<:AbstractArray{T,N}} at /Users/saxen/projects/Manifolds/src/ProductManifold.jl:124
  Manifolds.ProductArray(::Manifolds.ShapeSpecification{TRanges,TSizes}, ::TData<:AbstractArray{T,N}) where {TM, TRanges, TSizes, T, N, TData<:AbstractArray{T,N}} at /Users/saxen/projects/Manifolds/src/ProductManifold.jl:131
Stacktrace:
 [1] Manifolds.ProductArray(::Manifolds.ShapeSpecification{(1:3, 4:6, 7:9, 10:12),NTuple{4,Tuple{3}}}, ::Array{Float64,1}) at /Users/saxen/projects/Manifolds/src/ProductManifold.jl:131
 [2] top-level scope at none:0```

File: src/ProductManifold.jl [hunk 32]
```diff
@@ -0,0 +1,478 @@
...
+`TRanges` and `TSizes` statically define the relationship between representation
+of the product manifold and representations of point, tangent vectors
+and cotangent vectors of respective manifolds.
+
+# Constructor
+
+    ProductManifold(M_1, M_2, ..., M_n)
+
+generates the product manifold $M_1 \times M_2 \times \dots \times M_n$.
+Alternatively, the same manifold can be contructed using the `×` operator:
```
sdaxen: We should re-export × so our interface still works even if the user hasn't imported it from `LinearAlgebra`.

src/ProductManifold.jl
```diff
@@ -0,0 +1,478 @@
...
+end
+
+"""
+    submanifold(M::ProductManifold, i::Val)
+
+Extract the factor of the product manifold `M` indicated by indices in `i`.
+For example, for `i` equal to `Val((1, 3))` the product manifold constructed
+from the first and the third factor is returned.
+"""
+function submanifold(M::ProductManifold, i::Val)
```
sdaxen: Would it also be good to have a generic `submanifold(M::ProductManifold, i) = submanifold(M, Val((i...))` so users can get the type-stability but also use convenient `submanifold(M, 1:3)`?

src/ProductManifold.jl
sdaxen: Yeah, I noticed it's slow. I guess the reason for this constructor would be convenience for the new user who's not worried about speed or only intends to create a single point for reuse. The documentation/examples could make it clear that this is not the efficient way to do it if one will want to create multiple points.

I agree it makes sense to keep the shape specification separate from the manifold. I didn't see it, but do we have a test where alternate representations are used? It's not clear to me how to do that.

mateuszbaran:
OK, I've fixed that bug with products of more than three manifolds, re-exported `\times` and added a more friendly method to `submanifold`.

src/ProductManifold.jl
mateuszbaran: I don't know, I'm not that good at naming things. I think I'll just let you and @kellertuer pick names for this and `prod_point`.

sdaxen: Okay, here're my thoughts. Let's keep `submanifold`, since we may have more uses for that function later. I don't think we should name this `submanifold_point`, since one would expect that to return something similar to the original `ProductArray`. Instead, it should be `submanifold_component(s)`, since the user gets a container of parts. I don't know whether the function name should be plural.

src/Manifolds.jl
```diff
@@ -458,19 +492,24 @@ the assumption is to be optimistic.
...
 include("ProjectedDistribution.jl")
 
 export Manifold,
     IsDecoratorManifold,
-    Euclidean
-export manifold_dimension,
+    Euclidean,
+    Sphere
+export ×,
+    manifold_dimension,
```
sdaxen: `manifold_dimension` is exported twice

src/ProductManifold.jl
```diff
@@ -0,0 +1,495 @@
...
+
+function isapprox(M::ProductManifold, x, v, w; kwargs...)
+    return all(t -> isapprox(t...; kwargs...), ziptuples(M.manifolds, x.parts, v.parts, w.parts))
+end
+
+function representation_size(M::ProductManifold, ::Type{T}) where {T}
+    return (mapreduce(m -> sum(representation_size(m, T)), +, M.manifolds),)
+end
+
+manifold_dimension(M::ProductManifold) = mapreduce(m -> manifold_dimension(m), +, M.manifolds)
```
sdaxen: Don't need the anonymous function. Can just do `mapreduce(manifold_dimension, +, M.manifolds)`

sdaxen:
Will `ProductManifold`, `ProductArray`, etc be exported?

mateuszbaran:
I've exported `ProductManifold` now. I'm not sure yet about other types.

src/ProductManifold.jl
mateuszbaran: OK, I've renamed it to `submanifold_component`. Singular seems more appropriate, right now components can be accessed only one at a time since there is no obvious way to handle multi-component access for `ProductArray` (new `ProductArray`? a tuple with views? an array with views?)

BTW, how do we rename `product_point` now?

kellertuer: Yay! Picking names! `Supercalifragilistic_expialigetic` (sorry its late). I was liking the name `access(p,index)`or `access(p,indices)` (which would also get in handy if we figure out how to rename that to getindex and get really `p[i]`-things. Otherwise I am also fine with `manifold_component` without plural, since for a single index its a point on a manifold and even for a container of parts, that could be seen as a point on a submanifold that is still a product manifold; the consistent way for both is hence the singular.

src/ProductManifold.jl
mateuszbaran: I think new users generally should use `ProductMPoint` instead of `ProductArray`, it's more intuitive.   I need to improve documentation on it. There are tests for `ProductMPoint` at the end of the test file for product manifolds.

mateuszbaran:
I've cleaned up the interface a bit. I'd like to merge it and resolve remaining issues in separate PRs, it'll only get harder and harder to merge other PRs with the changes made here.

PR #25, opened 2019-07-03
[WIP] Lie groups

sethaxen:
See initial discussion in https://github.com/JuliaNLSolvers/Manifolds.jl/issues/1.

This PR offers 2 different ways to create a group. First, one can create a manifold that is also a group and get certain functionality for free. Second, one can decorate an existing manifold with an `AbstractGroupOperation`. Three such operations are provided. See `CompactSymplectic` and `SpecialOrthogonal` for examples.

I'm still working on the machinery for metrics. For now, the default implementation assumes the left invariant canonical metric, i.e. the metric inherited from some parent group such that the inner product at the identity element is the usual dot product and the inner product at any other element is obtained by left action of the element on the tangent vectors.


src/Group.jl
```diff
@@ -0,0 +1,397 @@
...
+function adjoint_representation_derivative(G::AbstractGroupManifold, v, w)
+    return lie_bracket(G, v, w)
+end
+
+function lie_bracket(G::AbstractGroupManifold, v, w)
+    error("lie_bracket not implemented on $(typeof(G)) for vectors $(typeof(v)) and $(typeof(w))")
+end
+
+function inner(G::AbstractGroupManifold, x, v, w)
+    return dot(left_action(G, x, v), left_action(G, x, w))
```
mateuszbaran: Is there any example where this gives different result from `dot(v, w)`?

src/Group.jl
```diff
@@ -0,0 +1,397 @@
...
+end
+
+"""
+    exp!(G::AbstractGroupManifold, y, v)
+
+Exponential map of tangent vector `v` at the identity element of group `G`.
+Result is saved to `y`.
+"""
+function exp!(G::AbstractGroupManifold, y, v)
```
mateuszbaran: I'm afraid adding these methods on top of what `Manifold` already has can lead to confusion and surprises. Maybe `exp_id!` would be a better name?

mateuszbaran: Similar comment for `exp`, `log`, `log!`.

src/CompactSymplectic.jl
```diff
@@ -0,0 +1,108 @@
...
+          y[vi] .= (θ ≈ 0 ? 1 - θ^2 / 6 : sin(θ) / θ) .* θu
+      end
+      return y
+end
+
+function log!(::CompactSymplectic1, v, y)
+    @assert length(v) == 4 && length(y) == 4
+    vi = SVector{3}(2:4)
+    @inbounds begin
+        sinθv = view(y, vi)
```
mateuszbaran: `view(y, 2:4)` is perfectly fine and can be used in broadcasting below, even as `sinθv .*= (θ ≈ 0 ? 1 + θ^2 / 6 : θ / sin(θ))`

mateuszbaran: Similarly in `exp!` above.

src/Group.jl
```diff
@@ -0,0 +1,288 @@
...
+
+left_action(::AbstractGroupManifold{MultiplicationOperation}, x, p) = x * p
+
+compose(::AbstractGroupManifold{MultiplicationOperation}, x...) = *(x...)
+
+/(::AbstractGroupManifold{MultiplicationOperation}, x, y) = x / y
+
+\(::AbstractGroupManifold{MultiplicationOperation}, x, y) = x \ y
+
+@traitfn function lie_bracket(G::GT, v, w) where {O<:MultiplicationOperation,
```
mateuszbaran: Wouldn't it make sense to add here another argument: the point `x` at which `v` and `w` are tangent?

src/Group.jl
```diff
@@ -0,0 +1,288 @@
...
+
+@traitfn function exp!(G::GT, y, v) where {O<:MultiplicationOperation,
+                                           GT<:AbstractGroupManifold{O};
+                                           IsMatrixGroup{GT}}
+    y .= exp(v)
+    return y
+end
+
+_log(x) = log(x)
+_log(x::StaticArray) = log(Matrix(x))
```
mateuszbaran: I think `::StaticMatrix` would be better here.

src/CompactSymplectic.jl
sdaxen: In every case I tested, benchmarks indicated that using an `SVector` and copying instead of viewing led to greater speed-ups, sometimes up to 30%. Not certain why, but I'm guessing the overhead of `view` here is more expensive than copying 3 elements with the speed-up from the indices being an `SVector.` `sinθv` here is only read, not modified.

mateuszbaran: Yes, sorry, I've confused `v` and `vi`.

src/Group.jl
sdaxen: I agree it could get messy. I'm wondering if a better approach is to define a special `Identity` type (similar to `LinearAlgebra.I`, but it's only an identity element, not a `UniformScaling`).  We can then keep `exp` and `log` but dispatch on whether the identity element is provide.

mateuszbaran: Yes, a special identity type similar to `LinearAlgebra.I` would be even better.

src/Group.jl
```diff
@@ -0,0 +1,397 @@
...
+    error("identity not implemented on $(typeof(G)) for point $(typeof(x))")
+end
+
+@doc doc"""
+    left_action(G::AbstractGroupManifold, x, p)
+
+For an element $p$ of some set, compute the left action of group element $x$
+on $p$, i.e. $x \cdot p$.
+"""
+function left_action(G::AbstractGroupManifold, x, p)
```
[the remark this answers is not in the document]
sdaxen: It would be nice to have a motivating example for this. For all the Lie groups and G-spaces I've used, the action on the set is the same as the action upon itself.

src/Group.jl
sdaxen: Done

src/Group.jl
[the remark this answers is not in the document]
mateuszbaran: Right, it wasn't clear from the docstring that you only consider actions on `G`. In general, you could consider for example actions of SO(2) on R^3 and then you need to additionally specify the axis of rotation somewhere. I have a similar type in development for FunManifolds and it's actually useful.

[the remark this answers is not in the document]
sdaxen: Ah, I see what you're saying. Yes, that would be useful.

src/Group.jl
mateuszbaran: Looks great, I'm just not sure if defining a global constant `e` is a good idea. Julia already has a constant `ℯ` (`\euler`) which looks very similar.

mateuszbaran: Sorry, never mind, there is no such constant :+1: .

sdaxen: Oh, there's no global constant `e` here (though I would love a global constant; it's just that as you mentioned `I` and `ℯ` are already taken.). I just use the letter `e` for a local variables as a convention.

sdaxen: Beat me to it. 👍

src/Group.jl
sdaxen: Yes, that's a good point

src/Group.jl
kellertuer: Since the left action brings you into the Lie algebra, there might be a more efficient way in there, I think; though I do not have an example at hand.

src/Group.jl
```diff
@@ -0,0 +1,319 @@
...
+# Adapted from `afoldl` in `operators.jl` in Julia base.
+# expand recursively up to a point, then switch to a loop.
+group_afoldl(op, G, a) = a
+group_afoldl(op, G, a, b) = op(G,a,b)
+group_afoldl(op, G, a, b, c...) = group_afoldl(op, G, op(G, a, b), c...)
+
+function group_afoldl(op,G,a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,qs...)
+    y = op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,op(G,a,b),c),d),e),f),g),h),i),j),k),l),m),n),o),p)
+    return op(G, y, reduce((x, y)->op(G, x, y), qs))
+end
```
kellertuer: Can you explain this a little, it looks confusing with so many operators, what happens for example if you call it with more than that?

src/Rotations.jl
```diff
@@ -13,6 +13,10 @@ generates the $\mathrm{SO}(n) \subset \mathbb R^{n\times n}$
 struct Rotations{N} <: Manifold end
 Rotations(n::Int) = Rotations{n}()
 
+show(io::IO, R::Rotations{N}) where {N} = print(io, "Rotations($(N))")
```
kellertuer: 👍 for individual `show`s

sdaxen:
Sorry for the slowness on this PR. Yesterday was a holiday, and I'm pretty soon some of the math is wrong (or makes too many assumptions) regarding Lie groups, so I've been revisiting the theory a bit. Will be able to work on this some more over the next few days.

kellertuer:
Don't worry, we are all working on that basically in our free time and I for example am still amazed how allthis started and how much is going on here in such a short time.

src/Group.jl
sdaxen: Yeah this is derived from `operators.jl` in Julia base. It's used so expressions like `+(a,b,c,d,e,f)` get efficiently inlined when there are up to 16 elements. In base, the elements in `qs` are iterated through. Here, we use `reduce`, which I think will be more efficient (still need to benchmark this specific choice).

mateuszbaran:
Yes, don't worry about that, you are making good progress. I won't be able to constantly work on this as much as I did this week either.

src/Group.jl
```diff
@@ -243,79 +363,85 @@ struct MultiplicationOperation <: AbstractGroupOperation end
 *(e::Identity{G}) where {G<:AbstractGroupManifold{MultiplicationOperation}} = e
 *(::Identity{G}, x) where {G<:AbstractGroupManifold{MultiplicationOperation}} = x
 *(x, ::Identity{G}) where {G<:AbstractGroupManifold{MultiplicationOperation}} = x
-*(e::E, ::E) where {G<:AbstractGroupManifold{MultiplicationOperation},E<:Identity{G}} = e
+*(e::E, ::E) where {G<:AbstractGroupManifold{MultiplicationOperation},E<:Identity{G}} = x
```
mateuszbaran: `x` is not defined here, right?

src/Group.jl
```diff
@@ -243,79 +363,85 @@ struct MultiplicationOperation <: AbstractGroupOperation end
...
 
 /(x, ::Identity{G}) where {G<:AbstractGroupManifold{MultiplicationOperation}} = x
 /(::Identity{G}, x) where {G<:AbstractGroupManifold{MultiplicationOperation}} = inv(x)
 /(e::E, ::E) where {G<:AbstractGroupManifold{MultiplicationOperation},E<:Identity{G}} = e
 
 \(x, ::Identity{G}) where {G<:AbstractGroupManifold{MultiplicationOperation}} = inv(x)
-\(::Identity{G}, x) where {G<:AbstractGroupManifold{MultiplicationOperation}} = x
+\(::Identity{G}, g) where {G<:AbstractGroupManifold{MultiplicationOperation}} = x
```
mateuszbaran: Similarly, `x` doesn't seem to be defined here either.

src/Group.jl
sdaxen: Added in a version of `lie_bracket` for left-invariant fields

mateuszbaran:
One more thing, I'd like actions on manifolds to be more general, I'll copy some of my work-in-progress code here.

First, I have a type for abstract actions:
```julia
"""
    ActionOnManifold

An abstract group action on a manifold.

All types implementing this interface must implement either `optimal_alignment`
(for construction minimization-based quotient manifold) or
`to_section(p::Point, a::ActionOnManifold)` and
`to_section(v::TangentVector, a::ActionOnManifold)` for section-based
quotient manifolds (or both).
"""
abstract type ActionOnManifold end
```

And then a specific action is implemented like, for example,
```julia
"""
    TranslationAction(dim)

Action of translation on a flat space by a `dim`-dimensional vector.
"""
struct TranslationAction <: ActionOnManifold
    dim::Int
end
```
And then we need an "apply action" method
```julia
apply_action(m::Manifold, p, x, a::TranslationAction)
```
where `x` represents here the translation vector and `p` in the translated point on manifold `m`.

Then you can implement `LeftAction` and `RightAction` as special cases of `ActionOnManifold`.

I've pushed my code mostly towards functional actions (like curve reparametrization) and quotient manifolds, so there is not much overlap with what you are doing here (but I'd definitely like to merge these two parts into a single framework, so that's why this more general approach to actions is quite important for me).

sdaxen:
General actions on manifolds is one of the motivating examples for https://github.com/JuliaNLSolvers/Manifolds.jl/pull/28, though I'm still trying to work out some of the logic there regarding inverses/partial functions before continuing.

Any right action can be constructed from a left action and vice versa, so my thought was to have `Left` or `Right` as a parameter when an action is created with the default that the right is defined by the left unless the default is overriden.

So the idea would be to have something roughly like

```julia
abstract type GroupAction{G,M,LoR} <: AbstractMap{M,M} end
struct GroupTranslation{G,LoR,ET} <: GroupAction{G,G,LoR}
    x::ET
end
apply_action(L::GroupTranslation{G,<:Left}, y) where {G} = compose(domain(L), L.x, y)
apply_action(L::GroupTranslation{G,<:Right}, y) where {G} = compose(domain(L), y, L.x)

```

and so on. This has the benefit of being able to apply the machinery we will develop for pushforwards/pullbacks to group actions.

 To make sure this would be sufficiently general, it would be nice to see some examples of the kinds of general actions you're interested in. All of my motivating cases are kinematic and therefore straightforward. :)

mateuszbaran:
You can look at my previous example of actions of SO(2) on R^3 where you need to specify the axis of rotation. Other actions I need are rotations of curves, translations, reparametrizations of curves (including SRSF/SRVF-based actions described here: https://www.springer.com/gp/book/9781493940189 ) and actions of products of these groups. Anyway, I think any approach that accommodates my first example should be general enough for all of these cases so you can just think about that one.

I definitely look forward to having pushforwards and pullbacks integrated with group actions :).

> ```julia
> abstract type GroupAction{G,M,LoR} <: AbstractMap{M,M} end
> struct GroupTranslation{G,LoR,ET} <: GroupAction{G,G,LoR}
>     x::ET
> end
> apply_action(L::GroupTranslation{G,<:Left}, y) where {G} = compose(domain(L), L.x, y)
> apply_action(L::GroupTranslation{G,<:Right}, y) where {G} = compose(domain(L), y, L.x)
> ```

Definition `GroupAction` looks OK but I think `GroupTranslation` should contain an object representing group `G`. `apply_action` should also be more general: we have basically four pieces in my example:

1. Point `y` on which the action acts.
2. Manifold to which `y` belongs.
3. Axis of rotation.
4. Angle of rotation.

By the convention we have in this package points 1 and 2 should be passed separately. Additionally, I need to be able to represent action (with axis) without angle, so we can either have a type that has axis but not angle and either pass angle as a separate argument or add yet another type that gathers the axis and angle.

sdaxen:
> Definition `GroupAction` looks OK but I think `GroupTranslation` should contain an object representing group `G`

Yes, good point. I wrote this quickly, and it doesn't follow all of the conventions used in #28.

>`apply_action` should also be more general: we have basically four pieces in my example:
> 
> 1. Point `y` on which the action acts.
> 2. Manifold to which `y` belongs.
> 3. Axis of rotation.
> 4. Angle of rotation.
> 
> By the convention we have in this package points 1 and 2 should be passed separately. Additionally, I need to be able to represent action (with axis) without angle, so we can either have a type that has axis but not angle and either pass angle as a separate argument or add yet another type that gathers the axis and angle.

My conception of maps so far (in #28) has been quite literal. A map is a callable with a domain and a codomain that stores any additional information it needs for computation. It may or may not have some sort of defined inverse. Because it stores the information needed for dispatch, as a callable, it does not need the lengthy conventions we use for our functions; it takes a single input point, which is a member of its domain, and produces a single output, which is a member of its codomain. If it takes multiple inputs, then the domain should be a `ProductManifold`, and the inputs should be packaged together in a way compatible with `ProductManifold`. If this conception isn't sufficiently powerful for these simple examples, then this is a good time to work out what we need.

From the perspective I proposed above and in #28 , if the group action is on a point in `ℝ³`, then the map's domain and codomain would be `ℝ³`. `SO(2)` action on a point around an axis could be implemented with two maps. First `ϕ: 𝕊² × SO(2) → SO(3)` takes the axis and 2D rotation matrix, constructing the corresponding 3D rotation matrix. A second generic map `ψ: ℝⁿˣᵐ → ℝⁿˣᵐ` computes the natural group action of `SO(n)` on `ℝⁿˣᵐ`. The resulting group action of `SO(2)` on `ℝ³` is then constructed with the axis and angle, calling `ϕ` during construction and storing the resulting `ψ`. It takes a lone argument, an element of `ℝ³`.  But I guess thinking about it this way makes it clear that we're really doing 2 different things. `SO(2)` only acts naturally on items in the `xy` plane. To make it act on items in `ℝ³`, we're essentially including it in `SO(3)` and then conjugating it with a rotation that takes the `z`-axis to the axis of rotation, at which point it can act on `ℝ³` naturally. If we construct `ϕ` from the axis-angle, then `SO(2)` doesn't even appear during the construction. I guess I wouldn't consider this an action of `SO(2)` on `ℝ³` so much as action of `SO(3)` on `ℝ³`, where the element of `SO(3)` is constructed from an element related to `SO(2)`.

Alternatively, a map could be defined from the product manifold `𝕊² × SO(2) × ℝ³  → ℝ³`, comprised of `ϕ` and `ψ`. I guess this is the most flexible way to do this, since you can wrap this to construct either of the approaches you mentioned. But I also wouldn't consider this map a group action, since its domain and codomain are not the same, and the domain includes items that are not being acted upon.

Does this sound right? Sorry for the long post; I just want to make sure I'm thinking about this right and that we're on the same page before proceeding.

mateuszbaran:
This is certainly a valuable way of viewing this but let me elaborate on one part I haven't focused on before. When you consider quotients of manifolds by an isometric action, the quotient typically can be sort-of equipped with a Riemannian metric (technically there are some other requirements like the action being free or orbits being closed but even when they are slightly violated you can get reasonable results). When working with such quotients I use two operations that are concerned about how a group acts but don't take any particular group element: `to_section` and `optimal_alignment`. I could describe them using translations and the previous example of rotations but to fully appreciate what they do we need a slightly more complicated example.

First, let `M` be the manifold of triangles on a plane with the usual euclidean metric you get when you represent them by 2x3 matrices of coordinates of their corners. Then, we have a group SE(2) that can act on these triangles by moving and rotating them. The action is isometric, so we can have a quotient. How do we work with points on such quotient manifold? It turns out `to_section` and `optimal_alignment` together with manifold structures of `M` and SE(2) are enough for most operations. `to_section` restricts us to a certain section of either the group that acts or a subgroup of it -- in this case, it's the subgroup of translation. We can translate all triangles such that their center is at (0, 0) and be done with this part of the problem. So it's basically a map `M \to M`, different for each action (not just a group though, one group can act on the same manifold in many ways that lead to different `to_section` maps, though maybe your approach of extending the group action would be enough? I don't know but I don't see a reason to limit myself here).

Now only rotations remain. Now, `optimal_alignment` is there to take care of it. given two centered triangles `t_1, t_2` , it computes the rotation (in general, an element of the group that acts) `g` that such that the result of action of `g` on `t_2` gives a point closest to `t_1`. Now, for example, distance between the quotients of `t_1` and `t_2` by this action can be computed as the distance between aligned points (or triangles in this case). This can be nicely extended to logarithmic and exponential maps for example. I strongly suspect that horizontal-vertical separation of tangent vectors has to be separately defined but I haven't worked this part out yet.

Trying to express this in terms of maps, `to_section` would be something like `(G \times M \to M) \to (M \to M)` that maps actions to maps on `M` and `optimal_alignment` would be like `(G \times M \to M) \to (M \times M \to G)`. That's why I think having a type for `(G \times M \to M)` that doesn't store elements of `G` is useful.

> Does this sound right? Sorry for the long post; I just want to make sure I'm thinking about this right and that we're on the same page before proceeding.

You shouldn't be sorry, I really appreciate your post (and mine is quite long too :wink:). I didn't previously think about this in terms of maps and I'm happy we can do this now.

sdaxen:
> Trying to express this in terms of maps, `to_section` would be something like `(G \times M \to M) \to (M \to M)` that maps actions to maps on `M` and `optimal_alignment` would be like `(G \times M \to M) \to (M \times M \to G)`. That's why I think having a type for `(G \times M \to M)` that doesn't store elements of `G` is useful.

It took me a while, but I think I finally understand what you're trying to do. I think a sufficiently flexible abstraction for a single group action would be something like `G \times M \to M`, where `G` is acting upon `M` to produce a new element `M`. Rather than `to_section` itself being a map from one map to another, it could just be a map from `M \to M`, typed on the type of action and storing a copy of the action map internally. This would be the recommended way to build a new map from an old one when simple composition is not possible (e.g. conjugation).

sdaxen:
One more thing. Even though a group may act on a manifold in many different ways, I think in some cases it makes sense to define a canonical action. e.g.,  `SO(2)` canonically acts on elements of `R^3` via rotation. More generally, it acts on any element of `R^n` by rotation in the `xy` plane. Canonical actions can be retrieved by something like `canonical_action(G, M)`, which would return a map. This would make it easier to construct direct and semi-direct products. While it'll probably require long signatures to exactly specify how a complicated (semi-)direct product behaves, with canonical actions, a user can create a new (semi-)direct product using canonical actions with the simple `×` and `⋉` operators.

mateuszbaran:
> Rather than `to_section` itself being a map from one map to another, it could just be a map from `M \to M`, typed on the type of action and storing a copy of the action map internally.

Yes, it definitely can be, that makes sense.

> e.g., `SO(2)` canonically acts on elements of `R^3` via rotation. More generally, it acts on any element of `R^n` by rotation in the `xy` plane

This doesn't really look canonical to me, the plane of rotation is selected in a very arbitrary way. Some canonical actions make sense (for example action of `SO(n)` on `R^n`) but I don't see a convincing use case for this.

> This would make it easier to construct direct and semi-direct products.

I think this is generally not relevant to direct products? I agree though that for semi-direct products some set of canonical actions can provide useful defaults.

sdaxen:
> > e.g., `SO(2)` canonically acts on elements of `R^3` via rotation. More generally, it acts on any element of `R^n` by rotation in the `xy` plane
> 
> This doesn't really look canonical to me, the plane of rotation is selected in a very arbitrary way. Some canonical actions make sense (for example action of `SO(n)` on `R^n`) but I don't see a convincing use case for this.

I don't think it's arbitrary. You an always extend a rotation matrix to higher dimensions by making it block-diagonal, where the upper block is a rotation matrix and the lower block is an identity matrix. So we could define a canonical rotation for `SO(n)` that acts on any `R^m` for `m \ge n` in this manner, where only the dimensions up to `n` experience the rotation. The reason I would call this canonical is because this is the behavior one would expect if they took a vector in `R^n` and appended `m-n` zeros to the end. The vector remains within the `n`-dimensional hyperplane, and so one would expect that if `SO(n)` acted on this new `m`-dimensional vector, it would do so by rotating within the `n`-dimensional hyperplane.

There are of course other ways to extend a rotation matrix (e.g. make the upper block the identity matrix), but I wouldn't all these canonical, because they might take the vector out of its original hyperplane.

> > This would make it easier to construct direct and semi-direct products.
> 
> I think this is generally not relevant to direct products? I agree though that for semi-direct products some set of canonical actions can provide useful defaults.

Here's the case I was thinking about. To form a direct product of two groups, their action on each other must be commutative, which makes the product satisfy the group axioms. We can construct a direct product specifying a specific group action between the two groups, or we can have an abbreviated notation that assumes the canonical group action. If a trait designates whether that action is commutative, then this would make it easy to check upon construction whether the canonical actions generated a true direct product. In fact, if the user ever tried to construct a semi-direct product from two commuting groups, a direct product would be returned, so this would make it easy and safe to construct arbitrary group products.

mateuszbaran:
> > > e.g., `SO(2)` canonically acts on elements of `R^3` via rotation. More generally, it acts on any element of `R^n` by rotation in the `xy` plane
> > 
> > 
> > This doesn't really look canonical to me, the plane of rotation is selected in a very arbitrary way. Some canonical actions make sense (for example action of `SO(n)` on `R^n`) but I don't see a convincing use case for this.
> 
> Sorry, I meant "`SO(2)` canonically acts on elements of `R^2` via rotation."
> 
> I don't think it's arbitrary. You an always extend a rotation matrix to higher dimensions by making it block-diagonal, where the upper block is a rotation matrix and the lower block is an identity matrix. So we could define a canonical rotation for `SO(n)` that acts on any `R^m` for `m \ge n` in this manner, where only the dimensions up to `n` experience the rotation. The reason I would call this canonical is because this is the behavior one would expect if they took a vector in `R^n` and appended `m-n` zeros to the end. The vector remains within the `n`-dimensional hyperplane, and so one would expect that if `SO(n)` acted on this new `m`-dimensional vector, it would do so by rotating within the `n`-dimensional hyperplane.
> 
> There are of course other ways to extend a rotation matrix (e.g. make the upper block the identity matrix), but I wouldn't all these canonical, because they might take the vector out of its original hyperplane.

I'm afraid that the action of `SO(n)` on `R^m` for `m>n` may hide some bugs (someone wants to rotate around an axis but forgets to make a different action and everything works but the axis is incorrect). I think adding a single function that embeds `SO(n)` in `SO(m)` as you described will make the code easier to understand. Is there any context where having to use this embedding is problematic?

> Here's the case I was thinking about. To form a direct product of two groups, their action on each other must be commutative, which makes the product satisfy the group axioms. We can construct a direct product specifying a specific group action between the two groups, or we can have an abbreviated notation that assumes the canonical group action. If a trait designates whether that action is commutative, then this would make it easy to check upon construction whether the canonical actions generated a true direct product. In fact, if the user ever tried to construct a semi-direct product from two commuting groups, a direct product would be returned, so this would make it easy and safe to construct arbitrary group products.

Good idea, I didn't think about that. 👍

sdaxen:
> I'm afraid that the action of `SO(n)` on `R^m` for `m>n` may hide some bugs (someone wants to rotate around an axis but forgets to make a different action and everything works but the axis is incorrect). I think adding a single function that embeds `SO(n)` in `SO(m)` as you described will make the code easier to understand. Is there any context where having to use this embedding is problematic?

🤷‍♂ I'm fine with not providing a canonical action for `SO(n)` on `R^m` for `m \ne n`. My point was just that if we did, the embedding I described would be it. That being said, I don't know if that action will be used that often anyways.

I think we also need a new name for group exponential. In general, this only coincides with the Riemannian exponential when the metric is bi-invariant, so we need to keep it distinct from `exp`. I was thinking `exp_group`, but I don't much care for the name.

mateuszbaran:
> I think we also need a new name for group exponential. In general, this only coincides with the Riemannian exponential when the metric is bi-invariant, so we need to keep it distinct from `exp`. I was thinking `exp_group`, but I don't much care for the name.

I have never used the group exponential before, so I can't help much here. Is there a non-contrived example of a Lie group that doesn't have a bi-invariant metric? What can I use Lie group exponential for?

sdaxen:
> I have never used the group exponential before, so I can't help much here. Is there a non-contrived example of a Lie group that doesn't have a bi-invariant metric? What can I use Lie group exponential for?

That's a good question. Most of my experience is with kinematic groups, and most literature on algorithms for groups only considers the general linear group (matrix manifolds with matrix multiplication as the operation) and doesn't state which properties of the groups are general. But there are other non-`GL(n)` examples (e.g. the group in #27) where it would be good to be generic so we have correct defaults. I'll need to get back with you on that when I have a chance to read up on this more. If anyone has recommended texts, I'd appreciate it.

mateuszbaran:
> Is there a non-contrived example of a Lie group that doesn't have a bi-invariant metric?

Apparently the distinction between left- and right-invariant metrics is important in practice: https://arxiv.org/abs/1401.3609 .

PR #28, opened 2019-07-06
[WIP] Initial implementation of maps

sethaxen:
This is a (very) rough start on a map interface that I started working on for providing group actions in #25. My vision for our maps is to provide some level of type checking at compile time, a more familiar mathematical interface for working with maps, various types of derivatives, and pushforwards/pullbacks. It will lay the groundwork for pullback metrics and and charts, though this PR will probably only introduce basic maps. The maps should have very little overhead (most of the existing maps essentially compile out). I'd appreciate feedback on the overall design.


src/Map.jl
```diff
@@ -0,0 +1,221 @@
...
+struct CompositeMap{D,CoD,F,G}
+    f::F
+    g::G
+end
+
+function CompositeMap(f::F, g::G) where {D,
+                                         I,
+                                         CoD,
+                                         F<:AbstractMap{I,CoD},
+                                         G<:AbstractMap{D,I}}
```
mateuszbaran: Why do you require codomain of `G` to be exactly equal to domain of `F`? I think it's enough if that codomain is a subset of domain of `F`.

src/Map.jl
```diff
@@ -0,0 +1,221 @@
...
+end
+
+(g::ShortestGeodesic)(t) = g.geodesic(t)
+
+show(io::IO, m::ShortestGeodesic) = print(io, "ShortestGeodesic(", m.domain, ", ", m.geodesic.Exp.point, ", ", m.geodesic.tvector, ")")
+
+codomain(g::ShortestGeodesic) = codomain(g.geodesic)
+
+function derivation(γ::AbstractCurve{MT}) where {MT}
+    dim = manifold_dimension(domain(γ))
```
mateuszbaran: You probably mean `codomain` here, right?

src/Map.jl
sdaxen: Largely for clarity. One could always use an inclusion map to bridge the gap, and since that would basically just be the identity function, it wouldn't add any overhead. Ultimately, these maps should be able to take `MPoint`s, `TVector`s, and such, and a custom `MPoint` for a submanifold won't necessarily be an `MPoint` for its parent manifold without an inclusion map, whose function might be to change `MPoint` types.

src/Map.jl
sdaxen: Yes, that's right.

src/Map.jl
```diff
@@ -0,0 +1,221 @@
...
+    TM = Euclidean(dim)
+    return FunctionMap(domain(γ), TM, t -> ForwardDiff.derivative(γ, t))
```
mateuszbaran: Some points:

1. I don't really like hardcoding ForwardDiff here (though I did the same thing in FunManifolds.jl).
2. To the main point. It won't work most of the time. Firstly, If that's a curve on an embedded manifold, you'll get something in the embedded space, not exactly `TM`. Secondly, the result of `ForwardDiff.derivative` might not even be in the tangent space. Look what I did in FunManifolds: https://github.com/mateuszbaran/FunManifolds.jl/blob/master/src/functional_transformations.jl .
3. I'm leaning towards having curve derivation being tangent bundle-valued to emphasize the idea that tangent spaces at different points are distinct vector spaces.

mateuszbaran:
Good start, maps would be good to have. I wonder what is the overhead of these abstractions though, it might be worth adding to the benchmark suite.

src/Map.jl
sdaxen: 1. I agree hardcoding `ForwardDiff` isn't ideal. `derivation` here is more of a proof of concept and has serious flaws as you mention. For one, `manifold_dimension` will be replaced by `representation_size` once #22 is merged.
2. For an embedded manifold, if the tangent space is defined as the linear subspace of the embedded space's tangent space that is tangent to the manifold at that point, then I believe it is true that the derivative of the curve will be in the tangent space at that point. I'll have to study your approach in `FunManifold.jl` more carefully.
3. Yes, I think this makes sense. It would be great if we had both a type for a tangent bundle but also a type for tangent spaces. The latter could be really simple and just store a manifold and a point and use `===` to check if two tangent spaces are indistinguishable. It's not ideal to be just using `Euclidean` everywhere.

mateuszbaran: 1. Yes, it's fine to have ForwardDiff hardcoded, it works and can always be improved later.
2. You might be right, I'm trying to remember why I did that projection in FunManifolds.jl and I think I've made a mistake there, sorry. Still, we'll have to invent something to allow Lie groups storing tangent vectors as elements of the Lie algebra.
3. I'll work on both of these things soon, so you might want to wait with derivatives until it's done.

sdaxen:
I just pushed some benchmarks for `Logarithm` and `Exponential` that I need help understanding. Namely, in the benchmark suite, these maps can be much less efficient than the corresponding `log` and `exp` functions. Even if the map has already been created, where as a callable, they are just wrappers of the `log` and `exp` functions, there is a loss in efficiency. Example:

| ID                                                                                                                           | time            | GC time | memory         | allocations |
|------------------------------------------------------------------------------------------------------------------------------|----------------:|--------:|---------------:|------------:|
| `["manifolds", "Euclidean{2} -- MVector", "Exp"]`                                                                            |  46.000 ns (5%) |         |  64 bytes (1%) |           3 |
| `["manifolds", "Euclidean{2} -- MVector", "instantiated_Exp"]`                                                               |  48.000 ns (5%) |         |  64 bytes (1%) |           3 |
| `["manifolds", "Euclidean{2} -- MVector", "exp"]`                                                                            |  19.293 ns (5%) |         |  64 bytes (1%) |           3 |
| `["manifolds", "Euclidean{2} -- MVector", "Log"]`                                                                            |  46.000 ns (5%) |         |  80 bytes (1%) |           3 |
| `["manifolds", "Euclidean{2} -- MVector", "instantiated_Log"]`                                                               |  47.000 ns (5%) |         |  80 bytes (1%) |           3 |
| `["manifolds", "Euclidean{2} -- MVector", "log"]`                                                                            |  19.960 ns (5%) |         |  80 bytes (1%) |           3 |

So there's a loss in efficiency, but it's in the calling of the map, not in the constructor. However, this is not what I see when I benchmark them manually. Here's an example:

```julia
julia> M = Euclidean(2)
Euclidean{Tuple{2}}()

julia> x = @MVector randn(2)
2-element MArray{Tuple{2},Float64,1,2}:
 1.3373546726197771
 0.6934958258194306

julia> v = @MVector randn(2)
2-element MArray{Tuple{2},Float64,1,2}:
 -1.7392391474145479
 -1.2145367177871123

julia> @benchmark exp(M, x, v)
BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  3
  --------------
  minimum time:     35.903 ns (0.00% GC)
  median time:      38.393 ns (0.00% GC)
  mean time:        45.497 ns (9.16% GC)
  maximum time:     1.820 μs (97.45% GC)
  --------------
  samples:          10000
  evals/sample:     992

julia> @benchmark Manifolds.Exponential(M, x)(v)
BenchmarkTools.Trial: 
  memory estimate:  80 bytes
  allocs estimate:  4
  --------------
  minimum time:     96.331 ns (0.00% GC)
  median time:      105.408 ns (0.00% GC)
  mean time:        123.362 ns (5.94% GC)
  maximum time:     4.239 μs (96.11% GC)
  --------------
  samples:          10000
  evals/sample:     944

julia> exp_map = Manifolds.Exponential(M, x)
Expₓ: 𝑇ₓ𝑀 ⟶ 𝑀,  𝑥 ∈ 𝑀
𝑀 = Euclidean{Tuple{2}}()
𝑥 = 2-element MArray{Tuple{2},Float64,1,2}:
 1.3373546726197771
 0.6934958258194306

julia> @benchmark exp_map(v)
BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  3
  --------------
  minimum time:     34.223 ns (0.00% GC)
  median time:      40.582 ns (0.00% GC)
  mean time:        57.338 ns (12.43% GC)
  maximum time:     5.017 μs (97.53% GC)
  --------------
  samples:          10000
  evals/sample:     993

julia> wrapped_Exp(M, x, v) = Manifolds.Exponential(M, x)(v)
wrapped_Exp (generic function with 1 method)

julia> @benchmark wrapped_Exp(M, x, v)
BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  3
  --------------
  minimum time:     34.430 ns (0.00% GC)
  median time:      38.553 ns (0.00% GC)
  mean time:        48.548 ns (10.28% GC)
  maximum time:     3.636 μs (97.90% GC)
  --------------
  samples:          10000
  evals/sample:     993
```

There's a pretty consistent pattern with these maps that when benchmarked in the REPL with creation of the map, there's a loss in efficiency, but if the map is pre-constructed or if the construction of the map and its calling take place in a function, then the result is as efficient as `exp` itself. Does anyone have any ideas why the benchmark suite and this don't agree?

mateuszbaran:
For me, results of benchmarks and from REPL more or less match. Did you delete your `tune.json` file before running with a new benchmark suite? I've noticed that it's good to remove that file every time the benchmark suite changes.

sdaxen:
Oh, no I hadn't deleted `tune.json`. Once I did, the suite agrees that `Exponential` and `Logarithm` have no overhead. Thanks @mateuszbaran! I'll add some tests for other types of maps.

I'm considering replacing `Injection`/`Bijection` and `Surjection` with `LeftInvertibleMap` and `RightInvertibleMap`, respectively, since that's more in-line with what they'll be used for (pullbacks mostly), and the terms "injection" and "surjection" technically preclude the map being a partial map (where not every element of the domain maps to the codomain). Unless we have a good interface for submanifolds/subsets, this could become cumbersome. But the inverse of an injection is a partial bijection, so if we don't distinguish between partial and total maps, then `Injection`/`Bijection` could be merged into a single `LeftInvertibleMap`.

mateuszbaran:
It's good that the problem with benchmarks is fixed 😃 .

Correct me if I'm wrong but if we are allowing partial maps, then every function has a (potentially partial) right inverse? So left-invertible maps should be a both-side-invertible perhaps?

sdaxen:
I decided to substantially simplify the interface. The specialized types `Bijections`, `Injections`, and such, were misleading, since nothing forces the map to actually have those properties, and any map can be partial. The purpose of those specialized types was to define various types of inverses and validate map properties while in use. However, I think it's better when providing a feature that uses maps to document what properties the map should have and leave it to the user to be sure their map does have those properties. In many cases, even a map that violates some of the assumptions might still do provide something useful.

Inverses are handled by `inv` and `pinv`. `inv` returns a true inverse and in general should not be defined, while `pinv` loosely returns a left or right inverse. I probably won't split that out into `linv` and `rinv` until it becomes useful (e.g. while implementing pushforwards/pullbacks, which aren't part of this PR).

Some of the interface will improve when #31 gets merged, in particular we'll get better support for tangent bundles and such, with the motivating example being `Exp` and `Log`.

What do you think about the `show`s used here? Maps can have really hideous type names, especially when using products and composition, so the goal is to have something cleaner and more mathematically explicit for visualizing in the REPL. Still have no idea how to show `CompositeMap` well.

I'd appreciate any feedback on this implementation. I'd like to get this merged soonish so that I can use it in #25.

src/Map.jl
```diff
@@ -0,0 +1,449 @@
...
+    manifold::MT
+    domain::MMT
+    codomain::TMT
+end
+
+const Log = RiemannianLogarithm
+
+function RiemannianLogarithm(M)
+    shape = representation_size(M, TVector)
+    TₓM = Euclidean(shape...)
```
mateuszbaran: This will have to be rewritten once #31 is merged.

sdaxen: Yep this is a placeholder (https://github.com/JuliaNLSolvers/Manifolds.jl/pull/28#issuecomment-515664204). :)

mateuszbaran:
> Inverses are handled by `inv` and `pinv`. `inv` returns a true inverse and in general should not be defined, while `pinv` loosely returns a left or right inverse. I probably won't split that out into `linv` and `rinv` until it becomes useful (e.g. while implementing pushforwards/pullbacks, which aren't part of this PR).

Actually, what can I safely use `pinv` for? I don't really get that function. Left and right inverses seem to be more natural.

> Some of the interface will improve when #31 gets merged, in particular we'll get better support for tangent bundles and such, with the motivating example being `Exp` and `Log`.

Yes, I don't really like the `Log` and `Exp` in this PR. If you want to merge this PR soon, could you perhaps put these functions in a separate PR? By the way, I'm generally more used to thinking about the logarithm as a function `x \in M \to ( M \to T_x M)` so that it's explicit in which tangent space vectors lie. Similarly for exponential map.

> What do you think about the `show`s used here? Maps can have really hideous type names, especially when using products and composition, so the goal is to have something cleaner and more mathematically explicit for visualizing in the REPL. Still have no idea how to show `CompositeMap` well.

I think it's generally a good idea but I'll have to think about what exactly should be printed a bit more.

sdaxen:
> Actually, what can I safely use `pinv` for? I don't really get that function. Left and right inverses seem to be more natural.

No, it's a good point. `pinv` isn't a safe or clear default. Even `LinearAlgebra.pinv` sometimes returns left inverse, sometimes right.  I've added `linv` and `rinv`.

> If you want to merge this PR soon, could you perhaps put these functions in a separate PR?

I'm fine with removing them for now. I'll have `Geodesic` and `ShortestGeodesic` source their corresponding functions for now.

> By the way, I'm generally more used to thinking about the logarithm as a function `x \in M \to ( M \to T_x M)` so that it's explicit in which tangent space vectors lie.

That's an interesting way to define it. I'll have to revisit when I implement these for real. There will be variants that are localized at a base point, (`Log_x`, etc), which would be the kind of map returned in your example.

src/Map.jl
```diff
@@ -0,0 +1,476 @@
...
+Get the manifold to which outputs to the map `m` belong. By default, this is
+assumed to be stored in the field `m.codomain`.
+"""
+codomain(m::AbstractMap) = m.codomain
+codomain(::AbstractMap{M,N}) where {M,N<:Euclidean} = N()
+
+@doc doc"""
+    inv(f::AbstractMap)
+
+For a _total_ map $f \colon M \to M$, return its true inverse $g$, such that
```
mateuszbaran: Shouldn't `inv` work for all invertible maps (not only those for which domain and codomain are equal?)

src/Map.jl
```diff
@@ -0,0 +1,476 @@
...
+AbstractCurve{M} = AbstractMap{RealScalars,M}
+
+@doc doc"""
+    AbstractField{M,T<:Tuple} = AbstractMap{M,Euclidean{T}}
+
+An alias for a generic field, a map from a point on a manifold `M` to real
+space of some dimension stored in `T`,
+i.e. $\phi \colon M \to \mathbb R^{m \times \dots \times n}$
+"""
+AbstractField{M,T<:Tuple} = AbstractMap{M,Euclidean{T}}
```
mateuszbaran: This is somewhat questionable, not all fields have to be represented by values on `Euclidean`. I'd say that any Hilbert space would do (and maybe even this is too restrictive). `AbstractField` can be useful but I'd leave codomain up to the user.

sethaxen:
Closing this since it's very stale and is no longer a work _in progress_

PR #30, opened 2019-07-08
Ensure log is defined at cutlocus

sdaxen:
In an earlier discussion (sorry, I can't find it), we'd discussed how the `log` map is technically not defined at the cut locus but that it made sense for our log maps to still return a valid velocity vector when mapping between a point and a point in its cut locus, but we wouldn't guarantee which vector was returned. `Sphere` and `Rotations(2)` both produced velocity vectors of 0 when mapping between points at a distance of pi. This PR fixes that. Along the way, I added the `vee` and `hat` maps for the Lie algebras so(3) and so(2), changed the convention for 2D rotation matrices to represent active rotations, sped up `Rotations(2)` `log!` by 3x, and added `Rotations(3)` to the benchmark.


kellertuer:
I thought for the sphere and the rotations we added an error, but I'd be fine with a consistent deterministic vector that corresponds to _one_ of the geodesics, too.

sdaxen:
> I thought for the sphere and the rotations we added an error, but I'd be fine with a consistent deterministic vector that corresponds to _one_ of the geodesics, too.

No, there was no error. They just returned 0 vectors. I'd prefer a deterministic vector too. `Sphere` currently returns a random vector. I hadn't thought of this, but we could get a deterministic vector by returning a vector projected from the `(1,0,...,0)` axis and scaled accordingly unless the point is along the axis, in which case it returns the same but for the `(0,1,...,0)` axis. `Rotations` returns deterministic vectors.

kellertuer:
As I said, I'd be fine with the (slight) inconsistency to return a deterministic vector from the set of possible vectors.

sdaxen:
Matrix logarithm has undefined behavior near pi. When the angle is approximately pi, a valid tangent vector without a norm of pi will sometimes be returned, sometimes a 0 matrix will be returned, and sometimes the matrix will contain all NaNs. What would you all think about a standard flag for `log` that requires that it always return a valid velocity vector. If the flag is not set, then behavior is undefined and we don't do any checks, but if the flag is set, then checks are turned on to detect when `log` is undefined and return a valid vector nonetheless.

mateuszbaran:
Why would anyone disable these checks? I don't think they would be particularly expensive.

sdaxen:
> Why would anyone disable these checks? I don't think they would be particularly expensive.

That's fair. I thought they'd be more expensive for `Rotations`, but e.g. for SO(4), the eigendecomposition takes ~3μs, while the matrix logarithm takes ~100μs. I guess I could see a possibility of having a manifold where the checks are more expensive, but we can cross that bridge when we come to it.

sdaxen:
I think this PR is ready to merge now. Let me know if any changes are needed.

mateuszbaran:
Yes, it's good, I think we can merge it.

kellertuer:
Could you write a little more. documentation for `wee` and `hat` in general? I always work on `SO(3)` just with matrices, so I am not yet sure, what _I_ can use that for, still in general, maybe in `Manifolds.jl` a general fallback with an error message and a proper general documentation what both functions aim for would be nice :) Does it “just” do reshapes to vectors or is there something else you aim for?

Same for ` cos_angle_4d_rotation_matrix` – a documentation would be great (and yes I know it's always not nice to document everything, but I like having an as complete as possible documentation).

Concerning the log – one could check for the same on the sphere.

sdaxen:
> Could you write a little more. documentation for `wee` and `hat` in general? I always work on `SO(3)` just with matrices, so I am not yet sure, what _I_ can use that for, still in general, maybe in `Manifolds.jl` a general fallback with an error message and a proper general documentation what both functions aim for would be nice :) Does it “just” do reshapes to vectors or is there something else you aim for?

I've only encountered `vee` and `hat` on `SO(n)` and `SE(n)`, though I think they could be of general use, as I've found with `SU(2)`. For matrix manifolds, they assume some set of basis matrices that let you represent the matrix vector space with a vector. The basis matrices can be chosen so that the vector has some nice properties. For example, for `SO(2)`, it's the angle of rotation, for `SO(3)`, is the axis-angle representation (and with the usual basis set, `SU(2)`'s `vee` map produces the same rotation vector as `SO(3)` for half of `SU(2)`), and for `SO(4)`, the first 3 elements represent the first plane rotation (analogous to `SO(3)`), and the last 3 elements in some sense represent the second plane rotation. `hat` makes it easy to construct a valid vector space matrix from a simple vector.

Perhaps there's also a use for these to construct vectors tangent to manifold such as `Sphere`, but I haven't thought about this, because how the bases at different points relate seems a bit arbitrary. It's a bit clearer with group manifolds (e.g. the `vee` map on the 3-sphere as a group at the identity produces the same vector as `SU(2)`, perhaps with some scaling factor).

> Same for ` cos_angle_4d_rotation_matrix` – a documentation would be great (and yes I know it's always not nice to document everything, but I like having an as complete as possible documentation).

Sure, I hadn't documented this because it's meant to be an internal function, but I can do that.

> Concerning the log – one could check for the same on the sphere.

Check for what specifically? This PR already defines log at the antipode of the point on the sphere.

kellertuer:
Sorry, I did not look careful enough for the sphere.

Concerning `hat` and `vee`, I was wondering, whether they have a meaning and use on arbitrary Riemannian manifolds – but I did not have the time to look for that, and will also not next week, because I am at the ICIAM in Valencia.

sdaxen:
> Concerning `hat` and `vee`, I was wondering, whether they have a meaning and use on arbitrary Riemannian manifolds...

I'll look into that and see if there's a general application.

Also, this isn't ready to merge just yet. For some reason the Travis CI builds stall on OS X for Julia 1 when testing `Rotations`.

sdaxen:
I've finally identified the test that's killing the build. It's the SO(4)/SO(5) matrix exponential with `StaticArray`, which for Julia 1.0 on OS X for some reason takes an extremely long time to compile. We didn't see this lag before this PR because the matrices at the cut locus hit a different branch point in `exp`. I've filed an issue for this (see https://github.com/JuliaArrays/StaticArrays.jl/issues/631). For now, I'll convert the matrices to plain `Matrix` so we can actually pass the tests.

sdaxen:
So I did some thinking about the generalization of the `hat`/`vee` maps. Given a matrix representation of a tangent space, we can construct a basis for this tangent space (while conventionally tangent vectors have raised indices, their bases have lowered indices): `v^{ij} = e_k^{ij} v^k`. The vector `v^k` is in some sense the dual of the original vector, but it's also contravariant. I'll call it a coordinate vector. One way to view the basis is that it acts on the matrix representation of the metric to produce a new matrix compatible with the coordinate vectors: 

```
v^{ij} = v^k e_k^{ij}\\
g(v,w) = v^{ji} w^{ki} g_{jk} = v^m w^n e_m^{ji} e_n^{ki} g_{jk} = v^m w^n g_{mn}\\
g_{mn} \ doteq g(e_m, e_n)
```

If the basis set is orthogonal, then `g_{mn}` is diagonal. If the basis set is orthonormal with respect to the metric, then `g_{mn}` it is the identity matrix, and the norm on the vector space with respect to the metric is equivalent to the Euclidean norm on its coordinate vector.

A basis element can be converted to a cobasis element using the musical isomorphisms, and this above reasoning follows on the cotangent space.

In general, we can construct an orthogonal basis from an initial basis using Gram-Schmidt orthogonalization. This is tedious and slow, however. On Lie groups, we have an advantage, in that the basis on the Lie algebra can be brought to any tangent space via the appropriate group action. There's a bit more theory to work out here, but there are several quantities in practical use of Lie groups that rely on expression of a tangent vector as an actual vector assuming some basis set (Jacobians of exponential parameterization and adjoint matrices, for one; see https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2843106/pdf/nihms-175123.pdf, for example). Again, most of my experience is with the groups in kinematics, so I can't speak to how widely these are used.

In this perspective, the `hat` and `vee` maps are simply identification with a basis and extraction of a coordinate vector assuming a basis. I could see the benefit of having some machinery for specifying a basis on a vector space, but this is probably best considered in relation to https://github.com/JuliaNLSolvers/Manifolds.jl/pull/31. I'm fine with not exporting `hat` and `vee` and restricting it to just the kinematic groups for now to finish up this PR.

sdaxen:
Unless anyone sees any issues, I think this PR is ready to merge.

mateuszbaran:
> In this perspective, the `hat` and `vee` maps are simply identification with a basis and extraction of a coordinate vector assuming a basis. I could see the benefit of having some machinery for specifying a basis on a vector space, but this is probably best considered in relation to #31. I'm fine with not exporting `hat` and `vee` and restricting it to just the kinematic groups for now to finish up this PR.

I think even a slightly more general transformation to an orthonormal basis of a vector space at a point can be useful. @kellertuer added (https://github.com/JuliaNLSolvers/Manifolds.jl/pull/27#issuecomment-511181170 ) a function that computes such bases. #31 is somewhat orthogonal to this issue as I'm not going to work on vector space bases in that PR. Charts will have much bigger impact here.

> Unless anyone sees any issues, I think this PR is ready to merge.

I'd like to have the tests pass on appveyor, otherwise it's OK to merge.

sdaxen:
The Appveyor failure is the same one documented in https://github.com/JuliaNLSolvers/Manifolds.jl/issues/29 and comes from numerical instability of geodesic numerical integration on 32-bit Windows, so it's not related to these changes. I pushed a few documentation changes to trigger a new build, but if it doesn't pass I don't see a reason to delay the merge.

mateuszbaran:
Right, it's the same failure. Maybe we should just skip this test on 32-bit Windows? At least until we have a proper solution?

sdaxen:
I'm fine with that. Do you know how to skip a test for a specific build?

mateuszbaran:
The easiest way would be to surround that test by an `if` that checks `Sys.ARCH` and `Sys.iswindows()`.

PR #31, opened 2019-07-08
[WIP] Taking vector bundles seriously

mateuszbaran:
I've started working on vector spaces and vector bundles. What still remains:

- [ ] operations on vectors and vector spaces (like tensor product, exterior product etc.)
- [ ] distributions
- [ ] broadcasting
- [ ] working out details of interface for vectors


mateuszbaran:
I've asked a question on discourse about that broadcasting and it looks like the `ProductArray` approach is what we should stick to.

kellertuer:
At first glance this looks like a very good start, thanks!

dkarrasch:
I came across this one very recently, and was wondering if that's worth being integrated/imported/...

https://discourse.julialang.org/t/ann-grassmann-jl-differential-geometric-algebra/21125

mateuszbaran:
That is a very interesting package but I doubt it's worth integrating it here. I've skimmed the readme and it has very few examples of actually computing something and a lot of notation abuse. It seems to be designed primarily with symbolic computation in mind.

mateuszbaran:
I've introduced `zero_vector` that is supposed to work for any vector spaces and now I'm thinking what should we do with `zero_tangent_vector`. Right now it's the preferred way to specify zero tangent vectors for manifolds, but it's a bit redundant since `zero_vector` can do the same thing and more. What do you think?

src/VectorBundle.jl
```diff
@@ -0,0 +1,233 @@
...
+    exp!(M.M, y.parts[1], x.parts[1], v.parts[1])
+    exp!(vsm, y.parts[2], x.parts[2], v.parts[2])
+    return y
+end
+
+function log!(M::VectorBundle, v, x, y)
+    vsm = VectorSpaceManifold(M.type, M.M, x.parts[1])
+
+    log!(M.M, v.parts[1], x.parts[1], y.parts[1])
+    log!(vsm, v.parts[2], x.parts[2], y.parts[2])
```
mateuszbaran: Needs to be fixed

kellertuer:
> I've introduced `zero_vector` that is supposed to work for any vector spaces and now I'm thinking what should we do with `zero_tangent_vector`. Right now it's the preferred way to specify zero tangent vectors for manifolds, but it's a bit redundant since `zero_vector` can do the same thing and more. What do you think?

I do not yet completely understand, what you are aiming for. For me it looks like that `zero_vector` (we could also just name it `zero` btw.) provides zero elements for tangent spaces and cotangent spaces? I would propose to define `zero_tangent_vector` as a fallback to the zero in tangent space type vector spaces since it might be easier to understand and to type.

While I know what you are aiming for with `VectorSpaceManifold` for me sounds like it's everything that is isomorphic to Rn for that manifold (it sounds like the manifold that _is_ a vector space). Would `VectorBundle` or `VectorBundleManifold` be the better name?

mateuszbaran:
> I do not yet completely understand, what you are aiming for. For me it looks like that `zero_vector` (we could also just name it `zero` btw.) provides zero elements for tangent spaces and cotangent spaces? I would propose to define `zero_tangent_vector` as a fallback to the zero in tangent space type vector spaces since it might be easier to understand and to type.

`zero_vector` (we can call it `zero`, makes sense) is supposed to work for any vector bundle, for example differential forms. Arrays representing such vectors may have different sizes. We can use `zero_tangent_vector` as a fallback for tangent spaces, this should make no problems.

> While I know what you are aiming for with `VectorSpaceManifold` for me sounds like it's everything that is isomorphic to Rn for that manifold (it sounds like the manifold that _is_ a vector space). Would `VectorBundle` or `VectorBundleManifold` be the better name?

Well, `VectorSpaceManifold`s are isomorphic to R^n in this implementation and `VectorBundle` exists as a separate type. I am growing less and less convinced about the usefulness of `VectorSpaceManifold` though, I'll probably remove it.

By the way, I'm going to change a bit about `vector_transport`, having `project_vector` as a default breaks the vector bundle (it needs a reversible vector transport).

kellertuer:
Thanks for the explanation. I would also favour to leave out `VectorSpaceManifold` and propose to take parallel transport as default (I might also like to have a `parallel_transport` function for that that falls back to vector transport with that.

src/VectorBundle.jl
mateuszbaran: fixed

src/VectorBundle.jl
```diff
@@ -0,0 +1,221 @@
...
+end
+
+function representation_size(M::VectorBundle)
+    len_manifold = prod(representation_size(M.M))
+    len_vs = prod(representation_size(M.M, M.type))
+    return (len_manifold + len_vs,)
+end
+
+TangentBundle(M::Manifold) = VectorBundle(TangentSpace, M)
+CotangentBundle(M::Manifold) = VectorBundle(CotangentSpace, M)
```
sdaxen: Would it be better to declare these as types too for dispatch? e.g.

```julia
TangentBundle{M} = VectorBundle{TangentSpace,M}
CotangentBundle{M} = VectorBundle{CotangentSpace,M}
foo(T::TangentBundle) = ...
```

src/VectorBundle.jl
```diff
@@ -0,0 +1,221 @@
...
+
+"""
+    bundle_projection(M::VectorBundle, x::ProductRepr)
+
+Bundle projection of point `x` from the vector bundle `M`.
+"""
+function bundle_projection(M::VectorBundle, x::ProductRepr)
+    return x.parts[1]
```
sdaxen: I don't really understand from the docstring what this is. It's a projection of the bundle to what? What does the point represent? Also, since it basically does the same thing as `submanifold_component`, perhaps it should call that in case it's specialized for some reason??

mateuszbaran: It's a projection from the bundle to the manifold it is defined on: https://en.wikipedia.org/wiki/Vector_bundle . I guess it can just call `submanifold_component`, right.

src/VectorBundle.jl
mateuszbaran: I've tried that and it looks like the name `TangentBundle` can refer either to the function I've defined or be an alias for `VectorBundle{TangentSpaceType, M}`, so one of these things needs a different name.

src/VectorBundle.jl
sdaxen: I think using the phrase "projection from the bundle to the manifold" in the docstring would be more descriptive.

src/VectorBundle.jl
```diff
@@ -0,0 +1,221 @@
...
+    inner(M::Manifold, VS::VectorSpaceType, x, v, w)
+
+Inner product of vectors `v` and `w` from the vector space of type `VS`
+at point `x` from manifold `M`.
+"""
+function inner(M::Manifold, VS::VectorSpaceType, x, v, w)
+    error("inner not defined for manifold of type $(typeof(M)), vector space of type $(typeof(VS)), point of type $(typeof(x)) and vectors of types $(typeof(v)) and $(typeof(w)).")
+end
+
+function inner(M::Manifold, VS::TCoTSpaceType, x, v, w)
```
sdaxen: Is this a suitable default for a `CotangentSpace`? Doesn't the cotangent vector `v` need to be converted to a tangent vector before defining the inner product? Actually, maybe we need a general method for interconverting the two. It could just error by default but for `MetricManifold` use the metric directly.

src/VectorBundle.jl
```diff
@@ -0,0 +1,221 @@
...
+
+Dimension of the vector space of type `VS` over manifold `M`.
+"""
+function manifold_dimension(M::Manifold, VS::VectorSpaceType)
+    error("manifold_dimension not implemented for manifold $(typeof(M)) and vector space of type $(typeof(VS)).")
+end
+
+manifold_dimension(M::Manifold, ::TCoTSpaceType) = manifold_dimension(M)
+
+function representation_size(M::Manifold, VS::TCoTSpaceType)
```
sdaxen: I think the point of these functions is to provide much of the same functionality for vector spaces as we provide for manifolds. I guess I'm just concerned about all of these different signatures. Wouldn't it be cleaner to define vector space as a `Manifold` typed on and containing the manifold `M`? I doubt that'll add much, if any, overhead, and it's more consistent. Then all of these functions could maintain the same signature as the rest of the interface.

src/VectorBundle.jl
sdaxen: This worked for me

```julia
TangentBundle{M} = VectorBundle{TangentSpaceType,M}
TangentBundle(M::Manifold) = VectorBundle(TangentSpace, M)

CotangentBundle{M} = VectorBundle{CotangentSpaceType,M}
CotangentBundle(M::Manifold) = VectorBundle(CotangentSpace, M)
```

```julia
julia> TM = Manifolds.TangentBundle(Sphere(2))
VectorBundle{Manifolds.TangentSpaceType,Sphere{2}}(Manifolds.TangentSpaceType(), Sphere{2}())

julia> typeof(TM)<:Manifolds.TangentBundle
true
```

benchmark/benchmarks.jl
```diff
@@ -123,9 +123,28 @@ function add_manifold_benchmarks()
...
+    pts_prod_mpoints = [Manifolds.ProductRepr(p[1], p[2]) for p in zip(pts_s2, pts_r2)]
     add_manifold(m_prod, pts_prod_mpoints, "ProductManifold with MPoint";
         test_tangent_vector_broadcasting = false)
+
+    # vector spaces and bundles
+    begin
+        T = MVector{3, Float64}
+        x = convert(T, [1.0, 0.0, 0.0])
+        TB = TangentBundle(s2)
+        MT = VectorSpaceManifold(Manifolds.TangentSpace, s2, x)
```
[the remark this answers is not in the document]
mateuszbaran: Thanks for catching it, I didn't check benchmarks recently.

src/VectorBundle.jl
```diff
@@ -0,0 +1,221 @@
...
+using the modifying operation.
+"""
+function similar_result(M::Manifold, VS::VectorSpaceType, f, x...)
+    T = similar_result_type(M, VS, f, x)
+    return similar(x[1], T)
+end
+
+
+"""
+    manifold_dimension(M::Manifold, VS::VectorSpaceType)
```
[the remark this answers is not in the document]
mateuszbaran: Right, calling it `vector_space_dimension` will be more clear.

I was thinking about treating vector spaces as manifolds (there was that `VectorSpaceManifold` type) but it doesn't seem to be particularly useful. It was also a bit problematic to provide reliable implementations of `inner`, `exp`, `log`, etc for vector bundles, since either there would be two sets of methods for these vector spaces (one for `VectorSpaceManifold` and one general like the ones I currently have) or all methods would have to work on `VectorSpaceManifold`. In the first case extending these methods could lead to inconsistencies and in the second case new objects of that type would have to be needlessly constructed in many methods of `VectorBundle`. How would you solve this issue?

src/VectorBundle.jl
mateuszbaran: This is a very good question. I was thinking that covectors will have the same representation as their duals and thus computing their inner product would be easy but this assumption is indeed wrong. We probably need a method like `dual!(M::Manifold, VS::VectorSpaceType, v, x, w)` that saves to `v` the dual representation of vector `w` from vector space of type `VS` at `x`.

src/VectorBundle.jl
mateuszbaran: I wrote about this above: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/31#discussion_r302963120 so maybe we can discuss it there.

src/VectorBundle.jl
mateuszbaran: Thanks, it doesn't work if the lines are in the reverse order and I didn't think of trying it this way.

src/VectorBundle.jl
kellertuer: For me inner on tangent space introduces an inner product on the cotangent space by the musical isomorphisms. While I am pro a conversion between tangents and cotangents, `dual` for me might also refer to a dual pairing between tangent and cotangent space, so we should discuss this representation-change-functions name maybe.

sdaxen: > For me inner on tangent space introduces an inner product on the cotangent space by the musical isomorphisms.

This is also how I interpreted it. If my math is correct, a vector is obtained from a covector by multiplication with the cometric, i.e. `v^j = g^{ji} v_i`. The inner product for vectors is `g_{ij} v^i w^j`. Therefore, the inner product in terms of covectors is simply `g^{ij} v_i w_j`.

We could by default raise an error message for covector inner products, and individual manifolds are free to implement them. For `MetricManifold`, we could compute the inner product by default using `inverse_local_metric `, much in the same way as the `inner` on the tangent space does with `local_metric` for these manifolds. This has the benefit of being consistent with what we already do.

src/VectorBundle.jl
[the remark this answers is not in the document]
sdaxen: To be honest, I haven't thought about it much, and I'm not certain I understand the problem or how you intend to use vector spaces. It seems to me that it's fine to by default not implement methods that aren't general for all vector spaces (e.g. `exp`). Although, correct me if I'm wrong, but vector spaces behave much like Euclidean space and therefore `exp` and `log` would simply be addition and subtraction by a vector, right?

Part of the problem might be that we currently abuse notation a bit. Each of our method signatures begins with a manifold, which is roughly interpreted as "related to this manifold". So even though the inner product is technically defined on the tangent bundle of a manifold, we define it on the manifold itself. Likewise, `exp` includes a point on the manifold and a point in its tangent space, which are different manifolds, but only the manifold with the point is provided. I think what we've done so far makes for a cleaner interface, and perhaps the approach you're using here is likewise preferred because it's cleaner and easier to remember.

src/VectorBundle.jl
mateuszbaran: Wouldn't it be better to implement the default covector inner product using the sharp isomorphism? And just make the musical isomorphisms raise an error by default?

sdaxen: Yes, good idea. I think what I proposed may be a better default for `MetricManifold`.

src/VectorBundle.jl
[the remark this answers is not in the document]
mateuszbaran: You are right, `exp`, `log`, etc. are simple to implement for vector spaces. The problem, though, is that someone might want to implement a vector space where `exp` is not performed by simple addition of arrays for whatever reason and he would expect vector bundles to use an appropriate method.

Your point on the abuse of notation is very appropriate here. The "related to this manifold" convention was mostly sufficient until now. We need a way to specify which vector spaces given vectors come from. I think the approach I'm using here will be insufficient once I get to operations on more than one vector space anyway, I'll look for a better one.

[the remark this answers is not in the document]
mateuszbaran: I've introduced a new type to solve this problem `ManifoldVectorSpace`. I'm not really happy with this name as it's neither a manifold nor a vector space, but essentially corresponds to the set of all fibers of a vector bundle. This might solve our problems, I guess we'll see when I add tensor products and exterior products.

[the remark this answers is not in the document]
kellertuer: I am not completely in all theory concerning fibres, but why then not just call it `VectorBundleFibers` then?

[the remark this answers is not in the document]
mateuszbaran: Thanks, that's a much better name :+1: .

src/VectorBundle.jl
kellertuer: Just something for discussion: Why is `M` the second type argument of `VectorBundle`? I would've thought it'll e the first (design-wise) Since I would read it as “VectorBundle over M of X-type”

mateuszbaran: `M` is second because I think it's more common to dispatch on the vector space type than on the manifold, so we can write for example `VectorBundle{TangentSpaceType}` instead of `VectorBundle{<:Manifold, TangentSpaceType}`. I generally try to order type arguments from most likely to dispatch on to least likely to dispatch on.

sdaxen:
Weird idea, but I wonder if it would be useful to adopt the convention of representing cotangent vectors as row vectors (more generally, that the musical isomorphisms contain a transposition). It's common to explain tangent vectors as being like column vectors and cotangent vectors as row vectors because then the action of multiplication followed by the trace produces a scalar and represents the action of the cotangent vector on a tangent vector. From this perspective, in Euclidean space the musical isomorphisms are transpose operations. In non-Euclidean space, the flat operation is left-multiplication by the metric followed by transposition. The corresponding sharp operation uses the inverse metric and transposition:

```latex
v^\flat = (G v)^T = v^T G^T = v^T G\\
(v^\flat)^\sharp = (v^\flat G^{-1})^T = (v^T G G^{-1})^T = v\\
v^\flat (w) = \operatorname{tr}((v^T G) w) = \operatorname{tr}(v^T G w) = g(v,w)
```

This has the benefit of producing cotangent vectors that can directly act on tangent vectors via usual multiplication. For actual vectors we then get some nice implicit checks. Like you can't add a vector and a covector. And you can't increment a point column vector with a momentum row vector without first converting it to a velocity vector. And with the usual layout of a Jacobian matrix, gradients, which are 1-forms, are row vectors of that matrix. So we would have some implicit checks in the 1D case. This would mean that the cotangent vector in SO(3) is just the negative of the tangent vector.

The major downside I see with this is odd behavior if the user doesn't follow the convention. Also, the product of a row and column vector is a 1-element vector, on which the trace operator isn't defined for some reason.

mateuszbaran:
I don't think this transposition would be particularly useful, acting on tangent vectors via multiplication isn't something that could be relied upon anyway. I'm also afraid it wouldn't interact well with `ProductArray`.

By the way, here is a quick update about the progress on this PR. After implementing the musical isomorphisms and thinking about operations on more general vector spaces I think the idea of describing all arguments for a function only by the first argument (manifold, vector bundle fibers etc.) won't work that well. It was nice when we always had one "main" manifold we are working on but here we need to be able to mix vectors from different vector spaces. Musical isomorphisms already show this to a degree -- the type of the first argument was selected in a completely arbitrary way.

Another thing, I'll most likely postpone implementation of representation of tensor products and exterior products until we have charts, as I couldn't really find a reason why one would like to have them represented using the ambient space.

sdaxen:
> After implementing the musical isomorphisms and thinking about operations on more general vector spaces I think the idea of describing all arguments for a function only by the first argument (manifold, vector bundle fibers etc.) won't work that well. It was nice when we always had one "main" manifold we are working on but here we need to be able to mix vectors from different vector spaces. Musical isomorphisms already show this to a degree -- the type of the first argument was selected in a completely arbitrary way.

I think this was one of the uses we discussed for `MPoint` and its like. In principle, it lets us decorate a point, (co)tangent vector, etc, with information about the manifold to which it belongs and other properties, similar to decorator manifolds. Unless we can come up with a good set of rules for function signatures, perhaps we should revisit that option.

mateuszbaran:
> think this was one of the potential uses we discussed for `MPoint` and its like. In principle, it lets us decorate a point, (co)tangent vector, etc, with information about the manifold to which it belongs and other properties, similar to decorator manifolds. Unless we can come up with a good set of rules for function signatures, perhaps we should revisit that option.

I've been thinking about it recently. It's not ideal to decorate everything, mainly for performance reasons but we'll have to either decorate points and vectors for certain functions or make very long argument lists where the type of an argument (for example vector, covector etc.) is determined by another argument, for example
```julia
flat_isomorphism!(M::Sphere, v, ::CotangentBundleFibers, x, w, ::TangentBundleFibers)
```
The second option might be a bit better for performance but it doesn't look idiomatic. A rule of thumb here might be that points or vectors need to be decorated whenever a function operates on points or vectors from more than one space (or, in the future, more than one chart). I think I'll try this approach.

mateuszbaran:
I've added a decorator for vectors from fibers of vector bundles that should solve the issue with musical isomorphisms. I think I won't work in this PR on some of the things I originally wanted. Broadcasting is a hard problem for vector bundles (at least without some additional assumptions). I'm not quite sure what to do with distributions, I'll probably just generalize `TVectorDistribution` and maybe do a distribution for cotangent spaces. Tensor and exterior products need to wait for charts.

Do you have anything specific you'd like to see in this PR?

sdaxen:
I haven't had a chance to inspect the recent additions carefully, but there are a few pieces for my other PRs that I'm hoping would be included here (and maybe they already are). First, I would like to be able to define a map whose domain/codomain is a (co)tangent space/bundle in the same way as for a manifold.

In a related manner, I'd like the same machinery to be usable for defining a Lie algebra, which should be a tangent space but also enable dispatch. I don't have an obvious use case for this right now, but I could see it becoming useful while defining maps for groups. An example is how we've implemented `exp` and such for `Rotations`, where the tangent vector is _always_ in the Lie algebra. It would be nice to be able to use dispatch to specify whether the tangent vector is in the tangent space at the point generally or in the algebra.

mateuszbaran:
> First, I would like to be able to define a map whose domain/codomain is a (co)tangent space/bundle in the same way as for a manifold.

For (co)tangent bundles it's already easy, you can just use the type `VectorBundle`. (co)tangent spaces (at a specified point) were removed, mostly as a result of performance concerns and trying to have a bit more streamlined interface. Maybe I should put it back. Do you need the full manifold interface or just a separate type for "vector space at a point"? I think the latter might be more clear as there is not much reason to actually use the manifold interface for vector spaces.

> In a related manner, I'd like the same machinery to be usable for defining a Lie algebra, which should be a tangent space but also enable dispatch. I don't have an obvious use case for this right now, but I could see it becoming useful while defining maps for groups. An example is how we've implemented `exp` and such for `Rotations`, where the tangent vector is _always_ in the Lie algebra. It would be nice to be able to use dispatch to specify whether the tangent vector is in the tangent space at the point generally or in the algebra.

That would probably be best expressed by putting an object of type `Identity` as the point at which the vector space is defined in that case.

sdaxen:
> Do you need the full manifold interface or just a separate type for "vector space at a point"? I think the latter might be more clear as there is not much reason to actually use the manifold interface for vector spaces.

Yes, I think just having the type should be enough. I can relax the requirement that the domain and codomain both be manifolds. That might provide a nice generalization for the maps anyways and also enable us to specify when outputs are scalars.

mateuszbaran:
I've added a type for vector space at a point and extended tangent vector distributions to arbitrary vector spaces. I think this PR is fairly good now.

sdaxen:
I'll try to take another detailed look at this tomorrow (it's quite late here).

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+    x::TX
+end
+
+"""
+    FVector(type::VectorSpaceType, data)
+
+Decorator indicating that the vector `data` from a fiber of a vector bundle
+is of type `type`.
+"""
+struct FVector{TType<:VectorSpaceType,TData}
```
sdaxen: I guess I would expect `TVector` and `CoTVector` to be subtypes of `FVector`. I get that this is more general. Would it make sense to add an `AbstractFVector` parent type, of which `TVector`, `CoTVector`, and `FVector` are all subtypes?

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+(-)(v1::FVector, v2::FVector) = FVector(v1.type, v1.data - v2.data)
+(-)(v::FVector) = FVector(v.type, -v.data)
+(*)(a::Number, v::FVector) = FVector(v.type, a*v.data)
+
+eltype(::Type{FVector{TType,TData}}) where {TType<:VectorSpaceType,TData} = eltype(TData)
+similar(x::FVector) = FVector(x.type, similar(x.data))
+similar(x::FVector, ::Type{T}) where {T} = FVector(x.type, similar(x.data, T))
+
+@doc doc"""
+    flat_isomorphism!(M::Manifold, v::FVector, x, w::FVector)
```
sdaxen: Since isomorphism is a property of the map that won't always be included in the name of isomorphisms, wouldn't it make more sense to simply name these functions `flat` and `sharp`? Also, I think it would be nice to have more documentation here. sharp and flat are defined such that they produce (co)vectors such that the usual covector action upon a vector is equivalent to the inner product (but with equations :)).

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
+
+"""
+    VectorSpaceType
+
+Abstract type for tangent space, cotangent space, their tensor products,
+exterior products etc.
```
sdaxen: Should be "spaces" and have a comma before "etc"

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+Every vector space `VS` is supposed to provide:
+* a method of constucting vectors,
```
sdaxen: "constructing"

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+* basic operations: addition, subtraction, multiplication by a scalar
+  and negation (unary minus),
+* zero_vector!(VS, v, x) to construct zero vectors at point `x`,
+* `similar(v, T)` for vector `v`,
+* `copyto!(v, w)` for vectors `v` and `w`,
+* `eltype(v)` for vector `v`,
+* `vector_space_dimension(::VectorBundleFibers{<:typeof(VS)})`.
+
+Optionally:
```
sdaxen: Many of these function names need backticks

sdaxen: Also `@ref` would be good

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+    M::TM
+end
+
+"""
+    VectorSpaceAtPoint(fiber::VectorBundleFibers, x)
+
+A vector space (fiber type `fiber` of a vector bundle) at point `x` from
+the manifold `fiber.M`.
+"""
+struct VectorSpaceAtPoint{TFiber<:VectorBundleFibers, TX}
```
sdaxen: Would be nice to have convenient types and constructors for `TangentSpaceAtPoint` and `CotangentSpaceAtPoint`, similar to `TangentBundleFibers` defined below.

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+norm(M::VectorBundleFibers{<:TangentSpaceType}, x, v) = norm(M.M, x, v)
+
+"""
+    vector_distance(M::VectorBundleFibers, x, v, w)
+
+Distance between vectors `v` and `w` from the vector space at point `x`
+from the manifold `M.M`, that is the base manifold of `M`.
+"""
+vector_distance(M::VectorBundleFibers, x, v, w) = norm(M, x, v-w)
```
sdaxen: Why not just `distance`? The type should make it clear it's a vector.

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+"""
+    inner(M::VectorBundleFibers, x, v, w)
```
sdaxen: We've generally been putting `inner` higher than `norm`.

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+Inner product of vectors `v` and `w` from the vector space of type `VS`
```
sdaxen: `VS` doesn't appear anywhere

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+with vector spaces of type `VS`. In contrast with `VectorBundle`, operations
+on `VectorBundleFibers` expect point-like and vector-like parts to be
+passed separately instead of being bundled together. It can be thought of
+as a representation of vector spaces from a vector bundle but without
+storing the point at which a vector space is attached (which is specified
+separately in various functions).
+"""
+struct VectorBundleFibers{TVS<:VectorSpaceType, TM<:Manifold}
+    VS::TVS
+    M::TM
```
sdaxen: I think it makes sense to use `base_manifold` to retrieve `M`.

sdaxen: Especially since in several places, the docstrings call `M` the "base manifold"

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+Dimension of the vector space of type `M`.
+"""
+function vector_space_dimension(M::VectorBundleFibers)
+    error("vector_space_dimension not implemented for vector space family $(typeof(M)).")
+end
+
+vector_space_dimension(M::VectorBundleFibers{<:TCoTSpaceType}) = manifold_dimension(M.M)
+
+function vector_space_dimension(M::VectorBundleFibers{<:TensorProductType})
```
sdaxen: I haven't tested this, but I think it can be implemented more efficiently with `mapreduce`?

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+Vector bundle on manifold `M` of type `type`.
+"""
+struct VectorBundle{TVS<:VectorSpaceType, TM<:Manifold} <: Manifold
+    type::TVS
+    M::TM
+    VS::VectorBundleFibers{TVS, TM}
+end
+
+function VectorBundle(VS::TVS, M::TM) where{TVS<:VectorSpaceType, TM<:Manifold}
```
sdaxen: "where "

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+CotangentBundleFibers(M::Manifold) = VectorBundleFibers(CotangentSpace, M)
+
+TangentBundle{M} = VectorBundle{TangentSpaceType,M}
+TangentBundle(M::Manifold) = VectorBundle(TangentSpace, M)
+
+CotangentBundle{M} = VectorBundle{CotangentSpaceType,M}
+CotangentBundle(M::Manifold) = VectorBundle(CotangentSpace, M)
+
+"""
+    bundle_projection(M::VectorBundle, x::ProductRepr)
```
sdaxen: Generally, I think `B` makes for a better default variable name  for bundles than `M`. It also changes `M.M` to `B.M`, which is a little nicer.

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+end
+
+function isapprox(M::VectorBundle, x, v, w; kwargs...)
+    return isapprox(M.M, v.parts[1], w.parts[1]; kwargs...) &&
+        isapprox(M.M, x.parts[1], v.parts[2], w.parts[2]; kwargs...)
+end
+
+manifold_dimension(M::VectorBundle) = manifold_dimension(M.M) + vector_space_dimension(M.VS)
+
+function inner(M::VectorBundle, x, v, w)
```
sdaxen: I don't understand what this function means. What is an inner product on a tangent bundle, for example? What is `x`, and why does it have parts?

src/VectorBundle.jl
```diff
@@ -0,0 +1,372 @@
...
+
+function distance(M::VectorBundle, x, y)
+    dist_man = distance(M.M, x.parts[1], y.parts[1])
+    vy_x = vector_transport(M.M, y.parts[1], y.parts[2], x.parts[1])
+    dist_vec = vector_distance(M.VS, x.parts[1], x.parts[2], vy_x)
+
+    return sqrt(dist_man^2 + dist_vec^2)
+end
+
+function exp!(M::VectorBundle, y, x, v)
```
sdaxen: Oh this is cool! I think we're missing `exp` and `log`? Also, would love to see this documented in both `exp` and `vector_transport`, for the users who want geodesic flow.

src/VectorBundle.jl
sdaxen: Oh, I see. `x` is a point with a tangent vector. I don't really have a concept for the tangent space to a vector bundle. Perhaps documentation could explain `v` and `w`?

src/Sphere.jl
```diff
@@ -76,6 +76,27 @@ function zero_tangent_vector!(S::Sphere, v, x)
     return v
 end
 
+function vector_transport!(M::Sphere, vto, x, v, y)
```
sdaxen: I wonder if a simpler interface is `vector_transport!(M, wto, x, v, w)`, which transports `w` from `x` to `exp(M, x, v)`. That would eliminate, for example, the `v_xy` computation below, since `vector_transport` will usually be coupled with `exp`, so a velocity will already exist.

kellertuer: If so, I would like to have both interfaces, if possible (maybe only with types?).

src/ArrayManifold.jl
```diff
@@ -76,6 +84,14 @@ function inner(M::ArrayManifold, x, v, w; kwargs...)
...
 end
 
+function exp(M::ArrayManifold, x, v; kwargs...)
+    is_manifold_point(M, x; kwargs...)
+    is_tangent_vector(M, x, v; kwargs...)
+    y = ArrayMPoint(exp(M.manifold, array_value(x), array_value(v)))
+    is_manifold_point(M, y; kwargs...)
+    return y
+end
+
```
kellertuer: This looks quite nice. The only question I have is: Do we want to check every time? I think we discussed a while back, that _with_ types we do checks, without types (so if you have just arrays) we don't. Can that be done, here, too?

src/DistributionsBase.jl
```diff
@@ -4,7 +4,7 @@ using Distributions
 struct MPointvariate <: VariateForm
```
kellertuer: Could you write a short documentation for the `VariateForms`?

src/ProductManifold.jl
```diff
@@ -386,14 +162,29 @@ function inverse_retract!(M::ProductManifold, v, x, y, method::InverseProductRet
...
+end
+
+function sharp_isomorphism!(M::ProductManifold, v::FVector{TangentSpaceType}, x, w::FVector{CotangentSpaceType})
+    vfs = map(u -> FVector(TangentSpace, u), v.data.parts)
+    wfs = map(u -> FVector(CotangentSpace, u), w.data.parts)
+    map(sharp_isomorphism!, M.manifolds, vfs, x.parts, wfs)
+    return v
+end
+
+
```
kellertuer: Great to see sharp and flat here! Could you add a short documentation?

src/VectorBundle.jl
kellertuer: I would also op for `flat` and `sharp` – while sharp could even be `♭`, for sharp, that's not so easy ;)

sdaxen: Eh, it's not so hard. `\sharp` is `♯`, not `#`, so it works.

kellertuer: Right, I forgot. And I wanted to start calling `#` octothorpe ( https://en.wikipedia.org/wiki/Number_sign )

sdaxen: Can we all please agree to call it _anything_ before "octothorpe" ;)

kellertuer: Yes, of course, that's for `#` not for `♯`.

src/Sphere.jl
mateuszbaran: Hmm... I think these two functions (transporting in a direction and transporting to a point) should have different names. We'll also need vector transport along a curve. How about `vector_transport_point!`, `vector_transport_tvector!` and `vector_transport_curve!`?

kellertuer: I don't mind having different names, but I would prefer names, that indicate _what_ we are doing instead of what you have to provide. What about

* `vector_transport_to!` for the point version, since we want to transport _to_ `p`
* `vector_transport_direction!` for the vector version, since the tangent vector indicates the direction? 
* `vector_transport_along!` for the curve version?

I am not that happy with the second name yet, but if we come up with good ones, we can also distinguish the to ways of specifying geodesics in the same naming scheme.

src/ArrayManifold.jl
mateuszbaran: I don't think I understand? This method of `exp` looks just like the other function for array manifolds.

src/VectorBundle.jl
mateuszbaran: Maybe, I just don't see yet where I could use `AbstractFVector`.

src/VectorBundle.jl
mateuszbaran: Default versions of `exp` and `log` seem to work OK (they are tested). What exactly do you want documented for these functions? I think `exp` for vector bundles could be just described in general documentation for vector bundles.

src/Sphere.jl
sdaxen: > ...I would prefer names, that indicate _what_ we are doing instead of what you have to provide. What about
> 
> * `vector_transport_to!` for the point version, since we want to transport _to_ `p`
> * `vector_transport_direction!` for the vector version, since the tangent vector indicates the direction? 
> * `vector_transport_along!` for the curve version?
> 
> I am not that happy with the second name yet, but if we come up with good ones, we can also distinguish the to ways of specifying geodesics in the same naming scheme.
> 

So the base case here would be `vector_transport_along!`, which would require specification of a curve, either a function or a map (#28). `vector_transport_to` would call this with `shortest_geodesic`, while `vector_transport_direction` would call this with `geodesic`, right? I think `shortest_geodesic` and `geodesic` are adequately descriptive, but I don't see how we can adopt that terminology here. Perhaps `vector_transport_along_geodesic!` instead of `vector_transport_direction`. Likewise, we could do `vector_transport_along_shortest_geodesic!`, but that's just ugly. 

I guess the generalization of this would be `fiber_transport_XXX` or `transport_fiber` (I prefer this order since it's the fiber that is being transported, not the other way around). Alternatively, we can just provide `transport` and use types to specify what is being transported and how. e.g. `transport!(::TangentBundleFibers, xto, vto, x, v, curve)`.

src/VectorBundle.jl
mateuszbaran: I don't think it makes much of a difference anyway. Allocating objects of type `VectorBundleFibers` is what takes the most time.

src/VectorBundle.jl
sdaxen: > Default versions of `exp` and `log` seem to work OK (they are tested).

Oh, of course you're right. I was thinking `VectorBundle` wasn't a `Manifold`, but it is.

> What exactly do you want documented for these functions?

TBH, I'm looking at the code for `log`, and I don't get what it's doing. Same with `exp`. Would be nice to have it in the documentation.

src/Sphere.jl
mateuszbaran: No, `vector_transport_along!`, is definitely not the base case. Usually it would be implemented by sampling the curve and calling `vector_transport_to!` many times. The base case would either be the `_to!` variant or the `_direction!` variant. I don't have a preference here, one has to do `log` and the other has to do `exp` anyway.

> I guess the generalization of this would be `fiber_transport_XXX` or `transport_fiber` (I prefer this order since it's the fiber that is being transported, not the other way around). Alternatively, we can just provide `transport` and use types to specify what is being transported and how. e.g. `transport!(::TangentBundleFibers, xto, vto, x, v, curve)`.

`transport_vector` is fine as a generalization, it just needs the same trick with `FVector` that I did with `flat!` and `sharp!`. By the way, fiber in this case refers to an entire vector space.

src/VectorBundle.jl
mateuszbaran: Vector bundles are manifolds, so they can have vector bundles themselves You can look here: https://en.wikipedia.org/wiki/Double_tangent_bundle . I'll make a better general documentation for vector bundles.

src/Sphere.jl
kellertuer: I would also be fine with the last idea of just naming in `transport` and use dispatch; and yes, I think the default would be the point/`_to` variant.

sdaxen: > No, `vector_transport_along!`, is definitely not the base case. Usually it would be implemented by sampling the curve and calling `vector_transport_to!` many times.

Oh, sorry, of course. I was thinking more from a math perspective than implementation. It's true that the curve variant would need something like Schild's ladder and couldn't be the default.

src/VectorBundle.jl
sdaxen: 🤯

sdaxen:
One small thing this PR is missing is default implementations of `sharp!` and `flat!` for `MetricManifold`s, as well as the new implementation of `inner` for covectors discussed in  https://github.com/JuliaNLSolvers/Manifolds.jl/pull/31#discussion_r303197348

mateuszbaran:
I'll add `sharp!` and `flat!` for metric manifolds. `inner` for covectors is already there: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/31/files#diff-81d0c6a713446be33242b8af95482d5eR205 .

sdaxen:
> I'll add `sharp!` and `flat!` for metric manifolds. `inner` for covectors is already there: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/31/files#diff-81d0c6a713446be33242b8af95482d5eR205 .

Yes, the comment I linked to proposes a more sensible default when you have a form for the local metric, which `MetricManifold` does. The current default would for `MetricManifold` compute the metric and its inverse, multiply both covectors by the inverse metric, then take the inner product with the metric. When in reality all that is needed is to take their inner product with the inverse metric. Instead of 4 matrix-vector products, only 2 are necessary, and we avoid computing the local metric an extra time.

mateuszbaran:
I think I've addressed all issues. Is there anything I should improve in this PR? I'd like to merge it finally (I'm going to focus on AD and power manifold next).

src/ArrayManifold.jl
kellertuer: Yes, but array manifolds was intended to explicitly turn on checks (i.e. loosing a little performance compared to just arrays). Here you always have the checks on. Should we do something about that here and not check things?

kellertuer:
Sorry, I forgot to answer your question on my comment (got a little lost in finishing my paper recently); I have answered, but despite this last point I am fine with merging. Looking forward to the power manifold.

src/ArrayManifold.jl
mateuszbaran: I think it would be better to discuss this in a separate issue, it shouldn't block this PR.

mateuszbaran:
OK, thanks :). I'll merge it then and that last point can be resolved separately.

Issue #33, opened 2019-08-02
Using `ProductArray` with large submanifold


sdaxen:
It would be nice to be able to use `ProductArray` with some views not being a `SizedAbstractArray`. My use case is a `ProductManifold` with one thousands-dimensional Euclidean submanifold and other smaller submanifolds. It's not feasible to represent the part of the array corresponding to the Euclidean submanifold with a `SizedAbstractArray`, so I would like it to be a normal `Array`. I know `ProductMPoint` respects the type of representation, but I'm passing the product array to another package that expects an `AbstractArray` input.

I can already custom create a `ProductArray` with one of the parts being a normal `view`, but this is not respected whenever a new `ProductArray` is created; only the shape specification and underlying data array are used to create the new `ProductArray`.

Would it be a good idea to have some logic baked into the package so users don't accidentally try to create a 20,000-dimensional sized array like I did? I'm thinking also of some of the `MetricManifold` methods like `christoffel_symbols_first` that return sized arrays by default.

mateuszbaran:
Very good point, I didn't think about that. I'll work on that soon.

mateuszbaran:
Fixed by #34

PR #34, opened 2019-08-05
Non-statically sized product arrays

mateuszbaran:
I've made some changes to allow `ProductArray`s to have views that are not statically sized (it was causing some performance issues, see #33) .

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -124,17 +124,16 @@ function add_manifold_benchmarks()
 
 
     m_prod = Manifolds.ProductManifold(s2, r2)
-    shape_s2r2 = Manifolds.ShapeSpecification(s2, r2)
-    reshaper_static = Manifolds.StaticReshaper()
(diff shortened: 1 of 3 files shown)


sdaxen:
If I'm understanding this correctly, this solution would force all views to be either usual views or static views. I wonder though if it's possible to have a mixture. In the example mentioned in #33, I have a thousands-dimensional Euclidean manifold and many smaller submanifolds, which I would like to represent with static views. This will be, I think a common use-case in statistics, where many variables are transformed to a single (scaled and rotated) Euclidean manifold with a single PD matrix as a metric and other variables may be represented as points on other manifolds, such as PD matrices, n-simplices, or n-spheres. Perhaps a solution is to permit multiple reshapers to be passed to a `ProductArray`?

mateuszbaran:
The new versions allows for different reshapers for different views, does it look OK?

sdaxen:
Looks great to me! I don't see any problems and am fine with merging. Nice work on doing this so fast.

PR #39, opened 2019-10-09
Some fixes for mutating rand!

mateuszbaran:
I've fixed mutating `rand!` for most distributions and the `Array` type. `rand!` doesn't seem to be compatible with `MArray` and `SizedArray` though.


mateuszbaran:
By the way, the problems with `MArray` and `SizedArray` are caused by this: https://github.com/JuliaStats/PDMats.jl/issues/96 .

kellertuer:
Thanks for continuation of our work, I got a little stuck in teaching preparations for the new semester :/

mateuszbaran:
Luckily I don't have much work with teaching this semester. I was a bit busy with other projects but the power manifold is coming along quite nicely and I'll make a WIP PR once this: https://github.com/JuliaRegistries/General/pull/4205 is merged.

Issue #35, opened 2019-09-11
Integration into Optim


antoine-levitt:
Hi,

I just checked back, it seems there has been lots of activity on this project, great! Is it ready for prime time? In particular, can we replace Optim's implementation of manifolds with this package? We can think about a deeper integration sometime in the future, but right now I think just supporting manifolds embedded in R^n using retraction and projection of a vector onto the tangent space for the Stiefel and sphere manifolds (the basic things implemented in https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/Manifolds.jl) with a stable-ish API will be enough.

@pkofod: or should we wait until the next-gen optim?

mateuszbaran:
Thanks for asking!

I think `Manifolds.jl` isn't quite ready yet. A few things we need to consider before using it in things like `Optim.jl`:

1. Stiefel manifold isn't done yet (but it shouldn't be problematic), also the power manifold is a work in progress. I want to do it right (that's why I'm working on hybrid dynamic-static arrays in this branch: https://github.com/JuliaNLSolvers/Manifolds.jl/tree/mbaran/power-manifold , I'll put them in a separate package soon).
2. Documentation isn't deployed yet and we need a better README.
3. `Manifolds.jl`, at least in its current shape, is a pretty heavy dependency for something as simple as projection and retraction. A large part of this is caused by manifold-valued distributions (which could be moved into a separate package) and the metric manifold.
4. Our API for basic operations (including retraction) looks pretty stable. Projection onto the tangent space might receive a new argument argument that indicates the method to be used (like for retractions) but API should be backward-compatible with the current one.

I think issues 1-3 should be solved before `Manifolds.jl` can be safely used in `Optim.jl` (or the new Optim). I need a few weeks to solve (1). Issue (3) needs some discussion, I'm not sure how important a low number of dependencies is for you.

pkofod:
> @pkofod: or should we wait until the next-gen optim?

I think it's fine to try it out in a system we know :)

pkofod:
> I'm not sure how important a low number of dependencies is for you.

Not important if it's "our own"

kellertuer:
Thanks for this topic, I was not able to continue working the last month (sadly), I will continue to work on the SPDs hopefully next week – but I am on @mateuszbaran s side considering the basic API, but I would love to have at least a proper, complete documentation as well as maybe a few more manifolds before integrating somewhere.

pkofod:
[...]

PR #41, opened 2019-10-15
Make OrdinaryDiffEq and ForwardDiff optional

sdaxen:
As discussed in #35, by far most of `Manifold`'s long load time (>20s!) is due to `OrdinaryDiffEq` as a dependency. Since `OrdinaryDiffEq` and `ForwardDiff` are only necessary for `MetricManifold` defaults, this PR makes them optional dependencies. On my machine running Julia 1.3 rc2, `Manifold`'s new load time is ~2s.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,11 +7,10 @@ version = "0.1.0"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Einsum = "b7d42ee7-0b51-5a75-98ca-779d3107e4c0"
-ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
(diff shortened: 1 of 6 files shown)


mateuszbaran:
Thanks, that's a very good idea 👍 . My only comment is that I think it would be nice to tell users somewhere that `exp` and `exp!` are defined for metric manifolds but they need to use `OrdinaryDiffEq` and `ForwardDiff`. Right now all they would see if they tried using these functions without `OrdinaryDiffEq` is that they are not defined and it's not really obvious what should be done about it.

sdaxen:
Good point. Are you thinking documentation or a nice error message? I think the latter should be achievable with a helper function and dispatch.

mateuszbaran:
I think a nice error message would be perfect, yes.

mateuszbaran:
Looks great now, thanks!

PR #45, opened 2019-11-15
more docs for vector bundles

mateuszbaran:
I've added more docs to `VectorBundle`. Looking at it again, I've noticed that projections a bit sloppily implemented and vector transport needs to be improved but it doesn't look urgent.


mateuszbaran:
@kellertuer do you think this documentation is OK or would you recommend some changes?

kellertuer:
Looks ok, I like that it includes math formulae for the documentation and that all input and output parameters are explained.

Issue #24, opened 2019-07-02
Behavior of `is_manifold_point`


mateuszbaran:
As suggested here: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/18#discussion_r297415302 we should have two different functions: one for testing whether a point belongs to a manifold that returns `true` or `false` and the other one that throws an error when the point does not belong to the manifold. The first one can be generically implemented using the second one, with a `try` block. I'm not sure though how to name these functions.

sdaxen:
Maybe something like `is_manifold_point` for the function that returns a `Bool` and `check_manifold_point` for the function that uses it.

kellertuer:
For me, both also have a slightly different goal: while `is_manifold_point` should just (maybe even with goal performance) return `true or `false`, the other (maybe also `validate_manifold_point`?) should provide a long error message that tries to return, _why_ it is not a point, for example for an Sphere point because its in the wrong space (Rn+2 instead of Rn+1) or its not unit...
So it might only be an idea for a default fallback to do the _is_ case with a try catch. I would not recommend to do the other way around (use _is_ to do _check_/_validate_), since just  ´false` would not provide any information.

Since it should throw then an error anyways (and there is a "non-erroring version“) I would also not care about performance for the `check`/`validate`/`whatevernamewefind_`.

kellertuer:
I think this is resolved with `ManifoldsBase.jl` in version 0.2.0 and can hence be closed?

mateuszbaran:
Technically merging #46 will solve it in this repository but yes, it's almost done.

PR #46, opened 2019-11-20
Transition to ManifoldsBase.jl

mateuszbaran:
This PR transitions `Manifolds.jl` to the interface from `ManifoldsBase.jl`. Simultaneously, it resolves issue #24 .


src/ProductManifold.jl
```diff
@@ -178,37 +178,55 @@ end
...
 Check whether `x` is a valid point on the [`ProductManifold`](@ref) `M`.
 
 The tolerance for the last test can be set using the ´kwargs...`.
 """
-function is_manifold_point(M::ProductManifold, x::ProductRepr; kwargs...)
-    return all(t -> is_manifold_point(t...; kwargs...), ziptuples(M.manifolds, x.parts))
+function manifold_point_error(M::ProductManifold, x::ProductRepr; kwargs...)
+    for t ∈ ziptuples(M.manifolds, x.parts)
+        err = manifold_point_error(t...; kwargs...)
+        err === nothing || return err
```
kellertuer: This would just return the first error occuring, right?

If one works on code and looks for errors that would mean one has to solve one error then gets the second, solves that ...
Is there a way to collect all errors and return a string containing all messages?

I think I would find that better to get all errors at once (let's say if you made a mistake in 4 components).

src/ProductManifold.jl
```diff
@@ -178,37 +178,55 @@ end
...
+    end
+    return nothing
+end
+
+function tangent_vector_error(M::ProductManifold, x::ProductArray, v::ProductArray; kwargs...)
+    perr = manifold_point_error(M, x)
+    perr === nothing || return perr
+    for t ∈ ziptuples(M.manifolds, x.parts, v.parts)
+        err = tangent_vector_error(t...; kwargs...)
```
kellertuer: this is also the same as above.

src/ProductManifold.jl
```diff
@@ -178,37 +178,55 @@ end
...
-end
-
-function is_tangent_vector(M::ProductManifold, x::ProductArray, v::ProductArray; kwargs...)
-    is_manifold_point(M, x)
-    return all(t -> is_tangent_vector(t...; kwargs...), ziptuples(M.manifolds, x.parts, v.parts))
+function tangent_vector_error(M::ProductManifold, x::ProductRepr, v::ProductRepr; kwargs...)
+    perr = manifold_point_error(M, x)
+    perr === nothing || return perr
+    for t ∈ ziptuples(M.manifolds, x.parts, v.parts)
+        err = tangent_vector_error(t...; kwargs...)
```
kellertuer: This is the same point as above for manifold points

kellertuer:
The Readme should link to the `ManifoldsBase` – in the long run the `Base` part should be reflected better in the documentation, but I can do that also later.

src/ProductManifold.jl
mateuszbaran: Right, only the user is informed only about the first returned error. Your suggestion is good but that requires some design changes. Can we assume that `manifold_point_error` always returns a string or `nothing` and `check_manifold_point` throws a `DomainError` if `manifold_point_error` returns something else than `nothing`? Or maybe we could even return an empty string from `manifold_point_error` if there is no error (instead of `nothing`).

kellertuer: Well one could leave that at nothing, collect an array of the inner `manifold_point_errors()`, filter out nothings and return nothing if the filtered result is empty. Otherwise join the strings (error messages), maybe with a `"\n"` in between. Since here we always only do strings, the check anyways has no choice than return a `DomainError`. What do you think about that?

mateuszbaran: That looks fine, I'll do it.

mateuszbaran: There is a small problem with this, `DomainError` should carry two things: the problematic value and an (optional) error message. If we merge two error messages, the value would have to be discarded. The current solution is not worse than what we had so I'd just open an issue about it and solve it later.

mateuszbaran:
Let's wait with this until https://github.com/JuliaNLSolvers/ManifoldsBase.jl/pull/15 is merged and 0.2.0 of ManifoldsBase.jl is tagged.

mateuszbaran:
I think it's done, I'll merge this if CI passes.

mateuszbaran:
Finally CI is happy :tada:

PR #49, opened 2019-11-25
Fixed Rank and Stiefel

kellertuer:
In order to keep me reminded to do stuff: This starts to transfer two manifolds (maybe a third), that I started with students and a colleague in Manopt.jl:

* fixed rank matrices
* (complex and real valued) Stiefel

the possibly third is the generalized Stiefel, which should easily (I hope) be includable into Stiefel.


kellertuer:
I think I transferrred / implemented all functions I could find in papers or on manopt. Feel free to already criticise any slow implementations – Documentation is also nearly done – but testing is still missing.

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+    \text{ s.t. }
+    U_x^\mathrm{T}U = 0_k,
+    V_x^\mathrm{T}V = 0_k
+\bigr\},
+````
+where $0_k$ is the $k\times k$ zero matrix. See [`UMVTVector`](@ref) for details.
+
+This representation follows
+> Bart Vandereycken: "Low-rank matrix completion by Riemannian Optimization,
+> SIAM Journal on Optiomoization, 23(2), pp. 1214–1236, 2013.
```
mateuszbaran: I think there are too many "o"s

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+> doi: [10.1137/110845768](https://doi.org/10.1137/110845768),
+> arXiv: [1209.3834](https://arxiv.org/abs/1209.3834).
+"""
+struct FixedRankMatrices{M,N,K,T} <: Manifold end
+FixedRankMatrices(m::Int, n::Int, k::Int, T::Type = Type{Real}) = FixedRankMatrices{m,n,k,T}()
```
mateuszbaran: I think we can just use `T::Type = Real`, as `Real` is already a type.

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+* `SVDMPoint(U,S,Vt,k)` for the svd factors to initialize the `SVDMPoint`,
+  stores its svd factors shortened to the best rank $k$ approximation
+"""
+struct SVDMPoint{T} <: MPoint
+    U::Matrix{T}
+    S::Vector{T}
+    Vt::Matrix{T}
+end
+SVDMPoint(A::Matrix,args...) = SVDMPoint(svd(A),args...)
+SVDMPoint(S::SVD,args...) = SVDMPoint(S.U,S.S,S.Vt,args...)
```
mateuszbaran: These constructors are a bit confusing, I'd suggest explicitly writing variants with and without `k`.

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+  best rank $k$ approximation
+* `SVDMPoint(U,S,Vt,k)` for the svd factors to initialize the `SVDMPoint`,
+  stores its svd factors shortened to the best rank $k$ approximation
+"""
+struct SVDMPoint{T} <: MPoint
+    U::Matrix{T}
+    S::Vector{T}
+    Vt::Matrix{T}
+end
+SVDMPoint(A::Matrix,args...) = SVDMPoint(svd(A),args...)
```
mateuszbaran:
```suggestion
SVDMPoint(A::AbstractMatrix,args...) = SVDMPoint(svd(A),args...)
```

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+* `SVDMPoint(S)` for an `SVD` object, stores its svd factors
+* `SVDMPoint(U,S,Vt)` for the svd factors to initialize the `SVDMPoint``
+* `SVDMPoint(A,k)` for a matrix `A`, stores its svd factors shortened to the
+  best rank $k$ approximation
+* `SVDMPoint(S,k)` for an `SVD` object, stores its svd factors shortened to the
+  best rank $k$ approximation
+* `SVDMPoint(U,S,Vt,k)` for the svd factors to initialize the `SVDMPoint`,
+  stores its svd factors shortened to the best rank $k$ approximation
+"""
+struct SVDMPoint{T} <: MPoint
```
mateuszbaran: To make it more compatible with `StaticArrays` and AD, I recommend more abstract typing here, something like
```julia
struct SVDMPoint{TU<:AbstractMatrix, TS<:AbstractVector, TVt<:AbstractMatrix} <: MPoint
    U::TU
    S::TS
    Vt::TVt
end
```

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+
+A tangent vector that can be described as a product $UMV^\mathrm{T}$, at least
+together with its base point, see for example [`FixedRankMatrices`](@ref)
+
+# Constructors
+* `UMVTVector(U,S,Vt)` store umv factors to initialize the `UMVTVector`
+* `UMVTVector(U,S,Vt,k)` store the umv factors after shortening them down to
+  inner dimensions $k$, i.e. in $UMV^\mathrm{T}$, $M\in\mathbb R^{k\times k}$
+"""
+struct UMVTVector{T} <: TVector
```
mateuszbaran: I'd suggest similar changes in this struct as for `SVDMPoint`

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+    if !isapprox(v.Vt*x.Vt', zeros(k,k))
+        return DomainError(norm(v.Vt*x.Vt-zeros(k,k)), "The tangent vector $(v) is not a tangent vector to $(x) on the fixed rank matrices since v.V'x.V is not zero.")
+    end
+end
+
+function inner(::FixedRankMatrices{M,N,k,T}, x::SVDMPoint, v::UMVTVector, w::UMVTVector) where {M,N,k,T}
+    return dot(v.U,w.U) + dot(v.M,w.M) + dot(v.Vt,w.Vt)
+end
+
+manifold_dimension(::FixedRankMatrices{M,N,k,T}) where {M,N,k,T <: Real} = (M+N-k)*k
```
mateuszbaran: If you change the constructor for `FixedRankMatrices` as suggested, this would just be
```julia
manifold_dimension(::FixedRankMatrices{M,N,k,Real}) where {M,N,k} = (M+N-k)*k
manifold_dimension(::FixedRankMatrices{M,N,k,Complex}) where {M,N,k} = 2*(M+N-k)*k
```

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+project the matrix $A\in\mathbb R^{m,n}$ or a [`UMVTVector`](@ref) (e.g. from
+another tangent space) onto the tangent space at $x$, further
+decomposing the result into $v=UMV$, i.e. a [`UMVTVector`](@ref) following
+Section 3 in 
+> Bart Vandereycken: "Low-rank matrix completion by Riemannian Optimization,
+> SIAM Journal on Optiomoization, 23(2), pp. 1214–1236, 2013.
+> doi: [10.1137/110845768](https://doi.org/10.1137/110845768),
+> arXiv: [1209.3834](https://arxiv.org/abs/1209.3834).
+"""
+function project_tangent!(::FixedRankMatrices{M,N,k,T}, vto::UMVTVector, x::SVDMPoint, A::Matrix) where {M,N,k,T}
```
mateuszbaran:
```suggestion
function project_tangent!(::FixedRankMatrices{M,N,k,T}, vto::UMVTVector, x::SVDMPoint, A::AbstractMatrix) where {M,N,k,T}
```

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
...
+"""
+function retract!(::FixedRankMatrices{M,N,k,T}, y::SVDMPoint, x::SVDMPoint, v::UMVTVector, ::PolarRetraction) where {M,N,k,T}
+    s = svd( x.U * Diagonal(x.S) * x.Vt + (x.U * v.M * x.Vt + v.U*x.Vt + v.U*v.Vt) )
+    y.U .= s.U[:,1:k]
+    y.S .= s.S[1:k]
+    y.Vt .= s.Vt[1:k,:]
+    return y
+end
+
+zero_tangent_vector(::FixedRankMatrices{m,n,k,T},x::SVDMPoint) where {m,n,k,T} = UMTVector(zeros(T,n,k), zeros(T,k,k), zeros(T,k,m))
```
mateuszbaran: The mutating variant would also be nice

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,177 @@
+using LinearAlgebra: diag, Diagonal, svd, SVD, rank, dot
+
+@doc doc"""
```
mateuszbaran: This manifold is missing many methods that I took for granted previously, some tests will have to be made optional.

src/Stiefel.jl
```diff
@@ -0,0 +1,199 @@
...
+````
+where $0_n$ is the $n\times n$ zero matrix.
+
+# Constructor
+    Stiefel(m,n,T=Real)
+
+generate the (real-valued) Stiefel manifold of $m\times n$ dimensional orthonormal matrices.
+"""
+struct Stiefel{M,N,T} <: Manifold end
+Stiefel(m::Int, n::Int,T::Type = Type{Real} ) = Stiefel{m,n,T}()
```
mateuszbaran: My comment for constructor of `FixedRankMatrices` is also relevant here (`Real` vs `Type{Real}`)

src/Stiefel.jl
```diff
@@ -0,0 +1,199 @@
...
+ I_n & {\bar x}^{\mathrm{T}}v\end{pmatrix}
+ \right)
+\begin{pmatrix}  \exp( -{\bar x}^{\mathrm{T}}v) \\ 0_n\end{pmatrix}$
+
+where $\exp$ denotes matrix exponential, and $I_n$ and
+$0_n$ are the identity matrix and the zero matrix of dimension $n \times n$,
+respectively.
+"""
+function exp!(M::Stiefel{m,n,T}, y, x, v) where {m,n,T}
+    y .= [x v] * exp([x'v-v'*v; one(zeros(T,n,n)) x'*v]) * [exp(-x'v); zeros(T,n,n)]
```
mateuszbaran: I really don't like this line but it's a tough one to fix. @dkarrasch , could you take a look at this?

src/Stiefel.jl
```diff
@@ -0,0 +1,199 @@
...
+for two points `x`, `y` on the [`Stiefel`](@ref) manifold `M` and return
+the resulting tangent vector in `v`. The computation follows Algorithm 1 in
+
+> T. Kaneko, S. Fiori, T. Tanaka: "Empirical Arithmetic Averaging over the
+> Compact Stiefel Manifold", IEEE Transactions on Signal Processing, 2013,
+> doi: [10.1109/TSP.2012.2226167](https://doi.org/10.1109/TSP.2012.2226167).
+"""
+function inverse_retract!(::Stiefel{M,N,T}, v, x, y, ::QRInverseRetraction) where {M,N,T}
+  A = x'*y
+  R = zeros(T,N,N)
```
mateuszbaran: I'd replace `T` with, probably, something like `typeof(one(eltype(x))*one(eltype(y))` to properly mix types of arguments.

src/Stiefel.jl
```diff
@@ -0,0 +1,199 @@
...
+
+@doc doc"""
+    inner(M,x,ξ,ν)
+
+compute the inner product for two tangent vectors `v`, `w` from the
+tangent space of `x` on the [`Stiefel`](@ref) manifold `M`. The formula reads
+````math
+ (v,w)_x = \operatorname{trace}({\bar v}^{\mathrm{T}}w).
+"""
+inner(::Stiefel{M,N,T}, x, v, w) where {M,N,T} = real(tr(v'*w))
```
mateuszbaran: Isn't that just `real(dot(v, w))`?

src/FixedRankMatrices.jl
kellertuer: Oh, it should actually be enough to do _just_ the mutating one, the other one is already generally given in `Manifolds`.

src/FixedRankMatrices.jl
kellertuer: That's why I did not directly do the tests, exp/log have to get optional for example.

src/Stiefel.jl
kellertuer: Indeed.

src/FixedRankMatrices.jl
```diff
@@ -78,14 +80,24 @@ together with its base point, see for example [`FixedRankMatrices`](@ref)
...
+    M::TM
+    Vt::TVt
 end
 UMVTVector(U,M,Vt,k::Int) = UMVTVector(U[:,1:k],M[1:k,1:k],Vt[1:k,:])
 UMVTVector(U,M,Vt) = UMVTVector{eltype(U)}(U,M,Vt)
 
+*(v::UMVTVector, s::N) where {N <: Number} = UMVTVector(v.U*s, v.M*s,  v.Vpt*s)
+*(s::N, v::UMVTVector) where {N <: Number} = UMVTVector(s*v.U, s*v.M, s*v.Vpt) 
+/(v::UMVTVector, s::N) where {N <: Number} = UMVTVector(v.U/s, v.M/s, v.Vt/s)
+/(s::N, v::UMVTVector) where {N <: Number} = UMVTVector(v.U/s, v.M/s, v.Vt/s)
```
mateuszbaran: I think this should be the definition of the `\` method, right?

kellertuer: I don't want to solve linear equations, hm, what do I want to do? I don't know, maybe we leave that line out until I know again.

dkarrasch: Since you don't need the `N` type parameter, you could simply write `s::Number`. And, yes, I think the order of arguments should be preserved:
```julia
*(v::UMVTVector, s::Number) = UMVTVector(v.U*s, v.M*s,  v.Vpt*s)
*(s::Number, v::UMVTVector) = UMVTVector(s*v.U, s*v.M, s*v.Vpt) 
/(v::UMVTVector, s::Number) = UMVTVector(v.U/s, v.M/s, v.Vt/s)
/(s::Number, v::UMVTVector) = UMVTVector(s/v.U, s/v.M, s/v.Vt)
```
just in case somebody comes along with noncommutative number types.

mateuszbaran: `/(s::Number, v::AbstractVector)` actually calculates a pseudoinverse of `v` and multiplies it by `s` so I don't think it's what you want to do here. https://github.com/JuliaLang/julia/blob/6d26f14ede64cd949e715a76816df989b8bdedfc/stdlib/LinearAlgebra/src/generic.jl#L1110
`\(s::Number, v::AbstractVector)` divides `v` by `s`.

kellertuer: Then I'll better switch to `\`, you're right.

src/FixedRankMatrices.jl
kellertuer: I found a way: I generalized all appearances of `log` (despite those that really test exp/log identity) to `inverse_retract` with a `default_inverse_retraction` that as a parameter defaults to `LogarithmicInverseRetraction`as a parameter. Now I just have to fix my inverse retractions, still.

kellertuer:
If anybody has an idea for the inverse retractions (both polar and QR) for `Stiefel` any ideas are welcome. I can only get one thing – lapack errors – when trying to do tests and I can't directly see why.

kellertuer:
Ha. Found it, adapted the utils and all old tests are also back up running.
Missing:

- [ ]  FixedRank Tests
- [ ] Complex Stiefel Test
- [ ] Check Documentation
- [ ] Document parameters of `test_manifold`.

src/Stiefel.jl
```diff
@@ -102,18 +103,27 @@ end
...
+Setting $m=x^\mathrm{T}y$ this yields $(-m)s + s(-m^\mathrm{T}) + 2I = 0$,
+a continuous-time algebraic Riccati equation (arec). This follows Algorithm 2 in
+
+> T. Kaneko, S. Fiori, T. Tanaka: "Empirical Arithmetic Averaging over the
+> Compact Stiefel Manifold", IEEE Transactions on Signal Processing, 2013,
+> doi: [10.1109/TSP.2012.2226167](https://doi.org/10.1109/TSP.2012.2226167).
 """
 function inverse_retract!(::Stiefel{M,N,T}, v, x, y, ::PolarInverseRetraction) where {M,N,T}
-  v .= y*lyap(x'*y, -2*one(x'*x)) - x
+   # arec (A,Q,R) solves A'X + XA + XRX + Q = 0
```
mateuszbaran: The arguments for `arec` are in a different order (A, R, Q). It's quite surprising that this solver is more numerically stable than `lyap` and `lyapc` (from `MatrixEquations`).

mateuszbaran: And Travis reports errors from `MatrixEquations`

kellertuer: Oh that's just a type on the comment. Indeed for the equation Q is 2I, so the call is correct and just the comment is wrong; thanks for noticing.

I have not tried `lyapc` – but `lyap` (from LinearAlgebra) was not that stable.

kellertuer: That I do not understand: What's restriction `*` that combined with `1.0.0` only leaves `1.0.0`whcih I actually _have_ installed and which is available? On my Julia 1.2 the tests ran fine – I think I still have to learn how to add packages, such that tests don't fail.

mateuszbaran: `MatrixEquations` requires at least Julia 1.2 (that's what they have in Project.toml)

mateuszbaran: Oh, and they only test on Julia 1.2. Not even Julia 1.3. That's unfortunate.

kellertuer: Hm, I don't have a solution for that, I was happy to find numerically a solution for the problem I was facing.

mateuszbaran: We could improve `MatrixEquations` and make a PR but maybe we could do with `layp` from `LinearAlgebra` for now? And just have greater tolerances in tests?

kellertuer: I am not sure what the PR would look like, but back to `lapis not that good of an idea, since it's more than tolerance – it's `NaN`s and `Inf`s that I fought most of the afternoon.

mateuszbaran: Right, if it's `NaN`s and `Inf`s then it looks like we need `MatrixEquations`. I'll try to find some time to fix the library. For now you can convert the first argument of `arec` to `Array` (since `Transposed` seems to cause trouble with `hvcat` used in `MatrixEquations`). Travis on MacOS and Julia 1.2 also complains about some numerical inaccuracy of `project_tangent` for Stiefel.

kellertuer: Cool, thanks; yes, `project_tangent` seems to be a tough one when it comes to accuracy, I already introduced a tolerance-variable for that, but ycould not find a way to get that more accurate. I will do the Array tomorrow somewhen.

mateuszbaran: Some problems seem to come from type piracy in LinearOperators which could be fixed on master. There are still some accuracy issues. Finally, I've opened this issue in `ManifoldEquations.jl` https://github.com/andreasvarga/MatrixEquations.jl/issues/3 . I think we have to wait for now.

kellertuer: I haven't completely checked the type piracy you mentioned – which types does that involve?

mateuszbaran: Type piracy causes this error:
```
Type Array{Float32,2}: Error During Test at /home/travis/build/JuliaNLSolvers/Manifolds.jl/test/stiefel.jl:18
  Got exception outside of a @test
  MethodError: hvcat(::Tuple{Int64,Int64}, ::Adjoint{Float32,Array{Float32,2}}, ::Array{Float32,2}, ::Array{Float32,2}, ::Array{Float32,2}) is ambiguous. Candidates:
    hvcat(rows::Tuple{Vararg{Int64,N} where N}, xs::Union{AbstractArray{T,1}, AbstractArray{T,2}}...) where T in Base at abstractarray.jl:1634
    hvcat(rows::Tuple{Vararg{Int64,N} where N}, ops::Union{AbstractArray{T,2} where T, LinearOperators.AbstractLinearOperator}...) in LinearOperators at /home/travis/.julia/packages/LinearOperators/La2Ey/src/LinearOperators.jl:639
    hvcat(rows::Tuple{Vararg{Int64,N} where N}, xs::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T...) in Base at abstractarray.jl:1633
  Possible fix, define
    hvcat(::Tuple{Vararg{Int64,N} where N}, ::Vararg{Union{AbstractArray{T,1}, AbstractArray{T,2}},N} where N)
  Stacktrace:
   [1] arec(::Adjoint{Float32,Array{Float32,2}}, ::Array{Float32,2}, ::Array{Float32,2}) at /home/travis/.julia/packages/MatrixEquations/MBfWx/src/riccati.jl:81
```
You can see that the method at `LinearOperators.jl:639` can be called with any matrices, without using any types defined by `LinearOperators.jl`. That's type piracy.

kellertuer:
I just fixed the problem with MatrixEquations by getting back to the `lyap`-approach.

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,193 @@
...
+> arXiv: [1209.3834](https://arxiv.org/abs/1209.3834).
+"""
+struct FixedRankMatrices{M,N,K,T} <: Manifold end
+FixedRankMatrices(m::Int, n::Int, k::Int, T::Type = Real) = FixedRankMatrices{m,n,k,T}()
+
+@doc doc"""
+    SVDMPoint <: MPoint
+
+A point on a certain manifold, where the data is stored in a svd like fashion,
+i.e. in the form $USV^\mathrm{T}$, where this structre stores $U$, $S$ and
```
sethaxen:
```suggestion
i.e. in the form $USV^\mathrm{T}$, where this structure stores $U$, $S$ and
```

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,193 @@
...
+$V^\mathrm{T}$. The storage might also be shortened to just $k$ singular values
+and accordingly shortened $U$ (columns) and $V^\mathrm{T}$ (rows)
+
+# Constructors
+* `SVDMPoint(A)` for a matrix `A`, stores its svd factors (i.e. implicitly $k=\min\{m,n\})
```
sethaxen: The latex in this list is garbled

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,193 @@
...
+
+@doc doc"""
+    retract!(M, y, x, v, ::PolarRetraction)
+
+compute an SVD-based retraction on the [`FixedRankMatrices`](@ref) manifold
+by computing
+````math
+    y = U_kS_kV_k^\mathrm{T},
+````
+where $U_kS_kV_k^\mathmr{T}$ is the shortened singular value decomposition $USV=x+v$,
```
sethaxen:
```suggestion
where $U_k S_k V_k^\mathrm{T}$ is the shortened singular value decomposition $USV=x+v$,
```

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,193 @@
+using LinearAlgebra: diag, Diagonal, svd, SVD, rank, dot
+import Base: \, /, +, -, *
+@doc doc"""
+    FixedRankMatrices{M,N,K} <: Manifold
```
sethaxen: Can you document what the default metric choice is?

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
+using LinearAlgebra: diag, qr, tr, svd, mul!, zeros, lyap
+
+@doc doc"""
+    Stiefel{m,n,T} <: Manifold
+
+The Stiefel manifold consists of all $m\times n$, $m\geq n$ orthonormal matrices, i.e.
+
+````math
+\mathcal M = \{ x \in \mathbb K^{m\times n} : \overline{x}^{\mathrm{T}}x = I_n \}
```
sethaxen: Is K for the field more common than F? I remember you'd said earlier that Germans use K.

sethaxen: What does the overline signify? Complex conjugation? If so, perhaps the superscript H for the adjoint would be more clear.

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
...
+````
+where $\mathbb K \in \{\mathbb R, \mathbb C\}$ and
+$I_n \in \mathbb R^{n\times n}$ denotes the unit matrix.
```
sethaxen:
```suggestion
$I_n \in \mathbb R^{n\times n}$ denotes the $n \times n$ identity matrix.
```

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
...
+
+The tangent space at a point $x\in\mathcal M$ is given by
+````math
+T_x\mathcal M = \{ v \in \mathbb{K}^{m\times n} : x^{\mathrm{T}}v + v^{\mathrm{T}}x=0_n\},
+````
+where $0_n$ is the $n\times n$ zero matrix.
+
+The metric is either inherited from $\mathbb R^{m,n}$ for the real-valued case
+or the one inherited from interpreting the complex valued entries in the Gaussian
+plane $\mathbb R^2$ and then ober all entries as before.
```
sethaxen: ober?

sethaxen: Are the imaginary entries interleaved with the reals or separated?
We should probably set as a convention that equipping a complex manifold with a `EuclideanMetric` means that Hermitian inner product is applied with the Hermitian metric as the identity matrix. Which is equivalent to this definition. If that's documented there, then it doesn't need to be here.

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
...
+    return y
+end
+
+@doc doc"""
+    inner(M,x,ξ,ν)
+
+compute the inner product for two tangent vectors `v`, `w` from the
+tangent space of `x` on the [`Stiefel`](@ref) manifold `M`. The formula reads
+
+````math
```
sethaxen:
```suggestion
```math
```

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
...
+ ````
+ """
+inner(::Stiefel{M,N,T}, x, v, w) where {M,N,T} = real(dot(v,w))
+
+@doc doc"""
+    inverse_retract!(M, v, x, y, ::PolarInverseRetraction)
+
+compute the inverse retraction based on a singular value decomposition
+for two points `x`, `y` on the [`Stiefel`](@ref) manifold `M` and return
+the resulting tangent vector in `v`. This follows the wolloing approach:
```
sethaxen: wolloing?

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
...
+From the Polar retraction we know that
+````math
+    \operatorname{retr}_x^{-1}y = ys - q 
+````
+if such a symmetric positive definite $n\times n$ matrix exists. Since $ys-q$ is
+    also a tangent vector at $x$ we obtain
+````math
+x^{\mathrm{T}}ys + s(x^{\mathrm{T}}y)^{\mathrm{T}} + 2\mathrm{I}_k = 0.
```
sethaxen:
```suggestion
x^{\mathrm{T}}ys + s(x^{\mathrm{T}}y)^{\mathrm{T}} + 2 I_k = 0.
```

src/Stiefel.jl
```diff
@@ -0,0 +1,239 @@
...
+    retract!(M, y, x, v, ::QRRetraction )
+
+compute the QR-based retraction [`QRRetraction`](@ref) on the
+[`Stiefel`](@ref) manifold `M`. With $QR = x + v$ the retraction reads
+````math
+y = \operatorname{retr}_xv = QD,
+````
+where D is a $m\times n$ matrix with 
+````math
+D = \operatorname{diag}\bigl(\operatorname{sgn}(R_{ii}+0,5)_{i=1}^n \bigr).
```
sethaxen: What is the `sgn` operator?

src/Stiefel.jl
kellertuer: Oh, sorry, I meant to use F (for Field), germans still use K (which stands for Körper that translates (literally) to body but is the German mathemetical term for a field). I will change that.

Yes overline transposed is the complex conjugate – I can add that. I have used that notation for the Grassmann manifold already, but I can also change that to H. I personally prefer `$\bar{\cdot}^{\mathrm{T}}$` but if we in general switch to `$\cdot^\mathrm{H}$` I won't mind.

src/Stiefel.jl
kellertuer: Oh, the ober is is a typo (though in German it means “upper” as a prefix, it's just b and v being close on the keyboard).

It is actually the Euclidean (Hermitian) Metric form the embedding. I will rephase that. The question is, whether we should introduce a `HermitianMetric` to emphasise that (the difference could be, that the Euclidean really only does transposed and the Hermitian does an adjoint? Not sure whether that's reasonable).

src/Stiefel.jl
kellertuer: Oh, hm, no. You can start code with any number of back tricks, and this one starts and ends with 4, as do all in that file. Should we default to 3? I don't have a strong preference, just that I have sometimes seen 3 to indicate something different (don't remember what, but that's when I switched to 4).

src/Stiefel.jl
kellertuer: *following

docs/src/manifolds/stiefel.md
```diff
@@ -0,0 +1,23 @@
...
+
+```@autodocs
+Modules = [Manifolds]
+Pages = ["Stiefel.jl"]
+Order = [:type, :function]
+```
+
+## Literature
+
+T. Kaneko, S. Fiori, T. Tanaka: “Empirical Arithmetic Averaging over the
```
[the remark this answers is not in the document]
kellertuer: Yes, we can do that. I am currently using a different format in doc strings but I think that is necessary since they should be readable on REPL. I will reformat these

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,193 @@
...
+The manifold of $m\times n$ real-valued matrices of fixed rank $k$, i.e.
+````math
+\mathcal M = \{ x \in \mathbb R^{m\times n} : \operatorname{rank}(x) = k \}
+````
+
+# Representation with 3 matrix factors
+A point $x\in\mathcal M$ can be stored using orthonormal matrices
+$U\in\mathbb R^{m\times k}$, $V\in\mathbb R^{n,k}$ as well as the $k$ singular
+values of $x = USV^\mathrm{T}$. To be precise, $U$ and $V$ are from the manifolds
+[`Stiefel`](@ref)`(n,k)` and [`Stiefel`](@ref)`(m,k)`, respectively; see
```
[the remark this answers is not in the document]
kellertuer: Hm, I don't get your point? They are orthonormal matrices of said format and that can also be phrased as that they are from said manifolds.

src/Stiefel.jl
kellertuer: I added a definition.

src/FixedRankMatrices.jl
kellertuer: Yes, of course.

src/Stiefel.jl
sethaxen: Hm, my misunderstanding. For some reason, this equation [isn't rendering]( https://julianlsolvers.github.io/Manifolds.jl/previews/PR49/manifolds/stiefel.html#ManifoldsBase.inner-Union{Tuple{T},%20Tuple{N},%20Tuple{M},%20Tuple{Stiefel{M,N,T},Any,Any,Any}}%20where%20T%20where%20N%20where%20M)

kellertuer: Ah, spaces upfront, especially at the closing ticks.

src/Stiefel.jl
sethaxen: TBH, I don't know enough about complex manifolds to have an informed opinion here. All my use cases use the same Euclidean Hermitian metric. We currently already interpret a `EuclideanMetric` in several ways, all unified by the fact that they have the usual Euclidean inner product. Are there any complex manifolds whose inner product is not Hermitian? If not, maybe the convention should be that all inner products are adjoints, which of course includes the usual transpose.

kellertuer: Of course there are, but then a `HermitianMetric` is just not the default metric ;)

My question was more related to the naming scheme, since `EuclideanMetric` might lead a user to think that it just does transpose (not adjoint) on complex numbers (which it of course does not).

src/Stiefel.jl
sethaxen: I figured it was that. :)

I don't have a strong preference here (my own writing switches between 4 notations for the adjoint, that's how undecided I am). The reason I (currently) like the H superscript is because it emphasizes its connection to the transpose (in nearly all cases I've encountered, the adjoint for complex matrices is more like the transpose for real than the transpose for complex is), and it's unique; I've never used the H superscript for anything else, whereas overlines have meanings in stats, asterisks have meanings in differential geometry, and apostrophes have meanings in calculus.

🤷‍♂ But I might feel differently in a week.

src/FixedRankMatrices.jl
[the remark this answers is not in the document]
sethaxen: It's different dimensions than you state here. They're swapped. Unless I'm missing something.

src/FixedRankMatrices.jl
```diff
@@ -0,0 +1,219 @@
...
+on $\mathbb R^{m\times n}$ to the tangent bundle.
+
+This representation follows
+> Bart Vandereycken: "Low-rank matrix completion by Riemannian Optimization,
+> SIAM Journal on Optimization, 23(2), pp. 1214–1236, 2013.
+> doi: [10.1137/110845768](https://doi.org/10.1137/110845768),
+> arXiv: [1209.3834](https://arxiv.org/abs/1209.3834).
+"""
+struct FixedRankMatrices{M,N,K,T} <: Manifold end
+FixedRankMatrices(m::Int, n::Int, k::Int, T::Type = Real) = FixedRankMatrices{m,n,k,T}()
```
dkarrasch: `Real` is an abstract type, which may be bad for performance. Maybe you need to introduce something like your own set of fields? As I was typing, I remembered related code from @Jutho

https://github.com/Jutho/TensorKit.jl/blob/ae270985e1d71bc982d912144d98ef8b560da9e2/src/spaces/vectorspaces.jl#L10-L19

So, depending on your type `T`, you may wish to put `ℝ` or `ℂ` as your last argument.

src/Stiefel.jl
```diff
@@ -0,0 +1,241 @@
...
+From the Polar retraction we know that
+````math
+    \operatorname{retr}_x^{-1}y = ys - q 
+````
+if such a symmetric positive definite $n\times n$ matrix exists. Since $ys-q$ is
+    also a tangent vector at $x$ we obtain
+````math
+x^{\mathrm{T}}ys + s(x^{\mathrm{T}}y)^{\mathrm{T}} + 2I_k = 0.
+````
+This can either be solved by a Lyaponov approach or a continuous-time
```
dkarrasch: Lyapunov

src/Stiefel.jl
```diff
@@ -0,0 +1,241 @@
...
+algebraic Riccati equation as described in 
+
+> T. Kaneko, S. Fiori, T. Tanaka: "Empirical Arithmetic Averaging over the
+> Compact Stiefel Manifold", IEEE Transactions on Signal Processing, 2013,
+> doi: [10.1109/TSP.2012.2226167](https://doi.org/10.1109/TSP.2012.2226167).
+
+This implementation follows the Loyapnov approach.
```
dkarrasch: Lyapunov

src/FixedRankMatrices.jl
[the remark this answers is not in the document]
kellertuer: I'm confused. U is n x k, where n ≥ k, V is m x k, where m ≥ k, which corresponds to the interface of Stiefel and the product with S k x k is valid, so where are they swapped exactly?

src/Stiefel.jl
```diff
@@ -0,0 +1,241 @@
...
+    D = Diagonal( sign.( sign.(d .+ convert(T, 0.5))) )
+    y .= zeros(M,N)
+    y[1:N,1:N] .= D
+    y .= qrfac.Q * D
+    return y
+end
+
+representation_size(::Stiefel{M,N,T}) where {M,N,T} = (M,N)
+
+zero_tangent_vector!(::Stiefel{M,N,T},v,x) where {M,N,T} = fill!(v,0)
```
dkarrasch: When types are not needed, then it's Julia style to omit them altogether.

src/Stiefel.jl
sethaxen: I think if it's clearly documented, there's not much risk of confusion. And it might be more in line with expectations. e.g. `dot`, which is basically just an alias for "Euclidean inner product", on complex matrices does an adjoint.

src/Stiefel.jl
```diff
@@ -0,0 +1,241 @@
...
+# Constructor
+    Stiefel(m,n,T=Real)
+
+generate the (real-valued) Stiefel manifold of $m\times n$ dimensional orthonormal matrices.
+"""
+struct Stiefel{M,N,T} <: Manifold end
+Stiefel(m::Int, n::Int,T::Type = Real) = Stiefel{m,n,T}()
+
+function check_manifold_point(S::Stiefel{M,N,T},x; kwargs...) where {M,N,T}
+    if (T <: Real) && !(eltype(x) <: Real)
```
dkarrasch: If you go with the above own-defined fields, then this turns into
```julia
if T === ℝ ....
```

src/FixedRankMatrices.jl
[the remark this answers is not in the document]
sethaxen: But above you say `$U\in\mathbb R^{m\times k}$` and `$V\in\mathbb R^{n,k}$`. Then here you swap `m` and `n`.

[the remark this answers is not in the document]
kellertuer: Oh. Thanks, I will fix that to match again.

src/Stiefel.jl
kellertuer: Ok. Then we just stay with `EuclideanMetric` and I will document that `EuclideanMetric` with a complex manifold means a Hermitian metric.

src/FixedRankMatrices.jl
kellertuer: While I like the notation that one gets there, the disadvantage is, that one can not “see” the `eltype(x)` of a point on a manifold from `M`, though I am also not 100% sure whether we need that somewhere explicitly.

kellertuer: Maybe @mateuszbaran could help deciding, which way we want to go?

dkarrasch: Oh, you mean `T` is meant to encode the `eltype` of manifold points? That's fine, but then the default should not be `Real`, but, say, `Float64`. I thought this `T` is decoupled from `eltype`s, and just meant to encode whether the manifold is real or complex.

kellertuer: Actually, I am not yet 100% sure, I had in mind to maybe use it that way. But you might also be right, that it might be enough to know whether it's Real or Complex, then we should of course follow your idea to provide field types.

mateuszbaran: `T` was not supposed to encode `eltype` of points on a manifold, it was supposed to just be a tag (effectively equivalent to the code from `TensorKit.jl`, although using specific types makes the intent more clear). I think that generally manifold objects should not store a precise `eltype` of points.

BTW, `TensorKit.jl` looks interesting. Maybe it would make sense to integrate `Manifolds.jl` and `TensorKit.jl` at some level?

kellertuer: Great. That's what I wanted to make sure :) So we mimic/adapt the types from `TensorKit.jl` in `Manifolds.jl` ?

mateuszbaran: Yes, we can adapt the types from `TensorKit.jl` (though maybe without these single-letter Unicode aliases).

kellertuer: Oh, I especially _like_ these... do you have a strongwomen opinion against that and why?

mateuszbaran: It's not a strong opinion, just a suspicion that someone might want to use these symbols for something else and there will be incompatibility. You can use them if you like, we can remove them if they actually cause problems.

kellertuer: You might be right. I will still start with them and we will follow your suggestion to remove them if they cause problems.

kellertuer: Since also manifolds from other PRs are involved in this, I would like to postpone this to after finishing this PR and do a separate one for the type-change, see #62.

mateuszbaran: OK, this can be done later.

kellertuer:
Since I would like to first do the type-change for Real/Complex. Can we open an issue for further retractions/exponential map/metrics for fixed rank so I can keep that in mind and finish that after the type changes? Then I would just finish a proper code coverage here and merge then.

mateuszbaran:
Sure, additional functionality can go into another PR.

kellertuer:
I will collect the suggestions in an Issue after this is merged, then.

kellertuer:
So I still have to increase code coverage; any idea why the inital codecov might have been gone? Can that be fixed to have a code coverage change back?

kellertuer:
Well, well. The idea of fixing this within this year might have to be posponed. My idea of “shortly” fixing the remaining tests this morning got directly stuck in the first test, since I might have misunderstood the projection operation and I can't get it to work for now.

sethaxen:
RE codecov, I didn't see how it got messed up, but I would just use it as a helper to identify missing tests rather than as a hard rule that coverage goes up. On one of my projects, codecov never shows certain code as being tested, even though it provably is. 🤷‍♂

kellertuer:
While I just fixed something (that I might have misread before), retractions are now tested also when exp/log are not (previously they were inside). However that breaks Grassmann retraction tests for now, so I am left to fix retractions for Grassmann/fixesRank/Stiefel and projection for fixedRank.

kellertuer:
[ ] Power manifold retractions fail (a little hard to debug, since they also have randomness within)
[ ] Stiefel project onto tangent does not yet work nor is it tested.

I till update this list to keep track of remaining things as soon as the codecoverage has done its job.

kellertuer:
All projections work, I adapted the new Literature/Footnote approach. So if the code coverage is fine I would like to merge this.

kellertuer:
Well well well, upgrading `ManifoldsBase` caused side effects that I did not have in mind. I implemented a few mor `injectivity_radius` – that I could find – and adapted the tests accordingly.

Issue #48, opened 2019-11-24
Simplify `MetricManifold` to not use `SimpleTraits`


kellertuer:
While working on a first example on a manifold with two different metrics in #27 I noticed that quite a few things are nice, with metric manifold but with the `hasMetric` trait it has a mayor drawbag:

All functions that currently have a fallback from `MetricManifold{M,G}` for the `hasMetric{M,G}` (let's call that default metric for now) trait have the large disadvantage, that _every_ new and non default metric has to implement _every_ function that has a fallback implemented. Also the fallbacks are all quite longish.

I propose the following:
* Remove the `hasMetric` trait and all `hasMetric` fallbacks
* Introduce conversions to the non `MetricManifold` case for the default metric and all metrics that should fall back to default implementations for certain functions
* these conversions do not “trigger” if one explicitly overwrites for example `exp(MetricManifold{M,G},x,v)` for a new metric `G` (but would trigger for `manifold_dimension` if that's not explicitly implemented for just mentioned type
* keep the `!hasMetric` implementations of for example `exp!`, `flat!`, and `sharp!` in `MatricManifold`, since __if__ you implement a new metric, you'd implement these anyways ( or let them fall back)

I will try this today on #27, but of course would love to hear other opinions on that, too.

kellertuer:
And a small addendum: why are `flat` and `sharp` written for `FVector`s of two types an not for `TVector` and `CoTVector` ? Do flat and sharp need typing there anyways, since it's the defaults for `MetricManifold`s?

mateuszbaran:
IIRC it was argued that along with the introduction of `ManifoldsBase.jl`, metric manifolds should be moved to THTT. I think this should make it possible to have fallbacks that don't have to be implemented for every new metric. Essentially I don't see the point of having the `!hasMetric` methods in the first place, they could just be the default fallback.

> And a small addendum: why are `flat` and `sharp` written for `FVector`s of two types an not for `TVector` and `CoTVector` ? Do flat and sharp need typing there anyways, since it's the defaults for `MetricManifold`s?

A few reasons:
1) Semantics of `TVector` and `CoTVector` suggest that they will be checked using but `FVector` leaves this unspecified.
2) `flat` and `sharp` can be defined for other vector bundles.
3) Not wrapping arguments in anything would make it too easy to put them in the wrong order when calling.

kellertuer:
[...]

PR #52, opened 2019-11-28
A decorator for Metrics that works without SimpleTraits

kellertuer:
As discussed in #48, I implemented the current decorator anew to cover the three cases we use more rigorously

* A manifold without a metric can still just be implemented
* A manifold with two different metrics that have closed forms can be easily implemented
* Adding a second metric to the case from the first point can easily be done, even with both being explicit `MetricManifolds` using the `is_default_metric`
* A metric with a `g` local tensor can be added to a manifold, with one restriction: It is not useful to set that one to default, since that would deactivate using the tensor again. Default should only be used for these implicit ones with closed formulae

The documentation is also already done, so I look forward to having feedback on this new model.


docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
 # Metric manifold
 
+A Riemannian manifold always consists of a [topological manifold](https://en.wikipedia.org/wiki/Topological_manifold) together with a smoothly varying metric $g$.
+
+However, often there is a implicitly assumed (daefault) metric, like the usual
```
mateuszbaran:
```suggestion
However, often there is an implicitly assumed (default) metric, like the usual
```

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+inner product on [`Euclidean`](@ref) space. This decorator takes this into
+account. Hence it is not necessary to acts as a (as transparent as possible)
+decorator, see [`is_decorator_manifold`](@ref) for details.
```
mateuszbaran: I don't really understand this sentence ("Hence it (...) for details"). Could you rephrase it?

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+
+This manifold servers two purposes: To implement different metrics (i.g. in closed
```
mateuszbaran:
```suggestion
This manifold decorator serves two purposes: to implement different metrics (e.g. in closed
```

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+form) for one [`Manifold`](@ref) and to provide a way to compute geodesics
+on manifolds, where this [`Metric`](@ref) does not yield closed formula.
```
mateuszbaran: I think eventually all other functions would also have a default implementation based on the metric, right? Like `exp`, `log`, parallel transport etc.

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+
+## Implement Different Metrics on the same Manifold
+
+In order to distinguish different metrics on one manifold, one can introduce
+two [`Metric`](@ref)s and use this type to dispatch on the metric, see 
+[`SymmetricPositiveDefinite`](@ref). To avoid overhead, one [`Metric`](@ref)
+can then be marked as being the default, i.e. the one that is used, when no
+[`MetricManifold`](@ref) is used. This avoids reimplementation of a first
+existing metric, access for the default metric functions (without an [`MetricManifold`](@ref)
+overhead) as well as the transparent fallback of the corresponding [`MetricManifold`](@ref)
```
mateuszbaran: Just to clarify, `MetricManifold` used in this way should not cause any runtime overhead.

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+account. Hence it is not necessary to acts as a (as transparent as possible)
+decorator, see [`is_decorator_manifold`](@ref) for details.
+
+This manifold servers two purposes: To implement different metrics (i.g. in closed
+form) for one [`Manifold`](@ref) and to provide a way to compute geodesics
+on manifolds, where this [`Metric`](@ref) does not yield closed formula. 
+
+Let's first look at the provided types.
+
+## types
```
mateuszbaran:
```suggestion
## Types
```

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+In order to distinguish different metrics on one manifold, one can introduce
+two [`Metric`](@ref)s and use this type to dispatch on the metric, see 
+[`SymmetricPositiveDefinite`](@ref). To avoid overhead, one [`Metric`](@ref)
+can then be marked as being the default, i.e. the one that is used, when no
+[`MetricManifold`](@ref) is used. This avoids reimplementation of a first
+existing metric, access for the default metric functions (without an [`MetricManifold`](@ref)
+overhead) as well as the transparent fallback of the corresponding [`MetricManifold`](@ref)
+with default metric to the (classical) implementations.
+
+## Implementaion of Metrics
```
mateuszbaran:
```suggestion
## Implementation of Metrics
```

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+## Implement Different Metrics on the same Manifold
+
+In order to distinguish different metrics on one manifold, one can introduce
+two [`Metric`](@ref)s and use this type to dispatch on the metric, see 
+[`SymmetricPositiveDefinite`](@ref). To avoid overhead, one [`Metric`](@ref)
+can then be marked as being the default, i.e. the one that is used, when no
+[`MetricManifold`](@ref) is used. This avoids reimplementation of a first
+existing metric, access for the default metric functions (without an [`MetricManifold`](@ref)
+overhead) as well as the transparent fallback of the corresponding [`MetricManifold`](@ref)
+with default metric to the (classical) implementations.
```
mateuszbaran: Maybe "undecorated" would be clearer than "classical"?

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+Order = [:type]
+```
+
+## Implement Different Metrics on the same Manifold
+
+In order to distinguish different metrics on one manifold, one can introduce
+two [`Metric`](@ref)s and use this type to dispatch on the metric, see 
+[`SymmetricPositiveDefinite`](@ref). To avoid overhead, one [`Metric`](@ref)
+can then be marked as being the default, i.e. the one that is used, when no
+[`MetricManifold`](@ref) is used. This avoids reimplementation of a first
```
mateuszbaran:
```suggestion
[`MetricManifold`](@ref) is used. This avoids reimplementation of the first
```

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,46 @@
...
+existing metric, access for the default metric functions (without an [`MetricManifold`](@ref)
```
mateuszbaran:
```suggestion
existing metric, access to the default metric functions (without an [`MetricManifold`](@ref)
```

Perhaps also "default metric functions" should be replaced by a more precise phrase, maybe "metric-dependent functions of the undecorated manifold"?

src/Metric.jl
```diff
@@ -27,23 +27,55 @@ abstract type LorentzMetric <: Metric end
...
+This means that any occurence of
+[`MetricManifold`](@ref)`(M,G)` where `is_default_metric(M,G) = Val{true}`
+falls back to just be called with `M` such that the [`Manifold`](@ref) `M`
+implicitly has this metric, for example if this was the first one implemented
+or is the one most commonly assumed to be used.
+"""
+is_default_metric(M::MMT) where {MMT <: MetricManifold} = is_default_metric(base_manifold(M),metric(M))
+# this automatically undecorates
+convert(::Type{MT},M::MetricManifold{<:MT,GT}) where {MT <: Manifold,GT} = base_manifold(M)
+# this should austomatically decorate at least for simple cases
```
mateuszbaran:
```suggestion
# this should automatically decorate at least for simple cases
```

mateuszbaran:
Probably the SPD manifold needs some changes since coverage says that some methods are not called, I guess the new metric manifold implementation makes some methods redundant?

docs/src/manifolds/metric.md
kellertuer: Well I hope so, yes, for now there is only exp implemented based on the metric tensor. I have not yet read how to do the remaining ones numerically – but also since my goal is more to implement different metrics

docs/src/manifolds/metric.md
kellertuer: Yes, it should not cause overhead. I'll add that.

kellertuer:
Interestingly I do a for loop over metric manifolds and (added a print temporarily) it seems to take always the first of the for loop entries, that's why SPD never checks LogCholesky. Have to look closer into that. However, Metrics should be covered better now.

kellertuer:
Just had a moment of time and finished the final remarks as well as took some time to increase test coverage (still a special case in rotations missing for example), so despite (slowly) increasing test coverage(even more); this is ready to merge.

kellertuer:
I'll wait a few days for everyone to check, but from implementations and testing as well as documentation I think this is finished.

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,51 @@
 # Metric manifold
 
+A Riemannian manifold always consists of a [topological manifold](https://en.wikipedia.org/wiki/Topological_manifold) together with a smoothly varying metric $g$.
+
+However, often there is an implicitly assumed (default) metric, like the usual
+inner product on [`Euclidean`](@ref) space. This decorator takes this into
+account. It is not necessary to use this decorator, if you imlement just one (or
+the first) metric. If you later introduce a second, the old (first) one can be
+used with or without a metricx. The decorator acts transparent in that sense;
```
mateuszbaran:
```suggestion
used with or without a metric. The decorator acts transparent in that sense;
```

src/Metric.jl
```diff
@@ -27,23 +27,54 @@ abstract type LorentzMetric <: Metric end
...
+[`MetricManifold`](@ref)(M,G) where `is_default_metric(M,G) = Val{true}`
+falls back to just be called with `M` such that the [`Manifold`](@ref) `M`
+implicitly has this metric, for example if this was the first one implemented
+or is the one most commonly assumed to be used.
+"""
+is_default_metric(M::Manifold,G::Metric) = Val(false)
+"""
+    is_default_metric(MM)
+
+indicate, whether the [`Metric`](@ref) `G` is the default metric for
```
mateuszbaran: I think in this docstring `MM.G` and `MM.M` should be used instead of just `G` and `M` (because the function only takes one argument called `MM`).

mateuszbaran:
Apart from these two docstring issues I think it can be merged 👍

sethaxen:
It's a holiday weekend here. I should be able to review by Tuesday.

src/Metric.jl
kellertuer: Thanks for carefully reading even the docstrings; you're right that makes the statements far clearer.

kellertuer:
Nevermind,if it gets Tuesday. I just had a (long) train ride yesterday and one (not so long) today (and one tomorrow...) so I finished this from my side (when I got too lazy to write the remaining few tests). Feel free to take your time to review this; but I would like to make sure, that you're ideas are still preserved as you intended them :)

src/Metric.jl
```diff
@@ -27,23 +27,54 @@ abstract type LorentzMetric <: Metric end
...
-[`local_metric`](@ref).
+Equip a [`Manifold`](@ref) explicitly with a [`Metric`](@ref) `G`. 
 
-# Constructor
+For a Metric Manifold, by default, assumes, that you implement the linear form
+from [`local_metric`](@ref) in order to evaluate the exponential map.
+
+If the corresponding [`Metric`](@ref) `G` yields closed form formulae for e.g.
+the exponential map and this is implemented directly (without solving the ode),
+you of course can still implement that directly.
```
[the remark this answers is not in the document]
kellertuer: Thanks for the grammar correction, I'll remove the "can" after "of course" then.

docs/src/manifolds/metric.md
```diff
@@ -1,7 +1,51 @@
 # Metric manifold
 
+A Riemannian manifold always consists of a [topological manifold](https://en.wikipedia.org/wiki/Topological_manifold) together with a smoothly varying metric $g$.
+
+However, often there is an implicitly assumed (default) metric, like the usual
+inner product on [`Euclidean`](@ref) space. This decorator takes this into
+account. It is not necessary to use this decorator, if you imlement just one (or
+the first) metric. If you later introduce a second, the old (first) one can be
```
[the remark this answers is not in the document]
kellertuer: I avoided the second "one" by rephrasing the sentence.

PR #55, opened 2019-12-04
[WIP] Fixes for SPD manifold and MMatrix

mateuszbaran:
I've made some changes trying to get SPD manifolds working with `MMatrix`. Let's see how this goes in CI.

Ref. https://github.com/JuliaArrays/StaticArrays.jl/issues/696 and https://github.com/JuliaArrays/StaticArrays.jl/issues/697


src/SymmetricPositiveDefinite.jl
```diff
@@ -215,7 +215,7 @@ denotes the lower triangular matrix with the diagonal multiplied by $\frac{1}{2}
 function exp!(M::MetricManifold{SymmetricPositiveDefinite{N},LogCholeskyMetric}, y, x, v) where N
     (l,w) = spd_to_cholesky(x,v)
     z = exp!(CholeskySpace{N}(),y,l,w)
-    mul!(y,z,z')
+    y .= z*z'
```
kellertuer: Are you sure this is necessary? I thought the old version was more memory efficient?

test/symmetric_positive_definite.jl
```diff
@@ -8,14 +8,15 @@ types = [ Matrix{Float32},
...
 for M in [M1, M2, M3]
     @testset "$(typeof(M))" begin
         for T in types
-            if M == M3 && T <: MMatrix
-                #TODO fix the issue that causes this failure
-                continue
+            exp_log_atol_multiplier = 8.0
+            if T <: MMatrix{3,3,Float64}
+                # eigendecomposition of 3x3 SPD matrices from StaticArrays is not very accurate
+                exp_log_atol_multiplier = 5.0e7
```
kellertuer: `e7`? sounds like _really_ inaccurate.

test/utils.jl
```diff
@@ -104,10 +104,10 @@ function test_manifold(M::Manifold, pts::AbstractVector;
         retract!(M, new_pt, pts[1], tv1)
         @test is_manifold_point(M, new_pt)
         for x ∈ pts
-            @test isapprox(M, zero_tangent_vector(M, x), log(M, x, x); atol = eps(eltype(x)) * exp_log_atol_multiplier)
-            @test isapprox(M, zero_tangent_vector(M, x), inverse_retract(M, x, x); atol = eps(eltype(x)) * exp_log_atol_multiplier)
+            @test isapprox(M, x, zero_tangent_vector(M, x), log(M, x, x); atol = eps(eltype(x)) * exp_log_atol_multiplier)
```
kellertuer: good that you noticed, indeed `x` should be added here in order to compare semantically tangent vectors and not points on `M`.

src/SymmetricPositiveDefinite.jl
dkarrasch: In-place multiplication is `mul!`!
```julia
using LinearAlgebra, BenchmarkTools
A = randn(5,5)
S = similar(A)
fun!(z, y) = z .= y*y'
@btime fun!($S, $A);
@btime mul!($S, $A, $A');
```
The first one (`fun!`) is _not_ allocation-free.

mateuszbaran: Well, `mul!(y,z,z')` gives wrong answer for `MMatrix`. I've reported it here: https://github.com/JuliaArrays/StaticArrays.jl/issues/697 . In this code `z` and `y` is the same matrix, `LinearAlgebra` handles it correctly but `StaticArrays` doesn't. Since one additional allocation won't slow this down too much, having a bit slower but correct implementation seems better until `StaticArrays` is fixed.

test/symmetric_positive_definite.jl
mateuszbaran: I guess `StaticArray` chooses fast over accurate but that's indeed quite extreme. The algorithm is good enough for `Float32` but for `Float64` one would expect something more accurate.

mateuszbaran:
> For example one of my optimization problems in on the product manifold `P(3)^{112x112x50}`, i.e. a very large product manifold of `3x3` SPDs, there it might really be beneficial to use `MMatrix` (the Matlab version on just one vertical slice is for example

That also looks like a great application for `PowerManifold` and `HybridArray`, it uses the efficient linear algebra from `StaticArrays`.

src/SymmetricPositiveDefinite.jl
dkarrasch: In `mul!`, the storage variable must not be aliased with any of the factors, as the docstring clearly states. There can't be an in-place multiplication "while you're multiplying". So your issue over at `StaticArrays.jl` is not an issue: there is simply no guarantee of what happens when `C` is aliased with either `A` or `B`. What happens if the first argument is different from the second and third for `MMatrix`, as in `mul!(y, z, z')`?

mateuszbaran: I didn't read that docstring and assumed that the `mul!` here is correct. It looks like `mul!` only accidentally works on `Array` here. The difference between `mul!` for `MMatrix` and `Matrix` only happens when there is aliasing. Here `y` and `z` are supposed to be the same matrix, it looks like introducing `z` is just misleading.

kellertuer: Oh. Indeed. With the line 217, `z` is the same as `y`, we should avoid.

dkarrasch: Yes, as suggested in https://github.com/JuliaNLSolvers/Manifolds.jl/pull/27#discussion_r349890992, for exactly that reason.

kellertuer: Oh! Sorry, I must have misread that (though I am not sure how that happened) but of course an `exp` on Cholesky helpes a lot in this case. Sorry for having missed that. One should definitely incorporate that comment here, then.

mateuszbaran: OK, so I'll change it to a non-mutating `exp`.

mateuszbaran: I've changed this so I think this PR can be merged once CI passes.

dkarrasch: Now you should write `mul!(y, z, z')` to have a truly non-allocating multiplication.

mateuszbaran: Of course, I forgot about it. Thanks!

mateuszbaran: Except it it doesn't work for `MMatrix` :/

dkarrasch: What? What's the type of `z`? Where does `lmul!` come from?

dkarrasch: Hm, I see. I'll double-check.

dkarrasch: This is really tedious... Seems like there is no triangular multiplication routines for `MMatrix`.

dkarrasch: Okay, it seems there is no in-place `mul!` for triangular `[S/M]Matrix`, only `*`, see https://github.com/JuliaArrays/StaticArrays.jl/blob/master/src/triangular.jl. So the line above should be okay for those types. It's suboptimal for `Base` arrays (and larger static matrices that fall back to BLAS), but maybe these are not much used in this context.

mateuszbaran: OK,, we can improve this another time.

PR #56, opened 2019-12-06
Grassmannian

kellertuer:
While waiting for the fix in Stiefel, I started the next manifold – Grassmannians.


src/Grassmann.jl
```diff
@@ -0,0 +1,263 @@
+using LinearAlgebra: svd, qr, Diagonal, det
+import LinearAlgebra: norm
+@doc doc"""
+    Grassmann{n,k} <: Manifold
```
mateuszbaran: This should mention `T`.

src/Grassmann.jl
```diff
@@ -0,0 +1,263 @@
...
+
+The Grassmann manifold $\operatorname{Gr}(n,k)$ consists of all $k$-dimensional
+subspaces of $\mathbb F^n$, where $\mathbb F \in \{\mathbb R, \mathbb C\}$ is
+either the real- (or complex-) valued subspaced of the $n$-dimensional (complex)
+space.
```
mateuszbaran: This description is slightly unclear. $\mathbb{C}^n$ is $2n$-dimensional, right? Could you reflect this in the description? Also, `subspaced` should be replaced with `subspaces`.

src/Grassmann.jl
```diff
@@ -0,0 +1,263 @@
...
+
+The manifold can be represented as
+````math
+\operatorname{Gr}(n,k) \coloneqq \bigl\{ \operatorname{span}(x)
+: x \in \mathbb F^{n\times k}, \bar{x}^\mathrm{T}x = I_k\},
+````
+where $\bar\cdot$ denotes the complex conjugate and $I_k$ is the $k\times k$
```
mateuszbaran: I haven't worked with Grassman manifolds before, is this actually an isometric embedding? It doesn't look like it but `inner` is just `dot`?

src/Grassmann.jl
```diff
@@ -0,0 +1,263 @@
...
+````
+where cosine and sine are applied element wise to the diagonal entries of $S$,
+as
+````math
+y = Q,
+````
+where $Q$ of the QR decomposition $z=QR$ of $z$. This last step is for numerical
+stability reasons.
+"""
+function exp(M::Grassmann{N,K,T},y, x, v) where {N,K,T}
```
mateuszbaran: That's `exp!`, right?

src/Grassmann.jl
```diff
@@ -0,0 +1,263 @@
...
+and the $\operatorname{atan}$ is meant elementwise.
+"""
+function log!(M::Grassmann{N,K,T}, v, x, y) where {N,K,T}
+    z = y'*x
+  	if det(z)≠0
+        d = svd( z\(y' - z*x'), full = false)
+        v .= d.V * atan.(Diagonal(d.S)) * (d.U')
+        return v   
+  	else
+   		throw( ErrorException("The points $x and $y are antipodal, thus these input parameters are invalid.") )
```
mateuszbaran: Maybe `DomainError` would be more appropriate?

src/Manifolds.jl
```diff
@@ -45,6 +45,7 @@ import ManifoldsBase: base_manifold,
     inverse_retract!,
     isapprox,
     is_manifold_point,
+    injectivity_radius,
```
mateuszbaran: Why is this in this place? There should be `injectivity_radius` a few lines above.

src/Manifolds.jl
```diff
@@ -59,7 +60,6 @@ import ManifoldsBase: base_manifold,
     representation_size,
     retract,
     retract!,
-    check_tangent_vector,
```
mateuszbaran: Why did you remove this?

src/Grassmann.jl
kellertuer: Yes, thanks for noticing.

src/Manifolds.jl
kellertuer: We can also add it again, I was thinking, the `is_tangent_vector` is now the interface (i.e. exported) and this one is internal, i.e. you can still call it with `Manifolds.check_tangent_vector` but the recommended way to check is the first mentioned method.

src/Grassmann.jl
kellertuer: Grassmann is a quotient of Stiefel, so you can use the inner product and it is a sub manifold of R^n (C^n) and inherits the inner product. Due to the quotient I am not sure about the isometry.

src/Manifolds.jl
mateuszbaran: Right, it makes more sense to not export it.

src/Manifolds.jl
kellertuer: Interesting, I must have missed that – we should maybe find a way to make this more structured (also the exports).

src/Grassmann.jl
mateuszbaran: AFAIK that's only true when the quotient has a global isometric section and we use it. In general we need horizontal-vertical decomposition of tangent vectors (see Section 1.5 of https://pdfs.semanticscholar.org/aad1/de01cdea203a3bb25db0cd0e45a611db23fe.pdf for example).

mateuszbaran: Unless you are careful enough to only store the horizontal part at all times?

src/Manifolds.jl
mateuszbaran: It's mostly in alphabetical order but probably it should be in alphabetical order *exactly*.

kellertuer:
Oh, after an hour of trying to get at least some test running I might also have merged the wrong branch, but I will only continue in a few days to try to get the tests running, something is still very strange, `exp` seems to produce only zero matrices for now, which makes everything fail tremendously.

src/Grassmann.jl
```diff
@@ -0,0 +1,263 @@
...
+y = Q,
+````
+where $Q$ of the QR decomposition $z=QR$ of $z$. This last step is for numerical
+stability reasons.
+"""
+function exp!(M::Grassmann{N,K,T},y, x, v) where {N,K,T}
+    d = svd(v)
+    z = x * d.V * cos.(Diagonal(d.S)) * (d.V)' + (d.U) * sin.(Diagonal(d.S)) * (d.V)'
+    # reorthonormalize 
+    y = qr(z).Q
```
mateuszbaran: This should probably be `copyto!`?

kellertuer:
I worked a little bit to at least habe exp/log defined, but since the subspaces are represented by a basis and that is not unique, the tests are still failing for exp/log and the retractions.

kellertuer:
I still have to find a more stable way to compute the distance; until now it's only accurate to 10^-7 for Float64.

mateuszbaran:
You don't have `isapprox` for tangent vectors on `Grassmanian`. I don't really understand the problem here. Could you expand a bit about the way horizontal-vertical separation of the tangent spaces of the quotient manifold is handled?

kellertuer:
Oh, I#ll check for that function tomorrow.

It's actually the laziest way of handling: none. You can have different matrices representing the same subspace. The only thing that is not allowed then is, to infere from the matrix representations x!=y that they represent different points (you'd have to check `distance(M,x,y)` instead)

As far as I recon this does not happen for tangents (within log, which restricts to some subspace), so that's why inner is still fine being inherited from the embedding and hence isapprox should just be Frobenius of the difference.

mateuszbaran:
> As far as I recon this does not happen for tangents (within log, which restricts to some subspace)

OK, that's a rather important assumption though. `inner` should work (or at least throw an error) even if you have two tangent vectors at the same point where said point is represented by two different matrices.

> The problem is, that the quotient yield numerical instabilities in said Frobenius norm for the Grasmann points (especially if you have two matrices being very different but representing the same subspace).

Right, this definitely sounds like it might be a problem.

src/Manifolds.jl
kellertuer: It should be on the main branch in the mean time, since I did that on another PR.

kellertuer:
On a train one has a little bit of time and not that much internet (in Germany), so I did some tests and it seems, the exponential map is the culprit and it's numerically instable. It get's worse if one leaves our QR (as written in the code of manopt).

kellertuer:
Exp and log seem to map to different representants so disabling that makes the tests run, but the power/product tests now fail after merging the master and I can't see why, yet.

kellertuer:
@mateuszbaran introducing the exports_log_atol` breaks quite a lot of tests in power/product and vector_fiber. Do you have an idea, what I did wrong? I don't see where this effect has come from.

mateuszbaran:
> @mateuszbaran introducing the exports_log_atol` breaks quite a lot of tests in power/product and vector_fiber. Do you have an idea, what I did wrong? I don't see where this effect has come from.

Specifying `atol` makes `rtol` default to 0 in `isapprox` so that might be an issue. You can try setting `rtol` to the default (`sqrt(eps)`) and see if this helps.

kellertuer:
> Specifying `atol` makes `rtol` default to 0 in `isapprox` so that might be an issue. You can try setting `rtol` to the default (`sqrt(eps)`) and see if this helps.

Could/Should we pass that out, too and set it by default? I will try something tomorrow then, but yes that might be the issue.

mateuszbaran:
We could pass that, at least for tests that would be failing otherwise.

kellertuer:
I will try to pass down rtol and atol for the exp/log then as a test and we can transfer further ones if required. I will follow the same rules as the original r/atols (not just using a factor) for this test.

kellertuer:
All old tests are running again, from Grassmann exp/log is currently disabled, I really don't see where my mistake is (formulae are exactly the same as in both pymanopt ([here](https://github.com/pymanopt/pymanopt/blob/master/pymanopt/manifolds/grassmann.py#L129)) and manopt (line 176-223 [here](https://www.manopt.org/reference/manopt/manifolds/grassmann/grassmannfactory.html)), though I haven't found whether they test the functions. In my short tests (example from `test/grassmann.jl`) exp seems to be not the inverse of log, not even in the sense that for `v2 = log(M,x,exp(M,x,v))` we would get `inner(M,x,v,v2)=0`. Both `v` and `v2` are at least valid tangents at x, though.

src/Grassmann.jl
kellertuer: No I am not that careful but the formulae are invariant to that part, so while it might not be equal element wise. For example any element from the horizontal part of the zero tangent vector still yields a norm of zero.

src/Grassmann.jl
kellertuer: I rephrased this.

mateuszbaran:
Could you give me a day or two to make a final review?

chakravala:
Have you considered using actual [Grassmann](https://github.com/chakravala/Grassmann.jl) algebra instead of matrix algebra?

It can achieve the same things, but without allocating any matrices.

kellertuer:
Sorry, I am not experienced enough in providing that, especially since I have more matrix-like applications in optimization in mind, where I will require the manifold this PR proposes. If you have time to provide such an interface here, we are happy to have such an interface and hope that `ManifoldsBase.jl` is flexible enough to incorporate that here, too.

src/Grassmann.jl
mateuszbaran: I think I see what happens. `log` is constructed carefully enough to make `dot` the correct inner product. Tangent vectors are tied to a particular representation of the point but no one actually expects anything more. I did my quotient manifolds a bit differently but this might be a better approach, at least when it can be used.

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
+using LinearAlgebra: svd, qr, diag, Diagonal, det
+import LinearAlgebra: norm
+@doc doc"""
+    Grassmann{n,k,T} <: Manifold
+
+The Grassmann manifold $\operatorname{Gr}(n,k)$ consists of all subspaces span
```
mateuszbaran:
```suggestion
The Grassmann manifold $\operatorname{Gr}(n,k)$ consists of all subspaces spanned
```

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+either the real- (or complex-) valued vectors. This yields all $k$-dimensional
+subspaces of $\mathbb R^n$ for the real-valued case and all $2k$-dimensional
+subspaces of $\mathbb C^n$ for the second.
+
+The manifold can be represented as
+````math
+\operatorname{Gr}(n,k) \coloneqq \bigl\{ \operatorname{span}(x)
+: x \in \mathbb F^{n\times k}, \bar{x}^\mathrm{T}x = I_k\},
+````
+where ${\bar\cdot}^{\mathrm{T}}$ denotes the complex conjugate transposed and
```
mateuszbaran:
```suggestion
where ${\bar\cdot}^{\mathrm{T}}$ denotes the complex conjugate transpose and
```

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+i.e the Grassmann manifold is the quotient of the [`Stiefel`](@ref) manifold and
+the orthogonal group $\operatorname{O}(k)$ of orthogonal $k\times k$ matrices.
+
+The tangent space at a point (subspace) $x$ is given by
+````math
+T_x\mathrm{Gr}(n,k) = \bigl\{
+ v \in \mathbb{F}^{n\times k} : 
+ {\bar v}^{\mathrm{T}}x + {\bar x}^{\mathrm{T}}v = 0_{k} \bigr\},
+````
+where $0_{k}$ denotes the $k\times k$ zero matrix.
```
mateuszbaran: It might be worth pointing out that different representations of the same point on Grassmanian lead to different tangent spaces.

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+    end
+    c = x'*x
+    if !isapprox(c, one(c); kwargs...)
+        return DomainError(norm(c-one(c)),
+            "The point $(x) does not lie on the Grassmann manifold of dimension ($(N),$(K)), because x'x is not the unit matrix.")
+    end
+end
+function check_tangent_vector(G::Grassmann{N,K,T},x,v; kwargs...) where {N,K,T}
+    t = check_manifold_point(G,x)
+    if (t != nothing)
```
mateuszbaran:
```suggestion
    if (t !== nothing)
```

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+    end
+    if (T <: Real) && !(eltype(v) <: Real)
+        return DomainError(eltype(v),
+            "The matrix $(v) is not a real-valued matrix, so it can not be a tangent vector to the Grassmann manifold of dimension ($(N),$(K)).")
+    end
+    if (T <: Complex) && !(eltype(v) <: Real) && !(eltype(v) <: Complex)
+        return DomainError(eltype(v),
+            "The matrix $(v) is neiter real- nor complex-valued matrix, so it can not bea tangent vector to the complex Grassmann manifold of dimension ($(N),$(K)).")
+    end
+    if any( size(v) != representation_size(G) )
```
mateuszbaran: Do we need the `any` here?

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+${\bar x}^\mathrm{T}y$, where ${\bar\cdot}^{\mathrm{T}}$ denotes the complex
+conjugate transposed. Then the distance is given by
+````math
+d_{\mathrm{GR}(n,k)}(x,y) = \operatorname{norm}(\operatorname{Re}(b)).
+````
+where
+
+$b_{i}=\begin{cases} 0 & \text{if} \; S_i≧1 \\ \operatorname{acos}(S_i) & \, \text{if} \; S_i<1 \end{cases}.$
+"""
+function distance(M::Grassmann{N,K,T}, x, y) where {N,K,T}
```
mateuszbaran: There is something wrong with indentation in this function (there are some tabs instead of spaces).

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+````
+"""
+function exp!(M::Grassmann{N,K,T},y, x, v) where {N,K,T}
+    if norm(M,x,v) ≈ 0
+        return (y .= x)
+    end
+    d = svd(v)
+    z =  x * d.V * Diagonal(cos.(d.S)) * d.Vt + d.U * Diagonal(sin.(d.S)) * d.Vt
+    # reorthonormalize
+    y = copyto!(y, Array(qr(z).Q) )
```
mateuszbaran:
```suggestion
    copyto!(y, Array(qr(z).Q) )
```
We don't need the assignment.

src/Grassmann.jl
```diff
@@ -0,0 +1,275 @@
...
+USV = ({\bar y}^\mathrm{T}x)^{-1} ( {\bar y}^\mathrm{T} - {\bar y}^\mathrm{T}x{\bar x}^\mathrm{T} ).
+````
+In this formula the $\operatorname{atan}$ is meant elementwise.
+"""
+function log!(M::Grassmann{N,K,T}, v, x, y) where {N,K,T}
+    z = y'*x
+    At = y' - z*x'
+    Bt = z\At
+    d = svd(Bt')
+    return (v .= d.U * Diagonal(atan.(d.S)) * d.Vt)
```
mateuszbaran: I think it's safer to return `v` on a separate line (this might return the RHS).

test/utils.jl
```diff
@@ -88,24 +96,27 @@ function test_manifold(M::Manifold, pts::AbstractVector;
...
+
+    tv1 = inverse_retract(M,pts[1],pts[2],default_inverse_retraction_method)
+    test_log_yields_tangent && @testset "is_tangent_vector" begin
         @test is_tangent_vector(M, pts[1], tv1; atol = eps(eltype(pts[1])))
         @test check_tangent_vector(M, pts[1], tv1; atol = eps(eltype(pts[1]))) === nothing
         @test check_tangent_vector(M, pts[1], tv1; atol = eps(eltype(pts[1]))) === nothing
     end
 
-    @testset "log/exp tests" begin
+    test_exp_log &&@testset "log/exp tests" begin
```
mateuszbaran:
```suggestion
    test_exp_log && @testset "log/exp tests" begin
```

src/Grassmann.jl
mateuszbaran: Similarly in point checking.

src/Grassmann.jl
kellertuer: well, they lead to different _representations_ of one and the same tangent space – but yes that could be added, you're right.

mateuszbaran: Yes, that's what I mean -- different representations.

Issue #57, opened 2019-12-08
Implement more generic distributions on Manifolds


sethaxen:
This issue narrows the discussion in https://github.com/JuliaNLSolvers/Manifolds.jl/issues/3 to implementing distributions on Manifolds. It also continues some discussion from https://discourse.julialang.org/t/rfc-taking-directional-orientational-statistics-seriously/31951 and on the Slack channel.

There are a few generic distributions on manifolds that we'll want, with examples in directional statistics noted where available and not obvious:
- [ ] `Dirac`: Dirac measure, has 0 density everywhere but at a fixed point.
- [ ] `Mixture`: alias of `Distributions.MixtureModel` typed on our supports and variates
- [ ] `Uniform`: similar to `Distributions.Uniform` but defined on manifolds.
- [x] `Product`: product distribution analogous to `Distributions.Product` on `ProductManifold`
- [x] `Power`: product distribution on sub-manifolds of the power-manifold
   - [ ] Probably need something more general as well for cases where one wants something like `Product` but on a `PowerManifold`. Useful for cases like Matrix von Mises-Fisher on the Steifel manifold which can be composed from a product of von Mises-Fisher distributions with different means and concentrations on the spherical submanifolds
- [x] `Projected`: distribution on ambient space projected to submanifold
   - e.g. Projected Normal
- [ ] `Restricted`, `Conditioned`, or `Intersection`: distribution on manifold resulting from intersection of a distribution on ambient space to the manifold
  - e.g. Fisher-Bingham which has as special cases von Mises-Fisher and many other directional distributions
- [ ] `Retracted`: distribution on tangent space at point retracted to manifold
   - e.g. tangential normal (aka Log-normal, but that's confusing)
- [ ] `Riemannian`??: Similar to `Retracted` but with adjustment for curvature (name chosen because "Riemannian normal" is the only well-studied example of this I know).
   - e.g. "spherical normal"
- [ ] `IsotropicDiffusion` or `Brownian`: Distribution resulting from brownian motion for a fixed time aka solution to the heat equation on the manifold
   - e.g. Spherical brownian motion distribution

There are plenty of Lie group specific ones that we can address when we've finished implementing Lie groups, i.e distributions resulting from quotient, actions, etc.

[...]

PR #58, opened 2019-12-08
Introduce Median and Mean

kellertuer:
As a first step to #57 this implements median and mean on manifolds. For variance (see for example [1607.08481](https://arxiv.org/pdf/1607.08481.pdf) we still need a good way to write matrices in tangent spaces (there it is the one with respect to an ONB, see end of Sec 3).

The thing still missing here (follows the next days) is testing both algorithms.


kellertuer:
I think we only should do the variance, when we have decided on more structure concerning distributions (i.e. how to model a tuple `(x,Sigma)` of mean/variance. So this one is ready to merge after some reviews.

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+    stop_atol=10^-7,
+    stop_iter=100
+    ) where {T}
+    iter = 0
+    y=x[1]
+    yOld = y
+    while ( ( iter==0 || distance(M,y,yOld) > stop_atol ) && (iter < stop_iter) )
+        iter += 1
+        yOld = y
+        y = exp(M,yOld,  1/2*sum( weights.*log.(Ref(M), Ref(yOld),x) )  )
```
mateuszbaran: I'd suggest using mutating `exp` and `log` here. Otherwise there will be a lot of allocations. Or maybe we can have two variants in case someone has some reasons for avoiding in-place operations (like `mean(::Manifold, x::AbstractVector, use_mutation::Val{true}`)

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+check the [`Manopt.jl`](https://manoptjl.org) package and use a solver therefrom.
+
+The algorithm is further described in 
+> Afsari, B; Tron, R.; Vidal, R.: On the Convergence of Gradient
+> Descent for Finding the Riemannian Center of Mass,
+> SIAM Journal on Control and Optimization (2013), 51(3), pp. 2230–2260, 
+> doi: [10.1137/12086282X](https://doi.org/10.1137/12086282X),
+> arxiv: [1201.0925](https://arxiv.org/abs/1201.0925">1201.0925)
+"""
+function mean(M::Manifold, x::Vector{T};
```
mateuszbaran: I'd use abstract type here `x::AbstractVector`.

src/Manifolds.jl
mateuszbaran: Also `sum(...)/2` instead of `1/2*sum(...)` would avoid promotion to `Float64`. Similarly in weight initialization.

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+faster.  For more stopping criterai and details check the
+[`Manopt.jl`](https://manoptjl.org) package and use a solver therefrom.
+
+The algorithm is further described in Algorithm 4.3 and 4.4 in 
+> Bačák, M: Computing Medians and Means in Hadamard Spaces.
+> SIAM Journal on Optimization (2014), 24(3), pp. 1542–1566,
+> doi: [10.1137/140953393](https://doi.org/10.1137/140953393),
+> arxiv: [1210.2145](https://arxiv.org/abs/1210.2145)
+"""
+function median(M::Manifold, x::Vector{T};
```
mateuszbaran: Comments from `mean` also apply here.

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+    n = length(x)
+    y=x[1]
+    yOld = y
+    order = 1:n
+    iter = 0
+    while ( ( iter==0 || distance(M,y,yOld) > stop_atol ) && (iter < stop_iter) )
+        iter += 1
+        λ = 1/(iter+1)
+        yOld = y
+        order = use_random ? randperm(n) : order
```
mateuszbaran: That's almost surely a type instability. Making `use_random` a `::Val` or initializing `order` by `collect(1:n)` would fix it.

test/statistics.jl
```diff
@@ -0,0 +1,15 @@
+include("utils.jl")
+
+@testset "Median and Mean" begin
```
sethaxen: This is fine for now, but this test will eventually be replaced, since we can implement a more efficient `Sphere`-specific one.

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
     return v
 end
 
+@doc doc"""
+    mean(M,x; weights=1/n*ones(n), stop_tol=10^-7, stop_iter=100 )
```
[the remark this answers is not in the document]
sethaxen: Can you special-case `mean`/`median` for `Euclidean` to forward to the usual functions?

src/Manifolds.jl
```diff
@@ -19,13 +19,17 @@ import Base: isapprox,
     +,
     -,
     *
+import Distributions: mean,
```
sethaxen: Distributions just reexports from Statistics. Let's import from there. See below re `mean!`

src/Manifolds.jl
sethaxen: Is the `Ref` necessary?

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+"""
+function mean(M::Manifold, x::Vector{T};
+    weights=1/length(x)*ones(length(x)),
+    stop_atol=10^-7,
+    stop_iter=100
+    ) where {T}
+    iter = 0
+    y=x[1]
+    yOld = y
+    while ( ( iter==0 || distance(M,y,yOld) > stop_atol ) && (iter < stop_iter) )
```
sethaxen: What if we used our own `isapprox` instead, since that's really what we're interested in? And then we can have the same kwargs for atol and rtol, etc.

sethaxen: Why not

```julia
for iter in 1:stop_iter
    vmean = sum(weights .* log.(Ref(M), Ref(yold), x))) / 2
    yold, y = y, exp(M, yold, vmean)
    isapprox(M, y, yold; kwargs...) && break
end
```
It's simpler and easier to read.

src/Manifolds.jl
sethaxen: And we don't need `T`

src/Manifolds.jl
kellertuer: Of course we can overload a a mutating one, I will check for that. Well you can not broadcast the `M` nor the `yOld` so I put them in `Ref`s?

kellertuer: > Also `sum(...)/2` instead of `1/2*sum(...)` would avoid promotion to `Float64`. Similarly in weight initialization.

You even notice the smallest details, thanks :)

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+        yOld = y
+        y = exp(M,yOld,  1/2*sum( weights.*log.(Ref(M), Ref(yOld),x) )  )
+    end
+    return y
+end
+
+@doc doc"""
+    median(M,x; weights=1/n*ones(n), stop_tol=10^-10, stop_iter=10000, use_rand = false )
+
+computes the Riemannian median of the vector `x`  of `n` points on the
```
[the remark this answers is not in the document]
kellertuer: I actually haven't seen much literature on the median; but it's the generalization similar to the mean (minimizes distances squared) just with distances. What would you call it?

src/Manifolds.jl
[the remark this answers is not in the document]
kellertuer: Concerning the name, I am against calling it **Frechet**, since thats the global minimizer (see https://en.wikipedia.org/wiki/Fréchet_mean ) which might not be unique on manifolds (due to positive curvature for example) and I am against calling it **Karcher**, since Karcher opposes that (see https://arxiv.org/pdf/1407.2087.pdf), so I prefer to call the function just `mean` and document it as Riemannian Center of Mass.

I can check for the weights and adopt that, surely. And yes, we of course can put that into a separate file and provide fallbacks for Euclidean, since a closed form is known there.

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+`x` of `n` points on the [`Manifold`](@ref) `M`. This function
+uses the gradient descent scheme. Optionally one can provide
+weights $w_i$ for the weighted Riemannian center of mass.
+The general formula to compute the minimizer reads
+````math
+\argmin_{y\in\mathcal M} \frac{1}{2}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}^2(y,x_i),
+````
+where $\mathrm{d}_{\mathcal M}$ denotes the Riemian [`distance`](@ref).
+
+Finally `stop_tol` denotes the minimal change between to iterates then to stop
```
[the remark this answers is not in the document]
kellertuer: That's just a typo and it should read _two_ iterates

src/Manifolds.jl
```diff
@@ -102,6 +106,94 @@ function hat(M::Manifold, x, vⁱ)
...
+    ) where {T}
+    n = length(x)
+    y=x[1]
+    yOld = y
+    order = 1:n
+    iter = 0
+    while ( ( iter==0 || distance(M,y,yOld) > stop_atol ) && (iter < stop_iter) )
+        iter += 1
+        λ = 1/(iter+1)
+        yOld = y
```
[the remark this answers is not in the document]
kellertuer: Yes, of course. That's merely a remedy of my old usage of camelCase (other than for types) and me being german, we capitalize everything (basically).

test/statistics.jl
kellertuer: Of course, this was just the first that came to my mind. With the Euclidean Fallbacks, one should check those, too, with a specific one for the sphere (you mean the online one?) we should check that, too and this test could then also be done on `Hyperbolic` or `SymmetricPositiveDefinite`, a switch to Hyperbolic would be quite easy actually.

src/Manifolds.jl
kellertuer: Yes, why not. I have to to a separate line for the logs anyway when switching to inplace-logs as @mateuszbaran proposed.

src/Manifolds.jl
kellertuer: I think it's easier to switch to `shuffle!` then.

src/Manifolds.jl
kellertuer: The `Ref`s are necessary – `M` does not have a length, and you don't want to broadcast `yold`.

src/Manifolds.jl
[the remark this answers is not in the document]
sethaxen: > Concerning the name, I am against calling it **Frechet**, since thats the global minimizer (see [https://en.wikipedia.org/wiki/Fréchet_mean](https://en.wikipedia.org/wiki/Fr%C3%A9chet_mean) ) which might not be unique on manifolds

Are you against calling _this_ function a `frechet_mean` (I agree since it's not) or calling _any_ function a `frechet_mean`. If the latter, I don't follow.

If the assumptions behind the COM apply to general manifolds and also give rise to a definition of median and variance, then I agree it makes sense to call the COM `mean`.

src/Manifolds.jl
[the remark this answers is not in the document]
sethaxen: I'd call it something as similar to whatever we call the mean as possible. e.g. if COM is just `mean`, then `median` is fine.

src/Manifolds.jl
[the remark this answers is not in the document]
sethaxen: You also have weird comma rules. I had many arguments with a German colleague while cowriting a paper. 😉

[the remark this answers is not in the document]
kellertuer: No, we don't ;) They are just different from English ones, so one can not easily use German grammar intuition on English sentences.

src/Manifolds.jl
kellertuer: I am not yet sure why one should disable mutation, but I switched to a mutating one for now.

src/Manifolds.jl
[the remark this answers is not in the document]
kellertuer: COM is quite often called (Riemannian) mean.

src/Manifolds.jl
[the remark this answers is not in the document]
kellertuer: I haven'T completely understood, how I should model weights here, sorry; can you elaborate on that? I thought just a weight vector can't be that complicated – reading your link, I might have to reconsider.

src/Manifolds.jl
kellertuer: I think this switch to is_approx makes one test now fail and I can't see why, have to check that later.

src/Manifolds.jl
mateuszbaran: OK, we can think about non-mutating mean and median when it's clear that we need it. The `CurveSpace` manifold in `FunManifolds.jl` doesn't support mutation but I doubt using `mean` or `median` there is a good idea.

kellertuer: And if needed, the step back to non-mutating is not that much to do.

src/Statistics.jl
```diff
@@ -0,0 +1,115 @@
...
+
+The algorithm is further described in 
+> Afsari, B; Tron, R.; Vidal, R.: On the Convergence of Gradient
+> Descent for Finding the Riemannian Center of Mass,
+> SIAM Journal on Control and Optimization (2013), 51(3), pp. 2230–2260, 
+> doi: [10.1137/12086282X](https://doi.org/10.1137/12086282X),
+> arxiv: [1201.0925](https://arxiv.org/abs/1201.0925">1201.0925)
+"""
+function mean(M::Manifold, x::AbstractVector; kwargs...)
+    y = x[1]
```
mateuszbaran: I'd use `similar_result(M, mean, x[1], x)` here.

src/Statistics.jl
```diff
@@ -0,0 +1,115 @@
...
+            weights = ones(length(x)) / length(x),
+            stop_iter=100,
+            kwargs...
+        ) where {T}
+    yold = y
+    v = zero_tangent_vector.(Ref(M), fill(y,length(x)))
+    for i=1:stop_iter
+        yold = y
+        log!.(Ref(M), v, Ref(yold), x)
+        y = exp(M, yold, sum( weights.*v ) / 2 )
```
mateuszbaran:
```suggestion
        exp!(M, y, yold, sum( weights.*v ) / 2 )
```

src/Statistics.jl
kellertuer: Yes, no problem.

src/Statistics.jl
```diff
@@ -0,0 +1,115 @@
...
+check the [`Manopt.jl`](https://manoptjl.org) package and use a solver therefrom.
+
+The algorithm is further described in Algorithm 4.3 and 4.4 in 
+> Bačák, M: Computing Medians and Means in Hadamard Spaces.
+> SIAM Journal on Optimization (2014), 24(3), pp. 1542–1566,
+> doi: [10.1137/140953393](https://doi.org/10.1137/140953393),
+> arxiv: [1210.2145](https://arxiv.org/abs/1210.2145)
+"""
+function median(M::Manifold, x::AbstractVector; kwargs...)
+    y = x[1]
```
mateuszbaran: I'd use `similar_result(M, median, x[1], x)` here.

src/Statistics.jl
```diff
@@ -0,0 +1,115 @@
...
+        ) where {T}
+    n = length(x)
+    yold = y
+    order = collect(1:n)
+    iter = 0
+    for i=1:stop_iter
+        λ = 1/(iter+1)
+        yold = y
+        use_random && shuffle!(order)
+        v = zero_tangent_vector(M,y)
```
mateuszbaran:
```suggestion
        zero_tangent_vector!(M, v, y)
```
`v` just needs to be allocated once before the loop (probably using non-mutating `zero_tangent_vector`).

src/Statistics.jl
```diff
@@ -0,0 +1,115 @@
...
+    return median!(M, y, x; kwargs...)
+end
+function median!(M::Manifold, y, x::AbstractVector;
+            weights= ones(length(x)) / length(x),
+            stop_iter=10000,
+            use_random = false,
+            kwargs...
+        ) where {T}
+    n = length(x)
+    yold = y
```
mateuszbaran: We can't really assume that `y` is a valid starting point. `copy(x[1])` should be, or perhaps we could specify the starting point in a keyword argument?

src/Statistics.jl
kellertuer: Hm, no, there is a problem: The function is not defined for these types of input?

src/Statistics.jl
kellertuer: Until now for all input we always assumed everything is valid (for exp and log for example) and fine and only checked for `ArrayManifold` that points actually are on the manifold (one can have a lot of fun with non-normed points on the sphere giving strange results). So I would opt to stay with that scheme here, too.

src/Statistics.jl
mateuszbaran: Right, this won't work. `similar_result(M, mean, x[1])` should be good enough.

kellertuer: Hm, throws the same error message (no method matching) and I am not an expert on the idea of `similar_result` here.

src/Statistics.jl
```diff
@@ -0,0 +1,116 @@
...
+function mean!(M::Manifold, y, x::AbstractVector, w::AbstractWeights = weights(ones(length(x)) / length(x));
+            stop_iter=100,
+            kwargs...
+        ) where {T}
+    yold = y
+    v = zero_tangent_vector.(Ref(M), fill(y,length(x)))
+    for i=1:stop_iter
+        yold = y
+        log!.(Ref(M), v, Ref(yold), x)
+        y = exp(M, yold, sum( values(w).*v ) / 2 )
```
mateuszbaran:
```suggestion
        exp!(M, y, yold, sum( values(w).*v ) / 2 )
```
Maybe?
Also, `values(w)` could be calculated before the loop.

src/Statistics.jl
mateuszbaran: But `y` is output and not input? If that's the starting point it should be in the docs (and non-mutating mean/median shouldn't use `similar_result`).

src/Statistics.jl
kellertuer: `values(w)` is just accessing the internal array of values within `AbstractWeights`.

src/Statistics.jl
kellertuer: So what do you propose to do? I am fine with documenting that the output y is also used as starting point

src/Statistics.jl
```diff
@@ -0,0 +1,116 @@
...
+    return y
+end
+median(M::Euclidean,x) = median(x)
+median(M::Euclidean,x,w) = median(x,w)
+@doc doc"""
+    var(M,x, corrected::Bool=true, mean=mean(M,x))
+
+compute the variance of the points in `x` on the [`Manifold`](@ref) `M`, i.e.
+````
+    \frac{1}{m}\sum_{i=1}^n d_{\mathcal M}(m,x_i),
```
mateuszbaran: I think that should be the squared distance? And you have two different things called `m`.

src/Statistics.jl
```diff
@@ -0,0 +1,116 @@
...
+        isapprox(M,y,yold; kwargs...) && break
+    end
+    return y
+end
+median(M::Euclidean,x) = median(x)
+median(M::Euclidean,x,w) = median(x,w)
+@doc doc"""
+    var(M,x, corrected::Bool=true, mean=mean(M,x))
+
+compute the variance of the points in `x` on the [`Manifold`](@ref) `M`, i.e.
```
mateuszbaran:
```suggestion
compute the variance of the points in vector `x` of length `n` on the [`Manifold`](@ref) `M`, i.e.
```

src/Statistics.jl
mateuszbaran: I see now, that's a bug. `Manifolds.jl` doesn't import `similar_result` from `ManifoldsBase.jl` but it should. I'm really surprised this didn't cause any other problems.

src/Statistics.jl
mateuszbaran: OK, then let's keep `values(w)` as is.

src/Statistics.jl
mateuszbaran: Right, I think using `y` as a starting point is the most reasonable option. The non-mutating variant should then supply a copy of the first point as `y` here then.

kellertuer: Tha's actually now missing, that `similar_result` works, it filly `y` basically with zeros instead of `x[1]`.

src/Statistics.jl
kellertuer: Thanks for noticing, I switched for the normalise to `n'` in the doc..

src/Statistics.jl
kellertuer: Interesting – I fixed it but still have to set `y` to `x[1]` then, because `similar` returns zero values in all entries.

src/Statistics.jl
kellertuer: Since similar would return something close to zero, I also copied the values over afterwards; however, the median now does something different with random and nonrandom (result-wise) which should not be the case; I can't see that currently.

But I documented that y is the start value and added start-value-arguments for the nonmutatnig ones.

src/Statistics.jl
mateuszbaran: Yes, since `mean!` is supposed to use `y` as a starting point, you should use `copy(x[1])` here (not just `x[1]` since it will be overwritten).

kellertuer: Still, I have some side-effect somewhere, at least in the median; despite that your proposed improvements seem to be very good :)

src/Statistics.jl
```diff
@@ -0,0 +1,137 @@
...
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = 1/(iter+1)
+        copyto!(yold,y)
+        use_random && shuffle!(order)
+        zero_tangent_vector!(M,v,y)
+        for i=1:n
+            t = min( λ * w[order[i]] / distance(M,y,x[order[i]]) , 1 )
+            log!(M, v, y, x[order[i]])
+            exp!( M, y, y, t* v)
```
mateuszbaran: I think we can't assume that overwriting the point at which the vector is tangent is safe. I'd suggest adding another temporary besides `yold`.

kellertuer: Oh that is actually fine. Line 115 computes a new `v`, which is only once used in line 116 to do the step, `y`is updated and directly after that (next loop) `v` is updated in line 115 again.

src/Statistics.jl
mateuszbaran: I think `exp!(M, y, y, ...)` might be an issue but it will still work for some manifolds and should cause issues on both random and non-random orderings. I don't see any other problems.

src/Statistics.jl
mateuszbaran: What I mean is that `exp!` should be able to assume that writing to the output doesn't modify any arguments. For example this should work:
```julia
function exp!(S::Sphere, y, x, v)
    θ = norm(S, x, v)
    fill!(y, 0)
    y .= cos(θ) .* x .+ usinc(θ) .* v
    return y
end
```
And it wouldn't called like this (`exp!( M, y, y, t* v)`).

kellertuer: Ah and if you pass to both `x` and `y` the same variable you get something strange, yes; hm, so I have to use the non-mutating one here?

kellertuer: Found the culprit, I mixed up loop indices at some point, too; but then with the non-mutating `exp` this again works.

mateuszbaran: Yes, a non-mutating `exp` is correct here (although a mutating one with another temporary would be faster).

kellertuer: for now I would like to stay with this one and if we need the mutating one (which would not work for one of your manifolds) we could still implement the mutating one with an optional parameter.

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+    n = length(x)
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
+    order = collect(1:n)
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/(2*i)
+        copyto!(yold,y)
+        use_random && shuffle!(order)
+        for j=1:n
```
sethaxen: Why not `for j in order` and replace all `order[j]` with `j`?

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+            t = min( λ * w[order[j]] / distance(M,y,x[order[j]]) , 1 )
+            log!(M, v, y, x[order[j]])
+            y = exp(M, y, t* v)
```
sethaxen:
```suggestion
            y = exp(M, y, v, t)
```

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
+    order = collect(1:n)
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/(2*i)
+        copyto!(yold,y)
+        use_random && shuffle!(order)
+        for j=1:n
+            t = min( λ * w[order[j]] / distance(M,y,x[order[j]]) , 1 )
```
sethaxen: Check `length(w)==n` above and then use `@inbounds` here and below

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+            log!(M, v, y, x[order[j]])
+            y = exp(M, y, t* v)
+        end
+        isapprox(M, y, yold; kwargs...) && break
+    end
+    return y
+end
+median(M::Euclidean, x::AbstractVector, w::AbstractWeights) = median(x,w)
```
sethaxen: Likewise move to `Euclidean`

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+median(M::Euclidean, x::AbstractVector) = median(x)
+@doc doc"""
+    var(M,x, corrected::Bool=true, mean=mean(M,x))
```
sethaxen: I like the elegance of this signature, but it's probably best to use the one from `StatsBase.var`: https://github.com/JuliaStats/StatsBase.jl/blob/master/src/moments.jl#L41 where these are keyword arguments. They also have `varm` for the case where the mean is known.

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+on the [`Manifold`](@ref) `M`, i.e.
+````
+    \frac{1}{n'}\sum_{i=1}^n d_{\mathcal M}^2(m,x_i),
+````
+where $\mathrm{d}_{\mathcal M}$ denotes the Riemannian [`distance`](@ref)
+and $n'=n-1$ if `corrected=true` and $n'=n$ otherwise.
+
+A mean might be provided and is otherwise computed.
+"""
+var(M::Manifold, x::AbstractVector, corrected::Bool=true, mean=mean(M,x)) = sum( distance.(Ref(M), Ref(mean), x).^2 )/(length(x)-Int(corrected))
```
sethaxen: Can you add the weighted case as well?

sethaxen: Add for `Euclidean` the default `var`

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+            stop_iter=100,
+            kwargs...
+        ) where {T}
+    yold = similar_result(M, mean, y)
+    copyto!(yold,y)
+    v = zero_tangent_vector.(Ref(M), fill(y,length(x)))
+    for i=1:stop_iter
+        copyto!(yold,y)
+        log!.(Ref(M), v, Ref(yold), x)
+        exp!(M, y, yold, sum( values(w).*v ) / 2 )
```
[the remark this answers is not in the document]
sethaxen: Could declare a `vmean` above and then do
```suggestion
        mean!(vmean, v, w)
        exp!(M, y, yold, vmean)
```

src/Statistics.jl
kellertuer: Do you have a formula for the weighted one? Especially, I am unsure whether it's weights or weights-squared in the formula (i.e. `(weights.*distance(Ref(M),Ref(mean),x)).^2` or `weights.*(distance(Ref(M),Ref(mean),x)).^2` within the sum. How does the correction term look like then?

sethaxen: See the StatsBase implementation of `_moment2`: https://github.com/JuliaStats/StatsBase.jl/blob/master/src/moments.jl#L234-L244. Here's an (untested) implementation of these functions based on theirs:

```julia
function var(
    M::Manifold,
    x::AbstractVector,
    w::AbstractWeights,
    mean = mean(M, x, w);
    corrected::Bool = false,  # for var corrected defaults to false
)
    wv = convert(Vector, w)
    sqdist = wv .* distance.(Ref(M), Ref(mean), x) .^ 2
    s = sum(sqdist)
    c = w isa Weights ? varcorrection(length(x), corrected) / w.sum :
        varcorrection(w, corrected)
    return c * s
end

function var(M::Manifold, x::AbstractVector, mean = mean(M, x, w); kwargs...)
    return var(M, x, weights(ones(length(x))), mean; kwargs...)
end

function std(
    M::Manifold,
    x::AbstractVector,
    w::AbstractWeights,
    mean = mean(M, x, w);
    corrected::Bool = true,
)
    return sqrt(var(M, x, w; corrected = corrected))
end

function std(M::Manifold, x::AbstractVector, mean = mean(M, x, w); kwargs...)
    return std(M, x, weights(ones(length(x))), mean; kwargs...)
end

function mean_and_var(
    M::Manifold,
    x::AbstractVector,
    w::AbstractWeights = weights(ones(length(x)));
    kwargs...,
)
    m = mean(M, x, w)
    v = var(M, x, w, m; kwargs...)
    return m, v
end

function mean_and_std(
    M::Manifold,
    x::AbstractVector,
    w::AbstractWeights = weights(ones(length(x)));
    kwargs...,
)
    m = mean(M, x, w)
    s = std(M, x, w, m; kwargs...)
    return s, v
end
```

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+between two iterates is small. For more stopping criteria
+check the [`Manopt.jl`](https://manoptjl.org) package and use a solver therefrom.
+
+The algorithm is further described in Algorithm 4.3 and 4.4 in 
+> Bačák, M: Computing Medians and Means in Hadamard Spaces.
+> SIAM Journal on Optimization (2014), 24(3), pp. 1542–1566,
+> doi: [10.1137/140953393](https://doi.org/10.1137/140953393),
+> arxiv: [1210.2145](https://arxiv.org/abs/1210.2145)
+"""
+function median(M::Manifold, x::AbstractVector, w::AbstractWeights = weights(ones(length(x)) / length(x)); x0=x[1], kwargs...)
```
sethaxen: You can actually instantiate weights as `weights(ones(length(x)))`, since they're relative, not absolute. Just make sure everything is divide by `sum(w)` (which is precomputed)

src/Statistics.jl
sethaxen: I made some fixes to the above snippet

src/Statistics.jl
sethaxen: Actually, since to be safe you'll need a `1/sum(w)` in here anyways, probably best have a `wv = convert(Vector, w) ./ sum(w)` above and use `wv` here.

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+computes the [`mean`](@ref) in-place in `y` where the initial value of `y` is
+the starting point of the algorithm.
+"""
+function mean!(M::Manifold, y, x::AbstractVector, w::AbstractWeights = weights(ones(length(x)) / length(x));
+            stop_iter=100,
+            kwargs...
+        ) where {T}
+    yold = similar_result(M, mean, y)
+    copyto!(yold,y)
+    v = zero_tangent_vector.(Ref(M), fill(y,length(x)))
```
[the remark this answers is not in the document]
kellertuer: Hm, the problem, even with my old version was, that I need _different_ copies. I previously had `v = fill(0,zero_tangent_vector(M,y)` which has the same effect as yours: one memory allocation and a vector where changing one element affects all elements. That's not, what we need. Your's with a `map(_ -> copy(v0), x)` works.

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+        ) where {T}
+    n = length(x)
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
+    order = collect(1:n)
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/(2*i)
+        copyto!(yold,y)
+        use_random && shuffle!(order)
```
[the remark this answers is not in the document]
kellertuer: this always seems strange, since this optional argument will only have an effect if random is true. What about one keyword argument that is `shuffle_rng` that defaults to `nothing` (no shuffle) and you can set it to an png to activate shuffling?

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+    return median!(M, y, x; kwargs...)
+end
+@doc doc"""
+    median!(M,y,x,w)
+
+computes the [`median`](@ref) in-place in `y` where the initial value of `y` is
+the starting point of the algorithm.
+"""
+function median!(M::Manifold, y, x::AbstractVector, w::AbstractWeights = weights(ones(length(x)) / length(x));
+            stop_iter=100000,
```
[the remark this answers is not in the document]
kellertuer: Yes. It's nondifferentiable, so you can no to a gradient based approach, either a sub gradient descent (which is less stable) or a proximal map based approach (like the cyclic proximal point here), still their convergence is a little slow.

src/Statistics.jl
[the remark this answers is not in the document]
kellertuer: The 2 stems from regularising the step size, the `n` is hidden in the weights (which sum op to one, but I till correct that as you proposed below).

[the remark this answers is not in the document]
kellertuer: Concerning the vmean – you mean in the sense of a mean on the tangent vectors? hm, yes one could do that.

src/Statistics.jl
```diff
@@ -0,0 +1,135 @@
...
+            use_random = false,
+            kwargs...
+        ) where {T}
+    n = length(x)
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
+    order = collect(1:n)
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/(2*i)
```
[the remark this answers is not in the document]
kellertuer: I think this is more readable; similar to the step size, this is merely for a slightly better convergence, on could do any `c/i` where `c` is a constant.

src/Statistics.jl
kellertuer: Neat idea.

src/Statistics.jl
kellertuer: With the optional arguments I am fine, but `varm` is incorporated here already and they (just) have far more cases and implementations that I think are not that necessary here? They really spend a lot of lines on different cases for `var` and `varm` and I don't feel like copying all of them here.

src/Statistics.jl
kellertuer: Hm, this has a problem with the default (non keyword) arguments, since one case of var occurs double (manifold, array, mean), which occurs in both cases. I think for the second var, the mean can not have a default.

kellertuer:
I hope I did nearly every of your remarks, also unified your snipped (that's for that) – just adding tests and documentation for all `std`-functions is missing. I hope to find a moment for this in the evening.

src/Statistics.jl
kellertuer: I now did the signatures you proposed below, just have to adapt documentation.

src/Statistics.jl
```diff
@@ -0,0 +1,197 @@
...
+    n = length(x)
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
+    order = collect(1:n)
+    (length(w) != n) && error("The number of weights ($(length(w))) does not math the number of points for the median ($(n)).")
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/i
+        copyto!(yold,y)
+        (shuffle_rng != nothing) && shuffle!(shuffle_rng, order)
```
mateuszbaran:
```suggestion
        (shuffle_rng !== nothing) && shuffle!(shuffle_rng, order)
```
This should be faster

src/Euclidean.jl
```diff
@@ -183,3 +185,8 @@ function normal_tvector_distribution(M::Euclidean{Tuple{N}}, x, σ) where N
     d = Distributions.MvNormal(zero(x), σ)
     return ProjectedFVectorDistribution(TangentBundleFibers(M), x, d, project_vector!, x)
 end
+
+mean(M::Euclidean, x::AbstractVector, w::AbstractWeights) = mean(x,w)
+mean(M::Euclidean, x::AbstractVector) = mean(x)
```
sethaxen: Should also have `var` here. And probably also `mean!`.

sethaxen: and median!

kellertuer: And I thought this would be one of the shortest PRs I''ll do, it's getting the longest, actually – but yes you're right.

src/Manifolds.jl
```diff
@@ -216,6 +225,12 @@ export base_manifold,
     log,
     log!,
     manifold_dimension,
+    mean,
+    mean!,
+    mean_and_var,
```
sethaxen: You can import `mean_and_var` and `mean_and_std` from StatsBase.

kellertuer: Ah, I looked for it a while and did not see it anywhere.

sethaxen: https://github.com/JuliaStats/StatsBase.jl/blob/master/src/moments.jl#L152-L166

src/Euclidean.jl
sethaxen: Sorry!

sethaxen:
> I hope to find a moment for this in the evening.

Ha! I see what you did there!

src/Euclidean.jl
kellertuer: The problem with `median!` is, that in `Statistics.jl` there is no two argument `median!` which I find very strange.

kellertuer:
> > I hope to find a moment for this in the evening.
> 
> Ha! I see what you did there!

Playing word games is mean.

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+@doc doc"""
+    mean(M,x; weights=Weights(ones(n),n), x0=x[1], stop_iter=100, kwargs... )
+
+computes the Riemannian center of mass also known as Karcher mean of the vector
+`x` of `n` points on the [`Manifold`](@ref) `M`. This function
+uses the gradient descent scheme. Optionally one can provide
+weights $w_i$ for the weighted Riemannian center of mass.
+The general formula to compute the minimizer reads
+````math
+\argmin_{y\in\mathcal M} \frac{1}{2}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}^2(y,x_i),
```
sethaxen: I think you need this
```suggestion
\operatorname{argmin}_{y\in\mathcal M} \frac{1}{2}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}^2(y,x_i),
```
Also perhaps should have a `\sum_i=1^n w_i` term in here to avoid confusion re the weights.

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+````
+where $\mathrm{d}_{\mathcal M}$ denotes the Riemannian [`distance`](@ref).
+
+Optionally you can provide `x0`, the starting point (by default set to the first
+data point). `stop_iter` denotes the maximal number of iterations to perform and
+the `kwargs...` are passed to [`isapprox`](@ref) to stop, when the inimal change
```
sethaxen:
```suggestion
the `kwargs...` are passed to [`isapprox`](@ref) to stop, when the minimal change
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+"""
+function mean!(M::Manifold, y, x::AbstractVector, w::AbstractWeights = Weights(ones(length(x)), length(x));
+            stop_iter=100,
+            kwargs...
+        ) where {T}
+    yold = similar_result(M, mean, y)
+    copyto!(yold,y)
+    v0 = zero_tangent_vector(M, y)
+    v = map(_ -> copy(v0), x)
+    #v = zero_tangent_vector.(Ref(M), fill(y,length(x)))
```
sethaxen: Let's delete this

src/Statistics.jl
kellertuer: On my side the \argmin renders perfectly fine? Even with half a space between the g and the m as it should be

src/Statistics.jl
kellertuer: Ah yes, a leftover.

src/Statistics.jl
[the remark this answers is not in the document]
sethaxen: Yes I like calling `vmean` on the tangent vectors better. It avoids any funny business with the weights, since we let StatsBase handle the usual weighted mean in the tangent space.
I still don't understand the 2, but I trust you. 🙂

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+        copyto!(yold,y)
+        log!.(Ref(M), v, Ref(yold), x)
+        exp!(M, y, yold, sum( values(w).*v ) / 2 )
+        isapprox(M,y,yold; kwargs...) && break
+    end
+    return y
+end
+
+@doc doc"""
+    median(M,x weights=1/n*ones(n); stop_iter=10000, shuffle=nothing )
```
sethaxen:
```suggestion
    median(M, x, weights=1/n*ones(n); stop_iter=10000, shuffle=nothing)
```

src/Statistics.jl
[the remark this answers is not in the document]
kellertuer: Everything works as intended now.

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
+using Random: shuffle!
+using StatsBase: AbstractWeights, Weights, values, varcorrection
+@doc doc"""
+    mean(M,x; weights=Weights(ones(n),n), x0=x[1], stop_iter=100, kwargs... )
```
sethaxen:
```suggestion
    mean(M, x, weights=Weights(ones(n),n); x0=x[1], stop_iter=100, kwargs... )
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+@doc doc"""
+    median(M,x weights=1/n*ones(n); stop_iter=10000, shuffle=nothing )
+
+computes the Riemannian median of the vector `x`  of `n` points on the
+[`Manifold`](@ref) `M`. This function is nonsmooth (i.e nondifferentiable) and
+uses a cyclic proximal point scheme. Optionally one can provide
+weights $w_i$ for the weighted Riemannian median. The general formula to compute
+the minimizer reads
+````math
+\argmin_{y\in\mathcal M}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}(y,x_i),
```
sethaxen: Same comment RE `\argmin`

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+````
+where $\mathrm{d}_{\mathcal M}$ denotes the Riemannian [`distance`](@ref).
+
+The cyclic proximal point can be run with a random cyclic order by setting
+`shuffle` to a random number generator, e.g. `GLOBAL_RNG`.
```
sethaxen:
```suggestion
`shuffle_rng` to a random number generator, e.g. `GLOBAL_RNG`.
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+
+Optionally you can provide `x0`, the starting point (by default set to the first
+data point). `stop_iter` denotes the maximal number of iterations to perform and
+the `kwargs...` are passed to [`isapprox`](@ref) to stop, when the inimal change
```
sethaxen:
```suggestion
the `kwargs...` are passed to [`isapprox`](@ref) to stop, when the minimal change
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+    w::AbstractWeights = Weights(ones(length(x)), length(x));
+    stop_iter=100000,
+    shuffle_rng = nothing,
+    kwargs...
+) where {T}
+    n = length(x)
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
+    order = collect(1:n)
+    (length(w) != n) && error("The number of weights ($(length(w))) does not math the number of points for the median ($(n)).")
```
sethaxen:
```suggestion
    (length(w) != n) && error("The number of weights ($(length(w))) does not match the number of points for the median ($(n)).")
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+@doc doc"""
+    median!(M,y,x,w)
+
+computes the [`median`](@ref) in-place in `y` where the initial value of `y` is
+the starting point of the algorithm.
+"""
+function median!(M::Manifold, y, x::AbstractVector,
+    w::AbstractWeights = Weights(ones(length(x)), length(x));
+    stop_iter=100000,
+    shuffle_rng = nothing,
```
sethaxen:
```suggestion
    shuffle_rng::Union{Nothing,AbstractRNG} = nothing,
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+    order = collect(1:n)
+    (length(w) != n) && error("The number of weights ($(length(w))) does not math the number of points for the median ($(n)).")
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/i
+        copyto!(yold,y)
+        (shuffle_rng !== nothing) && shuffle!(shuffle_rng, order)
+        for j in order
+            @inbounds t = min( λ * w[j]/w.sum, distance(M,y,x[j]) )
+            log!(M, v, y, x[j])
```
sethaxen:
```suggestion
            @inbounds log!(M, v, y, x[j])
```

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+    return y
+end
+@doc doc"""
+    var(M,x,w=Weights(ones(n),n),m=nothing; corrected=false, kwargs...)
+
+compute the (weighted) variance of a `Vector` `x` of `n` data points on the
+[`Manifold`](@ref) `M`, i.e.
+
+````math
+\frac{1}{c} \sum_{i=1}^n w_i d_{\mathcal M}^2 (x_i,m)
```
sethaxen: Define `c`

kellertuer: Ah yes, I did not see, how the correction works for the weighted case, so yes, one has to define that here, can you give me a hint what it is?

src/Statistics.jl
sethaxen: Add `wv = convert(Vector, w) ./ w.sum` here.

src/Statistics.jl
```diff
@@ -0,0 +1,257 @@
...
+    copyto!(yold,y)
+    order = collect(1:n)
+    (length(w) != n) && error("The number of weights ($(length(w))) does not math the number of points for the median ($(n)).")
+    v = zero_tangent_vector(M,y)
+    for i=1:stop_iter
+        λ = n/i
+        copyto!(yold,y)
+        (shuffle_rng !== nothing) && shuffle!(shuffle_rng, order)
+        for j in order
+            @inbounds t = min( λ * w[j]/w.sum, distance(M,y,x[j]) )
```
sethaxen: Then here use `wv[j]` so we just normalize weights once

src/Statistics.jl
sethaxen: Okay no worries then. Haven't render on my machine, was just using LaTeXiT.

src/Statistics.jl
kellertuer: what's `wv` here? It's not defined in that function?

src/Statistics.jl
kellertuer: It'S defined within `mathtools` or something that is loaded/implemented in MathJax.

src/Statistics.jl
sethaxen: I would just say it's the potential correction and then refer them to `StatsBase.varcorrection` for details. There are a few different corrections, and they'll be wanting to read those docs when they select a weights type anyways.

kellertuer: Good idea.

src/Statistics.jl
sethaxen: I define it in an earlier comment a few lines above. It's `wv = convert(Vector, w) ./ w.sum`. So then the weights are normalized once.

src/Euclidean.jl
sethaxen: Oh that is surprising. StatsBase doesn't have it either. Okay, then we should probably just have an overload that computes `median` and then copies it.

src/Statistics.jl
kellertuer: Hm, I can't find that in the Doc in the v0.32? To what should I link?

src/Euclidean.jl
kellertuer: I don'T follow you here. What do you mean? Can you propose a line of code?

src/Statistics.jl
sethaxen: Oh, this one documents it pretty well: https://juliastats.org/StatsBase.jl/stable/scalarstats/#Statistics.var

kellertuer: Ah, I will link to that then; I was looking for the `varcorrection`, then I will link that one.

src/Statistics.jl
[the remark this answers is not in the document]
sethaxen: I don't think so? There's no `w.sum` anywhere.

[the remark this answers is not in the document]
kellertuer: Then you might have finally have lost me here and have to explain much more. I worked on this quite a while so I think it does what it should. If you think it does not – please provide better code.

src/Euclidean.jl
sethaxen:
```julia
median!(M::Euclidean, y, x::AbstractVector, w::AbstractWeights) = copyto!(y, median(x, w))
median!(M::Euclidean, y, x::AbstractVector) = copyto!(y, median(x))
```

kellertuer: Thanks, I'll add that then.

src/Statistics.jl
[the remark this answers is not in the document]
sethaxen: Do you mind if I push some changes with tests later today? You can always revert if you disagree.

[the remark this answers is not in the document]
kellertuer: Feel free to do so, when I ask for code, I am completely fine with a push.

kellertuer:
Thanks for the update, though we seem to now have a collision in DiffEqBase. I will check your changes, soon.

src/Statistics.jl
```diff
@@ -227,34 +231,63 @@ function std(
...
+    corrected = false,
     kwargs...
 )
     m = mean(M, x, w; kwargs...)
-    v = var(M, x, w, m; kwargs...)
+    v = var(M, x, w, m; corrected = corrected, kwargs...)
     return m, v
 end
+
+function mean_and_var(
```
kellertuer: Can't this function be replaced with setting `w::AbstractWeights = ProbabilityWeights(ones(n), n)` in line 244 as it was before? That seems to be the only difference between both versions.

test/statistics.jl
```diff
@@ -1,47 +1,191 @@
...
-    @test mean(Euclidean(1),x) == mean(x)
-    @test mean(Euclidean(1),x,w) == mean(x,w)
-    @test median(Euclidean(1),x) == median(x)
-    @test median(Euclidean(1),x,w) == median(x,w)
-end
\ No newline at end of file
+import ManifoldsBase: manifold_dimension, exp!, log!, distance, zero_tangent_vector!
+
+struct TestStatsSphere{N} <: Manifold end
+TestStatsSphere(N) = TestStatsSphere{N}()
```
kellertuer: Is there a reason for introducing a new manifold here? All functions seem to be just fallbacks to the sphere.

test/statistics.jl
```diff
@@ -1,47 +1,191 @@
...
+            θ = π / 4
+            @test isapprox(M, mean(M, x), geodesic(M, x[1], x[2], θ))
+            test_mean(M, x, [1.0, 1.0, 0.0] / √2)
+            test_median(M, x, [1.0, 1.0, 0.0] / √2; rng = MersenneTwister(1212), atol = 10^-12)
+            test_var(M, x, θ^2 * 2)
+            test_std(M, x, θ * √2)
+        end
+    end
+
+    @testset "Euclidean fallbacks" begin
```
kellertuer: I think we should also test the mutating fallbacks.

test/statistics.jl
sethaxen: Yes, because this is a test for the statistics, not a test for the sphere. And I'll be overloading the sphere cases in the next PR.

src/Statistics.jl
sethaxen: The other difference is the change in `corrected`. Unweighted `var` has `corrected=true` and weighted `var` has `corrected=false`.

test/statistics.jl
sethaxen: Thanks for the suggestion. I wrote the tests and they methods they fall back to don't work with those types, so I removed them.

sethaxen:
> Thanks for the update, though we seem to now have a collision in DiffEqBase. I will check your changes, soon.

Oh weird. Maybe a compat issue?

I'm done with the major changes now. The bulk of the work was adding some more tests, which exposed some things that needed fixing.

Most of the Euclidean overloads didn't work and got removed. We could probably put some more work there to be sure that every Euclidean case is efficient, but I think it's fine for this PR.

We still have failing tests for `median`. Some are just a matter of relaxing the tolerance, but some are stranger. I was hoping you could take a look at those; I haven't read the referenced paper yet.

How do we expect the Riemannian median to behave on a Euclidean manifold for array points? Will it be approximately the Euclidean median in each element, or something more like a centroid, or something else entirely?

kellertuer:
Well the median is not that stable on the sphere and it depends really how long you let the CPPA run then (its far more stable on the Euclidean space or on SPDs).

test/statistics.jl
kellertuer: Ok.

src/Statistics.jl
kellertuer: Ah I did not see that, sorry; then tis is fine, too.

kellertuer:
I can not approve your changes since it's my PR; but they are fine ;) For vectors/matrices (Euclidean(3)) you get the centroid, yes.

sethaxen:
The tests on https://github.com/JuliaNLSolvers/Manifolds.jl/pull/60 which is a no-change copy of master are all failing with the same DiffEqBase issue, so some package was updated and broke everything for us.

kellertuer:
And I thing I just saw why the median is failing; the points are too far apart on the sphere.

kellertuer:
So the vector R1 case still fails for one (really just one) of the weighted means an I do not yet see why.

sethaxen:
Here's an MWE:

```julia
using Manifolds, Random, StatsBase
x = [[1.], [2.], [3.], [4.]]
w = pweights(ones(length(x)))
M = Euclidean(1)
median(M, x, w; atol=1e-20, shuffle_rng = MersenneTwister(52)) # [1.0]
```

Is the equation for `t` in the code for `median` right? It looks different from Algorithm 4.3's equation for `t_k^n`.

kellertuer:
Oh I have to carefully check, it depends on whether you take geodesics with g(0)=a, g(1)=b compared to g(0)=a, g(T)=b, T=dist(a,b). I will check.

kellertuer:
When #60 is merged, I would like to check code coverage and documentation again, but despite that, this should be done I hope.

kellertuer:
Was this a rebase? I was quite confused on my computer with that and would've preferred a merge.

sethaxen:
Yes, it was a rebase. Sorry, git still confuses me sometimes, and I thought this was the right way to do it.

kellertuer:
`git merge` would've been right, but nevermind, the `force push` is always a sign of something dangerous ;) – this way my git was shortly confused and I had to hardcore-reset to your copy.

src/Statistics.jl
```diff
@@ -0,0 +1,286 @@
...
+    (length(w) != n) && throw(DimensionMismatch("The number of weights ($(length(w))) does not match the number of points for the median ($(n))."))
+    v = zero_tangent_vector(M,y)
+    wv = convert(Vector, w) ./ w.sum
+    for i=1:stop_iter
+        λ =  .5 / i
+        copyto!(yold,y)
+        for j in 1:n
+            @inbounds t = min( λ * wv[j] / distance(M,y,x[j]), 1. )
+            @inbounds log!(M, v, y, x[j])
+            y = exp(M, y, v, t)
```
sethaxen: You know, we should probably use mutating `exp!` here

src/Statistics.jl
```diff
@@ -0,0 +1,286 @@
...
+the starting point of the algorithm.
+"""
+function median!(M::Manifold, y, x::AbstractVector,
+    w::AbstractWeights = (n = length(x); Weights(ones(n), n));
+    stop_iter=1000000,
+    kwargs...
+) where {T}
+    n = length(x)
+    yold = similar_result(M,median,y)
+    copyto!(yold,y)
```
sethaxen: This `copyto!` is unnecessary since the loop begins with that.

test/statistics.jl
```diff
@@ -0,0 +1,260 @@
...
+
+        @testset "zero variance" begin
+            x = fill([0.,0.,1.], 5)
+            test_mean(M, x, [0.,0.,1.])
+            test_median(M, x, [0.,0.,1.]; atol = 10^-12)
+            test_var(M, x, 0.0)
+            test_std(M, x, 0.0)
+        end
+
+        @testset "two points" begin
```
sethaxen:
```suggestion
        @testset "three points" begin
```

src/Statistics.jl
kellertuer: No, we discussed that already, since `y`is already the base point it should not be mutable, that might lead to unwanted side effects.

sethaxen: But we can introduce a `ytmp` allocated outside the loop that is mutated here and then copied to `y` on the next line. That would replace the allocation with a copy. I have an implementation that passes the tests if you don't mind me pushing.

kellertuer: If you think its worth the effort, yes, why not. I like the current version for its readability.

sethaxen:
I can't see anything else that needs changing. If all the tests pass and the test coverage looks good, I'm happy to merge.

Thanks for the hard work on this! I know it was more involved than we thought, but I'm pleased that we have this excellent implementation.

kellertuer:
Well a few things took longer than expected, indeed; well, if it now works ... :) I would like to check the final documentation again just once the tests passed, but despite that I am also for merging then. Thanks for the good team work.

sethaxen:
I've been working on implementing the spherical mean for a future PR, which would overload this implementation. However, under some conditions, we'd want to fall back to this default, but we can't if it's overloaded. Any ideas on a good way to do that? Best idea I have so far is to split these implementations into internal functions like `_mean_default` that are wrapped by the signatures here.

kellertuer:
I'd propose to introduce a `method=` keyword for the main interface (`mean` and `median`) which is an `AbstractMethod` (or such) and dispatch on that as a further variable after the weight. The current mean (and default) is the `GradientMethod`, the current median is then `CyclicProximalPointMethod` (I hope to provide `SubGradientMethod` and `RandomCyclicProximalPointMethod` then, since the first is sometimes better and I hope to find my mistake in a random version).

PR #60, opened 2019-12-12
Add compat for Requires

sethaxen:


sethaxen:
The tests are failing. I suspect it's to do with Requires 1.0, which was released 8 hours ago.

sethaxen:
Okay, it looks like this patch fixed the issue. i.e. adding an implicit upper bound on the Requires version. For what it's worth, I'm seeing this issue on a number of CompatHelper PRs to bump Requires compat to 1.0 (i.e. something about DiffEqBase, OrdinaryDiffEq, or StaticArrays being incompatible).

We should probably add upper-bounds on all dependencies to prevent this kind of thing from happening in the future and use CompatHelper to maintain the bounds.

sethaxen:
I don't have a great grasp on this, but my guess is once this PR is merged, we may be able to bump our upper bound to 1.0 as well: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/398

mateuszbaran:
Right, we should add upper bounds to all dependencies and set CompatHelper up.

> I don't have a great grasp on this, but my guess is once this PR is merged, we may be able to bump our upper bound to 1.0 as well: [JuliaDiffEq/DiffEqBase.jl#398](https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/398)

We also need to wait until DiffEqBase tags a new release with this PR merged. Technically we can add `Manifest.toml` that specifies an unregistered version but I don't think it's worth it in this case.

sethaxen:
> We also need to wait until DiffEqBase tags a new release with this PR merged. Technically we can add `Manifest.toml` that specifies an unregistered version but I don't think it's worth it in this case.

Yes, of course, it should be released first. 🙂

PR #61, opened 2019-12-13
Add geodesic interpolation mean and higher moments

sethaxen:
This PR adds the geodesic interpolation method for online mean computation.

It introduces a new order of default calls for `mean`, `mean!`, `median` and `median!`. This enables one to add a new non-default method for a manifold like

```julia
struct MyMethod <: AbstractMethod end

function mean!(M::MyManifold, y, x::AbstractVector, w::AbstractWeights, ::MyMethod; kwargs...)
    ...
end
```

To change the default for the manifold, one would instead implement

```julia
function mean!(M::MyManifold, y, x::AbstractVector, w::AbstractWeights, method::MyMethod = MyMethod(); kwargs...)
    ...
end
```

This should change the default for all calls to `mean` and `mean!` for `MyManifold`.


src/Sphere.jl
```diff
@@ -156,3 +156,75 @@ function normal_tvector_distribution(S::Sphere, x, σ)
...
+Compute the Riemannian mean (center of mass) of `x`. The algorithm is adapted
+from the below paper to handle the weighted mean. It uses repeated weighted
+geodesic interpolation and converges to the Riemannian mean when all `x` are
+within a distance of π/2 of the mean.
+
+If `shuffle_rng` is provided, it is used to shuffle the order in which the
+points are considered for computing the mean.
+
+For more information on the geodesic interpolation method, see the below paper:
+> Salehian H et al: An efficient recursive estimator of the
```
kellertuer: I would prefer to see all the authors. There is not really a need to shorten that (unless its more than 12).

src/Sphere.jl
```diff
@@ -156,3 +156,75 @@ function normal_tvector_distribution(S::Sphere, x, σ)
...
+> Fréchet mean on a hypersphere with applications to Medical Image Analysis,
+> Mathematical Foundations of Computational Anatomy (2015).
```
kellertuer: Could you add a doi? Do they have a freely available version (arXiv/HAL) we could add?

src/Sphere.jl
```diff
@@ -156,3 +156,75 @@ function normal_tvector_distribution(S::Sphere, x, σ)
...
+"""
+function mean!(
```
kellertuer: Would the geodesic interpolation maybe also something one could provide for all manifolds? Or is it really only useful here? It seems to do something reasonable on manifolds in general. but I can't follow this completely – it seems it is then only used to initialise the gradient descent and is not a mean itself? Then the name mean (with GeodesicInterpolationMethod) is misleading?

src/Statistics.jl
```diff
@@ -1,21 +1,41 @@
...
-`x` of `n` points on the [`Manifold`](@ref) `M`. This function
-uses the gradient descent scheme. Optionally one can provide
-weights $w_i$ for the weighted Riemannian center of mass.
-The general formula to compute the minimizer reads
+Compute the (optionally weighted) Riemannian center of mass also known as
+Karcher mean of the vector `x` of points on the [`Manifold`](@ref) `M`, defined
+as the point that satisfies the minimizer
 ````math
-\argmin_{y\in\mathcal M} \frac{1}{2}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}^2(y,x_i),
+\argmin_{y\in\mathcal M} \frac{1}{2 \sum w}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}^2(y,x_i),
```
kellertuer: Do you mean `\sum_i w_i`?

src/Statistics.jl
```diff
@@ -1,21 +1,41 @@
...
 ````
 where $\mathrm{d}_{\mathcal M}$ denotes the Riemannian [`distance`](@ref).
 
-Note that any provided weights are normalized to
-$\displaystyle\sum_{i=1}^n w_i=1$.
+In the general case, the [`GradientMethod`](@ref) is used to compute the mean.
+However, this default may overloaded for specific manifolds.
```
kellertuer: do you mean "may be overloaded" ?

src/Statistics.jl
```diff
@@ -28,21 +48,63 @@ The algorithm is further described in
...
-computes the [`mean`](@ref) in-place in `y` where the initial value of `y` is
-the starting point of the algorithm.
+Compute the [`mean`](@ref) in-place in `y`.
 """
-function mean!(M::Manifold, y, x::AbstractVector, w::AbstractWeights = (n = length(x); Weights(ones(n), n));
-            stop_iter=100,
-            kwargs...
-        ) where {T}
+mean!
+
```
kellertuer: This is quite a lot of cases, but I see you want to always make the method optional...still, can some of these cases be combined?

src/Statistics.jl
```diff
@@ -59,19 +121,38 @@ function mean!(M::Manifold, y, x::AbstractVector, w::AbstractWeights = (n = leng
...
-computes the Riemannian median of the vector `x`  of `n` points on the
-[`Manifold`](@ref) `M`. This function is nonsmooth (i.e nondifferentiable) and
-uses a cyclic proximal point scheme. Optionally one can provide
-weights $w_i$ for the weighted Riemannian median. The general formula to compute
-the minimizer reads
+Compute the (optionally weighted) Riemannian median of the vector `x` of points on the
+[`Manifold`](@ref) `M`, defined as the point that satisfies the minimizer
 ````math
-\argmin_{y\in\mathcal M}\sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}(y,x_i),
+\argmin_{y\in\mathcal M} \frac{1}{\sum w} \sum_{i=1}^n w_i\mathrm{d}_{\mathcal M}(y,x_i),
```
kellertuer: see note above, I think this should read `\sum_i w_i` instead of just ` \sum w`.

src/Statistics.jl
```diff
@@ -175,9 +300,10 @@ function var(
         m = mean(M, x; kwargs...)
     end
     n = length(x)
-    w = ProbabilityWeights(ones(n), n)
+    w = _var_weights(n)
```
kellertuer: what's `_var_weights`? I might have missed it...

src/Sphere.jl
sethaxen: Unfortunately, there's no DOI (that I can find). It looks like this was a pretty small conference (led by Xavier Pennec). The official posting of the paper is here:https://www-sop.inria.fr/asclepios/events/MFCA15/Papers/MFCA15_4_2.pdf, but I don't know how stable the site/URL is. If we have link checking turned on, it should be fine. Think I should just hyperlink the title to it?

src/Sphere.jl
sethaxen: The geodesic interpolation algorithm is based on the geometric interpretation of incremental mean calculation as a weighted interpolation along the geodesic between the existing mean and the new datapoint. I don't know whether this should converge to the mean on any manifold.

It's general enough that we could leave it as a general method that others can test on their manifold. On the sphere, the paper only proved convergence to the mean when all the datapoints are within a geodesic ball of radius pi/2 from the mean. Hence the default on the `Sphere` is to compute the mean this way, then check the condition. If it doesn't apply, the estimate is used to initialize gradient descent because from manual checking, even if the condition is violated, the mean tends to be close to the mean computed by gradient descent. And the algorithm tends to be quite a bit faster than the gradient descent approach.

src/Statistics.jl
sethaxen: Yes. I was using the abbreviated notation StatsBase uses for weight summation, but I can expand it.

src/Statistics.jl
sethaxen: Do you mean in the docstring or in the code?

src/Sphere.jl
mateuszbaran: If I understand the paper correctly, it's actually an estimator for the mean of the distribution from which the points `x` were drawn. They prove that on spheres it converges and is unbiased.

src/Manifolds.jl
```diff
@@ -89,6 +89,12 @@ using Random: AbstractRNG
 using UnsafeArrays
 using Einsum: @einsum
 
+abstract type AbstractMethod end
```
mateuszbaran: Could we perhaps have a less generic name for this type? Or maybe a docstring which explains what kinds of methods it covers?

src/Sphere.jl
sethaxen: I think though it is worth noting that it only converges to the same mean as the gradient descent algorithm asymptotically with increasing sample size. For low sample size, it may give a different answer. Perhaps even if the condition is satisfied, it should always be used to initialize the gradient descent approach?

src/Manifolds.jl
sethaxen: Yes this is not ideal. I was thinking something like `AbstractMeanMethod` but e.g. `GradientMethod` is a generic name that could be useful for other functions. I don't know about `CyclicProximalPointMethod` though.

kellertuer: `CyclicProximalPointMethod` is generic in the sense that is just describes the optimization approach like `GradientMethod`.

src/Sphere.jl
kellertuer: I think it was published as a book chapter, but you're right - that one is hard to track down.

src/Sphere.jl
kellertuer: There is for example also no convergence theory for the Gradient method on Hadamard manifolds or for the CPPA approach on the Sphere (a reason why I would like to introduce sub gradient).

Using it as an initialisation to a different method that aims to solve the same problem was, what got me wondering first, I think. Is it good enough in approximation? Then we don't need to continue with grad desc. Is it not good enough? Then we maybe should not call it mean?

src/Statistics.jl
kellertuer: in the code. Document them is fine I think

src/Manifolds.jl
sethaxen: What if we introduced a new `AbstractMethod` in ManifoldsBase that is the parent type for our other methods? Then we can have a subtype AbstractOptimizationMethod that these inherit from.

src/Sphere.jl
sethaxen: Well it's good enough under conditions, but those conditions have only been worked out for a single manifold. When called with that method explicitly requested, we trust that the user knows what they're doing, and I think that's the right way to go. Only in the default case where the user doesn't care what algorithm is used would we maybe fallback. So I don't think the method name is misleading as long as the documentation is clear.

src/Statistics.jl
sethaxen: I'll double check but this was the only way I could work out to make it easy to add new methods and change defaults for manifolds easily, since optional arguments cannot be followed by required arguments. It's a lot of cases here but makes it easier and less error prone to extend. Assuming we keep the code as is, are you fine with how these are documented?

src/Sphere.jl
kellertuer: Ok. But then let's not directly use it as a start for the gradient descent.

And of course, document it clearly.

src/Statistics.jl
kellertuer: Yes, documentation is fine I think.

src/Sphere.jl
sethaxen: > Ok. But then let's not directly use it as a start for the gradient descent.

In the general case yes (i.e. for `mean!(::Manifold, ::Any, ::AbstractVector, ::AbstractWeights, ::GeodesicInterpolationMethod; kwargs...)` , but for `mean!(::Sphere, ::Any, ::AbstractVector, ::AbstractWeights; kwargs...)` (i.e. the default for `Sphere`) where we know the conditions under which the estimate is good, it's important to fall back to gradient descent if the conditions are violated.

src/Statistics.jl
sethaxen: `_var_weights` and `_mean_weights` were utility functions to create unit weights of length `n`, but they've now been merged to `_unit_weights`. When the next version of StatsBase is released, we can make it return `StatsBase.UnitWeights`.

sethaxen:
After searching through the literature, it turns out that the geodesic interpolation method has been shown to converge for all of our implemented manifolds and even the ones @kellertuer is currently developing.

Once I have a nice test suite in place, I'm going to standardise the benchmarks in the linked literature for these manifolds. In particular, assuming the points are tangent-normally distributed, at what minimum sample size is the error sufficiently low and how much slower is gradient descent than pure interpolation or interpolation to seed the recurrence, as a function of the standard deviation? This will help us determine for which manifolds and under which conditions this algorithm should be the default.

src/Manifolds.jl
mateuszbaran: I think I prefer to just have `AbstractOptimizationMethod` and `AbstractEstimationMethod` here. They don't need a common base type, do they?

sethaxen:
So informal benchmark results are in, and the results are:

1. There's no minimum sample size at which geodesic interpolation (GI) becomes sufficiently accurate. At low sample size, its error in estimating the true mean is equivalent to that of gradient descent (GD).
2. Both GI and GD scale linearly with the sample size. GI is 5-15x faster across all manifolds I checked (`Euclidean`, `Sphere`, `Rotations`, `SymmetricPositiveDefinite`), both for narrow and broad distributions. This is after a change to the `GradientMethod` implementation that gave in some cases ~5x speed-ups.
3. When all of the points are within the geodesic ball in the docs, the error of the two estimators is indistinguishable. 
4. Computing GI and then initializing GD if any points are outside the geodesic ball (which is a very conservative check) is in the worst case 1.1x slower than GD and in most cases gives a 2-10x speed-up. It converges to essentially the same solution as GD.

In summary, for now there are no clear downsides to defaulting to GI for the applicable manifolds with the condition check and optional GD fallback.

src/Statistics.jl
```diff
@@ -75,15 +75,15 @@ between two iterates is small. For more stopping criteria check the
 
 The algorithm is further described in [[Afsari et al, 2013]](#Afsari2013).
```
kellertuer: How does this render in REPL? In `Documenter` this refers to the literature hopefully available at the end of the statistics page (otherwise the link is not valid) but in REPL?. For this situation I like a (maybe shortened a little bit) reference in a block quote (`>`s)

src/Statistics.jl
```diff
@@ -75,15 +75,15 @@ between two iterates is small. For more stopping criteria check the
...
+mean(::Manifold, args...)
 
 @doc doc"""
     mean!(M::Manifold, y, x::AbstractVector, [w::AbstractWeights]; kwargs...)
     mean!(M::Manifold, y, x::AbstractVector, [w::AbstractWeights], method::AbstractMethod; kwargs...)
 
 Compute the [`mean`](@ref) in-place in `y`.
 """
-mean!
+mean!(::Manifold, args...)
```
kellertuer: Is this a `nothing`-definition and why? Can the documentation also be attached to line 88f?

src/Statistics.jl
```diff
@@ -233,15 +233,15 @@ change between two iterates is small. For more stopping criteria check the
 The algorithm is further described in Algorithm 4.3 and 4.4 in
 [[Bačák, 2014]](#Bačák2014)
 """
-median
+median(::Manifold, args...)
```
kellertuer: Same question as above for the `mean`.

src/Statistics.jl
```diff
@@ -233,15 +239,15 @@ change between two iterates is small. For more stopping criteria check the
...
 
 @doc doc"""
     median!(M::Manifold, y, x::AbstractVector, [w::AbstractWeights]; kwargs...)
     median!(M::Manifold, y, x::AbstractVector, [w::AbstractWeights], method::AbstractMethod; kwargs...)
 
 computes the [`median`](@ref) in-place in `y`.
 """
-median!
+median!(::Manifold, args...)
```
kellertuer: same as above for `median`...

src/Statistics.jl
```diff
@@ -286,12 +292,12 @@ function median!(
     x0=x[1],
     stop_iter=1000000,
     kwargs...
-) where {T}
-    x0 === nothing || copyto!(y, x0)
+)
     n = length(x)
+    (length(w) != n) && throw(DimensionMismatch("The number of weights ($(length(w))) does not match the number of points for the median ($(n))."))
+    x0 === nothing || copyto!(y, x0)
```
kellertuer: I think we discussed earlier, that `y` might carry the initial value, but I don't see that documented here... Maybe with having `x0` anyways there is no need to assume that, it's either `x[1]` or set by the user to a proper initial value, so we should always copy it to `y` here I think? Is there a reason to actually actively pass `nothing`?

kellertuer:
If, in the end, you could check, that the number of untested lines (code coverage) does not increase, that would also be awesome :)

src/Statistics.jl
sethaxen: It's documented here https://github.com/JuliaNLSolvers/Manifolds.jl/blob/6fdb918ff35edfb00573a4edec74fd819ce27051/src/Statistics.jl#L69-L74

The use case for `x0=nothing` is when `y` already contains an initial value that you want to refine in-place without allocating a new vector. This isn't performance critical, so if this is confusing, we can change it.

Alternatively, we could do `y === x0 || copyto!(y, x0)`. Then for the use case I'm thinking of, one would call `mean!(M, y, x, w; x0=y)`, which would also avoid the allocation and copy.

src/Statistics.jl
sethaxen: We use a definition with args as opposed to what it was before so that

```julia
julia> M = MyManifold()
? mean!(M, y, x, w)
```
gives this specific help menu, not the generic one for `Statistics.mean!`. This is even more useful for specific manifolds (e.g. see the `mean(::Sphere, args...)` case).

I don't attach it to one of the below functions for cleanliness. It defines the general interface for the user, but placing it below made it harder for me the developer to follow the logic of the different `mean` and `mean!` calls because it interrupted their flow. I don't know of any problems that this would cause, but if there are any, I can change it.

src/Statistics.jl
sethaxen: It's very unattractive in REPL. I can switch it back. The block quotes look nice.

sethaxen:
Certainly, and that tests pass of course. Test suite coming.

sethaxen:
On another note, the online mean computation is a feature of the Welford algorithm for online mean/variance computation. I derived a geodesic interpolation version of Welford for the variance and implemented it, and it's very accurate for every manifold I informally tested it on. I'm somewhat surprised that from what I can tell, no one has published this extension thus far. I'm not planning to include it here (for most manifolds, variance computation is a fraction of the cost of mean); I'm just surprised that such a simple and efficient algorithm with a little geometric intuition works so well.

src/Statistics.jl
kellertuer: Hm, oh ;) That was me. I think your alternative is really less confusing, so I would opt for that.

src/Statistics.jl
kellertuer: hm, interesting. Until now that is not consistently implemented that way, but it seems reasonable to do so (maybe better in `ManifoldsBase` for most functions). Thanks for the explanation. It can of course stay that way then.

sethaxen: Yes I have been learning recently about tricks like this that are very useful.

I also am adopting a convention for `mean` that maybe makes sense to adopt everywhere. Namely, our pattern is to overload the mutating functions, which changes the defaults for the non-mutating functions that call them. However, most users will probably initially implement a function with mutating and then switch to non-mutating to make more efficient. If we only document the non-mutating overload, then help messages aren't useful. But if we document both overloads, we have a lot more work to make docstrings consistent. Instead, we can adopt the convention that when we overload a mutating default, we always only document the corresponding non-mutating function, unless the non-mutating is also overloaded for the same types.

See e.g. the `mean(::Sphere, args...)` case. Here I overloaded `mean!(::Sphere, args...; kwargs...)`, but I only overloaded the docstring for `mean(::Sphere, args...)`.

mateuszbaran: This sounds neat, although a single docstring for both mutating and non-mutating version would be a bit more complicated (it has to cover both possibilities).

kellertuer: I think it's not meant to represent both; as far as I understand doing a doc-string for `mean(::Sphere, args...)` repeating the formula (from the `mean!`) provides a doctoring for the mean (non mutating) on the Sphere (which you have anyways through our `ManifoldsBase`-fallback from `mean` to `mean!`). I think the idea for user experience is nice (to type `? mean(::Sphere)` to see the doc), but in documentation it a little redundant.

src/Statistics.jl
```diff
@@ -86,18 +86,22 @@ Compute the mean using the specified `method`.
...
 
 Optionally, provide `x0`, the starting point (by default set to the first data
-point). Set `x0` to `nothing` to use the intitial value of `y` as the starting
-point. `stop_iter` denotes the maximal number of iterations to perform and the
+point). `stop_iter` denotes the maximal number of iterations to perform and the
 `kwargs...` are passed to [`isapprox`](@ref) to stop, when the minimal change
 between two iterates is small. For more stopping criteria check the
 [`Manopt.jl`](https://manoptjl.org) package and use a solver therefrom.
 
+Optionally, pass `retraction` and `inverse_retraction` method types to specify
```
kellertuer: Similar to the nice idea in the Symmetric it would maybe be nice to point to `AbstractRetractionMethod` and  `AbstractInverseRetractionMethod`s here?

src/Statistics.jl
```diff
@@ -222,15 +220,15 @@ function mean!(
...
+        inverse_retract!(M, v, y, x[j], inverse_retraction)
+        retract!(M, ytmp, y, v, t, retraction)
         copyto!(y, ytmp)
     end
     return y
 end
 
 @doc doc"""
-    median(M::Manifold, x::AbstractVector, [w::AbstractWeights]; kwargs...)
+    median(M::Manifold, x::AbstractVector [, w::AbstractWeights]; kwargs...)
```
kellertuer: this is really just a minor remark, but here you have `[, ... ]` while in 244 you have `[ ... ,]` - is that intentional or would it maybe be cool to have that unified? In general I like the `[]`-notation.

src/Statistics.jl
```diff
@@ -348,172 +337,313 @@ function median!(
...
+    cov(M, x, w::AbstractWeights, m=mean(M, x, w); corrected=false, kwargs...)
 
-compute the variance of a `Vector` `x` of `n` data points on the
-[`Manifold`](@ref) `M`, i.e.
+Compute the (optionally weighted) covariance of a `Vector` `x` of `n` data
+points on the [`Manifold`](@ref) `M`, i.e.
 
 ````math
-\sqrt{\frac{1}{c} \sum_{i=1}^n d_{\mathcal M}^2 (x_i,m)},
+\frac{1}{c} \sum_{i=1}^n w_i (\log_{\mu} x_i) (\log_{\mu} x_i)^\mathrm{T},
```
kellertuer: This might be my only problem with all these nice commits here: What's the log-transposed? I think in general you can only do that if you have a basis (ONB) and the log returns a vector representation with respect to that basis (compare `Grassmann` for example where the tangent space does not have such a unique nice representer. What's otherwise the transpose?

src/Statistics.jl
sethaxen: That would be nice. They're currently not in our docs at all. I'm assuming because they're not exported by ManifoldsBase.

src/Statistics.jl
sethaxen: I ended up unifying under the `[, ... ]` notation. I think it's the most sensible in most cases, the exception being in multiline signatures. It's also the syntax the Julia docs use. I prefer the comma inside because if the argument isn't included in the signature, then the comma makes no sense.

src/Statistics.jl
sethaxen: That should probably be resolved in another PR.

src/Statistics.jl
sethaxen: This concept of covariance (as a bilinear form independent of ONB) is borrowed from [Pennec's paper](http://www-sop.inria.fr/asclepios/Publications/Xavier.Pennec/Pennec.JMIV06.pdf). You bring up some good points, and the more I think about it the less it makes sense. I'll need to think about this some more.

kellertuer: I think in the embedding (for example with the Figure 2 of the Sphere) that makes totally sense, I just haven't read the paper close enough to make sure he is just taking about that. If he means the general case – you can of course use it, Xavier Pennec knows his stuff :)

sethaxen: After reading a bit more, this seems complicated, and I'd like to work it out in another PR after we have good support for ONBs. I'll remove `cov` from this PR.

From what I can tell, the expression for covariance assumes one is in an exponential chart not necessarily given an ONB. But it seems the expression A) assumes the tangent vector representation is a column vector and B) assumes a Euclidean metric (it notes that the trace of the covariance matrix is the Riemannian variance, which is only always true when the inner product is Euclidean, which in an exponential chart happens when using an ONB).

This interpretation is bolstered by [this paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.475.1066&rep=rep1&type=pdf), where Pennec works out an example of the Riemannian normal on covariance matrices and defines the same covariance with respect to an orthonormal basis.

sethaxen:
I think I've incorporated all the changes that will be in this PR. The tests pass on my machine (OS X) and I notice they do for Travis too, but they fail on Linux. If one of you is on a linux, could you take a look? This is the line that fails: https://github.com/JuliaNLSolvers/Manifolds.jl/blob/625bf6e13237ffe6a9e61106b4b6f26c1ccf9254/test/rotations.jl#L161

mateuszbaran:
I'm on Linux but this tests passes locally.

src/Rotations.jl
```diff
@@ -41,6 +41,12 @@ inner(M::Rotations, x, w, v) = dot(w, v)
 
 norm(M::Rotations, x, v) = norm(v)
 
+function project_point!(M::Rotations, y, x)
```
kellertuer: can you add a small doctoring here mentioning the method of projection? That would be nice, then it shows um in the docs.

src/Rotations.jl
```diff
@@ -547,19 +553,20 @@ function normal_rotation_distribution(M::Rotations{N}, x, σ::Real) where N
...
+Compute the Riemannian [`mean`](@ref mean(M::Manifold, args...)) of `x` using
+[`GeodesicInterpolation`](@ref). If any `x` are not within
+$\frac{\pi}{2 \sqrt 2}$ of the estimated mean, then the estimate is used to
+initialize mean computation using the [`GradientDescent`](@ref).
 """
 mean(::Rotations, args...)
 
-function mean!(M::Rotations, y, x::AbstractVector, w::AbstractWeights; shuffle_rng = nothing, kwargs...)
-    mean!(M, y, x, w, GeodesicInterpolationMethod(); shuffle_rng = shuffle_rng, kwargs...)
+function mean!(M::Rotations, y, x::AbstractVector, w::AbstractVector; shuffle_rng = nothing, kwargs...)
```
kellertuer: Why the change back to an abstract vector for the weights?

src/Rotations.jl
```diff
@@ -547,19 +553,20 @@ function normal_rotation_distribution(M::Rotations{N}, x, σ::Real) where N
...
+    mean!(M, y, x, w, GeodesicInterpolation(); shuffle_rng = shuffle_rng, kwargs...)
     for i in eachindex(x)
         @inbounds if distance(M, y, x[i]) ≥ π/2/√2
-            return mean!(M, y, x, w, GradientMethod(); x0 = y, kwargs...)
+            return mean!(M, y, x, w, GradientDescent(); x0 = y, kwargs...)
         end
     end
     return y
```
kellertuer: For me this implementation looks very similar to the Grassmann one, just the injectivity_radius is different (within the if) – could we implement that in general and just set the default method for rotaition and Grassmann to this mix of GeodesicInterpolation and GradientDescent? This makes these lines to be just one line and we have just one implementation of this method.

src/Sphere.jl
```diff
@@ -158,19 +158,20 @@ function normal_tvector_distribution(S::Sphere, x, σ)
...
+Compute the Riemannian [`mean`](@ref mean(M::Manifold, args...)) of `x` using
+[`GeodesicInterpolation`](@ref). If any `x` are not within $\frac{\pi}{2}$
+of the estimated mean, then the estimate is used to initialize mean computation
+using the [`GradientDescent`](@ref).
 """
 mean(::Sphere, args...)
 
-function mean!(S::Sphere, y, x::AbstractVector, w::AbstractWeights; shuffle_rng = nothing, kwargs...)
-    mean!(S, y, x, w, GeodesicInterpolationMethod(); shuffle_rng = shuffle_rng, kwargs...)
+function mean!(S::Sphere, y, x::AbstractVector, w::AbstractVector; shuffle_rng = nothing, kwargs...)
```
kellertuer: The question to vector and weights also applies here.

src/Sphere.jl
```diff
@@ -158,19 +158,20 @@ function normal_tvector_distribution(S::Sphere, x, σ)
...
+    mean!(S, y, x, w, GeodesicInterpolation(); shuffle_rng = shuffle_rng, kwargs...)
     for i in eachindex(x)
         @inbounds if distance(S, y, x[i]) ≥ π/2
```
kellertuer: The question to a common implementation using a injectivity_radius instead of the magic number (pi/2) also applies here.

src/SymmetricPositiveDefinite.jl
```diff
@@ -444,3 +444,25 @@ function check_tangent_vector(M::SymmetricPositiveDefinite{N},x,v; kwargs...) wh
...
+"""
+    mean(M::SymmetricPositiveDefinite, x::AbstractVector[, w::AbstractWeights]; kwargs...)
+
+Compute the Riemannian
+[`mean`](@ref mean(::Manifold, ::AbstractVector, ::AbstractVector, ::GeodesicInterpolation))
+of `x` using [`GeodesicInterpolation`](@ref).
+"""
+mean(::SymmetricPositiveDefinite, args...)
+
+mean!(M::SymmetricPositiveDefinite, y, x::AbstractVector, w::AbstractVector; kwargs...) =
```
kellertuer: The above mentioned “setting a default” is exactly these lines here :)

src/SymmetricPositiveDefinite.jl
```diff
@@ -444,3 +444,25 @@ function check_tangent_vector(M::SymmetricPositiveDefinite{N},x,v; kwargs...) wh
...
+    mean!(M, y, x, w, GeodesicInterpolation(); kwargs...)
+
+function mean!(
```
kellertuer: This should be done in Metric.jl for each default metric (like here the `LinearAffine`) to fall back to the mean with the base manifold.

src/statistics.jl
```diff
@@ -1,9 +1,9 @@
-using StatsBase: AbstractWeights, Weights, ProbabilityWeights, values, varcorrection, covm
+using StatsBase: AbstractWeights, ProbabilityWeights, values, varcorrection
 
 _unit_weights(n::Int) = ProbabilityWeights(ones(n), n)
 
 @doc doc"""
-    GeodesicInterpolationMethod <: AbstractMethod
+    GeodesicInterpolation <: AbstractEstimationMethod
```
kellertuer: To some extend both GradientDescent and CyclicProximalPoint are also estimation Methods in this context (and it would avoid a conflict with Manopt that also will have an AbstractOptimizationMethod ;)). What do you think?

src/SymmetricPositiveDefinite.jl
```diff
@@ -444,3 +444,25 @@ function check_tangent_vector(M::SymmetricPositiveDefinite{N},x,v; kwargs...) wh
...
+    mean(M::SymmetricPositiveDefinite, x::AbstractVector[, w::AbstractWeights]; kwargs...)
+
+Compute the Riemannian
+[`mean`](@ref mean(::Manifold, ::AbstractVector, ::AbstractVector, ::GeodesicInterpolation))
+of `x` using [`GeodesicInterpolation`](@ref).
+"""
+mean(::SymmetricPositiveDefinite, args...)
+
+mean!(M::SymmetricPositiveDefinite, y, x::AbstractVector, w::AbstractVector; kwargs...) =
+    mean!(M, y, x, w, GeodesicInterpolation(); kwargs...)
```
kellertuer: The most efficient one should be cyclic proximal point, I think. SPDs are a Hadamard manifold, so the distance is convex globally and the results from Bacak apply.

src/Rotations.jl
sethaxen: Simply to have fewer type restrictions. `AbstractWeights` is necessary in some cases for dispatch (e.g. for `moment` to distinguish between optional mean `m` and optional weights) and in other places for variance correction (any method calling `var`). Anywhere else, its type is irrelevant, and there's no need to enforce it. Note that the docstrings still use `AbstractWeights`, and it's still the officially supported API.

src/statistics.jl
sethaxen: This was from a suggestion by @mateuszbaran . I guess it depends on whether you will want `GradientDescent` and `CyclicProximalPoint` for anything else in Manifolds besides for these stats functions.

src/SymmetricPositiveDefinite.jl
sethaxen: We don't currently have a `CyclicProximalPoint` version of `mean!`.

src/statistics.jl
kellertuer: Indeed that depends. My reasoning is, that I don't see that we will, so in that sense they are Estimatormethods at least here, I think?

src/SymmetricPositiveDefinite.jl
kellertuer: Oh. I will provide that in an upcoming PR then, there is just a slight change (the parameter t changes) compared to the median.

src/Rotations.jl
kellertuer: Ok, that sounds reasonable.

src/statistics.jl
mateuszbaran: If both are supposed to be used as estimation methods only, they should inherit from `AbstractEstimationMethod`, and `GradientDescent` should be renamed to `GradientDescentEstimation` or something like that. Ultimately I think both `AbstractEstimationMethod` and `AbstractOptimizationMethod` should be added to `ManifoldsBase.jl`.

src/Sphere.jl
sethaxen: The "magic number" is unique to each manifold and is more conservative than the injectivity radius. If enough of the samples fall outside of that ball, then the accuracy drops quite a bit for some manifolds, so we don't want to use injectivity radius here.

But I do agree a common implementation is good. I'll add a `GeodesicInterpolationWithinRadius` method that implements this with fallback and have the manifolds with finite radii default to that.

src/Rotations.jl
mateuszbaran: And BTW, this function is incorrect. It projects onto the orthogonal group (rotations and reflections). This causes some tests to fail.

I don't know if there is a consistent way to project onto SO(n), I've always just used an SVD or QR (QR is generally faster) with an arbitrary reflection if needed. Like here: https://github.com/mateuszbaran/FunManifolds.jl/blob/04543dc52784762016bf7090e2fcfad5f98dd172/src/special_orthogonal.jl#L261

src/SymmetricPositiveDefinite.jl
sethaxen: And of course, we should probably benchmark the two techniques. The geodesic interpolation approach performs exactly `n-1` `retract!` and `inverse_retract!` calls for `n=length(x)`. The Cyclic proximal point implementation of `median!` requires `m * n` `retract!` and `inverse_retract!` for `m≥2`, where `m` is the number of iterations performed until convergence is detected. If the `mean!` version of cyclic proximal point has similar structure, I don't see how it could be more efficient. It could, however, be more accurate.

src/statistics.jl
sethaxen: For now, I'll make them both `AbstractEstimationMethod` and move their definitions to stats, since estimators should only be used in that context. What's the procedure for getting functionality into ManifoldsBase but still having it here for testing? I recommend we adopt a continuous delivery approach.

src/Rotations.jl
sethaxen: Yes, thanks, the change to the test helped me realize that. I saw a number of references that used the SVD approach with a sign change to the vector corresponding to the smallest singular value if necessary. It's not entirely clear from the references I saw, but it sounds like this is more reliable than the QR approach.

Ultimately, it might be good to think of an interface for specifying projection methods. e.g. on `Sphere`, if you know the point should be very close to the sphere and are only projecting to correct minor numerical imprecision, then you can do `x .*= 2 / (1 + dot(x, x))` and forego the square root. And for `Rotations`, one could (maybe) use QR instead of SVD and forego the reflection.

sethaxen: My numerical experiments indicate that indeed the SVD approach always gives an element of SO(n) closer to the original point than the QR approach does, under the Frobenius norm.

sethaxen:
I think I've made all the requested changes, so as long as the tests pass, this PR is ready to merge.

I also found that Documenter has a really nice [syntax for footnotes](https://juliadocs.github.io/Documenter.jl/stable/showcase/#Footnotes-1) that shows the references within the docstring within the REPL but within the documentation adds superscript hyperlinks to footnotes with the references. I took the liberty of replacing all of our references with it.

Now if only we could build those refs from bibtex files. 🙂

src/Rotations.jl
kellertuer: One can get back from O(n) to SO(n) by dividing by the sign of the determinant, since all matrices in SO(n) have determinant +1.

src/statistics.jl
kellertuer: In the long run `AbstractEstimationMethod` should go to `ManifoldsBase.jl`, yes, and maybe `AbstractOptimizationMethod`, too – though I'd happily "host" that in `Manopt`, too :)

kellertuer:
There's already quite a few issues asking for BibTeX-Support in Documenter. I already tried starting that once, but was too lazy to get it really started. But yes, I also really hope for a BibTeX-Style reference implementation in Documenter.

src/Sphere.jl
kellertuer: With "magic number" I meant exactly that – to me (see the Sphere) that looked like the injectivity radius, that's why I would like to have a function for that. As a conservative idea I would be fine with a default of injectivity_radius/2, but I would like to have a name not a number here. I will check you update later, then :)

src/statistics.jl
```diff
@@ -1,5 +1,19 @@
 using StatsBase: AbstractWeights, ProbabilityWeights, values, varcorrection
 
+"""
+    GradientDescentEstimation <: AbstractEstimationMethod
+
+Method for optimizing using gradient descent.
```
kellertuer: This should now read "method for estimation using gradient descent".

src/statistics.jl
```diff
@@ -1,5 +1,19 @@
...
+"""
+struct GradientDescentEstimation <: AbstractEstimationMethod end
+
+"""
+    CyclicProximalPointEstimation <: AbstractEstimationMethod
+
+Method for optimizing using the cyclic proximal point technique.
```
kellertuer: optimizing -> estimating.

src/statistics.jl
```diff
@@ -244,6 +301,43 @@ function mean!(
...
+    w::AbstractVector,
+    method::GeodesicInterpolationWithinRadius;
+    shuffle_rng = nothing,
+    kwargs...,
+)
+    mean!(M, y, x, w, GeodesicInterpolation(); shuffle_rng = shuffle_rng, kwargs...)
+    radius = method.radius
+    injectivity_radius(M) ≤ radius && return y
+    for i in eachindex(x)
+        @inbounds if distance(M, y, x[i]) ≥ radius
```
kellertuer: Cool! That'S a good solution here, keeping it flexible to not necessarily be the injectivity radius but still yielding a very readable code :)

sethaxen: Yes, the idea is that if the radius provided is so large that it's (probably) beyond our ability to compute distances anyways, then we just skip the expensive check. But now I see that I should be passing `y` to `injectivity_radius` here. :)

kellertuer: Ah yes, of course you can use the local injectivity radius and the global one you use now might be a little too conservative.

src/statistics.jl
mateuszbaran: > What's the procedure for getting functionality into ManifoldsBase but still having it here for testing? I recommend we adopt a continuous delivery approach.

I think it should look like this:
1. Make changes to `ManifoldsBase.jl`.
2. Release a new version of `ManifoldsBase.jl`.
3. Use the new version of `ManifoldsBase.jl` in a PR to `Manifolds.jl`, lower-bounding the dependency.

Is the continuous delivery approach different?

kellertuer: I think that's a good way. And we can do that whenever we decide that some feature (function, type, thtt,...) should go to `ManifoldsBase.jl`.

src/Rotations.jl
mateuszbaran: Yes, exactly, the SVD approach returns the closest rotation in the Frobenius norm but QR is faster. We need an `AbstractProjectionMethod`.

> One can get back from O(n) to SO(n) by dividing by the sign of the determinant, since all matrices in SO(n) have determinant +1.

Dividing what? Dividing/multiplying a 2k by 2k matrix by -1 doesn't change the determinant.

kellertuer: Oh, indeed, I was a little short in thinking: Why changing all signs, one sign should be enough – but yes that of the determinant does not help here for even cases ...

src/statistics.jl
sethaxen: No, that's basically continuous delivery, with the added twist that in the CD paradigm, you usually tag a new release after every PR, though I sometimes wait if I plan on finishing 2 PRs that go together back-to-back. See e.g. https://white.ucc.asn.au/2019/09/28/Continuous-Delivery-For-Julia-Packages.html

kellertuer: From a certain status onwards, I would favour this mode (maybe for `ManifoldsBase.jl` already), our package here still has too many discussions (and not yet “foreign” PRs). But from a first public version onwards – yes, why not.

PR #63, opened 2019-12-17
Symmetric matrices

manuelweisser:
Implements the Symmetric matrices as a basic manifold.


src/Manifolds.jl
```diff
@@ -162,6 +164,10 @@ export ArrayManifold,
     ArrayTVector
 
 export Manifold,
+    IsDecoratorManifold,
```
kellertuer: This line can be omitted, we implemented the decorator without `SimpleTraits`.

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
...
+````
+"""
+manifold_dimension(M::SymmetricMatrices{N,Complex}) where {N,T} = N*(N+1)
+
+@doc doc"""
+    project_point!(M::SymmetricMatrices,x)
+    
+Implements the [`project_point`](@ref project_point!(M::Manifold, x)) of an arbitrary matrix `x` onto the manifold of [`SymmetricMatrices`](@ref), i.e. 
+
+$x_{new} = \frac{1}{2} \left( x_{old} + x_{old}^{\mathrm{T}} \right).$
```
kellertuer: For consistency this could also be set with ` ````math ` and ` ```` ` (where you don't have to indent internally I think).

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
...
+"""
+project_point!(M::SymmetricMatrices, x) = (x .= 1/2 * (x + transpose(x)))
+
+@doc doc"""
+    project_tangent!(M::SymmetricMatrices, w, x, v)
+    
+Implements the [`project_tangent!`](@ref project_tangent!(M::Manifold, w, x, v)) of the matrix `v` onto the tangent space of the manifold of [`SymmetricMatrices`](@ref),
+i.e. the [`SymmetricMatrices`](@ref) itself. The result is stored in `w`:
+
+$w = \frac{1}{2} \left( v + v^{\mathrm{T}} \right).$
```
kellertuer: see last comment.

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
...
+    return y
+end
+
+@doc doc"""
+    log!(M::SymmetricMatrices, v, x, y)
+Computes the logarithmic map from the manifold point `x` to the manifold point `y`, both lying on the [`SymmetricMatrices`](@ref) manifold `M`. 
+The result is saved in `v`. 
+For the [`SymmetricMatrices`](@ref), the logarithmic map is simply the difference, i.e. 
+````math
+    v .= y-x
```
kellertuer: the dot does not belong to the math formula.

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
...
+
+@doc doc"""
+    manifold_dimension(M::SymmetricMatrices{n,T}) where {n,T <: Real}
+
+Return the dimension of the [`SymmetricMatrices`](@ref) matrix `M` with real-valued entries, i.e. 
+````math
+    \frac{n(n+1)}{2}
+````
+"""
+manifold_dimension(M::SymmetricMatrices{N,Real}) where {N,T} = div(N*(N+1),2)
```
sethaxen: Here and below we don't need the `T` type parameters

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
...
+````math
+    n(n+1)
+````
+"""
+manifold_dimension(M::SymmetricMatrices{N,Complex}) where {N,T} = N*(N+1)
+
+@doc doc"""
+    project_point!(M::SymmetricMatrices,x)
+    
+Implements the [`project_point`](@ref project_point!(M::Manifold, x)) of an arbitrary matrix `x` onto the manifold of [`SymmetricMatrices`](@ref), i.e.
```
sethaxen: I didn't know that syntax for `@ref`. Nice find!

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
+@doc doc"""
+    SymmetricMatrices{N,T} <: Manifold
+
+The [`Manifold`](@ref) $ Sym (N)$ consisting of the `T`-valued, symmetric matrices of size $ n\times n$, i.e. the set
```
sethaxen:
```suggestion
The [`Manifold`](@ref) $ \operatorname{Sym} (N)$ consisting of the `T`-valued, symmetric matrices of size $ n\times n$, i.e. the set 
```

src/Symmetric.jl
sethaxen: And also for `exp!`

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
+@doc doc"""
+    SymmetricMatrices{N,T} <: Manifold
```
sethaxen: I prefer the name `Symmetric` for consistency with what we already have for `SymmetricPositiveDefinite`. Are there any downsides to renaming it?

src/Symmetric.jl
```diff
@@ -0,0 +1,213 @@
...
+"""
+project_tangent!(M::SymmetricMatrices, w, x, v) = (w .= 1/2 * (v + transpose(v)))
+
+
+@doc doc"""
+    inner(M::SymmetricMatrices, x, w, v)
+
+compute the inner product of the two tangent vectors `w,v` from the tangent
+plane at `x` on the [`SymmetricMatrices`](@ref) `M` using the restriction of the
+metric from the embedding, i.e. $ (v,w)_x = v^\mathrm{T}w $.
```
sethaxen: Can you document this default metric in the docstring of `SymmetricMatrices`? Also I think this equation needs an `\operatorname{tr}`.
@kellertuer, what do you think about interpreting `EuclideanMetric` for matrix manifolds as being this metric? It seems be the only sensible interpretation to me, as our definition of `Euclidean` includes matrices that use that metric and then here we could specify the default metric as euclidean.

kellertuer: What do you mean by “this” metric? The metric from the Euclidean space? The `v^Tw`? yes we could do that. I would call that the metric from the embedding, though.

also `$ (v,w)_x = v^\mathrm{T}w $.` should read
````
\[
(v,w)_x = \operatorname{tr}(v^{\mathrm{T}}w)
\]
````
here (@manuelweisser)

Comment (sethaxen) [reply src/Symmetric.jl]: Actually, nevermind I take that back. I thought we had given `Rotations` and `Sphere` the `EuclideanMetric` default, interpreted as them inheriting the metric from the embedding, but I see now that we didn't, so it's fine not to specify a default metric here.

File: src/Manifolds.jl [hunk 10]
```diff
@@ -140,6 +140,8 @@ include("Euclidean.jl")
 include("CholeskySpace.jl")
 include("Rotations.jl")
 include("Sphere.jl")
+include("Symmetric.jl")
+include("ProjectedDistribution.jl")
```
sethaxen: This seems to be a redundant include (It's already included on line 143)

src/Symmetric.jl
manuelweisser: Did I get that comment right, @kellertuer  ? I changed that to 
```suggestion
using the restriction of the
metric from the embedding, i.e. 
````math
(v,w)_x = \operatorname{tr}(v^{\mathrm{T}}w)
````
```

src/Symmetric.jl
manuelweisser: I chose SymmetricMatrices (instead of Symmetric), because I think the name is not as ambiguous, as "Symmetric" could also mean symmetric functions (or anything symmetric).

kellertuer: I would also opt for SymmetricMatrices, since `Symmetric` is the command from `LinearAlgeba` that symmetrises a matrix.

src/Symmetric.jl
kellertuer: Concerning the math, yes :)

src/Symmetric.jl
sethaxen: Oh of course, I didn't think of that. I completely agree.

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+
+# Constructor
+
+    SymmetricMatrices(n::Int,T::Type = Real)
+
+"""
+struct SymmetricMatrices{N,T} <: Manifold end
+SymmetricMatrices(N::Int,T::Type = Real) = SymmetricMatrices{N,T}()
+
+function representation_size(::SymmetricMatrices{N,T}) where {N,T}
```
sethaxen: No need for `T` here

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+
+@doc doc"""
+    manifold_dimension(M::SymmetricMatrices{n,Real})
+
+Return the dimension of the [`SymmetricMatrices`](@ref) matrix `M` with real-valued entries, i.e. 
+````math
+    \frac{n(n+1)}{2}
+````
+"""
+manifold_dimension(M::SymmetricMatrices{N,Real}) where {N,T} = div(N*(N+1),2)
```
sethaxen: Here `T` is an unused type parameter

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+
+@doc doc"""
+    manifold_dimension(M::SymmetricMatrices{n,Complex})
+
+Return the dimension of the [`SymmetricMatrices`](@ref) matrix `M` with complex-valued entries, i.e. 
+````math
+    n(n+1)
+````
+"""
+manifold_dimension(M::SymmetricMatrices{N,Complex}) where {N,T} = N*(N+1)
```
sethaxen: Same comment about `T`

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+@doc doc"""
+    project_point!(M::SymmetricMatrices,x)
+    
+Implements the [`project_point`](@ref project_point!(M::Manifold, x)) of an arbitrary matrix `x` onto the manifold of [`SymmetricMatrices`](@ref), i.e. 
+````math
+x_{new} = \frac{1}{2} \left( x_{old} + x_{old}^{\mathrm{T}} \right).$
+````
+The value of the variable `x` is overwritten.
+"""
+project_point!(M::SymmetricMatrices, x) = (x .= 1/2 * (x + transpose(x)))
```
sethaxen:
```suggestion
project_point!(M::SymmetricMatrices, x) = (x .= (x .+ transpose(x)) ./ 2)
```

sethaxen: The broadcasted approach is roughly twice as fast in my checks.

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+    project_tangent!(M::SymmetricMatrices, w, x, v)
+    
+Implements the [`project_tangent!`](@ref project_tangent!(M::Manifold, w, x, v)) of the matrix `v` onto the tangent space of the manifold of [`SymmetricMatrices`](@ref),
+i.e. the [`SymmetricMatrices`](@ref) itself. The result is stored in `w`:
+````math
+w = \frac{1}{2} \left( v + v^{\mathrm{T}} \right)
+````
+
+"""
+project_tangent!(M::SymmetricMatrices, w, x, v) = (w .= 1/2 * (v + transpose(v)))
```
sethaxen:
```suggestion
project_tangent!(M::SymmetricMatrices, w, x, v) = (w .= (v .+ transpose(v)) ./ 2)
```

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+Computes the exponential map [`exp!`](@ref exp!(M::Manifold, y, x, v, t=1)) of the tangent vector `v` at
+manifold point `x` lying on the [`SymmetricMatrices`](@ref) manifold `M`. 
+The result is saved in `y`. 
+For the [`SymmetricMatrices`](@ref), the exponential map is simply the sum, i.e. 
+````math
+    y = x + v
+````
+"""
+function exp!(M::SymmetricMatrices, y, x, v)
+    y .= x + v
```
sethaxen:
```suggestion
    y .= x .+ v
```

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+
+@doc doc"""
+    vector_transport_to!(M::SymmetricMatrices, vto, x, v, y)
+Computes the [`vector_transport_to!`](@ref vector_transport_to!(M::Manifold, vto, x, v, y, ProjectionTransport())) 
+of the tangent vector `v` from the manifold point `x` on the [`SymmetricMatrices`](@ref) `M`
+to the manifold point $y\in M$. The result is stored in `vto`.
+Since the metric is inherited from the embedding space, this is just the identity.
+"""
+function vector_transport_to!(M::SymmetricMatrices, vto, x, v, y, ::ParallelTransport)
+    vto .= v
```
sethaxen: Here you can just use `copyto!`, which will do no copy if `v` and `vto` are the same object.

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+    log!(M::SymmetricMatrices, v, x, y)
+Computes the logarithmic map from the manifold point `x` to the manifold point `y`, both lying on the [`SymmetricMatrices`](@ref) manifold `M`. 
+The result is saved in `v`. 
+For the [`SymmetricMatrices`](@ref), the logarithmic map is simply the difference, i.e. 
+````math
+    v = y-x
+````
+"""
+function log!(M::SymmetricMatrices, v, x, y)
+    v .= y-x
```
sethaxen:
```suggestion
    v .= y .- x
```

src/Symmetric.jl
manuelweisser: Oh, I did not know you can leave that out. Thanks

kellertuer:
```suggestion
manifold_dimension(M::SymmetricMatrices{N,Real}) where {N} = div(N*(N+1),2)
```
I think you missed a `T` here, still

src/Symmetric.jl
```diff
@@ -0,0 +1,217 @@
...
+
+# Constructor
+
+    SymmetricMatrices(n::Int,T::Type = Real)
+
+"""
+struct SymmetricMatrices{N,T} <: Manifold end
+SymmetricMatrices(N::Int,T::Type = Real) = SymmetricMatrices{N,T}()
+
+function representation_size(::SymmetricMatrices{N,T}) where {N}
```
kellertuer:
```suggestion
function representation_size(::SymmetricMatrices{N}) where {N}
```
You missed another `T` that made the tests fail.

src/Symmetric.jl
manuelweisser: How can I leave out T here? Because the following code results in an (UndefVar) error: 
`function representation_size(::SymmetricMatrices{N,T}) where {N}
    return (N,N)
end`

kellertuer: Yes, then `T` is unknown, just really leave it out completely writing line 24 as
````Julia
function representation_size(::SymmetricMatrices{N}) where {N}
````

kellertuer:
Let's wait for the tests (the symmetric ones seem to not fail anymore) and wait for the test coverage – if that is fine we can merge this.

src/Symmetric.jl
sethaxen: Yes when ending type parameters are left out then they are implicitly added to the end so `::SymmetricMatrices{N}` is `SymmetricMatrices{N,T} where {T}`, which makes dispatch really convenient if one has many type parameters but most commonly dispatches on the first few.

sethaxen:
Looks good to me!

PR #64, opened 2019-12-18
Add AD backend, defaulting to FiniteDifferences

sethaxen:
After conversations on Slack, this PR adds some basic AD backend functionality to Manifolds.

All internal functions that require AD will just work now, using FiniteDifferences as the default. When ForwardDiff becomes available, it becomes an optional backend. We might consider switching the default to ForwardDiff when it becomes available, but since that can be triggered by any dependency, it might be jarring to the user. The downside to keeping FD as the default is that it's quite slow.

Example usage:

```julia
julia> using Manifolds, LinearAlgebra

julia> Manifolds.adbackends()
1-element Array{Symbol,1}:
 :finitedifferences

julia> Manifolds.adbackend()
:finitedifferences

julia> Manifolds._jacobian(x->[dot(x,x)], ones(3))
1×3 Array{Float64,2}:
 2.0  2.0  2.0

julia> using ForwardDiff

julia> Manifolds.adbackends()
2-element Array{Symbol,1}:
 :finitedifferences
 :forwarddiff      

julia> Manifolds.adbackend!(:forwarddiff)
:forwarddiff

julia> Manifolds.adbackend()
:forwarddiff

julia> Manifolds._jacobian(x->[dot(x,x)], ones(3))
1×3 Array{Float64,2}:
 2.0  2.0  2.0

julia> Manifolds._jacobian(x->[dot(x,x)], ones(3))
1×3 Array{Float64,2}:
 2.0  2.0  2.0

julia> using FiniteDifferences

julia> Manifolds._jacobian(x->[dot(x,x)], ones(3), forward_fdm(2, 1))
1×3 Array{Float64,2}:
 2.0  2.0  2.0
```


mateuszbaran:
This looks like a good starting point. I'm quite surprised that you find FD quite slow and that you've set such big atol in tests. FD should be quite fast here. One thing I'm worried about is that these functions that calculate jacobians and gradients don't really look type stable, so maybe it causes some slowdowns? Inspecting the code for FD-based gradients and jacobians using Cthulhu.jl could reveal something.

sethaxen:
> I'm quite surprised that you find FD quite slow ... FD should be quite fast here.

Well, fast and slow are relative. 🙂 It can be an order of magnitude or more slower than ForwardDiff by my tests. None of the functions we currently require AD for are performance critical though (unless someone is spending a lot of time numerically integrating geodesics).

> I'm quite surprised ... that you've set such big atol in tests.

Can you elaborate? Which tests, and what would you set the atol to?

> One thing I'm worried about is that these functions that calculate jacobians and gradients don't really look type stable, so maybe it causes some slowdowns?

What looks type unstable about the functions? Here's a more complicated example derived from the one you sent me. Type inference works just fine on this:

```julia
julia> using Manifolds, LinearAlgebra, BenchmarkTools

julia> M = Sphere(2);

julia> d = Manifolds.uniform_distribution(M, randn(representation_size(M)));

julia> x, y, = rand(d), rand(d);

julia> function logdetjac_log(M::Manifold, x, y; backend = :default)
           J = Manifolds._jacobian(y, backend) do z
               v = Manifolds.zero_tangent_vector(M, x)
               log!(M, v, x, project_point(M, z))
               return v
           end
           N = manifold_dimension(M)
           jsvd = svd(J)
           jr = view(Diagonal(jsvd.S) * jsvd.Vt, 1:N, :)
           G = jr * transpose(jr)
           return logdet(G) / 2
       end
logdetjac_log (generic function with 1 method)

julia> @inferred logdetjac_log(M, x, y)
1.4503089495038706

julia> @btime logdetjac_log(M, x, y)
  107.694 μs (990 allocations: 54.41 KiB)
1.4503089495038706
```

It does fail if I try to take the gradient of this function, but that's 2nd order AD:

```julia
julia> function logdetjac_log_withgrad(M, x, y)
           return (logdetjac_log(M, x, y), Manifolds._gradient(y -> logdetjac_log(M, x, y), y))
       end
logdetjac_log_withgrad (generic function with 1 method)

julia> @inferred logdetjac_log_withgrad(M, x, y)
ERROR: return type Tuple{Float64,Array{Float64,1}} does not match inferred return type Tuple{Float64,Any}
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at REPL[114]:1

julia> @btime logdetjac_log_withgrad(M, x, y)
  4.099 ms (36853 allocations: 1.97 MiB)
(1.4503089495038706, [0.9791312408054148, -2.4844908992735593, -0.04532829899342151])
```

mateuszbaran:
> Well, fast and slow are relative. 🙂 It can be an order of magnitude or more slower than ForwardDiff by my tests. None of the functions we currently require AD for are performance critical though (unless someone is spending a lot of time numerically integrating geodesics).

It was just a small remark. For some functions with a small number of inputs and outputs I actually had to micro-optimise `ForwardDiff` before it was faster than FD. Anyway, yes, outside of this one case AD was never my bottleneck.

> Can you elaborate? Which tests, and what would you set the atol to?

By default `isapprox` defaults to `rtol` equal to square root of machine epsilon (about 1.5e-8 for 'Float64') and `atol` to 0. The old tests were using the default values. I'm talking about the tests in `metric.jl`.

> What looks type unstable about the functions? Here's a more complicated example derived from the one you sent me. Type inference works just fine on this:

Sometimes the result type is inferred but some internal types are not and it can still hurt the performance. Things like `_jacobian(f, x) = _jacobian(f, x, Val(adbackend()))` have types that are constructed (`Val`) from externally obtained data (`adbackend`) that further accesses global state. Sometimes the compiler is smart enough to optimise it, sometimes it's not. Anyway, it's not a big problem, just something that should be kept in mind.

mateuszbaran:
And about this potential type instability, having it here (dispatch connected to global state) in some form is probably unavoidable anyway.

sethaxen:
> By default `isapprox` defaults to `rtol` equal to square root of machine epsilon (about 1.5e-8 for 'Float64') and `atol` to 0. The old tests were using the default values. I'm talking about the tests in `metric.jl`.

Yes, I had to set the `atol`s because many of those values are 0. ForwardDiff gets exactly 0, so there were no problems before, but FiniteDifferences just gets something close to 0.

> Sometimes the result type is inferred but some internal types are not and it can still hurt the performance. Things like `_jacobian(f, x) = _jacobian(f, x, Val(adbackend()))` have types that are constructed (`Val`) from externally obtained data (`adbackend`) that further accesses global state. Sometimes the compiler is smart enough to optimise it, sometimes it's not. Anyway, it's not a big problem, just something that should be kept in mind.

> And about this potential type instability, having it here (dispatch connected to global state) in some form is probably unavoidable anyway.

Yeah I don't see a great way around it in this case. But I think this is fine for an initial PR, and we can refine it later if necessary. Do you see any obstacles to merging?

mateuszbaran:
> Yeah I don't see a great way around it in this case. But I think this is fine for an initial PR, and we can refine it later if necessary. Do you see any obstacles to merging?

Right, the PR is fine. Implementation details can always be changed later if we find a better solution.

PR #68, opened 2019-12-22
[WIP] Lie groups and actions

mateuszbaran:
I've started the implementation of Lie groups and actions, since the previous PR (#25) is a bit stale and I think too big to finish in one go. The main part of the functionality that I would like to have in this PR is mostly done. The code is based on @sethaxen 's code from PR #25 and my earlier attempts at providing an interface for Lie group actions.

A few things to consider:
  1. I've encountered at least one useful monoidal action with Riemannian geometry on a manifold so I'm not totally sure the "group" terminology is perfect. 
  2. I'm going to add the translation action here.
  3. `translate_diff` might be important enough to include here, I'm not sure.
  4. Names `translate` and `Identity` might be a little too generic to be used here.
  5. I'll add more tests.


kellertuer:
Nice start at first glance. Can we have a symbol for the group action on the manifold, for example `⊕`, `⊗` or something like that? I would be in favour of that for readability.

mateuszbaran:
What would actually be the (two) arguments for such a symbol? `apply_action` has three arguments.

kellertuer:
Hm, then `apply_action` is not the thing I have in mind. To be more precise, for example for the `Rotations` the group action is concatenation of rotations  or their matrix multipilcation `x*y`for short. Not all group actions might have such a nice form so I would like to have a function `⊕(x,y)` for the Lie group action like the one just explained.

mateuszbaran:
Do you maybe have `compose_left` in mind? It also has three argument anyway.

What are `x` and `y` in `⊕(x,y)`? They can't be just any arrays, we wouldn't know what they represent. Do you want a wrapper/decorator for group elements?

kellertuer:
They are points on a manifold, so group elements for Lie groups. the group operation of the lie group is what I want to `⊕`

mateuszbaran:
But how do I tell if two 2x2 matrices are from the rotation group or the translation group for R^{2x2}? The group operation is different in both cases.

kellertuer:
Ah, well then I was again too stuck in my typed points. You're right – without typed points it's not possible, sad, I would've really liked above syntax.

src/groups/group.jl
```diff
@@ -0,0 +1,326 @@
...
+    copyto!(y, inv(x))
+    return y
+end
+
+inv(::AbstractGroupManifold{MultiplicationOperation}, x) = inv(x)
+
+compose_left(::AbstractGroupManifold{MultiplicationOperation}, x, y) = x * y
+function compose_left!(::AbstractGroupManifold{MultiplicationOperation}, z, x, y)
+    #TODO: z might alias with x or y, we might be able to optimize it if it doesn't.
+    copyto!(z, x*y)
```
sethaxen: Nice! I like how already this optimizes. e.g. if `x` is identity` and `y=z`, then this is a no op.

src/groups/group.jl
```diff
@@ -0,0 +1,326 @@
...
+"""
+    compose_left(G::AbstractGroupManifold, xs...)
+
+Compose elements of `G` using their left translation upon each other.
+This composition is left-associative.
+"""
+compose_left(G::AbstractGroupManifold, xs...) = _group_afoldl(compose_left, G, xs...)
+compose_left(G::AbstractGroupManifold, x) = x
+function compose_left(G::AbstractGroupManifold, x, y)
+    z = similar_result(G, compose_left, x, y)
```
sethaxen: How does this behave if both `x` and `y` are identity?

src/groups/group_action.jl
```diff
@@ -0,0 +1,90 @@
...
+"""
+abstract type AbstractActionOnManifold end
+
+"""
+    base_group(A::AbstractActionOnManifold)
+
+The group that acts in action `A`.
+"""
+function base_group(A::AbstractActionOnManifold)
+    error("Function base_group is not yet defined for type $(typeof(A)).")
```
sethaxen: Can we keep these errors consistent with those in ManifoldsBase? e.g. this would be 
```julia
error("base_group not implemented for type $(typeof(A)).")
```

src/groups/group_action.jl
```diff
@@ -0,0 +1,90 @@
...
+end
+
+"""
+    action_on(A::AbstractActionOnManifold)
+
+The manifold the action `A` acts upon.
+"""
+function action_on(A::AbstractActionOnManifold)
```
sethaxen: This name is confusing, because it doesn't return the action, it takes the action. I don't have a better idea. The technical name for the manifold upon which a group acts is I believe a G-manifold.

src/groups/group_action.jl
```diff
@@ -0,0 +1,90 @@
...
+    error("Function action_on is not yet defined for type $(typeof(A)).")
+end
+
+"""
+    apply_action(A::ActionOnManifold, x, a)
```
sethaxen: I prefer `apply` and let dispatch indicate that an action is being applied. See https://github.com/invenia/BlueStyle#function-naming.

src/groups/group_action.jl
```diff
@@ -0,0 +1,90 @@
...
+
+Apply action `a` of type `A` to the point `x`.
```
sethaxen: This indicates that `a::A`, but I don't think that's what you mean?

src/groups/group_action.jl
```diff
@@ -0,0 +1,90 @@
...
+    y = similar_result(A, apply_action, x, a)
+    apply_action!(A, y, x)
+    return y
+end
+
+"""
+    optimal_alignment(A::AbstractActionOnManifold, x1, x2)
+
+Calculate an action element of action `A` that, when acted
+upon `x1`, gets the element closest to `x2`.
```
sethaxen: I prefer "... that acts upon `x1` to produce the element closest to `x2`."
Also, by what definition of "closest"? Using the metric on the group manifold or the g-manifold? Or something else?

src/groups/group_action.jl
```diff
@@ -0,0 +1,90 @@
...
+
+"""
+    center_of_orbit(
+        A::AbstractActionOnManifold,
+        pts,
+        q
+        [, mean_method::AbstractEstimationMethod = GradientDescentEstimation()]
+    )
+
+Calculate an action element of action `A` that constitutes mean element of orbit
```
sethaxen: WOuld be good to have some formulas to define the orbit here.

src/groups/rotation_action.jl
```diff
@@ -0,0 +1,39 @@
+@doc doc"""
+    RotationAction(M::Manifold, SOn::SpecialOrthogonal)
+
+Space of actions of the [`SpectialOrthogonal`](@ref) group `\mathrm{SO}(N)`
+on a Euclidean-like manifold `M` of dimension `N`.
+"""
+struct RotationAction{TM<:Manifold,TSO<:SpecialOrthogonal} <: AbstractActionOnManifold
```
sethaxen: I believe in the earlier issue we had discussed that there was no canonical action of a rotation group on a manifold, with the example we discussed being action of `SO(2)` on `ℝ³`. The methods below would apply just fine if typed on something like `RotationAction{Euclidean{Tuple{N}},SpecialOrthogonal{N}}`, or in fact, any `Euclidean` where the first dimension is `N`.

src/groups/euclidean_group.jl
```diff
@@ -0,0 +1,17 @@
+@doc doc"""
+    EuclideanGroup{N} <: GroupManifold{Euclidean{N},AdditionOperation}
+
+Euclidean group `\mathrm{R}(N)` represented by translation arrays.
```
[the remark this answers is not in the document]
mateuszbaran: Right, I'll rename it.

src/groups/group.jl
```diff
@@ -0,0 +1,326 @@
...
+abstract type AbstractGroupOperation end
+
+@doc doc"""
+    AbstractGroupManifold{<:AbstractGroupOperation} <: Manifold
+
+Abstract type for a Lie group, a group that is also a smooth manifold with a
+smooth binary operation.
+`AbstractGroupManifold`s must implement at least [`inv`](@ref),
+[`group_id`](@ref), [`compose_left`](@ref), and [`translate_diff`](@ref).
+Group manifolds by default assume a left-invariant canonical metric.
```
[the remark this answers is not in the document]
mateuszbaran: This is a good question actually. I would prefer to make a minimal implementation in this PR so perhaps the question of left- or right-invariance should be left aside and all metric-related functions should be just forwarded to the decorated manifold. I'd also leave the Lie-theoretic exponential map out for now. What do you think about it?

src/groups/group.jl
```diff
@@ -0,0 +1,326 @@
+"""
+    AbstractGroupOperation
+
+Abstract type for smooth binary operations on elements of a Lie group.
```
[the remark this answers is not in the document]
mateuszbaran: Good idea, I'll add an equation here.

src/groups/group.jl
```diff
@@ -0,0 +1,326 @@
...
+end
+
+inv(::AbstractGroupManifold, e::Identity) = e
+
+function group_id!(G::AbstractGroupManifold, x)
+    error("group_id! not implemented on $(typeof(G)) for point $(typeof(x))")
+end
+
+@doc doc"""
+    group_id(G::AbstractGroupManifold, x)
```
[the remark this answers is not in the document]
mateuszbaran: I'd prefer to avoid overloading `Base.identity` here but it shouldn't cause any obvious problems.

src/groups/group.jl
```diff
@@ -0,0 +1,326 @@
...
+    compose_left(G::AbstractGroupManifold, xs...)
+
+Compose elements of `G` using their left translation upon each other.
+This composition is left-associative.
+"""
+compose_left(G::AbstractGroupManifold, xs...) = _group_afoldl(compose_left, G, xs...)
+compose_left(G::AbstractGroupManifold, x) = x
+function compose_left(G::AbstractGroupManifold, x, y)
+    z = similar_result(G, compose_left, x, y)
+    compose_left!(G, z, x, y)
```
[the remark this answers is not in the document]
mateuszbaran: I think I'll just replace _group_afoldl with a loop then.

src/groups/group.jl
mateuszbaran: Yes, sometimes it optimises but this should be non-allocating when `z` doesn't alias with `x` and `y`. I'll see if using `Base.dataids` would make sense.

src/groups/group.jl
mateuszbaran: Good catch, this case should have a separate method.

src/groups/group_action.jl
mateuszbaran: We can rename it to `g_manifold` then?

src/groups/group_action.jl
mateuszbaran: Right, it's a misleading description.

src/groups/group_action.jl
mateuszbaran: OK, I'll rephrase it. The metric from the g-manifold is used to define "closest". I'll add it to the docstring.

src/groups/rotation_action.jl
mateuszbaran: I was thinking about using this action in three cases:
1. action of SO(n) on R^n,
2. action of SO(n) on [landmark manifolds on R^n](https://academic.oup.com/blms/article-abstract/16/2/81/325787?redirectedFrom=fulltext),
3. action of SO(n) on manifolds of curves on R^n.

I should probably tighten method signatures here.

src/groups/group_action.jl
mateuszbaran: So we could rename this to `g_manifold`?

mateuszbaran:
I think this PR is feature-complete after these recent changes. New features can be added in other PRs. I'll just add some tests to improve the coverage.

src/groups/group.jl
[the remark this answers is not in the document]
sethaxen: Yeah probably best for now. How the Lie exponential could be implemented/used in general (not just for matrix manifolds) and the question of the metric was what hung up the last PR.

src/groups/group.jl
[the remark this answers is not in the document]
sethaxen: I think if we're already overloading `Base.inv` then there's no less reason to overload `Base.identity`.

src/groups/group_action.jl
sethaxen: I think so. It's an annoying name, but it's consistent with the usual terminology.

src/groups/rotation_action.jl
sethaxen: I don't know anything about landmark manifolds on R^n, but the other two sound good. I think additionally you could have action of SO(n) on SO(n).

sethaxen: These don't all necessarily need to be in this PR.

src/groups/translation_group.jl
```diff
@@ -0,0 +1,17 @@
+@doc doc"""
+    TranslationGroup{N} <: GroupManifold{Euclidean{N},AdditionOperation}
+
+Translation group `\mathrm{R}(N)` represented by translation arrays.
```
sethaxen: I prefer the notations `\mathrm{T}(N)`, `\mathrm{T}^N`, or `\mathbb{R}^N`

src/groups/group.jl
```diff
@@ -0,0 +1,514 @@
...
+The identity element of the group `group`.
+"""
+struct Identity{G<:AbstractGroupManifold}
+    group::G
+end
+
+(e::Identity)(x) = identity(e.group, x)
+
+@doc doc"""
+    inv(G::AbstractGroupManifold, x)
```
sethaxen: This docstring needs to move to `inv`, or doc references error.

src/groups/group.jl
```diff
@@ -0,0 +1,514 @@
...
+
+function identity!(G::AbstractGroupManifold, y, x)
+    error("identity! not implemented on $(typeof(G)) for points $(typeof(y)) and $(typeof(x))")
+end
+
+@doc doc"""
+    identity(G::AbstractGroupManifold, x)
+
+Identity element $e$, such that for any element $x$,
+$x \cdot e = e \cdot x = x$. The returned element is of a similar type to `x`.
```
sethaxen: Use the same notation for group operation here as in the docstring for `AbstractGroupOperation`

src/groups/rotation_action.jl
```diff
@@ -0,0 +1,39 @@
+@doc doc"""
+    RotationAction(M::Manifold, SOn::SpecialOrthogonal)
+
+Space of actions of the [`SpecialOrthogonal`](@ref) group `\mathrm{SO}(N)`
```
sethaxen:
```suggestion
Space of actions of the [`SpecialOrthogonal`](@ref) group $\mathrm{SO}(N)$
```

src/groups/group_action.jl
```diff
@@ -0,0 +1,99 @@
...
+    g_manifold(A::AbstractActionOnManifold)
+
+The manifold the action `A` acts upon.
+"""
+function g_manifold(A::AbstractActionOnManifold)
+    error("g_manifold not implemented for $(typeof(A)).")
+end
+
+"""
+    apply(A::ActionOnManifold, x, a)
```
sethaxen:
```suggestion
    apply(A::AbstractActionOnManifold, x, a)
```

src/groups/group_action.jl
```diff
@@ -0,0 +1,99 @@
...
+
+Apply action `a` to the point `x`. The action is specified by `A`.
```
sethaxen:
```suggestion
Apply action `a` to the point `x` with the rule specified by `A`.
```

src/groups/group_action.jl
```diff
@@ -0,0 +1,99 @@
...
+    copyto!(y, optimal_alignment(A, x1, x2))
+    return y
+end
+
+@doc doc"""
+    center_of_orbit(
+        A::AbstractActionOnManifold,
+        pts,
+        q
+        [, mean_method::AbstractEstimationMethod = GradientDescentEstimation()]
```
sethaxen: This uses 2 different notations for optional arguments. Should be either
`, mean_method::AbstractEstimationMethod = GradientDescentEstimation()` or
`[, mean_method::AbstractEstimationMethod]` with documentation for the default behavior when `mean_method` isn't given.

src/groups/group_action.jl
```diff
@@ -0,0 +1,99 @@
...
+function apply(A::AbstractActionOnManifold, x, a)
+    y = similar_result(A, apply, x, a)
+    apply!(A, y, x)
+    return y
+end
+
+@doc doc"""
+    optimal_alignment(A::AbstractActionOnManifold, x1, x2)
+
+Calculate an action element of action `A` acts upon `x1` to produce
```
sethaxen:
```suggestion
Calculate an action element of action `A` that acts upon `x1` to produce
```

src/groups/group.jl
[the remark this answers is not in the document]
mateuszbaran: I've made an issue for that: #75.

src/groups/group.jl
mateuszbaran: Right, so many methods are needed here...

src/groups/rotation_action.jl
mateuszbaran: I've opened an issue for this: #76

src/groups/group.jl
```diff
@@ -0,0 +1,514 @@
+@doc doc"""
+    AbstractGroupOperation
+
+Abstract type for smooth binary operations $\mu$ on elements of a Lie group $G$.
```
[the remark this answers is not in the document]
mateuszbaran: Do you actually mean group actions on sets or just group operations? I've changed it to $\circ$ for group operations but for now I've left $\cdot$ for other group actions.

src/groups/group.jl
```diff
@@ -0,0 +1,514 @@
...
+$x \cdot e = e \cdot x = x$. The returned element is of a similar type to `x`.
+"""
+function identity(G::AbstractGroupManifold, x)
+    y = similar_result(G, inv, x)
+    identity!(G, y, x)
+    return y
+end
+
+"""
+    compose_left(G::AbstractGroupManifold, xs...)
```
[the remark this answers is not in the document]
mateuszbaran: There could be `compose_right`. Would it be OK to call the two-argument operation just `compose` and leave `compose_left` as the function for an arbitrary number of arguments?

src/groups/group_action.jl
```diff
@@ -0,0 +1,99 @@
+"""
+    AbstractActionOnManifold
```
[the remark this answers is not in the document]
mateuszbaran: That's a good idea, I'll rename it.

src/groups/group.jl
```diff
@@ -0,0 +1,514 @@
...
+end
+function compose_left!(::GT, z, x::Identity{GT}, y) where GT<:AbstractGroupManifold{MultiplicationOperation}
+    copyto!(z, y)
+    return z
+end
+function compose_left!(::GT, z, x, y::Identity{GT}) where GT<:AbstractGroupManifold{MultiplicationOperation}
+    copyto!(z, x)
+    return z
+end
+function compose_left!(G::GT, z, x::Identity{GT}, y::Identity{GT}) where GT<:AbstractGroupManifold{MultiplicationOperation}
```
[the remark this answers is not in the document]
mateuszbaran: That function currently is not a part of our interface, we have `identity(G, z)` and `identity!(G, y, x)`. Do you think it's a good idea to add `identity!(G, z)` somehow?

src/groups/group.jl
mateuszbaran: I think it's OK now.

src/groups/translation_group.jl
mateuszbaran: I'll change it to `\mathrm{T}(N)` then.

src/groups/group.jl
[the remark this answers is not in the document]
sethaxen: Yes, I mean for just group operations. Sometimes $\cdot$ is used for both, but I prefer the explicit indication of composition with $\circ$.

src/groups/group.jl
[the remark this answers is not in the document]
sethaxen: The terms composition, group operation, and group law are used interchangeably to describe the binary group operation. Composition therefore takes `$\circ: g,h \to g \circ h$`. I've rarely seen a version of composition defined `$\circ: g,h \to h \circ g$`.

Left and right translation, on the other hand, _are_ commonly defined by designating one of the elements as the element being translated and the other as the element doing the translating. (left/right _actions_ often have more complicated definitions by requiring that left-action be left-associative and right-action be right-associative, a property that is trivially satisfied for translations).

So I think it makes the most sense to have a single `compose` function that is binary. But because composition is by definition associative, there's an argument to be made for permitting it to have arbitrary numbers of arguments, like the [associative scalar functions in base](https://github.com/JuliaLang/julia/blob/89a51fb98485219680c7106e94cf75fd8069f3f8/base/operators.jl#L512-L534), which was the rationale for the [`compose` function in the original PR](https://github.com/JuliaNLSolvers/Manifolds.jl/blob/43426c7c9009d6af926a382219302289188101fa/src/Group.jl#L91-L109).

Anyways, those are just my 2 cents. In general, I think we should avoid functions with explicit `_left` and `_right` in the names, since we have `LeftAction` and `RightAction`.

src/groups/group.jl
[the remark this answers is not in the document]
sethaxen: In the original comment, I also meant to replace $\mu$ with $\circ$. It doesn't matter much, sometimes $m$ or $\mu$ is used to denote the group law, but I think here it makes more sense to use consistent notation.

sethaxen:
What's the long-term plan for supporting right actions? I assume by existence and naming of `LeftAction` and `RightAction` there's a plan for `AbstractGroupAction` to support some typing by direction of action. What if we had `AbstractGroupAction{TD<: ActionDirection}`?

For `TranslationAction`, the type parameter can be ignored.

In general, left and right actions can be converted into one another using the inverse. Because the matrix groups all have canonical left actions, I think it makes the most sense to define the default right action in terms of the left.

Pirating the notation for left translation, the left and right actions can be written with their associative properties as 

```math
\begin{align}
L_h (L_g (x)) &= h \cdot (g \cdot x) = (h \circ g) \cdot x = L_{h \circ g} (x)\\
R_h (R_g (x)) &= (x \cdot g) \cdot h = x \cdot (g \circ h) = R_{g \circ h} (x)
\end{align}
```

i.e. applying successive left actions has the same affect as applying successive left translations to group elements and then applying the resulting action, while applying right actions has the same effect but for right translations. We can satisfy the latter property by defining the default right action as the inverse left action of the element, i.e. `$R_g (x) = L_{g^{-1}} (x)$`. Then,
```math
R_h (R_g (x)) = h^{-1} \cdot (g^{-1} \cdot x) = (h^{-1} \circ g^{-1}) \cdot x = (g \circ h)^{-1} \cdot x = R_{g \circ h)} (x),
```
and we satisfy the necessary associativity.

Incidentally, for `RotationAction` in kinematics this is the difference between intrinsic/extrinsic rotations or body-fixed/space-fixed reference frames. If I define a sequence of rotations of a frame `F` with respect to some global frame `O`, then they compose by left translation and act via the left action on `F`. The same sequence of rotations expressed wrt `F` compose by right translation and act via the kind of right action defined here.

src/groups/group.jl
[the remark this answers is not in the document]
mateuszbaran: OK, I'll change that.

src/groups/group.jl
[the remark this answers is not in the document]
mateuszbaran: > The terms composition, group operation, and group law are used interchangeably to describe the binary group operation. Composition therefore takes `$\circ: g,h \to g \circ h$`. I've rarely seen a version of composition defined `$\circ: g,h \to h \circ g$`.

In this notation `compose_left` and `compose_right` would only differ for >= 3 arguments. `compose_left(G, x, y, z) = (x \circ y) \circ z`,  `compose_left(G, x, y, z) = x \circ (y \circ z)`. Essentially left fold vs right fold. I know they should give the same answer (by associativity) but that's where the `left` came from. Moreover, I've written *should* because in some cases like the group of reparametrizations of a curve things slightly break due to discretization.

I think I'll (for now) rename `compose_left` to `compose` and leave only the binary case. Other variants can be added later.

src/groups/group.jl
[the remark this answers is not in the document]
mateuszbaran: I'm not sure. in this case `z` could be an invalid input and things could go wrong anyway. What do you think `identity!(G, y, x)` should be able to assume about `x`?

mateuszbaran:
> What's the long-term plan for supporting right actions? I assume by existence and naming of `LeftAction` and `RightAction` there's a plan for `AbstractGroupAction` to support some typing by direction of action. What if we had `AbstractGroupAction{TD<: ActionDirection}`?

Good question, I didn't think about it that much. I like the idea of a type parameter. We also don't currently have action composition and it's certainly necessary for that. I'll implement it in this PR.

mateuszbaran:
I've added composition of actions that respects action direction.

sethaxen:
I like the composition of actions!

I think as implemented (with action as a type parameter of `AbstractGroupAction`), it might make it a bit awkward to define canonical right actions in terms of left (which I think we should have): e.g.

```julia
switch_direction(::LeftAction) = RightAction()
switch_direction(::RightAction) = LeftAction()

function switch_direction(A::RotationAction{TM,TSO,TAD}) where {TM,TSO,TAD}
    return RotationAction(A.M, A.SOn, switch_direction(TAD()))
end

function apply!(A::AbstractGroupAction{<:RightAction}, y, x, a)
    G = base_group(A)
    apply!(swap_direction(A), y, x, inv(G, a))
    return y
end
```

This would require a `switch_direction` implementation for every action that just switches that type parameter. But the action direction as a type is necessary for composition of actions.

mateuszbaran:
I've pushed some changes regarding left/right rotation actions. I don't really want to go far in this direction, I've only used left actions. BTW, I'm not sure the canonical right action is particularly useful here.

mateuszbaran:
For rotations multiplication by rotation matrix from the right and the canonical right action are not the same thing IIUC

sethaxen:
>For rotations multiplication by rotation matrix from the right and the canonical right action are not the same thing IIUC

No, they're not. I've detailed [here](https://github.com/JuliaNLSolvers/Manifolds.jl/pull/68#issuecomment-569471037) what I believe the right canonical action should be defined as and a common use case. The currently implemented right action on `RotationAction` is a no-op since it only works for `SO(1)`.

I don't want to hold up this PR. After this is merged, I can submit another PR with the changes I'm thinking of, and we can see if you agree.

sethaxen:
As long as all the tests pass and the coverage looks good, this looks good to me.

mateuszbaran:
OK, should I remove my implementation of the right rotation action then or will you handle that in your PR?

sethaxen:
Either way doesn't matter. I'll be replacing it in that PR.

mateuszbaran:
Thanks for the review, I think tests and coverage should be OK after the last commit. I'll merge it if tests pass.

PR #70, opened 2019-12-23
introduce concrete Field type for Manifolds

### LINKED ISSUE #62: Use concrete types for the field a manifold is build upon
The additional type `T` we use to distinguish real- and complex-valued manifolds should be a concrete type, for example following https://github.com/Jutho/TensorKit.jl/blob/ae270985e1d71bc982d912144d98ef8b560da9e2/src/spaces/vectorspaces.jl#L10-L19 as suggested here https://github.com/JuliaNLSolvers/Manifolds.jl/pull/49#discussion_r358119733
and here https://github.com/JuliaNLSolvers/Manifolds.jl/pull/49#discussion_r358122101

kellertuer:
to close #62 .

### DIFF
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -105,6 +105,22 @@ using UnsafeArrays
 using Einsum: @einsum
 
 """
+    Field
+
+An abstract type to represent the field matrix manifolds are build upon,
+following the idea of [TensorKit](https://github.com/Jutho/TensorKit.jl)
+in order to hace concrete field types to dispatch on. The two most common
+field types are `RealNumbers` (`ℝ` for short) and `ComplexNumbers` (`ℂ`).
+"""
+abstract type Field end
+
+struct RealNumbers <: Field end
+struct ComplexNumbers <: Field end
+
+const ℝ = RealNumbers()
+const ℂ = ComplexNumbers()
+
+"""
     AbstractEstimationMethod
 
 Abstract type for defining statistical estimation methods.
@@ -236,7 +252,8 @@ export Manifold,
     TVector,
     CoTVector,
     SVDMPoint,
-    UMVTVector
+    UMVTVector,
+    ℝ, ℂ
 # decorator manifolds
 export ArrayManifold,
     ArrayMPoint,
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -105,17 +105,17 @@ using UnsafeArrays
 using Einsum: @einsum
 
 """
-    Field
+    AbstractField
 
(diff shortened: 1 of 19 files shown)


src/Manifolds.jl
```diff
@@ -104,6 +104,22 @@ using FiniteDifferences
...
 
+"""
+    Field
+
+An abstract type to represent the field matrix manifolds are build upon,
+following the idea of [TensorKit](https://github.com/Jutho/TensorKit.jl)
+in order to hace concrete field types to dispatch on. The two most common
+field types are `RealNumbers` (`ℝ` for short) and `ComplexNumbers` (`ℂ`).
+"""
+abstract type Field end
```
mateuszbaran: Maybe we should call this `AbstractField`?

test/euclidean.jl
```diff
@@ -2,14 +2,16 @@ include("utils.jl")
...
+    Ec = Manifolds.Euclidean(3;field=ℂ)
     EM = Manifolds.MetricManifold(E,Manifolds.EuclideanMetric())
     @test is_default_metric(EM) == Val{true}()
     @test is_default_metric(E,Manifolds.EuclideanMetric()) == Val{true}()
     x = zeros(3)
     @test det_local_metric(EM,x) == one(eltype(x))
     @test log_local_metric_density(EM,x) == zero(eltype(x))
     @test project_point!(E,x) == x
-
+    @test manifold_dimension(Ec) == 2*manifold_dimension(E)
```
mateuszbaran: I think we should run full tests on `Ec`.

src/Manifolds.jl
kellertuer: I don't mind and it would be consistent with our naming scheme, you're right.

test/euclidean.jl
kellertuer: No problem, I can do a few numbers for that manifold (though Euclidean really has a lot of tests already currently).

mateuszbaran: Calling `test_manifold` should be enough I think.

mateuszbaran:
I've added some more tests for the complex Euclidean case. It looks like there were problems in the test suite and random points and tangent vectors don't work right now in the complex case.

kellertuer:
Code coverage is a little strange, since it says, that `Euclidean` was roduced but the homepage says it's still 100%. Is it ok to merge this?

mateuszbaran:
Coverage sometimes behaves strangely, I think it's still OK to merge this. I don't see any changes that could actually reduce the coverage of `Euclidean`.

PR #71, opened 2019-12-23
Introduces the Manifold Hyperbolic Space.

kellertuer:
As well as a corresponding algorithm for the mean.


kellertuer:
Hm, sadly there was again an update breaking some dependency.

mateuszbaran:
Wow, a few more manifolds and we'll have more of them than your `Manopt.jl` and `pymanopt` :tada:.

I'm not sure why that error shows but one thing we can do is relaxing the bound on `Requires.jl` (the newest version is 1.0.0)

kellertuer:
Well this one is transcribed from `Manopt.jl`, since as soon as we have all of them here, I can port that to use `ManifoldsBase` (though that changes quite some stuff). So my first aim is to get all of them over here from time to time.

An the list I am sometimes looking at is the one from manopt (so the third, Matlab toolbox), I think we might not surpass them, though.

kellertuer:
Oh theres indeed just `Circle` missing (to `Manopt.jl`), which I will do next then, but both on R (angles) and on C (which is nearly line ´Sphere(1)` just in the complex plane.

kellertuer:
Concering the Verion conflict its Compat and ReverseDiff fighting over FunctionWrappers so I fear we have to wait?

mateuszbaran:
Yes, the Matlab Manopt has a lot of manifolds and optimization algorithms. I think we have better API though :slightly_smiling_face: .

mateuszbaran:
> Concering the Verion conflict its Compat and ReverseDiff fighting over FunctionWrappers so I fear we have to wait?

That's unfortunate, I'd just wait a few days. We don't have to merge any PRs right now.

kellertuer:
Concerning the matlab manopt – yes he still has a few more than me, but his toolbox is widely recognized, 5 years old and does quite well in matlab (though our API is better, but that might also just be the language – Matlab is not much fun). He is aware of this project here and said, he will propose to his students to also take a look :)

sethaxen:
We really need to get our own lower-bounds set up, along with CompatHelper. I'll submit a PR sometime this week probably. I'm just concerned I'll set the lower-bound too high.

sethaxen:
I don't have time to review this today, but it looks like it's finally time to use `LorentzMetric`!
```julia
struct MinkowskiMetric <: LorentzMetric end
```

kellertuer:
Yes, we can of course add that line and make it the default metric, though I have really no other metric in mind one could employ on this `Hyperbolic`, so it would be merely for completeness (as we do for the EuclideanMetric in Eucliean).

mateuszbaran:
> I don't have time to review this today, but it looks like it's finally time to use `LorentzMetric`!
> 
> ```julia
> struct MinkowskiMetric <: LorentzMetric end
> ```

Are you sure it's right? I may be wrong by a Lorentz metric is supposed to be non-positive definite when expressed in chart coordinates while what we have here is just an embedding of the hyperbolic manifold in a Minkowski space.

kellertuer:
I am not 100% sure but in the embedding the Minkowski metric _is_ non-positive definite; just when restricting it to one (embedded) tangent space it becomes positive definite.

mateuszbaran:
> I am not 100% sure but in the embedding the Minkowski metric _is_ non-positive definite; just when restricting it to one (embedded) tangent space it becomes positive definite.

Yes, that's what I think. The Minkowski metric isn't positive definite but the metric of the hyperbolic manifold is. It might be worth clarifying in the docstring, right now someone might think that the metric of the hyperbolic manifold isn't positive-definite.

kellertuer:
I clarified the MinkowksyMetric in both the Hyperbolic and the MinkowskiMetric docstring and hopefully added the last remaining tests. If they succeed this is ready to merge.

mateuszbaran:
Could we also have a point projection? This: https://math.stackexchange.com/a/3176477 looks reasonable and is quite simple.

src/Hyperbolic.jl
```diff
@@ -0,0 +1,240 @@
...
+Return the dimension of the hyperbolic space manifold $\mathbb H^n$, i.e. $n$.
+"""
+manifold_dimension(::Hyperbolic{N}) where {N} = N
+
+"""
+    mean(
+        M::Hyperbolic,
+        x::AbstractVector,
+        [w::AbstractWeights,]
+        method = CyclicProximalPointEstimationMethod();
```
mateuszbaran: This line does not apply to this method.

kellertuer: I don't follow you? The default we use on hyperbolic is the newly implemented mean using the CPP?

kellertuer:
Of course I can add the projection of a point as soon as I have understood wherer \dot in that note is the inner product or the Minkowsky inner product.

src/Hyperbolic.jl
mateuszbaran: This is a bit confusing because the method of `mean!` below does not have the `method` argument. After looking more closely which method calls which one it should have the intended behaviour though.

kellertuer: hm? I still don't follow you, the `statistics` has exactely this method and this line introduces a default if you don't provide a method (which is the here newly introduces CPP for mean)...

kellertuer:
> Could we also have a point projection? This: https://math.stackexchange.com/a/3176477 looks reasonable and is quite simple.

Reading this link closely, this is projection onto a _curve_ (given by p0,v0) on the hyperboloid.

src/Hyperbolic.jl
kellertuer: To be precise, the one thing I _do_ overwrite (and hence the documentation) is the _default_ method for estimation. It is changed form the grad desc to the CPP. and that's what's documented.

mateuszbaran:
We can probably project point `x` onto a curve (geodesic) defined by the tip of the hyperboloid `p0=(0,0,...,0,1)` and tangent vector projected from `x-p0` at `p0`?

src/Hyperbolic.jl
mateuszbaran: Yes, I understand. I was just confused but it's OK now.

kellertuer:
I am not sure whether that yields the projection which is the actual projection – I looked for a reference, but haven't found one yet.

kellertuer:
All new lines enjoy test coverage. So we could merge this and as soon as we find new formulae or figure out the projection, we could also do a new PR?

mateuszbaran:
Sure, projection can be done in another PR

PR #72, opened 2019-12-24
Introduces the manifold of the Circle

kellertuer:
This is a first step towards introducing the circle, though I am (again) struggling with our `similar_result` which seems to not like points on the manifold being plain `Float64`s.

This PR is based on #70 since it already employs said types.


mateuszbaran:
`similar_result` can't work with plain immutable numbers. Mutating variants don't work either so that's OK I think? If you want non-mutating variant that work with plain numbers I you'd have to implement them separately (which, BTW, is most likely needed for Zygote compatibility anyway).

kellertuer:
Ah, I see – that's not a problem, the formulae are really short and easy anyways. I will add them.

kellertuer:
It was a little more work, when I realized, that the points and tangents of `Circle` (both real and complex) are indeed `immutable` so there is no sense to write `mutating` functions. I adopted that and saw that in the tests thats just affecting 3 places (added another variable for that). And locally there wasn't even a package conflict (in Julia 1.3).

mateuszbaran:
Are you sure immutable numbers are the best option here? A one-element `SVector` should be just as fast and the code would then be compatible with mutation using `MVector` for example.

BTW, `FunctionWrappers.jl` is not marked as compatible with `Compat.jl` v3.1 so this might be a problem here.

kellertuer:
That the problem is somewhere between `FunctionWrapper.jl`, `Compat.jl` (3.1) and `ReverseDiff` (which wants to have the wrapper), I know, I don't know how to resolve that.

Concerning the type – I don't know exactely either, I have to check `SVector` and`MVector` for that since I haven't worked with them. It would make the compability toother manifolds easier. For that it would be nice to have a mutating type; however I always feel strange when using vectors of dimension 1 instead of numbers, also (or maybe especially) when initializing such data or later extracting the numbers again for an export.

mateuszbaran:
Instead of one-element vector, we can also use `Ref` in some places. I've just checked and the code of the circle manifold should be able to assume that the value is accessible by dereferencing with no index:

```julia
julia> using StaticArrays

julia> a = Ref(10)
Base.RefValue{Int64}(10)

julia> a[] = 12
12

julia> a[]
12

julia> b = MVector(10)
1-element MArray{Tuple{1},Int64,1,1} with indices SOneTo(1):
 10

julia> b[] = 12
12

julia> b[]
12
```
This looks nice enough I think.

kellertuer:
that looks fine. What do you think about keeping the current functions for `<:Real` (respectively `<:Complex`)? That way one _could_ use it for floats but not all things will work with that (the mean is overwritten currently but maybe others might not exist for float). Our default will be your suggestion and that works as expected with everything (AD not yet since you can not derive `rem` that easily), but a simple application might be fine with the (lets say reduced) Circle and Floats.

mateuszbaran:
That's a good plan, I think we can keep the current functions.

kellertuer:
This one is just missing a few more tests for the immutable datatypes introdued. I will work on that the next days somewhen.

PR #78, opened 2019-12-30
Use Hermitian inner product and reduce number of tests

sethaxen:
Following https://github.com/JuliaNLSolvers/ManifoldsBase.jl/pull/18, don't constrain inner products for complex manifolds to be real.

### DIFF
--- a/test/grassmann.jl
+++ b/test/grassmann.jl
@@ -13,10 +13,11 @@ include("utils.jl")
             @test_throws DomainError is_tangent_vector(M, [1. 0.; 0. 1.; 0. 0.],ones(3,2),true)
             @test injectivity_radius(M) == π/2
         end
-        types = [Matrix{Float64},
-                 MMatrix{3, 2, Float64},
-                 Matrix{Float32},
-                 MMatrix{3, 2, Float32}]
+        types = [
+            Matrix{Float64},
+            MMatrix{3, 2, Float64},
+            Matrix{Float32},
+        ]
         @testset "Type $T" for T in types
             x = [1.0 0.0; 0.0 1.0; 0.0 0.0]
             v = [0.0 0.0; 0.0 0.0; 0.0 1.0]
@@ -67,9 +68,6 @@ include("utils.jl")
         end
         types = [
             Matrix{ComplexF64},
-             MMatrix{3, 2, ComplexF64},
-             Matrix{ComplexF32},
-             MMatrix{3, 2, ComplexF32},
         ]
         @testset "Type $T" for T in types
             x = [0.5+0.5im 0.5+0.5im; 0.5+0.5im -0.5-0.5im; 0.0 0.0]
@@ -89,7 +87,7 @@ include("utils.jl")
                 test_reverse_diff = false,
                 retraction_methods = [PolarRetraction(), QRRetraction()],
                 inverse_retraction_methods = [PolarInverseRetraction(), QRInverseRetraction()],
-                exp_log_atol_multiplier = eltype(T) === ComplexF32 ? 5000.0 : 10.0,
+                exp_log_atol_multiplier = 10.0,
                 is_tangent_atol_multiplier = 10.0,
             )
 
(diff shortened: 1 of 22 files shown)


test/grassmann.jl
```diff
@@ -85,7 +89,7 @@ include("utils.jl")
                 test_reverse_diff = false,
                 retraction_methods = [PolarRetraction(), QRRetraction()],
                 inverse_retraction_methods = [PolarInverseRetraction(), QRInverseRetraction()],
-                exp_log_atol_multiplier = 10.0,
+                exp_log_atol_multiplier = 5000.0,
```
mateuszbaran: Could you set a lower multiplier for `ComplexF64`?

sethaxen:
Any tips on how to fix the failing x86 Windows builds? Even the Float64 checks fail there, and the tolerances look really low still.

mateuszbaran:
These Windows x86 builds have tolerance issues in many tests and slow down AppVeyor CI quite a bit. Maybe we should just get rid of them? I don't think there is anyone using 32-bit Windows for numerical computing.

sethaxen:
For Julia 1, it looks like we're sometimes exceeding the maximum time limit of 50 minutes. There are a few ways we can handle this:
- Remove some test cases
- Don't test against Julia 1
- Permit failures against Julia 1
- Split tests into different stages
- Other solutions?

@kellertuer @mateuszbaran what are your thoughts?

mateuszbaran:
The time limit issue doesn't look good and will only get worse with more manifolds. I think we should keep testing against Julia 1.0 for now. What do you mean by splitting tests into different stages?

For now, the easiest solution would be removing `Float32` `MArray` tests, `StaticArrays` compiles slowly and the `Float32` case isn't that useful. Long term, we may have to split the package similarly to `DifferentialEquations`.

kellertuer:
How would that splitting work? We would do `BasicManifolds.jl` and such? Should we then found our own Organisation (JuliaManifolds or something)?

In general I think we don't have to check all combinations, so if it works for `Float64`, `Float32` and `MArray{Float64}` I'd be fine (similarly for other decorator-style types).

sethaxen:
> The time limit issue doesn't look good and will only get worse with more manifolds. I think we should keep testing against Julia 1.0 for now. What do you mean by splitting tests into different stages?

We currently have 2 build stages: tests and documentation. We can further split the tests into several groups; each group would have its own stage (see https://docs.travis-ci.com/user/build-stages/define-steps/). The downside here is that the total wallclock time for testing and the number of jobs will go up.

I'm fine with reducing the number of array/number types that we test against. It's a band-aid, but a good one for now.

mateuszbaran:
Yes, something like `BasicManifolds`, `GroupManifolds` etc. I don't recommend rushing it, coordinating changes across many packages is more difficult than working in a single repository. `ManifoldsBase.jl` is already separated which will help.

Removing unnecessary tests is much easier than making new packages.

> Should we then found our own Organisation (JuliaManifolds or something)?

Maybe? JuliaNLSolvers maintainers keep setting up everything we need though.

> We currently have 2 build stages: tests and documentation. We can further split the tests into several groups; each group would have its own stage (see https://docs.travis-ci.com/user/build-stages/define-steps/). The downside here is that the total wallclock time for testing and the number of jobs will go up.

Nice, I didn't know about this. BTW, I think we can remove Julia 1.1 to reduce the test time. Only the latest stable and LTS releases of Julia are officially supported and that's 1.0 and 1.3 currently.

kellertuer:
Yes, let's not rush. For more than the current 2 packages, I would be in favour of an own organisation, I am sure we can manage to set that up, too. 

And of course, it's simpler to stay with 2 packages as long as we can, and moving a few parts over to base would also be an idea, as soon as certain parts of the interface are more stable.

sethaxen:
Travis's OS X builds seem to hang in the queue today. When they finish running, I'll merge this PR unless there are any objections to the reductions I've made to the tests.

Issue #81, opened 2019-12-31
Enhancements to group actions


sethaxen:
`apply` and `translate` have a lot of similarities. Both apply elements of a group to act on a manifold, with `translate` being a special case where the action is the group operation. For semidirect product groups, where one must specify how one group acts on the other, we will need actions to have some of the extra functionality that translations have, namely inverse actions, differential of the action, and differential of the inverse action. Moreover, while the signature for translations puts the element doing the translating on the left, while the signature for actions puts the element acting on the right.

I'd like to unify these 2 interfaces. I'm thinking of making a `GroupOperationAction{TG,TAD}` that wraps a group. `apply!` with this action would always forward to `translate!`. Likewise the new `inverse_apply!`, `apply_diff!`, and `inverse_apply_diff!` would forward to their corresponding translation interfaces. I'd also like to make the two signatures more similar by moving the action element to the left.

@mateuszbaran how does this sound?

mateuszbaran:
This sounds right, I saw the similarity but did not think about it that much. Canonical action of a group on itself is a good thing to have. The group interface itself remains separate and that's good. I'm OK with moving the action element to the left.

I'd be happy to see these semidirect products, I've implemented a few explicitly but having it partially automated would be nice 🙂 .

Issue #69, opened 2019-12-22
Syntax for equipping a manifold with structure


sethaxen:
We current have `MetricManifold`, which equips a `Manifold` with a `Metric`. And soon thanks to #68 we will have `GroupManifold`, which equips a manifold with an `AbstractGroupOperation`. I'd like to propose a simple syntax for "equipping" a manifold with a structure:

```julia
(g::Metric)(M::Manifold) = MetricManifold(M, g)
(op::AbstractGroupOperation)(M::Manifold) = GroupManifold(M, op)
```

This lets us have a convenient notation for equipping:
```julia
M = Rotations(3) |> MultiplicationOperation() |> LeftInvariantCanonicalMetric()
```

which is equivalent to
```julia
M = MetricManifold(GroupManifold(Rotations(3), MultiplicationOperation()), LeftInvariantCanonicalMetric())
```

but much clearer/more pleasant to read. Of course, we could choose another symbol.

mateuszbaran:
Good idea, the pipe syntax looks quite nice here.

PR #82, opened 2020-01-01
Add direct and semidirect product groups

sethaxen:
Following up #68, this PR implements some of the suggestions in #81. And since we now have actions on groups, this PR implements the semidirect product group, which is just an alias for a `GroupManifold` with a `SemidirectProductOperation`, which wraps two groups and an action by one group on the other. A simple example is given using `SpecialEuclidean(N)`.

This PR doesn't yet add any tests, and it breaks some old tests.

I haven't figured out how to implement the differential of right translation for the semidirect product group, but I'm also not convinced that the current differential code does what it says it does (I think it came from my old PR #25). @mateuszbaran do you have a use case for the differentials? I'm not certain why I implemented them. 🙂


mateuszbaran:
Great, I'll look at the code in a moment.

> I haven't figured out how to implement the differential of right translation for the semidirect product group, but I'm also not convinced that the current differential code does what it says it does (I think it came from my old PR #25). @mateuszbaran do you have a use case for the differentials? I'm not certain why I implemented them. 🙂

The differentials are from your old PR, yes. I don't really need them for anything but they looked quite important so I decided to put them in my PR (who knows, maybe I'll need them one day). BTW, I didn't verify your code for differentials.

mateuszbaran:
Hmm... I just realised I may need these differentials of group operations and actions after all. They may solve some of my problems with quotient manifolds. Anyway it doesn't look important enough to require it for all group operations and actions.

sethaxen:
> BTW, I didn't verify your code for differentials

I'm not convinced that any of the implemented differentials do what the docstrings claim. The docstrings claim that the action of the differential transports a tangent vector from a point to a translated point. But take the `MultiplicationOperation` implementation as an example:

https://github.com/JuliaNLSolvers/Manifolds.jl/blob/086ba4ea67405ca179b64f80f430158210d963e9/src/groups/group.jl#L460-L466

If `v \in T_y G`, then `x * v` will in general not be in `T_x G`, as the docstring claims. The exception is when `y = e`, where `e` is the identity element, but even in that case, I'm not confident enough regarding the conditions that must be satisfied for that statement to be true. This differential certainly isn't correct for `SpecialOrthogonal`, where the tangent space at any point is represented using the Lie algebra. I'd prefer that we remove implementations until we've verified the math/code.

mateuszbaran:
Yes, right, differentials for these left translations should be a no-op for `SpecialOrthogonal` and `transpose(x) * v * x` looks right for the right translation. I think it's a good idea then to just remove these definitions for differentials on abstract group operations and leave just the correct implementations for specific groups. The scaling group is also multiplicative but has a different differential for example.

sethaxen:
> I think it's a good idea then to just remove these definitions for differentials on abstract group operations and leave just the correct implementations for specific groups.

To clarify, are you recommending removing the differential implementations for `AbstractGroupManifold{MultiplicationOperation}` and `AbstractGroupManifold{AdditionOperation}`?

mateuszbaran:
Yes, I think we should remove `translate_diff` and `inverse_translate_diff` for `AbstractGroupManifold{MultiplicationOperation}`. The methods of these two functions for `AbstractGroupManifold{AdditionOperation}` are much safer defaults so maybe we can leave that.

src/groups/semidirect_product_group.jl
```diff
@@ -0,0 +1,148 @@
...
+end
+
+function translate_diff(
+    G::SemidirectProductGroup,
+    x,
+    y,
+    v,
+    conv::ActionDirection = LeftAction(),
+)
+    vout = similar_result(base_manifold(G), x, y, v)
```
mateuszbaran:
```suggestion
    vout = similar_result(base_manifold(G), translate_diff, x, y, v)
```

src/groups/special_orthogonal.jl
```diff
@@ -14,20 +14,47 @@ end
 
 # optimized inverseion for the special orthogonal group
```
mateuszbaran: This comment doesn't apply here after this change, it should be removed or moved to `inv(G::SpecialOrthogonal, x)`

src/groups/semidirect_product_group.jl
```diff
@@ -64,24 +64,27 @@ function show(io::IO, G::SemidirectProductGroup)
     )
 end
 
+_subarray(::SemidirectProductGroup, x, ::Val{i}) where {i} = submanifold_component(x, i)
+_subarray(G::SemidirectProductGroup, x, i::Int) = _subarray(G, x, Val(i))
```
mateuszbaran: Instead of this function I think it's a better idea to directly call `submanifold_component`. Probably with new methods that support `Val(i)` indexing.

BTW, I don't think the second method of `_subarray` helps under any circumstances. Either everything is inlined and constant-propagated or not and wrapping an argument in `Val` doesn't improve performance.

src/groups/special_euclidean.jl
```diff
@@ -39,3 +39,32 @@ function SpecialEuclidean(n)
 end
 
 show(io::IO, ::SpecialEuclidean{n}) where {n} = print(io, "SpecialEuclidean($(n))")
+
+function _subarray(::SpecialEuclidean{n}, x::AbstractMatrix, ::Val{1}) where {n}
```
mateuszbaran: Type parameters usually start with a capital letter.

src/groups/semidirect_product_group.jl
```diff
@@ -107,14 +111,15 @@ function translate_diff!(G::SemidirectProductGroup, vout, x, y, v, conv::LeftAct
     N = submanifold(PG, 1)
     H = submanifold(PG, 2)
     A = G.op.action
-    nx, hx = submanifold_component.(Ref(x), (1, 2))
-    ny, hy = submanifold_component.(Ref(y), (1, 2))
-    nv, hv = submanifold_component.(Ref(v), (1, 2))
-    nvout, hvout = submanifold_component.(Ref(vout), (1, 2))
+    nx, hx = _subarray.(Ref(G), Ref(x), Val.((1, 2)))
```
mateuszbaran: I think a simple `ntuple(i -> _subarray(G, x, Val(i)), Val(2))` would be cleaner here

src/manifolds/Euclidean.jl
```diff
@@ -231,7 +237,7 @@ end
 Return the array dimensions required to represent an element on the
 [`Euclidean`](@ref) `M`, i.e. the vector of all array dimensions.
 """
-@generated representation_size(::Euclidean{N}) where N = Tuple(N.parameters...)
+Base.@pure representation_size(::Euclidean{N}) where {N} = tuple(N.parameters...)
```
mateuszbaran: As far as I understand `Base.@pure`, it's supposed to refer to the entire function, not just one method. So this is incorrect.

sethaxen: Not quite. It can be applied to a single method, but once a `Base.@pure` method for a function has been called then no new methods may be added (I think; see [this discussion](https://discourse.julialang.org/t/pure-macro/3871/2)). Either way, as written here this is dangerous, so I'll fix it.

src/groups/special_euclidean.jl
sethaxen: @kellertuer's recent refactor adopted the convention that dimension types for some manifolds are lower-case. This follows that convention, but I see now that it's not universal. We should be consistent either way.

src/groups/semidirect_product_group.jl
sethaxen: The issue with `submanifold_component` is that we can't dispatch on the manifold. Given the name, perhaps that's something we should change.

It would also be good to have something like `submanifold_components` that dispatches on a manifold and for `ProductArray` and `ProductRepr` returns the `parts` field. For the SE(N) case with matrix input, it would return views of the two subarrays. For a group manifold with `Identity` input, it would return a tuple with `Identity` of the subgroups.

These custom methods are necessary either way though, because there's structure outside the subarrays in the matrix that needs to be enforced for the matrix to still be affine.

src/manifolds/Euclidean.jl
sethaxen: Looks like we already have `size_to_tuple` for this, so we can just replace it.

mateuszbaran: Yes, either way it's incorrect to use `@pure` here.

src/groups/special_euclidean.jl
mateuszbaran: Yes. if we're going to use lower case letters for dimension then it should be consistent (although I prefer type parameters starting with a capital letter).

sethaxen: I don't have a strong preference here, only for consistency. There are a few docstrings that refer to the dimension, and in cases like those, to avoid confusion, I think the type parameter for the dimension in the docstring at least should be lower case.
@kellertuer thoughts?

src/groups/semidirect_product_group.jl
mateuszbaran: > The issue with `submanifold_component` is that we can't dispatch on the manifold. Given the name, perhaps that's something we should change.

I think it's better to add a new argument to `submanifold_component` then and simplify `_subarray`.

> It would also be good to have something like `submanifold_components` that dispatches on a manifold and for `ProductArray` and `ProductRepr` returns the `parts` field. For the SE(N) case with matrix input, it would return views of the two subarrays. For a group manifold with `Identity` input, it would return a tuple with `Identity` of the subgroups.

OK, we can have that.

sethaxen:
The feature set here should be more or less complete, though it needs quite a few more tests, and I'm sure there are some bugs that need fixing. Still, a review would be nice to see if any of the new functionality needs a major rewrite.

I implemented trait-based forwarding of group functions to the decorated manifold until an implementation is found. I also co-opted our existing `base_group` to behave like `base_manifold` but stop when an `AbstractArrayGroup` implementation is found.

I added semantic checking with `ArrayManifold` (points 1 and 2 discussed [here](https://github.com/JuliaNLSolvers/Manifolds.jl/issues/84#issuecomment-570528610)). When the group API stabilizes sufficiently to move to ManifoldsBase, then those functions can move over there.

I also added `ProductManifold` variants of the group functions. In general, this means a `ProductManifold` with group members is functionally a direct product group. However, for additional features (like `Identity`), one can explicitly build a `ProductGroup`.

I added new `submanifold_component` signatures and a new `submanifold_components`. These are now used everywhere internally, which simplifies the code a bit. This also lets us have multiple product representations, and in fact a user could specify their own; they no longer are tied to our `.parts` field requirement. As an example, SE(N) supports an affine matrix representation.

Lastly, I added the piping syntax for decoration discussed in #69.

mateuszbaran:
Thanks, sounds great :slightly_smiling_face: . I'll make a review soon.

test/groups/group_operation_action.jl
```diff
@@ -0,0 +1,43 @@
...
+    A_right = GroupOperationAction(G, RightAction())
+
+    types = [Matrix{Float64}]
+
+    @test g_manifold(A_left) === G
+    @test base_group(A_left) == G
+    @test repr(A_left) == "GroupOperationAction($(repr(G)), LeftAction())"
+    @test repr(A_right) == "GroupOperationAction($(repr(G)), RightAction())"
+
+    a_pts = []
```
mateuszbaran: This line can be removed?

test/groups/groups_general.jl
```diff
@@ -29,8 +24,8 @@ end
         @test_throws ErrorException compose(G, x, x)
         @test compose(G, eg, x) == x
         @test compose(G, x, eg) == x
-        @test compose!(G, x, eg, x) == x
-        @test compose!(G, x, x, eg) == x
+        @test_throws ErrorException compose!(G, x, eg, x)
```
mateuszbaran: Why is an error expected here?

sethaxen: There are no longer any defaults for `compose!` for identity elements. This is because once one overloads `compose!` for a new operation, one needs to reimplement these special cases to resolve ambiguity, so the general defaults can never be used.

sethaxen:
> I'm going to add the scaling group in the future and we'll see how it interacts with these product groups.

Sounds great! I have a local implementation of the `CircleGroup`, which I'll probably push for testing mixing different types of groups for `ProductGroup`.

The semidirect product group might need a refactor at some point to make it easier to add general affine subgroups. e.g. one should be able to take the direct product of scale and rotation or rotation and scale and then take the semidirect product (i.e. `T(n) ⋊_θ (SO(n) × Scale(n))`) and use affine representations without a lot of extra work.

src/product_representations.jl
```diff
@@ -229,20 +244,26 @@ struct ProductRepr{TM<:Tuple}
...
-end
-
 (+)(v1::ProductRepr, v2::ProductRepr) = ProductRepr(map(+, v1.parts, v2.parts)...)
 (-)(v1::ProductRepr, v2::ProductRepr) = ProductRepr(map(-, v1.parts, v2.parts)...)
-(-)(v::ProductRepr) = ProductRepr(map(-, v.parts))
-(*)(a::Number, v::ProductRepr) = ProductRepr(map(t -> a*t, v.parts))
+(-)(v::ProductRepr) = ProductRepr(map(-, submanifold_components(v)))
+(*)(a::Number, v::ProductRepr) = ProductRepr(map(t -> a*t, submanifold_components(v)))
+
+function isapprox(x::ProductRepr, y::ProductRepr; kwargs...)
```
mateuszbaran: Are you sure this is needed? `ProductRepr` should generally be used in point or tangent vector `isapprox` for a specific manifold.

sethaxen: Good point. Fixed.

sethaxen:
I think this PR is ready to merge now.

src/groups/array_manifold.jl
```diff
@@ -0,0 +1,154 @@
+array_value(e::Identity) = e
```
kellertuer: This overall is a nice addition, since it does the same as we propose for the array manifold anyways – proper types and more tests during computations.

src/groups/circle_group.jl
```diff
@@ -0,0 +1,58 @@
+@doc doc"""
```
kellertuer: So this is the same as the real-valued circle with the addition group? I think that could just shortly be mentioned, since we have the addition group anyways.

src/groups/group_action.jl
```diff
@@ -10,67 +10,135 @@ abstract type AbstractGroupAction{AD<:ActionDirection} end
...
+function apply(A::AbstractGroupAction, a, x)
     y = similar_result(A, apply, x, a)
-    apply!(A, y, x)
-    return y
+    return apply!(A, y, a, x)
 end
 
-function compose(A::AbstractGroupAction{LeftAction}, a, b)
-    return compose(base_group(A), a, b)
+"""
```
kellertuer: For readability one could switch the words `inverse` and `apply`, also for the diff cases later. Just as an idea.

src/product_representations.jl
```diff
@@ -229,20 +244,18 @@ struct ProductRepr{TM<:Tuple}
...
-    map(copyto!, x.parts, y.parts)
+    map(copyto!, submanifold_components(x), submanifold_components(y))
     return x
 end
 
-function submanifold_component(x::ProductRepr, i::Integer)
-    return x.parts[i]
-end
-
 (+)(v1::ProductRepr, v2::ProductRepr) = ProductRepr(map(+, v1.parts, v2.parts)...)
```
mateuszbaran: I think you could also change these methods of `+` and `-` to use `submanifold_components` for consistency.

src/groups/circle_group.jl
sethaxen: Without overloading `compose` for the real circle group case to also take the modulo, then `AdditionOperation` applied to the real circle will not be the same group.
The "right" way to do this would be to have
```julia
struct CircleGroup{𝔽,O} <: AbstractGroupManifold{O}
    manifold::Circle{𝔽}
end
```
and then have different operations for the two cases. But then none of the `GroupManifold` forwards apply. Forwarding metric functions from `AbstractGroupManifold` is no less difficult than implementing the general forwarding default proposed in #84.

But I also think we should interpret `AdditionOperation` and `MultiplicationOperation` literally as being `+` and `-`, so a new operation `AdditionWithModulo` would be necessary.

src/groups/group_action.jl
sethaxen: I like this idea. Also because it makes like functions near one another in alphabetical order. We would also need to change `inverse_translate => translate_inverse` and `inverse_translate_diff => translate_diff_inverse` (or `translate_inverse_diff`). @mateuszbaran what do you think?

mateuszbaran: I don't have a preferred word order here, feel free to choose whichever looks the best for you.

src/groups/circle_group.jl
kellertuer: Yes, you're right, it's actually addition modulo, sorry.

src/groups/group_action.jl
sethaxen: Actually, maybe this isn't best. e.g. we already have `inverse_retract` so until now, `inverse_` has been used as a prefix not a suffix.

kellertuer: As I think for `exp` of the exponential_map for me `retract` is more like `retraction`(`_map`) so then the `inverse` is an adjective to that so in front while for  the `apply` it's more an adverb, so there it's after the verb, but yes it's inconsistent in the sense that `retract` in its form as written is also a verb of course.

sethaxen: I don't follow the argument. We have two maps, a retraction and an action. The retraction map is not explicitly represented as an object, but its action can be called with the verb `retract`. The action on a manifold is also a map; it is, however, explicitly represented, so we have this helper verb function `apply` to actually call its action. I don't see a major difference between these two functions.

kellertuer: And that's ok, I see that my feeling is not consistent and it stays a feeling. I liked the inverted version of `apply_inverse(::Action)`, but that's merely a feeling and not even an oppinion.

sethaxen:
Okay, once the test suite passes, I'll merge this.

Issue #74, opened 2019-12-27
Unify documentation


kellertuer:
In an effort to get the documentation into a unified format I am asking for a few ideas and comments.

currentyl we are often documenting `exp!`, which I would like to avoid/refactor to provide the user with a consistent documentation. I would like to propose the following – on the example of `exp` but indeed for all functions with both mutating and non-mutating forms as well as all functions that might have formulae involved that are nice to have in the documentation.

As @sethaxen demonstrated nicely I would like to add a docstring to `exp(M,args...)` for the manifold `M`that documents the mathematical formula and maybe a reference if it's not a textbook manifold/metric. With that all non-mutating functions have a nice documentation that shows up in the docs. The documentation should always also list the `t` argument as a case.
This yields a nice overview over available functions in a consistent format.

Additionally either in the page for the API (`ManifoldsBase.jl`) or on a seperate page (implementation details, or how-to-implement-your-onw-manifold.md) we could list the approach of implementing a mutating one that the non-mutating one uses (including the `similar_result` approach.

If you agree to this, I can use my time travelling to do a unifying rework on the documentation once the next manifolds are finished. I will then also look that the docs and most implementations follow the [Code style Blue](https://github.com/invenia/BlueStyle) guidelines (shall we add their icon/banner then in the readme?)

mateuszbaran:
Yes, I think unifying the style of documentation is a good idea. I agree with this approach. I can make some documentation on adding new manifolds.

We can officially use the blue style guidelines (and I like that banner). As a part of updating our code to this style we should probably rename files to lower_case_with_underscores, although I'd wait with it until we have fewer open PRs as renaming files may cause lots of conflicts.

sethaxen:
I like this plan!

> As a part of updating our code to this style we should probably rename files to lower_case_with_underscores, although I'd wait with it until we have fewer open PRs as renaming files may cause lots of conflicts.

Is this part of blue style? The general rule I like is that a filename might be camel-case if a single type is declared in it, but that would rule out a number of our manifolds cases, since auxiliary types are sometimes declared. So I agree that for uniformity, snake-case might be the right way to go.

I also am sitting on a massive PR to standardize docstrings/code formatting and make them follow blue style, but I'm sitting on that until things settle a little more.

mateuszbaran:
> > 
> 
[...]

Issue #32, opened 2019-07-14
Structure Code & References in Documentation


kellertuer:
I noticed two things we do differently in coding
(1) until now I always wrote `export`, `using/import`s at the beginning of a file, kind of like a table of contents, now we have a little bit of a mix, since some files contain the exports at the end
(2) Since we collect all functions for a manifold in a file, wouldn't it be nice to order functions alphabetically? I can just rearrange everything, but I'd like to have an opinion what you think about this

Furthermore, I'd like to provide references for formulae, not necessarily in single function documentations, but especially on the documentation page of a manifold (I also started to illustrate SPDs with an image in #27). Is there a preferred way how to do references? For example we could try `BibTeX.jl` but for now I just did a markdown list, see e.g. http://www.manoptjl.org/stable/solvers/cyclicProximalPoint/#Literature-1). The advantage of something more automated would be a more consistent format than the manual list.

sdaxen:
> (1) until now I always wrote `export`, `using/import`s at the beginning of a file, kind of like a table of contents, now we have a little bit of a mix, since some files contain the exports at the end

I've only ever written packages with all of the imports/exports in the main source file, though I can see the clarity for doing it in each file. Is there a nice way to test that all the right imports/exports are there?

> (2) Since we collect all functions for a manifold in a file, wouldn't it be nice to order functions alphabetically? I can just rearrange everything, but I'd like to have an opinion what you think about this

An alternative to ordering alphabetically that we've sort of being doing is ordering by related functionality. So `exp` and `log` are together, with `retract` near `exp`. `inner`, `norm`, `distance`, etc all go together so that it's easy to switch back and forth while developing.

That being said, while that approach may make the code clearer, it does nothing for the API documentation, which is alphabetical. Perhaps it is a better idea to sort alphabetically and frequently use `@ref`.

A compromise would be to have common sections for each manifold, and then within each section they're alphabetized. I guess I don't have a strong preference, so long as we're consistent.

👍 for documenting references. I like the idea of an automated approach that reads from a source bibtex file. Is that what `BibTeX.jl` does?

kellertuer:
> I've only ever written packages with all of the imports/exports in the main source file, though I can see the clarity for doing it in each file. Is there a nice way to test that all the right imports/exports are there?

[...]

PR #83, opened 2020-01-01
Unifies and refactors documentation

kellertuer:
Following the discussions at #32 and #74, leaving out literature for a future PR.


kellertuer:
I have one small problem with `retract`: In order to document for example an `SVD`-based retraction I have to provide the full signature and hence overwrite a version (since the `method` is the last argument), which I find not so nice  any idea on that? (For an example see Hyperpolic Line 253).

Despite that I was able to provide the general ones in the form e.g. `exp(::Hyperbolic, ::Any...)`, which avoids an unused variable name – which I think is nice.

Finally the double footnotes (to be addressed in a next PR) currently also generate double entries in the Literature (footnotes) section.

kellertuer:
I have a short question on the `fixed_random_rotation`, there is no randomness involved,so what is this function for? Does it project? The name is misleading.

Further for me the names `_gradient` and `_jacobian` indicate that that might be internal functions, yet they have docstrings (the only two ones not used). Should they be added to the `internals.md`?

Finally from the recent master merge `compose` is referenced, but since none of the defaults does have a docstring, there is also a warning left. What's a good default doc and which function should get that?

mateuszbaran:
> I have a short question on the `fixed_random_rotation`, there is no randomness involved,so what is this function for? Does it project? The name is misleading.

Do you mean `_fix_random_rotation`? This is somewhat explained in the docstring of `normal_rotation_distribution`. I agree that the name is misleading but it's an internal function anyway.

> Further for me the names `_gradient` and `_jacobian` indicate that that might be internal functions, yet they have docstrings (the only two ones not used). Should they be added to the `internals.md`?

There is nothing wrong with having docstrings for internal functions and yes, `internals.md` is there to collect such docstrings.

> Finally from the recent master merge `compose` is referenced, but since none of the defaults does have a docstring, there is also a warning left. What's a good default doc and which function should get that?

I think `compose(G::AbstractGroupManifold, x, y)` can have a docstring, something similar to the docstring of `compose!(G::AbstractGroupManifold, z, x, y)`. I must've missed this.

kellertuer:
Thanks for the fast response, I will refactor the doc for `compose´ then with the next manifold and add `_gradient` and `_jacobian` to the internals.

Oh and yes I meant `fixed_random_rotation` it seems to correct something and was near `rand`, where it is also used but not documented, the somewhat mentioned might only be the case if you know which part is speaking about that, it's not explicitly mentioned.

src/Sphere.jl
```diff
@@ -169,7 +169,7 @@ end
 Compute the Riemannian [`mean`](@ref mean(M::Manifold, args...)) of `x` using
 [`GeodesicInterpolationWithinRadius`](@ref).
 """
-mean(::Sphere, args...)
+mean(::Sphere, ::Any)
```
sethaxen: What's the benefit of `::Any` over `args...`? e.g. if one entered
`?mean(Sphere(2), x, w)`, does it still show this docstring in the REPL?

src/manifolds/Circle.jl
```diff
@@ -8,13 +8,21 @@ $\lvert z\rvert = 1$.
 
     Circle(f=ℝ)
 
-generates the Circle represented by angles, which alternatively can be set to use the field
-`f=ℂ` to obtain the Circle as a manifold embedded in ℂ.
+generates the ℝ-valued Circle represented by angles, which
+alternatively can be set to use the [`AbstractField`](@ref) `f=ℂ` to obtain the Circle
+represented by ℂ-valued Circle of unit numbers.
```
sethaxen:
```suggestion
represented by `ℂ`-valued `Circle` of unit numbers.
```

src/manifolds/Circle.jl
```diff
@@ -8,13 +8,21 @@ $\lvert z\rvert = 1$.
 
     Circle(f=ℝ)
 
-generates the Circle represented by angles, which alternatively can be set to use the field
-`f=ℂ` to obtain the Circle as a manifold embedded in ℂ.
+generates the ℝ-valued Circle represented by angles, which
+alternatively can be set to use the [`AbstractField`](@ref) `f=ℂ` to obtain the Circle
```
sethaxen:
```suggestion
alternatively can be set to use the [`AbstractField`](@ref) `f=ℂ` to obtain the `Circle`
```

src/manifolds/Circle.jl
```diff
@@ -8,13 +8,21 @@ $\lvert z\rvert = 1$.
...
+represented by ℂ-valued Circle of unit numbers.
 """
 struct Circle{F} <: Manifold where {F <: AbstractField} end
 Circle(f::AbstractField=ℝ) = Circle{f}()
 
+@doc doc"""
+    check_manifold_point(M::Circle, x)
 
+checks whether `x` is a point on the [`Circle`](@ref) `M`.
+For thge real-valued case, `x` is an angle and henceit checks that $x \in [-\pi,\pi)$.
```
sethaxen:
```suggestion
For the real-valued case, `x` is an angle and hence it checks that $x \in [-\pi,\pi)$.
```

src/manifolds/Circle.jl
```diff
@@ -23,12 +31,23 @@ function check_manifold_point(M::Circle{ℝ},x; kwargs...)
...
-
-function check_tangent_vector(M::Circle{ℝ},x,v; kwargs...)
+"""
+    check_tangent_vector(M::Circle, x, v)
+
+check whether `v` is a tangent vector in the tangent space of `x` on the
+[`Circle`](@ref) `M`.
+For the real-valued case represented by angles all `v` are valid, since the tangent space is
+the whole real line.
+For the complex-alued case `v` has to lie on the line parallel to the tangent line at `x`
```
sethaxen:
```suggestion
For the complex-valued case `v` has to lie on the line parallel to the tangent line at `x`
```

src/manifolds/Circle.jl
```diff
@@ -23,12 +31,23 @@ function check_manifold_point(M::Circle{ℝ},x; kwargs...)
...
+in the complex place, i.e. the inner product is zero.
```
sethaxen:
```suggestion
in the complex plane, i.e. the inner product is zero.
```

src/manifolds/Circle.jl
```diff
@@ -186,17 +205,35 @@ Compute the Riemannian [`mean`](@ref mean(M::Manifold, args...)) of `x` on the
...
 project_point!(::Circle{ℝ}, x) = (x .= sym_rem(x))
 project_point!(::Circle{ℂ}, x) = (x .= x/sum(abs.(x)))
 
+@doc doc"""
+    project_tangent(M::Circle, x, v)
+
+project a value `v` onto the tangent space of the point `x` on the [`Circle`](@ref) `M`.
+
+For the real-valued case this is just the identity.
+For the complex valued case `v` is projeted onto the line in the complex line
```
sethaxen:
```suggestion
For the complex valued case `v` is projected onto the line in the complex plane
```

src/manifolds/Euclidean.jl
```diff
@@ -46,183 +29,232 @@ the [`Euclidean`](@ref) manifold itself, or the [`Sphere`](@ref), where every
 tangent space (as a plane in the embedding) uses this metric (in the embedding).
 
 Since the metric is independent of the field type, this metric is also used for
-the Hermitian metrics, i.e. metrics that are analogous to the Euclidean but where
-the field type of the manifold is `ℂ`.
+the Hermitian metrics, i.e. metrics that are analogous to the `EuclideanMetric`
```
sethaxen: Perhaps it's clearer to say that the metric defines an inner product equivalent to the Frobenius inner product, implemented by `dot`, and give the equation (and its interpretation as complexifying. Then manifolds that use this metric can reference this docstring instead of re-explaining it. This also generalizes to other fields. e.g. quaternions.

src/manifolds/FixedRankMatrices.jl
```diff
@@ -207,23 +232,30 @@ function project_tangent!(
...
-representation_size(F::FixedRankMatrices{M,N}) where {M,N} = (M,N)
 
 @doc doc"""
-    retract!(M, y, x, v, ::PolarRetraction)
+    representation_size(M::FixedRankMatrices{m,n,k})
 
-compute an SVD-based retraction on the [`FixedRankMatrices`](@ref) manifold
-by computing
+returns the element size of a point on the [`FixedRankMatrices`](@ref) `M`, i.e.
+thge size of matrices on this manifold $(m,n)$.
```
sethaxen:
```suggestion
the size of matrices on this manifold $(m,n)$.
```

src/manifolds/Hyperbolic.jl
```diff
@@ -116,10 +129,15 @@ function flat!(M::Hyperbolic, v::FVector{CotangentSpaceType}, x, w::FVector{Tang
     return v
 end
 
+@doc doc"""
+    injectivity_radius(M::Hyperbolic[, x])
+
+returns the infjectivity radius on the [`Hyperbolic`](@ref), which is always $\infty$.
```
sethaxen:
```suggestion
returns the injectivity radius on the [`Hyperbolic`](@ref), which is always $\infty$.
```

src/Sphere.jl
kellertuer: oh, it should be `::Any...` then its just the type of the old `args...` without having an unused variable name (as the first one does not have the `M`).

kellertuer:
> This quite a lot of work, and I didn't get to look at everything. RE docstrings, I prefer the convention that function docstrings begin with an imperative sentence. e.g.
> "Compute the Riemannian mean" as opposed to a sentence with a lower case first letter (e.g. "compute the Riemannian mean") or a description (e.g. "computes the Riemannian mean"). 
until now I prefered the last one, so we should agree on one of those

> There are a few cases where a non-mutating docstring is attached to a mutating function. That should be avoided, as Documenter uses the signature to which the docstring is attached for URLs and for `@ref`s.
Oh, can you point them out carefully? I had hoped to introduce the non-mutating versions everywhere.

mateuszbaran:
> > This quite a lot of work, and I didn't get to look at everything. RE docstrings, I prefer the convention that function docstrings begin with an imperative sentence. e.g.
> > "Compute the Riemannian mean" as opposed to a sentence with a lower case first letter (e.g. "compute the Riemannian mean") or a description (e.g. "computes the Riemannian mean").
> 
> until now I prefered the last one, so we should agree on one of those

There is a quite detailed description how docstrings should be written in the Julia documentation: https://docs.julialang.org/en/v1/manual/documentation/ , so we could probably follow it a bit more closely than before.

sethaxen:
> There is a quite detailed description how docstrings should be written in the Julia documentation: https://docs.julialang.org/en/v1/manual/documentation/ , so we could probably follow it a bit more closely than before.

Thanks for the link; yes, I had not seen most of those recommendations. They also recommend the imperative style and a single sentence describing the method.

> 7. Use backticks to identify code and equations.
> Julia identifiers and code excerpts should always appear between backticks ` to enable highlighting. Equations in the LaTeX syntax can be inserted between double backticks \``. Use Unicode characters rather than their LaTeX escape sequence, i.e. \``α = 1\`` rather than \``\\alpha = 1\``.

We haven't been doing this yet, and it would certainly make the docstrings more REPL-friendly. But I only think we should do this if the browser renders the unicode the same as the LaTeX.

sethaxen:
Now that I'm thinking of it, we should probably have a wiki page where we establish notation/style conventions that we've specifically adopted (i.e. that arose out of discussions like this).

sethaxen:
> Oh, can you point them out carefully?

I had only noticed one so far, but I didn't look very carefully. I also don't know if this was introduced on this PR; there were already a  few: https://github.com/JuliaNLSolvers/Manifolds.jl/blob/d07dca6660b87845a9d05d8f90913deb03c2e0ea/src/Metric.jl#L285-L302

> I had hoped to introduce the non-mutating versions everywhere. And yes of course it should not be mixed, I am aiming for documenting all nopn-mutating and then having a how-to-build-your-own-manifold explaining the details (mutating versus non-mutating).

That sounds great!

kellertuer:
A wiki page? Usually one provides a contribution markdown file for that as far as I read. We could specify all that in there.

kellertuer:
Lookes shortly, something  like this https://github.com/atom/atom/blob/master/CONTRIBUTING.md – I can start setting that up tomorrow within this PR :)

mateuszbaran:
A markdown file with contributing guidelines is a good idea. A wiki page would also be OK though we don't use the wiki functionality here at the moment so a simple file would be appropriate.

kellertuer:
> > Oh, can you point them out carefully?
> 
> I had only noticed one so far, but I didn't look very carefully. I also don't know if this was introduced on this PR; there were already a few:

I haven't been to this file, but I'll carefully check. The ones I worked on are the ones that are in ? manifolds/` now and I am not yet 100% sure whether that folder should cover just the basic ones or also the ones that decorate (metric) or extend (product).

kellertuer:
> A markdown file with contributing guidelines is a good idea. A wiki page would also be OK though we don't use the wiki functionality here at the moment so a simple file would be appropriate.

I am for a contributing.md, since we don't use the wiki and I don't see where/why we should start to do so.

src/manifolds/Euclidean.jl
sethaxen: Also, since `Euclidean` supports multidimensional arrays, it would be more accurate to define the Frobenius norm for `Euclidean` equivalently as the element-wise sum of the Hadamard product of the conjugate:

`g_x(v,w) = \sum_{i,\dots,j} \overline{v}_{i \dots j} w_{i \dots j}`

instead of using the Hermitian transpose, which presumes no more than 2 dimensions.

I think more accurately, `dot` implements a Hilbert-Schmidt operator inner product, since any numeric type is free to overload `conj` (it doesn't need to be complex conjugation). e.g. Quaternions.jl uses `conj` for quaternionic conjugation, and `dot` applied to arrays of quaternions still works. Manifolds on non-real/non-complex manifolds will probably be rare, so I don't think we need to document this.

sethaxen:
> I am for a contributing.md, since we don't use the wiki and I don't see where/why we should start to do so.

That sounds fine.

> I have one small problem with `retract`: In order to document for example an `SVD`-based retraction I have to provide the full signature and hence overwrite a version (since the `method` is the last argument), which I find not so nice any idea on that? (For an example see Hyperpolic Line 253).

I didn't follow. Can you elaborate what the problem is with an example?

kellertuer:
> > I have one small problem with `retract`: In order to document for example an `SVD`-based retraction I have to provide the full signature and hence overwrite a version (since the `method` is the last argument), which I find not so nice any idea on that? (For an example see Hyperpolic Line 253).
> 
> I didn't follow. Can you elaborate what the problem is with an example?

The two functions afftected are `retract` and `vector_transport`. I am currently not able to link to lines in this PR (no links on the line numbers?) so please look at `CholeskySpace.jl` Line 169 (new version). This was my first idea and is not precise enough, since it should only document the `::ParallelTransport`-case. Compare that to `Hyperbolic.jl` line 253. This one _does_ document the right case, but has 5 new lines, which are (up to the type of the manifold and passing down the method) completely the same as the definition of `vector_transport` (non-mutating) in the `ManifoldsBase`. So just for documentation we produce redundancy. I don't like that, but I don't have a better solution for now.

sethaxen:
> The two functions afftected are `retract` and `vector_transport`. I am currently not able to link to lines in this PR (no links on the line numbers?) so please look at `CholeskySpace.jl` Line 169 (new version). This was my first idea and is not precise enough, since it should only document the `::ParallelTransport`-case. Compare that to `Hyperbolic.jl` line 253. This one _does_ document the right case, but has 5 new lines, which are (up to the type of the manifold and passing down the method) completely the same as the definition of `vector_transport` (non-mutating) in the `ManifoldsBase`. So just for documentation we produce redundancy. I don't like that, but I don't have a better solution for now.

For reference, these are the lines (within the most recent commit you can link to the lines):

https://github.com/JuliaNLSolvers/Manifolds.jl/blob/d07dca6660b87845a9d05d8f90913deb03c2e0ea/src/manifolds/CholeskySpace.jl#L162-L174

This case you can replace with `vector_transport_to(::CholeskySpace, ::Any, ::Any, ::Any, ::ParallelTransport)`.

https://github.com/JuliaNLSolvers/Manifolds.jl/blob/d07dca6660b87845a9d05d8f90913deb03c2e0ea/src/manifolds/Hyperbolic.jl#L241-L257

And this one you can remove the implementation and just use `vector_transport_to(::Hyperbolic, ::Any, ::Any, ::Any, ::ParallelTransport)`.

Attaching docstrings in this way doesn't override methods unless you define them, so the defaults remain. e.g.

```julia
julia> using Markdown: @doc_str

julia> bar(x) = 3
bar (generic function with 1 method)

julia> @doc doc"""
           foo(x)
       """
       bar(::Any)
bar

help?> bar(4)
  foo(x)

julia> bar(4)
3
```

kellertuer:
Thanks for the help, I haven't had the idea to use the `::Any`, idea and at first sight wasn't sure that would work (because _basically_ you do something like defining them) so thanks for also illustrating that. I am just finishing my todays manifold code rework (providing a lot more docs for the Sphere).

kellertuer:
The base manifolds are refactored. For the last, `SymmetricPositiveDefinite` I am struggling a little bit to filter out the right manifold_dimension in the `@docs`, since I would like to split that page by metrics afterwards. The same holds for `representation_size`. Somehow _all_ are printed though the signature ist quite precise. Any idea welcome.

Should I also move the others into manifolds? `ProductManifold`, etc?
Refactoring/codestyling them is my next goal.

sethaxen:
> For the last, `SymmetricPositiveDefinite` I am struggling a little bit to filter out the right manifold_dimension in the `@docs`, since I would like to split that page by metrics afterwards. The same holds for `representation_size`. Somehow _all_ are printed though the signature ist quite precise. Any idea welcome.

I encountered this while working on the docs for `SymmetricPositiveDefinite` as well. The only solution I found was to refactor to use `@autodocs` instead of `@docs`; IIRC, `@autodocs` rendered just the correct docstring.

To do that, you need to be able to filter by type of the function, which doesn't work when you want to filter by the signature of the function. An alternative is to put the implementations for each of the alternative metrics in their own .jl files and then you can `@autodocs` those files in separate blocks. This would probably be cleanest by putting them in a subdirectory within `manifolds`.

> Should I also move the others into manifolds? `ProductManifold`, etc?

I think so, that's how we have it divided in the documentation (though there we also divide into basic, combined, and decorators; we might want to do that here, but I don't think it's necessary).

kellertuer:
> To do that, you need to be able to filter by type of the function, which doesn't work when you want to filter by the signature of the function. An alternative is to put the implementations for each of the alternative metrics in their own .jl files and then you can `@autodocs` those files in separate blocks. This would probably be cleanest by putting them in a subdirectory within `manifolds`.

Exactely, I want to filter by signature, i.e. only the `manifold_dimension` for `SymmetricPositiveDefinite`...

> > Should I also move the others into manifolds? `ProductManifold`, etc?
> 
> I think so, that's how we have it divided in the documentation (though there we also divide into basic, combined, and decorators; we might want to do that here, but I don't think it's necessary).

Ok, just wanted to make sure.

kellertuer:
Found a short moment to refactor symmetricpositivedefinite. The files are now smaller (with a little long filenames maybe) but I like that metric specific functions to one metric are nicely summarized in one file. And the doc now looks gorgeous. Tomorrow I will continue with the other ones and also do further renaming of files to fit our naming scheme.

kellertuer:
I had a short half an hour to start with Power and noticed two points that I might have missed when we started this:

1) for retraction and inverse_retraction it should suffice to allow for the usual retractions, they can just be passed down and work elementwise, since exp and log also just work elementiwse. So why do we have a specific type to encapsualte the retraction (or inverse) type? For me that seems to be unnecessary

2) the injectivity radius is implemented too complicated (maybe still coming from ProductManifold?). Since in every entry we have the same manifold, there is no need to check and look for the min (for the global version), and in both (local and global) the result is too conservative, I think it can be increased by the factor `sqrt(prod(N))` where `N` is the `TSize` array. What do you think?

I will continue with this and also documenting the basic functions soon.

mateuszbaran:
Nice progress, thanks!

> 1. for retraction and inverse_retraction it should suffice to allow for the usual retractions, they can just be passed down and work elementwise, since exp and log also just work elementiwse. So why do we have a specific type to encapsualte the retraction (or inverse) type? For me that seems to be unnecessary

We can just propagate given retraction method, yes. I'm not sure why I've added `PowerRetraction` and `InversePowerRetraction`.


> 2\. the injectivity radius is implemented too complicated (maybe still coming from ProductManifold?). Since in every entry we have the same manifold, there is no need to check and look for the min (for the global version), and in both (local and global) the result is too conservative, I think it can be increased by the factor `sqrt(prod(N))` where `N` is the `TSize` array. What do you think?

I agree that the global version is too complicated. I don't know though how did you get that `sqrt(prod(N))` factor, could you explain? I think minimum is appropriate since we can have points on a power manifold that differ only in one component and then the injectivity radius of the powered manifold (the wrapped one) would apply, no?

kellertuer:
> We can just propagate given retraction method, yes. I'm not sure why I've added `PowerRetraction` and `InversePowerRetraction`.

Great, I'll change that; I also only noticed that when I tried to document retraction.

> 
> > 2. the injectivity radius is implemented too complicated (maybe still coming from ProductManifold?). Since in every entry we have the same manifold, there is no need to check and look for the min (for the global version), and in both (local and global) the result is too conservative, I think it can be increased by the factor `sqrt(prod(N))` where `N` is the `TSize` array. What do you think?
> 
> I agree that the global version is too complicated. I don't know though how did you get that `sqrt(prod(N))` factor, could you explain? I think minimum is appropriate since we can have points on a power manifold that differ only in one component and then the injectivity radius of the powered manifold (the wrapped one) would apply, no?

Oh, it's not points but tangent vectors that might only differ in one component (so it's the zero matrix of tangents despite one entry), you're right; still for the global one the min is not necessary since they are (globally on the base) all the same (for product you need that).

My idea was that, like in R^n and matrices the diagonal distance is something like `length(N)*min( N( n .> 1 ))`(the product was too optimistic) since the norm gives you this. But this only holds for distances and not for tangent vectors, so you're right, we sadly can not improve, so this has to stay as it is.

kellertuer:
I don't like the current form of representation size, since it just returns one array and one can not see the power form. To be precise for an 4x5 array of 3x3 matrices this returns `(3,3,4,5)` Can we do something that one sees the power manifold structure? Maybe `([3,3],4,5)`?

further the retract has still a dispatch problem for a specific retraction with the defaults in base:

````julia
  MethodError: inverse_retract!(::PowerManifold{Sphere{2},Tuple{5}}, ::HybridArray{Tuple{3,StaticArrays.Dynamic()},Float64,2,2,Array{Float64,2}}, ::HybridArray{Tuple{3,StaticArrays.Dynamic()},Float64,2,2,Array{Float64,2}}, ::HybridArray{Tuple{3,StaticArrays.Dynamic()},Float64,2,2,Array{Float64,2}}, ::ManifoldsBase.LogarithmicInverseRetraction) is ambiguous. Candidates:
    inverse_retract!(M::PowerManifold, v, x, y, method::R) where R<:AbstractInverseRetractionMethod in Manifolds at /Users/ronny/Repositories/Software/Manifolds.jl/src/manifolds/PowerManifold.jl:212
    inverse_retract!(M::Manifold, v, x, y, method::ManifoldsBase.LogarithmicInverseRetraction) in ManifoldsBase at /Users/ronny/Repositories/Software/ManifoldsBase.jl/src/ManifoldsBase.jl:212
  Possible fix, define
    inverse_retract!(::PowerManifold, ::Any, ::Any, ::Any, ::ManifoldsBase.LogarithmicInverseRetraction)
````
Maybe that's why you introduced the Powerretraction as less as I do like it?

**Comment by mateuszbaran** [comment]
> I don't like the current form of representation size, since it just returns one array and one can not see the power form. To be precise for an 4x5 array of 3x3 matrices this returns `(3,3,4,5)` Can we do something that one sees the power manifold structure? Maybe `([3,3],4,5)`?

`representation_size` for power manifold could return an object of a special type that prints differently, though for compatibility it'd still have to behave like an array of integers but it could be printed differently. I can think about it later when this is merged and I'm finished with ONBs.



> further the retract has still a dispatch problem for a specific retraction with the defaults in base:
> 
> ```julia
>   MethodError: inverse_retract!(::PowerManifold{Sphere{2},Tuple{5}}, ::HybridArray{Tuple{3,StaticArrays.Dynamic()},Float64,2,2,Array{Float64,2}}, ::HybridArray{Tuple{3,StaticArrays.Dynamic()},Float64,2,2,Array{Float64,2}}, ::HybridArray{Tuple{3,StaticArrays.Dynamic()},Float64,2,2,Array{Float64,2}}, ::ManifoldsBase.LogarithmicInverseRetraction) is ambiguous. Candidates:
>     inverse_retract!(M::PowerManifold, v, x, y, method::R) where R<:AbstractInverseRetractionMethod in Manifolds at /Users/ronny/Repositories/Software/Manifolds.jl/src/manifolds/PowerManifold.jl:212
>     inverse_retract!(M::Manifold, v, x, y, method::ManifoldsBase.LogarithmicInverseRetraction) in ManifoldsBase at /Users/ronny/Repositories/Software/ManifoldsBase.jl/src/ManifoldsBase.jl:212
>   Possible fix, define
>     inverse_retract!(::PowerManifold, ::Any, ::Any, ::Any, ::ManifoldsBase.LogarithmicInverseRetraction)
> ```
> 
> Maybe that's why you introduced the Powerretraction as less as I do like it

`PowerRetraction` is one way to solve it, another one is defining the method mentioned in the error:

```julia
inverse_retract!(M::PowerManifold, v, x, y, method::ManifoldsBase.LogarithmicInverseRetraction) = invoke(inverse_retract!, Tuple{PowerManifold, Any, Any, Any, AbstractInverseRetractionMethod}, M, v, x, y, method)
```
(not tested but should work)

kellertuer:
Hm, yes we can discuss the representation size later.

For the other issue, you proposal would mean we have to add such a line for every Retraction implemented, so then I would really favour the special type you used, though it looks strange at first.

kellertuer:
When I just checked the ProductManifold I noticed, there we indeed _need_ an extra retraction type so for consistency I think it's ok to have that for power, too. With Product finished, the manifolds are done, just the Metric decorator missing and the single ad/stats... files have to be checked (Metric might still be moved).

kellertuer:
I think this is nearly done. 
* all Manifolds are now CamelCase and in `manifolds/`
* all other files are CamelCase if their main purpose is the type their name indicates, otherwise snake_case
* all files have first their types then their functions, where in each block the order is alphabetic (up to the situations where structs depend on others)
* in nearly all lines we are not at 92 characters
* all functions documented are now the non-mutating ones in the general case as @sethaxen introduced
* I documented quite a few functions additionally
* changes `using`, `import` and `export` to code blue style.
* added a `CONTRIBUTIONS.md` documents this style
* add the code blue banner

Only for `project_vector` I wasn't sure, there is no non-mutating version yet? Should we have that? And there is exactely one docstring missing, for `compose`.

mateuszbaran:
The refactor looks really nice! I'll make a small review soon.

> Only for `project_vector` I wasn't sure, there is no non-mutating version yet? Should we have that?

Yes, I think it was just oversight.

> And there is exactely one docstring missing, for `compose`.

Right, you can make one based on the docstring for mutating `compose!`.

src/manifolds/VectorBundle.jl
```diff
@@ -538,8 +369,162 @@ Notation:
...
     return w
 end
+
+"""
+    project_vector(B::VectorBundleFibers, x, w)
+
+Project vector `w` from the vector space of type `B.VS` at point `x`
+and save the result to `v`.
+"""
+function project_vector(B::VectorBundleFibers, x, w)
```
mateuszbaran: I think you should leave the erroring default version of `project_vector` (it's for vector spaces other than the tangent space)

CONTRIBUTING.md
```diff
@@ -0,0 +1,65 @@
...
+non-mutating variant, see for example
+[exp!](https://julianlsolvers.github.io/Manifolds.jl/latest/interface.html#ManifoldsBase.exp!-Tuple{Manifold,Any,Any,Any})
+and
+[exp](https://julianlsolvers.github.io/Manifolds.jl/latest/interface.html#Base.exp-Tuple{Manifold,Any,Any}).
+The non-mutating one always falls back to use the mutating one, so in most cases it should
+suffice to implement the mutating one, i.e.
+[exp!](https://julianlsolvers.github.io/Manifolds.jl/latest/interface.html#ManifoldsBase.exp!-Tuple{Manifold,Any,Any,Any})
+in this case.
+
+Note that since the first argument is _always_ the [`Manifold`](https://julianlsolvers.github.io/Manifolds.jl/latest/interface.html#ManifoldsBase.Manifold), the mutable element is always the second argument in the signature.
```
mateuszbaran:
```suggestion
Note that since the first argument is _always_ the [`Manifold`](https://julianlsolvers.github.io/Manifolds.jl/latest/interface.html#ManifoldsBase.Manifold), the mutated argument is always the second one in the signature.
```

src/manifolds/VectorBundle.jl
kellertuer: This one is similar to `exp` and `exp!` (see https://github.com/JuliaNLSolvers/ManifoldsBase.jl/blob/master/src/ManifoldsBase.jl#L337 ), where the non-mutating falls back to the mutating one and (ony) this throws the error.

CONTRIBUTING.md
```diff
@@ -0,0 +1,65 @@
...
+In the example we have `exp(M, x, v)` for the exponential map and `exp!(M, y, v, x)` for the mutating one, that stores the result in `y`.
+
+On the contrary the user will most likely look for the non-mutating version, when looking for help, so it we recommend to add the document string to the non-mutating one, where all different signatures should be collected in one string. This can best be archieved by adding a docstring to the function with arbitrary signature despite the first being your manifold:
```
mateuszbaran:
```suggestion
On the contrary the user will most likely look for the non-mutating version, when looking for help, so we recommend to add the documentation string to the non-mutating one, where all different signatures should be collected in one string. This can best be achieved by adding a docstring to the method with a general signature with the first argument being your manifold:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,65 @@
...
+````julia
+    struct MyManifold <: Manifold end
+    @doc doc"""
+        exp(M::MyManifold, x, v)
+    
+    Describe the function.
+    """
+    exp(::MyManifold, ::Any...)
+````
+
```
mateuszbaran: W could also have here a section about missing methods

### Add a missing method

Not all functions from our interface have methods for all manifolds. Even a single new method is a good contribution.

docs/src/interface.md
```diff
@@ -11,7 +11,12 @@ we happily look forward to a
 [Pull Request](https://github.com/JuliaNLSolvers/Manifolds.jl/compare) to add it here.
 
 ```@autodocs
-Modules = [ManifoldsBase]
+Modules = [Manifolds, ManifoldsBase]
 Pages = ["ManifoldsBase.jl"]
 Order = [:type, :function]
 ```
+
+As a first manifold, a reduced simplified version of [`Euclidean`](@ref) the interface provides
```
mateuszbaran:
```suggestion
`DefaultManifold` is a simplified version of [`Euclidean`](@ref) and demonstrates a basic interface implementation.
```

src/distributions.jl
```diff
@@ -2,62 +2,62 @@
...
+struct FVectorvariate <: VariateForm
 end
 
 """
-    FVectorvariate
+    FVectorSupport(space::Manifold, VectorBundleFibers)
 
-Structure that subtypes `VariateForm`, indicating that a single sample
-is a vector from a fiber of a vector bundle.
+Value support for tangent vector-valued distributions (values from tangent
```
mateuszbaran: It's true that at the moment `FVectorSupport` is used only for tangent vectors but the interface is supposed to work for any fiber of any vector bundle.

src/distributions.jl
```diff
@@ -2,62 +2,62 @@
...
+    x::T
 end
 
 """
-    MPointSupport(manifold::Manifold)
+    FVectorDistribution{TSpace<:VectorBundleFibers, T}
 
-Value support for manifold-valued distributions (values from given
-manifold `M`).
+An abstract distribution for tangent vectors at point of type `T`
```
mateuszbaran: Similarly to my `FVectorSupport` comment

src/distributions.jl
```diff
@@ -2,62 +2,62 @@
...
 end
 
 """
-    FVectorSupport(space::Manifold, VectorBundleFibers)
+    MPointSupport(manifold::Manifold)
 
-Value support for tangent vector-valued distributions (values from tangent
-space at point `x` from given manifold).
+Value support for manifold-valued distributions (values from given
+manifold `M`).
```
mateuszbaran:
```suggestion
manifold `manifold`).
```

src/manifolds/Euclidean.jl
```diff
@@ -10,219 +10,259 @@ Euclidean vector space $\mathbb R^n$.
 
     Euclidean(n)
 
-generates the $n$-dimensional vector space $\mathbb R^n$.
+Generate the $n$-dimensional vector space $\mathbb R^n$.
 
     Euclidean(n₁,n₂,...,nᵢ; field=ℝ)
 
-generates the (2*)$n_1n_2\cdot\ldots n_i$-dimensional vector space $\mathbb F^{n_1, n_2, \ldots, n_i}$, whose
+Generate the (2*)$n_1n_2\cdot\ldots n_i$-dimensional vector space
```
mateuszbaran: I'm not sure if the `(2*)` part is clear enough here. Maybe we should write the dimension at the end of this docstring, after `field` is mentioned?

src/distributions.jl
mateuszbaran: Or change the argument name for `MPointSupport` to `M`.

src/manifolds/VectorBundle.jl
```diff
@@ -314,78 +107,52 @@ struct VectorBundle{TVS<:VectorSpaceType, TM<:Manifold} <: Manifold
...
 
 """
-    bundle_projection(B::VectorBundle, x::ProductRepr)
+    FVector(type::VectorSpaceType, data)
 
-Projection of point `x` from the bundle `M` to the base manifold.
-Returns the point on the base manifold `B.M` at which the vector part
-of `x` is attached.
+Decorator indicating that the vector `data` from a fiber of a vector bundle
+is of type `type`.
```
mateuszbaran: This word order seems more correct:

Decorator indicating that the vector `data` is from a fiber of a vector bundle
of type `type`.

src/projected_distribution.jl
```diff
@@ -16,29 +16,27 @@ function ProjectedPointDistribution(M::Manifold, d::Distribution, proj!, x::TRes
...
+    return MPointSupport(d.manifold)
+end
 
 """
     ProjectedFVectorDistribution(type::VectorBundleFibers, x, d, project_vector!)
 
 Generates a random vector from ambient space of manifold `type.manifold`
 at point `x` and projects it to vector space of type `type` using function
-`project_vector!`.
+[`project_vector`](@ref).
```
mateuszbaran: It uses the mutating variant so while the non-mutating one should have better docs, it's less precise.

src/manifolds/VectorBundle.jl
mateuszbaran: Sorry, I meant the mutating erroring variant:
```julia
"""
    project_vector!(B::VectorBundleFibers, v, x, w)
Project vector `w` from the vector space of type `B.VS` at point `x`
and save the result to `v`.
"""
function project_vector!(B::VectorBundleFibers, v, x, w)
    error("project_vector! not implemented for vector space family of type $(typeof(B)), output vector of type $(typeof(v)) and input vector at point $(typeof(x)) with type of w $(typeof(w)).")
end
```
It's still needed.

src/projected_distribution.jl
kellertuer: as for all other mutating ones, it actually oes not have a docstring itself (it also would up to 4-5 characters be a copy of the non-mutating docstring).

kellertuer: What do you think about the following: We document somewhere quite prominently, that it is often preferable to use a mutating version, but the documentation always refers to the (then uniquely documented) non-mutating version? That's how we currently document, so I would prefer to continue that way – but yes the docs could mention this approach more prominently, maybe within the page for the interface.

mateuszbaran: I was talking about this particular docstring for `ProjectedFVectorDistribution` -- it should say something like
```
projects it to vector space of type `type` using function `project_vector!`. See
[`project_vector`](@ref) for documentation.
```

I think everything else is fine.

kellertuer: Ah, soryy, then I misunderstood this. I thought you wanted to have an explicit docstring for `project_vector!`(which would up to an additional signature parameter just be the copy of the nonmutating one), but you proposed change looks nice, it says explicitly, the function is using the mustating one, while one should check the docs of the non-mutating one. Nice solution, will add that tomorrow.

src/manifolds/VectorBundle.jl
kellertuer: Ah, I might have forgotten the more general one, since I might have mistaken the special one (tangent) to be the mutating one and accidentally deleted your general one (I'll omit the docs however).

src/distributions.jl
kellertuer: I think I did not change that line. Do you propose to change it? If so, what would be your suggestion?

mateuszbaran: Right, it was like that before. I think something like this would be better:

Value support for vector bundle fiber-valued distributions (values from a fiber of a vector bundle at point `x` from the given manifold).
For example used for tangent vector-valued distributions.

kellertuer:
I would like to merge this not that in a hurry but not too far in the future, because it introduces quite some changes, though merely in reordering things. If I should merge some of the open features in here, merge them faster ;)

sethaxen:
I'm not planning to merge any of my PRs before this is merged. I think it will be easier to merge them on this than the other way around. I haven't had a chance to look at recent changes, but I trust you and would also like to see this merged very soon. :)

mateuszbaran:
I also think we should merge this PR first 👍 .

PR #86, opened 2020-01-04
Orthonormal basis support

mateuszbaran:
Moved from https://github.com/JuliaNLSolvers/ManifoldsBase.jl/pull/19 .

I'm not sure how to properly implement Gram-Schmidt here, projected vectors won't be linearly independent.


mateuszbaran:
How do we deal with bases of complex spaces? For example if I have `Euclidean(2;field=ℂ)`, do we return the real-coefficient basis `[(1,0), (0,1), (im,0), (0,im)]` or the complex-coefficient basis `[(1,0), (0,1)]` ?

sethaxen:
> I'm not sure how to properly implement Gram-Schmidt here, projected vectors won't be linearly independent.

Can you elaborate on the problem with an example?

> How do we deal with bases of complex spaces? For example if I have `Euclidean(2;field=ℂ)`, do we return the real-coefficient basis `[(1,0), (0,1), (im,0), (0,im)]` or the complex-coefficient basis `[(1,0), (0,1)]` ?

I prefer the real-coefficient basis. I like the idea of an ONB letting one represent any vector space with a real column vector with length the same as manifold dimension, which would also make Jacobians, etc a lot easier.

mateuszbaran:
> > I'm not sure how to properly implement Gram-Schmidt here, projected vectors won't be linearly independent.
> 
> Can you elaborate on the problem with an example?

The variants of Gram-Schmidt I know assume linear independence and clearly the procedure must be modified when vectors are not linearly independent. You can take for example three vectors [1,1], [0,eps], [1+eps,1] for a small positive eps. How do you end with one basis vector?

> I prefer the real-coefficient basis. I like the idea of an ONB letting one represent any vector space with a real column vector with length the same as manifold dimension, which would also make Jacobians, etc a lot easier.

OK, I'll use the real-coefficient basis then.

mateuszbaran:
I think I won't add projected ONBs for non-real embedded manifolds in this PR. It looks too complicated.

sethaxen:
> The variants of Gram-Schmidt I know assume linear independence and clearly the procedure must be modified when vectors are not linearly independent. You can take for example three vectors [1,1], [0,eps], [1+eps,1] for a small positive eps. How do you end with one basis vector?

Yes, I ran into this issue in the code snippet I posted on the slack channel a while back. For embedded manifolds I first projected to the tangent space and then peformed Gram-Schmidt. Before accepting a vector, I checked if it's orthonormal to all accepted vectors. Here's a code-snippet using `dot`:

```julia
function gs(E; kwargs...)
    Ξ = [E[1] ./ √dot(E[1], E[1])]
    for i in 2:length(E)
        Ei = E[i]
        for j in 1:length(Ξ)
            Ei = Ei .- dot(Ei, Ξ[j]) .* Ξ[j]
        end
        Ξi = Ei ./ √dot(Ei, Ei)
        indep = true
        for j in 1:length(Ξ)
            if !isapprox(dot(Ξi, Ξ[j]), 0.0; kwargs...)
                indep = false
                break
            end
        end
        indep && push!(Ξ, Ξi)
    end
    return Ξ
end

julia> E = [[1,1], [0,eps()], [1+eps(),1]]
3-element Array{Array{Float64,1},1}:
 [1.0, 1.0]                  
 [0.0, 2.220446049250313e-16]
 [1.0000000000000002, 1.0]   

julia> gs(E)
1-element Array{Array{Float64,1},1}:
 [0.7071067811865475, 0.7071067811865475]
```

sethaxen:
Here's an implementation of Gram-Schmidt using our manifold functions that should work for complex manifolds:

```julia
using Manifolds

function gram_schmidt(M::Manifold, x, E::AbstractVector; kwargs...)
    N = length(E)
    Ξ = empty(E)
    dim = manifold_dimension(M)
    N < dim && @warn "Input only has $(N) vectors, but manifold dimension is $(dim)."
    K = 0
    @inbounds for n in 1:N
        Ξₙ = project_tangent(M, x, E[n])
        for k in 1:K
            Ξₙ .-= real(inner(M, x, Ξₙ, Ξ[k])) .* Ξ[k]
        end
        nrmΞₙ = norm(M, x, Ξₙ)
        if nrmΞₙ == 0
            @warn "Input vector $(n) has length 0."
            @goto skip
        end
        Ξₙ ./= nrmΞₙ
        for k in 1:K
            if !isapprox(real(inner(M, x, Ξₙ, Ξ[k])), 0; kwargs...)
                @warn "Input vector $(n) is not linearly independent of output basis vector $(k)."
                @goto skip
            end
        end
        push!(Ξ, Ξₙ)
        K += 1
        K == dim && return Ξ
        @label skip
    end
    @warn "gram_schmidt only found $(K) orthonormal basis vectors, but manifold dimension is $(dim)."
    return Ξ
end
```

A demo using the above example:

```julia
julia> M = Euclidean(2);

julia> E = [[1,1], [0,eps()], [1+eps(),1]];

julia> Ξ = gram_schmidt(M, [0.0 0.0], E)
┌ Warning: Input vector 2 is not linearly independent of output basis vector 1.
└ @ Main REPL[195]:20
┌ Warning: Input vector 3 is not linearly independent of output basis vector 1.
└ @ Main REPL[195]:20
┌ Warning: gram_schmidt only found 1 orthonormal basis vectors, but manifold dimension is 2.
└ @ Main REPL[195]:29
1-element Array{Array{Float64,1},1}:
 [0.7071067811865475, 0.7071067811865475]

julia> gram_schmidt(M, [0.0 0.0], E; atol = 1e-3)
2-element Array{Array{Float64,1},1}:
 [0.7071067811865475, 0.7071067811865475] 
 [-0.7071067811865474, 0.7071067811865477]
```

And a demo on complex `Grassmann`:

```julia
julia> using LinearAlgebra

julia> M = Grassmann(3, 2, ℂ);

julia> x = [1.0 0.0; 0.0 1.0; 0.0 0.0];

julia> dim = manifold_dimension(M)
4

julia> Ξ = gram_schmidt(M, [1.0 0.0; 0.0 1.0; 0.0 0.0], [randn(ComplexF64, 3, 2) for i=1:dim]; atol=1e-10);

julia> all(is_tangent_vector.(Ref(M), Ref(x), Ξ))
true

julia> [real(inner(M, x, Ξ[i], Ξ[j])) for i=1:dim,j=1:dim] ≈ I
true

julia> v = project_tangent(M, x, randn(ComplexF64, 3, 2))
3×2 Array{Complex{Float64},2}:
      0.0+0.0im           0.0+0.0im     
      0.0+0.0im           0.0+0.0im     
 -1.26956-0.033828im  0.60129+0.984574im

julia> v2 = real.(inner.(Ref(M), Ref(x), Ref(v), Ξ))
4-element Array{Float64,1}:
  0.013840730908482701
 -1.5834545453191948  
  0.4812559772479172  
 -0.4524836143839637  

julia> sum(prod, zip(v2, Ξ)) ≈ v
true

```

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,185 @@
...
+along the curve.
+
+See also: [`inverse_represent_in_basis`](@ref), [`basis`](@ref)
+"""
+function represent_in_basis(M::Manifold, x, v, B::AbstractBasis)
+    error("represent_in_basis not implemented for manifold of type $(typeof(M)) a point of type $(typeof(x)), tangent vector of type $(typeof(v)) and basis of type $(typeof(B)).")
+end
+
+function represent_in_basis(M::Manifold, x, v, B::PrecomputedOrthonormalBasis)
+    return map(vb -> inner(M, x, v, vb), B.vectors)
```
sethaxen: Here and below, this will need a call to `real` to get real coefficients on manifolds not on a real field.

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,185 @@
...
+end
+
+"""
+    inverse_represent_in_basis(M::Manifold, x, v, B::AbstractBasis)
+
+Convert a one-dimentional vector of coefficients in a basis `B` of
```
sethaxen:
```suggestion
Convert a one-dimensional vector of coefficients in a basis `B` of
```

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,185 @@
...
+function inverse_represent_in_basis(M::Manifold, x, v, B::AbstractBasis)
+    error("inverse_represent_in_basis not implemented for manifold of type $(typeof(M)) a point of type $(typeof(x)), tangent vector of type $(typeof(v)) and basis of type $(typeof(B)).")
+end
+
+function inverse_represent_in_basis(M::Manifold, x, v, B::PrecomputedOrthonormalBasis)
+    # quite convoluted but:
+    #  1) preserves the correct `eltype`
+    #  2) guarantees a reasonable array type `vout`
+    #     (for example scalar * `SizedArray` is an `SArray`)
+    vt = v[1] * B.vectors[1]
```
sethaxen:
```suggestion
    vt = v[1] .* B.vectors[1]
```

mateuszbaran:
> Here's an implementation of Gram-Schmidt using our manifold functions that should work for complex manifolds:

Cool, thanks! I think you can replace `@goto` with `continue` there.

The other problem with projected ONBs here is that we currently have to API for getting a basis of the ambient space. This could be tricky if someone wanted to get a projected ONB for a product of a real and a complex manifold.

> RE euclidean basis vectors, we can use a generated function to create those for a speed boost.

If we wanted to optimise it, I'd probably start with removing the need to allocate basis vectors one by one.

> Does the SVD orthogonalization work for non-real manifolds?

It should work with minor modifications (we need to prepare a real matrix for SVD by expanding complex vectors into real and imaginary parts).

> I'm rethinking my earlier comment about the field of the coefficients. It might be a good idea to 
have a type parameter that determines whether the field of the coefficients is the same as that of the manifold or whether it's real. I suspect some users with complex/quaternionic manifolds might want ONBs that still give them complex/quaternionic coefficients. And when an ONB doesn't apply in the non-real coefficient case, we can just leave it non-implemented.

OK, that's reasonable.

src/Euclidean.jl
```diff
@@ -170,20 +170,36 @@ end
 
 function basis(M::Euclidean{<:Tuple,F}, x, B::ArbitraryOrthonormalBasis) where F
     vecs = [_euclidean_basis_vector(x, i) for i in eachindex(x)]
-    return PrecomputedOrthonormalBasis(vecs)
+    if F == ℝ
```
sethaxen: I'd prefer that this is implemented as an overload of `basis` rather than burying this check within a function.

sethaxen:
> > Here's an implementation of Gram-Schmidt using our manifold functions that should work for complex manifolds:
> 
> Cool, thanks! I think you can replace `@goto` with `continue` there.

One of them but not the other (it's in an inner loop but needs to break the outer loop).

> The other problem with projected ONBs here is that we currently have to API for getting a basis of the ambient space. This could be tricky if someone wanted to get a projected ONB for a product of a real and a complex manifold.

Why is this a problem? Currently if one has a product of a real and complex manifold, their tangent vectors will already have to be promoted to be of mixed type. The basis vectors here would not be special.

> > Does the SVD orthogonalization work for non-real manifolds?
> 
> It should work with minor modifications (we need to prepare a real matrix for SVD by expanding complex vectors into real and imaginary parts).

We should probably have a utility function like `realify` and `fieldify` (names TBD) that take a number/array of numbers in some field and expand one dimension by a factor of `field_dimension` #90 to get an equivalent real representation, or for `fieldify` to do the reverse.

src/Euclidean.jl
mateuszbaran: Ok, I'll change that.

mateuszbaran:
> > 
> 
> One of them but not the other (it's in an inner loop but needs to break the outer loop).

Yes, right.

> Why is this a problem? Currently if one has a product of a real and complex manifold, their tangent vectors will already have to be promoted to be of mixed type. The basis vectors here would not be special.

That's a problem because I use  `_euclidean_basis_vector` to get basis vectors and I don't know which ones (if any) correspond to real or complex dimensions. Consider for example `Euclidean(1;field=ℂ) \times Euclidean(1)`. It has 3 basis vectors: (1,0), (i,0), (0,1). All I have is a manifold object and a point `x` which may be just a pair of real numbers. I need to figure out that the ambient space is ℂ \times ℝ to generate the basis but we don't have an interface for that yet.

> 
> We should probably have a utility function like `realify` and `fieldify` (names TBD) that take a number/array of numbers in some field and expand one dimension by a factor of `field_dimension` #90 to get an equivalent real representation, or for `fieldify` to do the reverse.

Yes, that would be useful but I think it's out of scope of this PR.

src/orthonormal_bases.jl
mateuszbaran: I've fixed this one but where is the other real needed below? I don't see it.

sethaxen:
> That's a problem because I use `_euclidean_basis_vector` to get basis vectors and I don't know which ones (if any) correspond to real or complex dimensions. Consider for example `Euclidean(1;field=ℂ) \times Euclidean(1)`. It has 3 basis vectors: (1,0), (i,0), (0,1). All I have is a manifold object and a point `x` which may be just a pair of real numbers. I need to figure out that the ambient space is ℂ \times ℝ to generate the basis but we don't have an interface for that yet.

I think it's useful to think about the minimal function set necessary for a user to get basis vectors on a `Euclidean` manifold for a custom field with a type not in Base. Take as an example a quaternion field using `Quaternions.Quaternion`. How could a user pass a real point and get back `Quaternion` basis vectors if their field was the quaternions?

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,186 @@
...
+
+function basis(M::ArrayManifold, x, B::AbstractPrecomputedOrthonormalBasis)
+    N = length(B)
+    M_dim = manifold_dimension(M)
+    N == M_dim || throw(ArgumentError("Incorrect number of basis vectors; expected: $M_dim, given: $N"))
+    for i in 1:N
+        vi_norm = norm(M, x, B.vectors[i])
+        isapprox(vi_norm, 1) || throw(ArgumentError("vector number $i is not normalized (norm = $vi_norm)"))
+        for j in i+1:N
+            dot_val = inner(M, x, B.vectors[i], B.vectors[j])
```
sethaxen: This one needs a `real`

mateuszbaran: Right, of course :+1:

mateuszbaran:
I've added sphere and product ONBs (product ONBs have a special representation). I'll think about non-real bases again when real bases are better developed.

src/manifolds/Rotations.jl
```diff
@@ -307,6 +307,10 @@ Tangent vectors are represented by matrices.
 """
 inner(M::Rotations, x, w, v) = dot(w, v)
 
+function inverse_represent_in_basis(M::Rotations, x, v, B::ArbitraryOrthonormalBasis) where N
+    return hat(M, x, v)
```
sethaxen: There's a missing factor of √2 to make this an ONB.

```suggestion
    T = Base.promote_eltype(x, v)
    return hat(M, x, v) ./ sqrt(T(2))
```

src/manifolds/Rotations.jl
```diff
@@ -553,6 +557,10 @@ where tangent vectors are represented by elements from the Lie group
 project_tangent(::Rotations, ::Any...)
 project_tangent!(M::Rotations, w, x, v) = w .= (v .- transpose(v))./2
 
+function represent_in_basis(M::Rotations, x, v, B::ArbitraryOrthonormalBasis) where N
+    return vee(M, x, v)
```
sethaxen:
```suggestion
    T = Base.promote_eltype(x, v)
    return vee(M, x, v) .* sqrt(T(2))
```

src/manifolds/Sphere.jl
```diff
@@ -114,6 +114,16 @@ metric from the embedding, i.e. $ (v,w)_x = v^\mathrm{T}w $.
 """
 @inline inner(S::Sphere, x, w, v) = dot(w, v)
 
+function inverse_represent_in_basis(M::Sphere{N}, x, v, B::ArbitraryOrthonormalBasis) where N
+    if isapprox(x[end], -1)
```
sethaxen: This is a minor point, but I prefer that the privileged point be `(1,...,0)`, not `(0,...,-1)`. On most spheres, there's no reason to prefer one point over another, but for `Sphere(1)` and `Sphere(3)` identified as the circle and compact symplectic groups, the point `(1,...,0)` is privileged. For consistency, I think that point should be privileged for all spheres.

src/manifolds/Sphere.jl
```diff
@@ -233,6 +243,15 @@ Project the point `v` onto the tangent space at `x` on the [`Sphere`](@ref) `M`.
 project_tangent(::Sphere, ::Any...)
 project_tangent!(S::Sphere, w, x, v) = (w .= v .- dot(x, v) .* x)
 
+function represent_in_basis(M::Sphere{N}, x, v, B::ArbitraryOrthonormalBasis) where N
```
sethaxen: Can you add a docstring with how this method works?

src/manifolds/Sphere.jl
```diff
@@ -114,6 +114,16 @@ metric from the embedding, i.e. $ (v,w)_x = v^\mathrm{T}w $.
 """
 @inline inner(S::Sphere, x, w, v) = dot(w, v)
 
+function inverse_represent_in_basis(M::Sphere{N}, x, v, B::ArbitraryOrthonormalBasis) where N
+    if isapprox(x[end], -1)
+        return vcat(x, 0)
```
sethaxen:
```suggestion
        return vcat(v, 0)
```

test/utils.jl
```diff
@@ -273,6 +276,50 @@ function test_manifold(M::Manifold, pts::AbstractVector;
...
+        if !isa(btype, ProjectedOrthonormalBasis)
+            v1 = inverse_retract(M, x, pts[2], default_inverse_retraction_method)
+            vb = represent_in_basis(M, x, v1, btype)
+
+            @test represent_in_basis(M, x, v1, b) ≈ represent_in_basis(M, x, v1, btype)
+            @test inverse_represent_in_basis(M, x, vb, b) ≈ inverse_represent_in_basis(M, x, vb, btype)
+        end
+    end
+
+    for btype ∈ (basis_types_to_from..., basis_types_vecs...)
```
sethaxen: Should also check for orthonormality here for `AbstractOrthonormalBasis`.

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,186 @@
...
+    return B
+end
+
+function _euclidean_basis_vector(x, i)
+    y = zero(x)
+    y[i] = 1
+    return y
+end
+
+function basis(M::Manifold, x, B::ProjectedOrthonormalBasis{:svd})
```
sethaxen: You could define a default `represent_in_basis` and `inverse_represent_in_basis` for `ProjectedOrthonormalBasis` like this:
```julia
represent_in_basis(M::Manifold, x, v, B::ProjectedOrthonormalBasis) = represent_in_basis(M, x, v, basis(M, x, B))
inverse_represent_in_basis(M::Manifold, x, v, B::ProjectedOrthonormalBasis) = inverse_represent_in_basis(M, x, v, basis(M, x, B))
```
with clear documentation that it's more efficient to precompute if one intends to reuse.

src/manifolds/Rotations.jl
```diff
@@ -553,6 +557,10 @@ where tangent vectors are represented by elements from the Lie group
 project_tangent(::Rotations, ::Any...)
 project_tangent!(M::Rotations, w, x, v) = w .= (v .- transpose(v))./2
 
+function represent_in_basis(M::Rotations, x, v, B::ArbitraryOrthonormalBasis) where N
```
kellertuer: while this reads ok, the inverse reads really strange. Can we think about another name? `basis_coefficients_of(M,x,v,B)` maybe? for the inverse I only had the idea of `compose` but that name is taken right?

kellertuer: Also, docstrings would be nice :)

mateuszbaran: `compose` is taken for groups, right. I don't really have a better ideas for a name.

I'll add more docstrings later.

src/orthonormal_bases.jl
mateuszbaran: I've been thinking about it but this default is so inefficient that I'd like to steer people away from doing this. People who don't want to store the basis on their side should use `ArbitraryOrthonormalBasis` (or some other efficient method we may have in the future).

It could be a good idea, though, to implement `ArbitraryOrthonormalBasis` using projected `basis` if we don't have a better idea.

test/utils.jl
mateuszbaran: Probably by using `inverse_represent_in_basis` on (1, 0, ..., 0), (0, 1, 0, ..., 0), ..., (0, 0, ..., 0, 1) and checking returned vectors, right?

src/manifolds/Rotations.jl
kellertuer: Yes, I also don't have a very good name, but pondering around on the wikipedia page ( https://en.wikipedia.org/wiki/Basis_(linear_algebra) ) maybe we could call

* `represent_in_basis` -> `get_coordinates` (in the sense: get coordinates of `v` with respect to the basis `B`)
* `inverse_represent_in_basis` -> `get_vector` (in the sense: get vector of the coordinates `w` with respect to the basis `B`, though this name sounds a little too generic.

mateuszbaran: OK, `get_coordinates` and `get_vector` sound better. I'll change the code to use these names.

mateuszbaran:
@kellertuer Do you need a more complete support for ONBs on a circle? Or a different one? It turned out to be surprisingly tricky.

kellertuer:
No, I would love, just for completeness, to have the one with the curvature, but last week was quite busy, so I was not able to keep up, what the current status here is concerning that function.

mateuszbaran:
So far I've updated the `tangent_orthonormal_basis` function you've made for SPD to the new API. It returns that curvature. The PR is still a work-in-progress, I'll leave a message when I'm satisfied with the changes.

src/manifolds/Circle.jl
```diff
@@ -115,6 +115,23 @@ function flat!(::Circle, v::FVector{CotangentSpaceType}, x, w::FVector{TangentSp
 end
 flat(M::Circle, x::Number, w::FVector{TangentSpaceType}) = FVector(CotangentSpace,w.data)
 
+
+function get_coordinates(M::Circle{ℝ}, x, v, B::ArbitraryOrthonormalBasis)
+    return v
+end
+
+function get_coordinates(M::Circle{ℂ}, x, v, B::ArbitraryOrthonormalBasis)
+    return @SVector [imag(v[1] / x[1])]
```
sethaxen: This knocks off some compute cycles
```suggestion
    v, x = v[1], x[1]
    w = imag(v) * real(x) - real(v) * imag(x)
    return @SVector [w]
```

src/manifolds/Circle.jl
```diff
@@ -211,7 +228,7 @@ mean modulo 2π.
 mean(::Circle, ::Any)
 mean(::Circle,x::Array{<:Real},w::AbstractVector; kwargs...) = sym_rem(sum(w.*x))
 
-@inline norm(::Circle, x, v) = sum(abs.(v))
+@inline norm(::Circle, x, v) = sum(map(abs, v))
```
sethaxen: Should be able to do this
```suggestion
@inline norm(::Circle, x, v) = sum(abs, v)
```

src/manifolds/Euclidean.jl
```diff
@@ -40,6 +40,16 @@ This metric is the default metric for example for the [`Euclidean`](@ref) manifo
 """
 struct EuclideanMetric <: RiemannianMetric end
 
+function basis(M::Euclidean{<:Tuple, ℝ}, x, B::ArbitraryOrthonormalBasis)
+    vecs = [_euclidean_basis_vector(x, i) for i in eachindex(x)]
+    return PrecomputedOrthonormalBasis(vecs)
+end
+
+function basis(M::Euclidean{<:Tuple, ℂ}, x, B::ArbitraryOrthonormalBasis)
```
sethaxen: I don't think this corresponds to the basis used above. That basis should be
`[im*x]`.

sethaxen: Sorry, wrong file. I meant this as the ONB for the `Circle` case.

src/manifolds/Circle.jl
```diff
@@ -15,6 +15,12 @@ represented by `ℂ`-valued `Circle` of unit numbers.
 struct Circle{F} <: Manifold where {F <: AbstractNumbers} end
 Circle(f::AbstractNumbers=ℝ) = Circle{f}()
 
+function basis(M::Circle{ℝ}, x, B::DiagonalizingOrthonormalBasis)
+    sbv = sign(B.v[1])
+    vs = @SVector [@SVector [sbv == 0 ? 1 : sbv]]
```
sethaxen: Probably want to use `one(sbv)` for type-stability.

src/manifolds/Euclidean.jl
```diff
@@ -25,6 +25,8 @@ The dimension of this space is $k \dim_ℝ 𝔽$, where $\dim_ℝ 𝔽$ is the
 struct Euclidean{N<:Tuple,T} <: Manifold where {N, T<: AbstractNumbers} end
 Euclidean(n::Vararg{Int,N};field::AbstractNumbers=ℝ) where N = Euclidean{Tuple{n...},field}()
 
+const ArbitraryOrDiagonalizingBasis = Union{ArbitraryOrthonormalBasis, DiagonalizingOrthonormalBasis}
```
sethaxen: Might be good to put this in `Manifolds.jl`

mateuszbaran: Maybe `orthonormal_bases.jl` is an even better place?

mateuszbaran:
@sethaxen I've made some rough changes regarding complex-coefficient bases. Could you take a look if this looks reasonable?

I've modified your Gram-Schmidt and added it as an option, although it currently doesn't generate correct bases when basis scalar and manifold scalar differ. I'm not sure how to fix that, I don't like any of my ideas.

mateuszbaran:
> I haven't had a chance to look closely, but one idea I had was a function like this:
> 

Yes, well, this approach looks more or less appropriate. The problem with it (and some of the ideas I had) is that it looks like a lot of work that doesn't look particularly important.

src/manifolds/Circle.jl
```diff
@@ -115,6 +121,35 @@ function flat!(::Circle, v::FVector{CotangentSpaceType}, x, w::FVector{TangentSp
 end
 flat(M::Circle, x::Number, w::FVector{TangentSpaceType}) = FVector(CotangentSpace,w.data)
 
+
+function get_coordinates(M::Circle{ℝ}, x, v, B::ArbitraryOrthonormalBasis)
+    return v
+end
+
+function get_coordinates(M::Circle{ℝ}, x, v, B::DiagonalizingOrthonormalBasis)
```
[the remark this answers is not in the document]
mateuszbaran: I've spent a few hours trying to support scalar points here. It breaks some assumptions in other methods and tests. Things like converting to coordinates and back should return a value of the original type.

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,302 @@
+
+"""
+    AbstractBasis{F<:AbstractNumbers}
```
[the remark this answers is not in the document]
mateuszbaran: OK, I'll change that.

src/manifolds/Circle.jl
```diff
@@ -67,7 +73,7 @@ end
 
 Compute the inner product of two (complex) numbers with in the complex plane.
 """
-complex_dot(a,b) = dot(real.(a),real.(b)) + dot(imag.(a),imag.(b))
+complex_dot(a,b) = dot(map(real, a), map(real, b)) + dot(map(imag, a), map(imag, b))
```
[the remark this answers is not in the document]
mateuszbaran: I tried that but some tests fail after this change.

mateuszbaran:
I think the PR looks more or less OK. I think it has all diagonalizing bases from `manopt.jl` and quite well developed support for real-coefficient bases. It even has partial support for complex-coefficient bases, although properly supporting it would require quite a bit of additional work and I'd prefer to focus on different issues in the near future.

Do you see any issues that should be addressed here?

kellertuer:
This looks fine and despite my short technical command (which is nearly unrelated to the usual content) all looks fine :)

src/Manifolds.jl
```diff
@@ -21,6 +21,7 @@ using Einsum: @einsum
 using FiniteDifferences
 using HybridArrays
 using LinearAlgebra
+using ManifoldsBase
```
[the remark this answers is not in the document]
mateuszbaran: I'll revert this, I've added it as a temporary fix and it's not really needed.

docs/src/orthonormal_bases.md
```diff
@@ -0,0 +1,21 @@
+# Orthonormal bases
+
+Following functions and types provide support for orthonormal bases of the tangent space of different manifolds.
```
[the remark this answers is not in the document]
mateuszbaran: O, I'll add an equation.

docs/src/orthonormal_bases.md
```diff
@@ -0,0 +1,21 @@
...
+
+The main types are:
+* [`ArbitraryOrthonormalBasis`](@ref) that is designed to work when no special properties of the tangent space basis are required. It is designed to make [`get_coordinates`](@ref) and [`get_vector`](@ref) fast.
+* [`DiagonalizingOrthonormalBasis`](@ref) that diagonalizes the curvature tensor and makes the curvature in selected direction equal to 0.
+* [`ProjectedOrthonormalBasis`](@ref) that projects a basis of the ambient space and orthonormalizes projections to obtain a basis in a generic way.
+* Subtypes of [`AbstractPrecomputedOrthonormalBasis`](@ref) that work by storing (either explicitly or implicitly) a precomputed basis at a certain point. [`ArbitraryOrthonormalBasis`](@ref) doesn't require precomputing but [`DiagonalizingOrthonormalBasis`](@ref) and [`DiagonalizingOrthonormalBasis`](@ref) usually do.
+
+The main functions are:
```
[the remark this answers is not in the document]
mateuszbaran: OK, I'll prefix all of them with `get_`.

src/manifolds/Circle.jl
```diff
@@ -67,7 +73,7 @@ end
 
 Compute the inner product of two (complex) numbers with in the complex plane.
 """
-complex_dot(a,b) = dot(real.(a),real.(b)) + dot(imag.(a),imag.(b))
+complex_dot(a, b) = dot(map(real, a), map(real, b)) + dot(map(imag, a), map(imag, b))
```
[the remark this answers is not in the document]
mateuszbaran: Making this change breaks some tests:
```julia
Complex Circle Basics: Error During Test at /home/mateusz/repos/Manifolds.jl/test/circle.jl:65
  Got exception outside of a @test
  MethodError: no method matching isapprox(::SArray{Tuple{1},Float64,1,1}, ::Int64)
  Closest candidates are:
    isapprox(!Matched::Missing, ::Any; kwargs...) at missing.jl:90
    isapprox(!Matched::DoubleFloat{T}, ::F; atol, rtol, nans, norm) where {T<:Union{Float16, Float32, Float64}, F<:Real} at /home/mateusz/.julia/packages/DoubleFloats/0E72N/src/extras/misc.jl:23
    isapprox(!Matched::Number, ::Number; atol, rtol, nans) at floatfuncs.jl:274
    ...
  Stacktrace:
   [1] log!(::Circle{ℂ}, ::MArray{Tuple{1},Complex{Float64},1,1}, ::SArray{Tuple{1},Complex{Float64},1,1}, ::SArray{Tuple{1},Complex{Float64},1,1}) at /home/mateusz/repos/Manifolds.jl/src/manifolds/Circle.jl:228
   [2] top-level scope at /home/mateusz/repos/Manifolds.jl/test/circle.jl:82
   [3] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [4] top-level scope at /home/mateusz/repos/Manifolds.jl/test/circle.jl:66
   [5] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [6] top-level scope at /home/mateusz/repos/Manifolds.jl/test/circle.jl:4
```

src/orthonormal_bases.jl
```diff
@@ -0,0 +1,325 @@
...
+end
+function get_vector(M::DefaultManifold, x, v, ::ArbitraryOrthonormalBasis)
+    return reshape(v, representation_size(M))
+end
+
+function basis(M::DefaultManifold, x, ::ArbitraryOrthonormalBasis)
+    return PrecomputedOrthonormalBasis([_euclidean_basis_vector(x, i) for i in eachindex(x)])
+end
+
+function basis(M::Manifold, x, B::ProjectedOrthonormalBasis{:gram_schmidt,F}; kwargs...) where F
```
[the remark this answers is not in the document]
mateuszbaran: This also works for the complex coefficient basis of tangent spaces of the complex Grassman manifold (it's tested).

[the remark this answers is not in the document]
sethaxen: Hmm, it shouldn't, since it's a different inner product, hence a different meaning of "orthonormal". I suspect it's passing because the test only looks at the real part of the inner product. An orthonormal vector with complex coefficients will have a complex inner product between non-identical basis vectors of `0+0im`. For real coefficients, IIRC, the imaginary part is usually non-zero.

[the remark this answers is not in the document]
mateuszbaran: Right, that's possible. I'll change the signature.

mateuszbaran:
> Very nice work! Just a few comments, mainly to do with consistency with the number system of the coefficients and efficient indexing with `@inbounds`.
> 
> A few thoughts I had of things I'd like to see that can wait for future PRs:
> 
> * Interconversion between bases and their dual. (i.e. using `sharp`/`flat`, may require some specification RE whether the basis is for the vector or covector space)
> * introduction of `cov` using an ONB.

Thanks! I think I've addressed all issues except that one about `complex_dot`. Interconversion and `cov` should be relatively easy to implement using the API we have here.

Issue #90, opened 2020-01-05
Add function for field dimension


sethaxen:
Several manifolds are defined over fields, where the dimension of the manifold is dependent on the dimension of the field, i.e. the size of the smallest real space that the field can be identified with (frankly, I'm surprised Julia base doesn't have such a function). Rather than reimplementing `manifold_dimension` for every field/manifold combination, implement whenever possible a generic `manifold_dimension` that depends on the dimension of the field. Then add the following methods

```julia
field_dimension(::RealNumbers) = 1
field_dimension(::ComplexNumbers) = 2

manifold_dimension(::Euclidean{N,𝔽}) where {N,𝔽} = *(field_dimension(𝔽), N.parameters...)
manifold_dimension(::SymmetricMatrices{N,𝔽}) where {N,𝔽} = field_dimension(𝔽) * div(N*(N+1),2)
manifold_dimension(::Stiefel{M,N,𝔽}) where {M,N,𝔽} = field_dimension(𝔽) * (M*N - div(N*(N-1), 2)) - N
manifold_dimension(::FixedRankMatrices{M,N,k,𝔽}) where {M,N,k} = field_dimension(𝔽) * (M+N-k) * k
manifold_dimension(::Grassmann{N,K,𝔽}) where {N,K,𝔽} = field_dimension(𝔽)*K*(N-K)
```

To introduce a new field, e.g. quaternions, a user can then just add
```julia
struct QuaternionNumbers <: AbstractField end
field_dimension(::QuaternionNumbers) = 4
```
(To be honest, we should probably just add `QuaternionNumbers`, since adding it doesn't require us to pick any one implementation of quaternions)

kellertuer:
I think this is fine, for those manifolds where the dimensions differ more than just by this factor two, one can still overwrite both cases individually. The `manifold_dimensions` of course belong into the  seperate files (so maybe wait for my nearly finished rework).

PR #91, opened 2020-01-06
Add real_dimension and quaternion number system

sethaxen:
Implements #90. We can wait to merge this until #83 has been merged.

Note that this is just a first step towards handling fields for which types do not exist in Julia base, with quaternions as a motivating example. e.g. We'll want to find an interface for checking that a point on the manifold is compatible with the field of the manifold that doesn't require us to know the type of the point. We'll also want a way to realify and fieldify arrays.

Likewise, we'll want some better support for working with manifolds in terms of their field. Right now we do this a bit haphazardly. e.g. for complex numbers, we return the Hermitian inner product, but we return the real `angle` that comes from realifying the tangent space (we could return angles in the field as well, though they're harder to interpret). Likewise we define the dimension as over the realified manifold, not wrt its own field (the ratio is a factor of `field_dimension`). And for ONB's in #86 we're currently only considering vectors with real coefficients, which likewise comes from realifying the manifold.

A future PR will try to sort out all of this by defining an interface that lets one customize the field that these functions are computed with respect to (defaulting to `RealNumbers`).


mateuszbaran:
That's a good idea, although I'd like to note that quaternions are not a field. We could maybe call it a "scalar structure" or something.

sethaxen:
Or perhaps we rename AbstractField to AbstractNumberSystem and
field_dimension to number_dimension? That's also consistent with the names
of our current fields.

sethaxen:
Actually I agree, scalar is better, since manifolds will have charts in a
vector space. So perhaps AbstractScalar. And I think field_dimension should
be real_dimension for clarity. Should we rename RealNumbers to RealScalars?

kellertuer:
...I am also for `RealScalars` since it bears more consistency.

sethaxen:
Upon further thought, I'm going to go with `AbstractNumbers`, `RealNumbers`, ..., `QuaternionNumbers`, because scalars are also fields, so that distinction is no better.

sethaxen:
I think this PR is feature-complete.

src/Manifolds.jl
```diff
@@ -222,13 +204,13 @@ export AbstractEstimationMethod, GradientDescentEstimation, CyclicProximalPointE
...
 export base_manifold, bundle_projection, christoffel_symbols_first, christoffel_symbols_second,
     christoffel_symbols_second_jacobian, complex_dot, det_local_metric, distance,
-    einstein_tensor, exp, exp!, flat, flat!, gaussian_curvature, geodesic, hat, hat!,
-    injectivity_radius, inner, inverse_local_metric, inverse_retract, inverse_retract!,
-    isapprox, is_decorator_manifold, is_default_metric, is_manifold_point,
-    is_tangent_vector, isapprox, inner, kurtosis, local_metric, local_metric_jacobian,
-    log, log!, log_local_metric_density, manifold_dimension, metric, mean, mean!,
-    mean_and_var, mean_and_std, median, median!, moment, norm, normal_tvector_distribution,
-    one, project_point, project_point!, project_tangent, project_tangent!,
+    einstein_tensor, exp, exp!, real_dimension, flat, flat!, gaussian_curvature, geodesic,
```
mateuszbaran: `real_dimension` is not in alphabetical order

src/manifolds/Euclidean.jl
```diff
@@ -129,10 +129,12 @@ log_local_metric_density(M::MetricManifold{<:Manifold,EuclideanMetric}, x) = zer
...
 
 Return the manifold dimension of the [`Euclidean`](@ref) `M`, i.e.
-the product of all array dimensions.
+the product of all array dimensions and the [`real_dimension`](@ref) of the
+underlying number system.
 """
-@generated manifold_dimension(::Euclidean{N,ℝ}) where {N} = *(N.parameters...)
-@generated manifold_dimension(::Euclidean{N,ℂ}) where {N} = 2*( *(N.parameters...) )
+@generated function manifold_dimension(::Euclidean{N,𝔽}) where {N,𝔽}
+    return *(real_dimension(𝔽), N.parameters...)
```
mateuszbaran: I advise extracting the call to `real_dimension` outside of the generated function. Calling it here makes all method of `real_dimension` defined after this method of `manifold_dimension` potentially inaccessible.

It should be something like 
```julia
@generated _product_of_dimensions(::Euclidean{N}) where N = *(N.parameters...)
function manifold_dimension(M::Euclidean{N,𝔽}) where {N,𝔽}
    return real_dimension(𝔽) * _product_of_dimensions(M)
end
```

mateuszbaran:
Apart from these two small issues this PR looks good.

Issue #65, opened 2019-12-18
[RFC] Graph manifold


kellertuer:
I would like to use the holidays to think about modelling another manifold that extends the `PowerManifold` to some extend: Manifold-valued data on a graph. The idea stems from [this](http://dx.doi.org/10.1137/17M1118567) paper and at that time was implemented in Matlab but I was never happy with the implementation. I started some ideas in Julia, but I would like to do that efficiently. It's actually a nice interplay of data on vertices and edges on a graph.

# The setting
Let G be a graph with V its set of vertices and E the set of edges (each edge having a start point end an end point from V) and M be a manifold. Then we have four types of data

* manifold-valued data on the vertices, i.e. f: V-> M
* manifold-valued data on the on the edges, i.e. g: E -> M
* tangent vectors on edges, i.e. W: E->TM
* tangent vectors on vertices, i.e. Z: V-> TM (which might not that much be used)

If we call the `GraphManifold` GM, the idea is, that for example `log(GM, f, g)` for two manifold-valued data sets on the vertices yields tangent vectors on the edges, namely for an edge (v,w) in E we get the term log_{f_v} g_w. We can then mimic (see literature above) the Graph Laplace by producing Z: V -> TM with Z(v) = sum of all edges that start in v of these logs (note they are all log_f(v) - terms. And for example then perform a `exp(GM,f,Z)`

# RFC

I am not 100% sure how such an approach works for manifold-valued data on the edges, but I am curious to ponder about that 

My main question for comments is: What is a good and efficient way to represent this? the data f can be done similar to the Power manifold, but the GM also has to store its graph (adjacency matrix for example). Here I would like to ask for ideas/comments what to use – [LightGraphs.jl](https://github.com/JuliaGraphs/LightGraphs.jl) for example?

mateuszbaran:
I wonder how this relates to the noncommutative discrete differential geometry as described here: https://phorgyphynance.wordpress.com/ or here: https://ncatlab.org/ericforgy/files/mexico.pdf ?

Regarding efficient representation, what kinds of relatively complex operations would be the most common? Adding/deleting edges? Computing graph `log`s and `exp`s? Something else?

kellertuer:
Summing over all adjacent edges to assign a vector to a vertex is maybe also something. The `exp(GM,f,Z)` can be seen as a classical one on the power manifold |V|.

kellertuer:
[...]

PR #92, opened 2020-01-12
GraphManifold and Torus

kellertuer:
This is the branch to work on #65, where quite some things will be similar to `PowerManifold`, just on edges/vertices and I am not 100% sure yet about efficiency. the discuses `incident_log` is available in a first version.


src/manifolds/GraphManifold.jl
```diff
@@ -0,0 +1,89 @@
...
+
+# Fields
+* `G` is an `AbstractSimpleGraph`
+* `M` is a [`Manifold`](@ref)
+* `T` is a [`GraphManifoldType`](@ref), i.e. whether data is given on vertices
+or edges.
+"""
+struct GraphManifold{
+        G<:AbstractSimpleGraph,
+        M<:Manifold,
```
mateuszbaran: You could constrain it to `M<:PowerManifold` I think.

src/manifolds/GraphManifold.jl
```diff
@@ -0,0 +1,89 @@
...
+each node the sum of the [`log`](@ref)s to incident nodes is summed.
+"""
+function incident_log(M::GraphManifold{G,TM,VertexGraph}, x) where {G <: AbstractSimpleGraph,TM<:Manifold}
+    v = zero_tangent_vector(M,x)
+    return incident_log!(M, v, x)
+end
+function incident_log!(M::GraphManifold{G,TM,VertexGraph}, v, x) where {G <: AbstractSimpleGraph,TM<:Manifold}
+    for e in edges(M.graph)
+        v[src(e)] += log(M.manifold, x[src(e)], x[dst(e)])
+    end
```
mateuszbaran: This function should return `v`.

src/manifolds/GraphManifold.jl
```diff
@@ -0,0 +1,89 @@
...
+
+check whether `x` is a valid point on the [`GraphManifold`](@ref)
+"""
+check_manifold_point(::GraphManifold,::Any)
+
+function check_manifold_point(
+    M::GraphManifold{G,TM,VertexGraph},
+    x;
+    kwargs...) where {G <: AbstractSimpleGraph,TM<:Manifold}
+    for i in nv(M.graph)
```
mateuszbaran: This could just call `check_manifold_point` on the `PowerManifold`. I guess a lot of operations could be just forwarded to `PowerManifold`.

mateuszbaran: Of course, there could be some additional checks here, like whether the power manifold has the number of components that matches the graph.

src/manifolds/GraphManifold.jl
kellertuer: Neat idea, would spare quite a few definitions.

src/manifolds/GraphManifold.jl
kellertuer: yeah, with the idea to the `PowerManifold` that sounds right, and the additional check is then just the matching size. Very neat. This also means a lot of functions are already implemented (in Power).

kellertuer:
I am still suck in the sub typing, since I can not properly subtype it without the `TSize`? And I am not sure how to infer that.

mateuszbaran:
Sure, having an `AbstractProductManifold` sounds like a good solution here.

kellertuer:
Don't worry, when reworking (that went really smooth by the way) i noticed two docstrings also having `Product` instead of `Power` in them ;)

mateuszbaran:
Thanks for fixing these docstrings, I'm glad that reworking went smoothly 🙂 . We have had a huge progress during the last few weeks.

kellertuer:
Yes, we had; and it was really just deciding, that the `GraphManifold` can not be dispatched per size, so the `AbstractPowerManifold{M}` has just one type, which just exploded `representation _size` and `manifold_dimension`, but those are still one-liners for graphs and the rest directly works for Graphs by just adding `Abstract` to the definitions. So really smooth :), which I think might indicate having a well-thought design.

sethaxen:
I haven't had a chance to review this, but why not decorate the `PowerManifold`? There are a few other places that might make sense for creating a new manifold that is topologically a power manifold, e.g. the `Torus`.

kellertuer:
> I haven't had a chance to review this, but why not decorate the `PowerManifold`? There are a few other places that might make sense for creating a new manifold that is topologically a power manifold, e.g. the `Torus`.

The Torus _is_ basically just a Power Manifold, namely `PowerManifold(Circle),2)`.  And the implementation here is now nearly that and there's two argumentfs for the current approach

1.  I did not want to store the size of the manifold twice. For the `GraphManifold` where the data lives on the vertices, the dimension of the `Graphmanifold` is number-of-vertices (`nv(G)`) times the dimension of the inner manifold. If I would set the field `manifold` of the `GraphManifold` to be the corresponding `PowerManifold`, I'd not only have to initialize that one to be `PowerManifold(M, nv(G))`, I'd even have to update that when I change my Grph in beween. In the current form, if you add vertex to the graph, you just update `G` and add a point to the data

2. If I would do this decorating, I would have to write at least a line for every function (`exp`, `log`, transports, dimension, representation size, onbs, statistics...) to “pass it down to the inner power manifold. With the `AbstractPowerManifold` (which is not the decorating approach, mor like: typewise `PowerManifold` and `GraphManifold` are siblings and their parent is the abstract one), _all_ functions of `PoweManifold` are _directly_ available for `GraphManifold` (since they rae now implemented for the abstract case), despite _two_: `representation_size` and `manifold_dimension`, where the implementation for the `PowerManifold` employs its ' TSize` field and `GraphManifold` either its number of vertices or its number of edges. So that's 6 lines per `GraphManifoldType` and no passing down as described in the first argument.

And thinking of that, basically one should now do `Torus <: AbstractPowerManifold`, and the dimension of the torus would play the game of the number of vertices and would also directly inherit all functions from the abstract one. You just have to implement the same 6 lines for the above two functions, all others the torus directly gets from `AbstractPowerManifold` (and `Circle` of course. Yay!

kellertuer:
Oh and since I noticed that it's really easy to do the torus, yes 24 lines of code, I just shortly did that to illustrate. The torus should be complete (up to documentation and testing), since we have the methods from `AbstractPowerManifold` now and the `Circle` works for non-mutating data.

kellertuer:
I (again) struggle a little bit to get anything to run with a `PowerManifold` in testing since my points on a graph manifold as I imagined them just throw errors (even if I do 6-element array instead of 3x2), I haven't figured out why. Despite that – testing (graph and torus) and documenting torus is the only things left on this PR.

src/manifolds/GraphManifold.jl
```diff
@@ -0,0 +1,194 @@
+
+using LightGraphs
+import LightGraphs: AbstractSimpleGraph, is_directed
+import SimpleWeightedGraphs: AbstractSimpleWeightedGraph
```
mateuszbaran: Do you actually use the functionality of `SimpleWeightedGraphs`? It looks like it was designed for different purposes.

src/manifolds/GraphManifold.jl
```diff
@@ -0,0 +1,194 @@
...
+    representationsize(N::GraphManifold{G,M,VertexManifold})
+
+returns the representation size of a point on the [`GraphManifold`](@ref) on the vertices of $G=(V,E),
+which is the representation size of `M` with $\lvert V \rvert$ added.
+"""
+function representation_size(M::GraphManifold{<: AbstractSimpleGraph,<: Manifold, VertexManifold})
+    return (representation_size(M.manifold)..., nv(M.graph))
+end
+@doc doc"""
+    representationsize(N::GraphManifold{G,M,EdgeManifold})
```
mateuszbaran:
```suggestion
    representation_size(N::GraphManifold{G,M,EdgeManifold})
```

src/manifolds/GraphManifold.jl
```diff
@@ -0,0 +1,194 @@
...
+````math
+d_{\mathcal N} = \lvert E \rVert d_{\mathcal M}.
+````
+"""
+function manifold_dimension(M::GraphManifold{<: AbstractSimpleGraph, <: Manifold, EdgeManifold})
+    return manifold_dimension(M.manifold)*ne(M.graph)
+end
+
+@doc doc"""
+    representationsize(N::GraphManifold{G,M,VertexManifold})
```
mateuszbaran:
```suggestion
    representation_size(N::GraphManifold{G,M,VertexManifold})
```

src/manifolds/GraphManifold.jl
kellertuer: The edges have weights and I want to use exactly weights from edges in the `incident_log` if you habe a weighted graph, so I think I am using that exactly what it is introduced for. If you have a graph without weights, the `incident_log` assumes weights to be one.

kellertuer:
> Ah, right, I haven't properly documented how to make points on a power manifold. 

I added more detail on the stuff I stumbled upon today in #93, because it should be discussed – and I think can be solved, it's merely a design and interface decision.

src/manifolds/GraphManifold.jl
mateuszbaran: OK, I see. It would be a good idea to add a test with a weighted graph then, and maybe a reference to what these weights may be useful for.

kellertuer: Of course. My goal (from a paper to adapt the example) is that the graph models a surface where the edges are weighted by distance and the `incident_log` models a graph laplacian, of course that needs to be documented and tested, yes – will of course do that in the `md` then.

kellertuer:
For now this should work with the current power representations, though I still have to write a test for the `incident_log`, I hope the general test already works.

kellertuer:
Hm, fighting again the compatibility issues with the packages, somehow I mixed up the Manifest and I am not yet able to fix such issues because I do not understand where they come from and what one has to do to fix that. Locally this does not appear (why ever). Any ideas welcome.

kellertuer:
...and one thing I also have to check (which functions I'd have to overwrite) is a product manifold of immutable types, some dimensions xo not yet fit this (i.e. the representation size, the inner one, basically has to be `(,)`.

mateuszbaran:
> Hm, fighting again the compatibility issues with the packages, somehow I mixed up the Manifest and I am not yet able to fix such issues because I do not understand where they come from and what one has to do to fix that. Locally this does not appear (why ever). Any ideas welcome.
> 
> Edit: Since I really can under no circumstances reproduce the errors any of the test versions in the CI give us, I have no clue how to resolve this, even with updated packages the tests fail during installation for any Julia version on the CI; while on my machine everything works fine with Julia 1.3.

I'm not an expert in solving these issues but maybe reverting to the old `Manifest.toml` (the one we have on master branch) would help?

> ...and one thing I also have to check (which functions I'd have to overwrite) is a product manifold of immutable types, some dimensions xo not yet fit this (i.e. the representation size, the inner one, basically has to be `(,)` for the real circle.

`representation_size` should be `(1,)` for the real circle, and only the one-element array approach works with the multidimensional array power manifold representation. Nested representation won't need `representation_size` at all.

kellertuer:
Well I had hoped the torus would work with data like `[0, 0]` for the `Torus(2)`, currently it'll just work with `[[0.],[0.]]`, which is a little odd, that's what I have to check still.

Problem from the old Manifest ist, that I somehow have to add a few packages, most prominently `LightGraphs`, but maybe mixing both might work, I will try tomorrow.

mateuszbaran:
> Well I had hoped the torus would work with data like `[0, 0]` for the `Torus(2)`, currently it'll just work with `[[0.],[0.]]`, which is a little odd, that's what I have to check still.

Yes, that's a bit unfortunate, although using a 1-element `StaticVector` inside gives the same memory layout as the flat array and we need less special-casing here. I'm not sure how much special casing would be needed to make `[0,0]` work though.

> Problem from the old Manifest ist, that I somehow have to add a few packages, most prominently `LightGraphs`, but maybe mixing both might work, I will try tomorrow.

You could try pinning all packages from `Project.toml` at the exact versions from the `Manifest.toml` that is in master and then add new packages.

kellertuer:
Sorry, I am running out of ideas and can't figure out a way to get packages working, especially since _all_ commits I pushed here do work locally on my side, while some `getindex` seems to be missing sometimes in Julia 1.0 and `FunctionWrappers` just does strange things on versioning, that I do not understand nor can I find something in some documentation somewhere how to solve this (we might reach 1000 test runs very fast this way though).

mateuszbaran:
These package issues look terrible. Maybe someone on slack would have some ideas?

kellertuer:
Well it's a dependency of a dependency of a dependency of a dependency (.....) that is too restrictive as it seems and that's for at least two packages, but somehow this only happens in CI (at least I really can not reproduce here), so it's all just guessing for the last 1.5 hours where I usually planned to do at least something – now I will probably just got home after the next fail.

kellertuer:
Finally it's running again. Resolving that is really not much fun and partly just guessing :/

kellertuer:
While I increased test coverage and all new stuff is covered, the Grassmannian throws some basis vector warnings I don't understand, because I don't even see where `get_basis` is called; and for the `Torus` while it seems well defined and inherits a lot of functions of the `AbstractProduct`, I haven't found a(ny) way to define a point on that manifold... `[0.,0.]` would be nice to have but does not work, `SVector.([0. 0.])` also does not seem to work on some reshape,  so that's still an open point – document and illustrate how this could be used (or having product manifolds of `Circle()` with elements on the manifold being numbers to phrase differently).

kellertuer:
I also opened a topic on discourse to discuss yesterdays fun-with-dependencies. https://discourse.julialang.org/t/schematic-way-to-resolve-dependency-issues-in-a-package-development/33512

mateuszbaran:
> Finally it's running again. Resolving that is really not much fun and partly just guessing :/

Great! I hope someone describes a better way to solve such issues in your thread on discourse.

> While I increased test coverage and all new stuff is covered, the Grassmannian throws some basis vector warnings I don't understand, because I don't even see where `get_basis` is called; and for the `Torus` while it seems well defined and inherits a lot of functions of the `AbstractProduct`, I haven't found a(ny) way to define a point on that manifold... `[0.,0.]` would be nice to have but does not work, `SVector.([0. 0.])` also does not seem to work on some reshape, so that's still an open point – document and illustrate how this could be used (or having product manifolds of `Circle()` with elements on the manifold being numbers to phrase differently).

These warnings with Grassmann manifolds are, for now, expected. The Gram-Schmidt orthonormalization reports them. We could probably disable them by default.

I'll work on the product manifold today, I'll post here some details later.

kellertuer:
If that's expected for now – ok :) I saw that I can add a few tests still to fix a few linesin Stiefel and Symmetric, will add that today somewhen and if I find time look into starting the docs.

mateuszbaran:
Quick update: to make a point on the torus `PowerManifold(Circle(), 3)` you can use `a = [1.0 2.0 3.0]`. Later, `a = [[1.0], [2.0], [3.0]]` will also work. `a = [1.0, 2.0, 3.0]` can't work with the `representation_size` of `Circle` we currently have.

mateuszbaran:
I think it would make sense to change the `representation_size` of `Circle` to `()` (so that `a = [1.0, 2.0, 3.0]` is the correct way to make a point on a torus) but it would require some adjustments.

kellertuer:
Yes, we can and  try. I am still confusing the versions with and without comma from time to time, though. As long as we document the version as a column vector should still be fine (I have to check the Torus representation_size, still)

mateuszbaran:
I've changed the `representation_size` of `Circle` to `()` in #98 , so in that PR `a = [1.0, 2.0, 3.0]` is the correct way to make a point on a torus 🙂 .

kellertuer:
The documentation I just started has then to be adapted, but despite that - great!

kellertuer:
All tests are fine, I have full code coverage and the documentaion is updated, so if all is fine, this could be (squash) merged.

Issue #89, opened 2020-01-04
Invariant metrics on groups


sethaxen:
I think the default behavior of metric-dependent functions like `inner`, `exp`, and `log` on a `GroupManifold` should be to forward to the base manifold, i.e. by default groups use the default metric of the base manifold. They can be equipped with a generic metric like any manifold, but they can also be equipped a custom metric using the group structure.

Given a group `G`, an automorphism `τ` of the group on itself (i.e. left or right-translation), and a metric `g` on the Lie algebra, we can define a `τ`-invariant metric with inner product

```
g_x(v, w) = g_e(dτ_x^{-1} v, dτ_x^{-1} w),
```

We can thus evaluate this inner product everywhere and numerically integrate the geodesics. When the metric is bi-invariant, then we can also define the Riemannian logarithm and exponential in terms of the same maps at the tangent space, which in this case coincide with the Lie logarithm and exponential:

```math
\begin{aligned}
\exp_x &= τ_x \circ \exp_e \circ dτ_x^{-1}\\
\log_x &= dτ_x \circ \log_e \circ τ_x
\end{aligned}
```

The proposed new types are:

```julia
struct InvariantMetric{G,D} <: Metric
    metric::G
    direction::D
end
function InvariantMetric(metric::Metric, conv::ActionDirection = LeftAction())
    return InvariantMetric{typeof(G),typeof(conv)}(metric, conv)
end
const LeftInvariantMetric{G} = InvariantMetric{G,RightAction}
const RightInvariantMetric{G} = InvariantMetric{G,LeftAction}
```

where `G` is a metric on the Lie algebra.

When one knows that a given metric on a given group is bi-invariant, one can simply overload `exp!` to point to `gexp!` or `group_exp!` or whatever we name it and likewise `log!` to `glog!` or `group_log!`.

mateuszbaran:
I like that by default `GroupManifold` forwards manifold-related operations to decorated manifold. Having left- and right-invariant metrics as separate decorators looks like the correct approach.

> When one knows that a given metric on a given group is bi-invariant, one can simply overload `exp!` to point to `gexp!` or `group_exp!` or whatever we name it and likewise `log!` to `glog!` or `group_log!`.

This is, I think, the problematic part of the proposal. Both left- and right- invariant inner products define some metrics on the manifold, sometimes different ones. If the metric is not bi-invariant, the group exponential will again be different from Riemannian exponential.

[...]

Issue #75, opened 2019-12-28
Add the Lie-theoretic exponential map


mateuszbaran:
As discussed in https://github.com/JuliaNLSolvers/Manifolds.jl/pull/68#discussion_r361746072

sethaxen:
The first decision here seems to be what to call it. Options I can think of are:
- `gexp`
- `expg`
- `group_exp`
- `exp_group`

I prefer `expg` by analogy to `expm` in some languages being the matrix exponential, but I'm not thrilled with any of them.

The second decision is where to implement it. For `AdditionOperator`, it will coincide with the Riemannian exponential, and we can just forward it.
For `MultiplicationOperator`, it should default to calling `exp` for `Matrix`/`Number` inputs and otherwise be unimplemented. This would cover most textbook groups with a multiplication operation.

Analogously, I prefer `logg` (though `glog` sounds great, and now I remember I have some glög left over in the fridge).

kellertuer:
We can of course call it `glögg` (even `gløgg` is close by), I have ö on my keyboard directly, no problem.

While `expm` is short and one knows it's the matrix Exponential, I would vote for `group_exp` for readability. Another possibility is to use the `exp(::LieGroup{M}, ::Any...)` signature maybe, where `LieGroup <: GroupManifold` or we do it with `GroupManifold`? And where I got to names, should we maybe call `GroupManifold` `LieManifold` ? For me, Lie just sounds mor familiar.

mateuszbaran:
Sure, we can have `glögg` as an alias for Lie logarithm 😄 . As primary names, `group_exp` and `group_log` sound fine (although I also like `lie_exp` and `lie_log`).

kellertuer:
I also like `lie_exp`, but maybe then we should really go for `LieManifold` for consistency?

mateuszbaran:
Right, it would be inconsistent to sometimes use `Lie` and sometimes `group`. Both options, `Lie` and `group`, are fine for me.

kellertuer:
I am fine with a decision pro-`group`, but my personal favour would be `Lie`.

sethaxen:
[...]

PR #96, opened 2020-01-17
Add `group_exp`, `group_log`, and invariant metrics

sethaxen:
This PR will add the `group_exp` and `group_log` functions detailed in #75, along with implementations for our existing groups.

It will also implement invariant metrics discussed in #89.

Finally, it also implements the syntax in #69.


sethaxen:
I made a few changes from the implementation discussed in #89. Namely, I added the trait-like functions `has_invariant_metric` and `has_biinvariant_metric`. These can be used to annotate that an existing metric when added to an existing group manifold is left- or right-invariant (bi-invariance can be deduced from these). These won't be used for anything for the moment.

On the other hand, one might want to take existing inner product at the identity element and translate it around. That can be done by wrapping an existing metric in `InvariantMetric`, with the added caveat that defining a metric as invariant doesn't make it so.

This isn't enough to do anything useful though. To numerically integrate the geodesic currently requires an implementation of `local_metric`. We previously noted that numerical integration of geodesics didn't work in embedded coordinates, but now that we have ONBs, we should be able to change that, but that's for a future PR.

I think this PR is more or less feature-complete and just needs a few more tests and docstrings/refs, so it's ready for review.

mateuszbaran:
Great! I'll take a look at the PR.

src/groups/group.jl
```diff
@@ -554,6 +554,215 @@ function inverse_translate_diff!(
...
+    group_exp(G::AbstractGroupManifold{MultiplicationOperation}, v)
+
+For number and `AbstractMatrix` types of `v`, compute the usual numeric/matrix exponential,
+
+````math
+\exp_e = \sum_{n=0}^\infty \frac{1}{n!} v^n.
+````
+"""
+group_exp(M::Manifold, v) = group_exp(M, v, is_decorator_manifold(M))
+group_exp(M::Manifold, v, ::Val{true}) = group_exp(M.manifold, v)
```
mateuszbaran:
```suggestion
group_exp(M::Manifold, v, ::Val{true}) = group_exp(base_manifold(M), v)
```
We should probably use `base_manifold` in such places.

src/groups/group.jl
```diff
@@ -554,6 +554,215 @@ function inverse_translate_diff!(
...
+        x,
+        v,
+        method::GroupExponentialRetraction{<:ActionDirection},
+    )
+
+Compute the retraction using the group exponential "translated" to any point on the
+manifold. With a group translation $τ_x$ in a specified direction, the retraction is
+
+````math
+\operatorname{retr}_x = τ_x \circ \exp_e \circ (\mathrm{d}τ_x)_x^{-1}
```
mateuszbaran: We should probably use different symbols for Riemannian and group exponentials?

mateuszbaran: Similarly for logarithmic maps.

src/groups/metric.jl
```diff
@@ -0,0 +1,132 @@
...
+
+"""
+    has_biinvariant_metric(G::AbstractGroupManifold) -> Val
+
+Return `Val(true)` if the metric on the manifold is bi-invariant, that is, if the metric
+is both left- and right-invariant (see [`has_invariant_metric`](@ref)).
+"""
+function has_biinvariant_metric(M::Manifold)
+    return Val(
+        (
```
mateuszbaran: Why do you need these additional round brackets? It looks like to would return `Val((true,))` or `Val((false,))` instead of `Val(true)` or `Val(false)`.

src/groups/group.jl
sethaxen: The issue here is that `base_manifold` strips away all of the decorators, including `GroupManifold` and is therefore only appropriate for functions that are defined on the topological manifold. But I see now that it would probably be consistent to use the `is_decorator_group`/`base_group` pattern used for the other group-specific manifolds defined above.

mateuszbaran: Right, `is_decorator_group`/`base_group` is the right solution :+1:

src/groups/group.jl
sethaxen: I'm reluctant to use anything too different, to avoid changing from standard convention. I thought giving it a base `e` would be sufficient, but I see now that one can of course have the Riemannian exponential with base `e`. We could capitalize the map name (e.g. `Exp`, and `Log`) or remove the base point entirely, but ultimately, the context will need to make clear which is meant.

src/groups/metric.jl
sethaxen: > Why do you need these additional round brackets?

Unsuccessful attempt to avoid dangling `&&` sign. Will remove. The comma is outside the inner parenthesis, so the two expressions are equivalent.

src/groups/group.jl
mateuszbaran: OK, it should be clear from the text description of formulas which one is used where.

src/groups/group.jl
sethaxen: After looking into this further I was wrong, the `is_decorator_group`/`base_group` pattern is reserved for interactions with the `Identity` element, which only makes sense in the context of its corresponding `AbstractGroupManifold`. The group-specific functions are designed so that they can be called on any manifold, though in general, they will only be implemented for `AbstractGroupManifold`. This lets one call e.g. `compose` or `group_exp` on a `ProductManifold`, which will of course error if any submanifolds are not groups but work just fine if they all are.

So for consistency, I'd like to keep this as written here. Do you have specific concerns with that? As far as I can tell, stripping off one decorator at a time this way until an implementation is reached has no more overhead than our `base_manifold` function does, and we already require that decorators have a `manifold` field that contains the decorated manifold.

mateuszbaran: This was just a minor concern that maybe we shouldn't require all manifold decorators to have a `.manifold` field. We can keep the current version.

sethaxen: I agree. The Julian way would be to have a function like `decorated_manifold(M::Manifold) = M.manifold` that can be overloaded if one used a different field name or maybe even doesn't store the decorated manifold within the decorator. We can handle that in a future PR though.

sethaxen:
Alright, this PR should be feature-complete. As long as the test suite passes, I think it's ready to merge.

src/groups/product_group.jl
```diff
@@ -285,7 +285,7 @@ function group_log(M::ProductManifold, q::ProductRepr)
     return ProductRepr(map(group_log, M.manifolds, submanifold_components(M, q))...)
 end
 function group_log(M::ProductManifold, q)
-    X = zero_tangent_vector(G, q)
+    X = zero_tangent_vector(M, q)
```
mateuszbaran: This should probably be replaced by `allocate_result`.

src/groups/special_euclidean.jl
```diff
@@ -119,20 +108,42 @@ function affine_matrix(G::SpecialEuclidean{n}, p) where {n}
 end
 affine_matrix(::SpecialEuclidean{n}, p::AbstractMatrix) where {n} = p
 @generated function affine_matrix(::GT, ::Identity{GT}) where {n,GT<:SpecialEuclidean{n}}
-    return SDiagonal{n}(I)
+    s = maybesize(Size(n, n))
+    s isa Size && return SDiagonal{n,Float64}(I)
```
mateuszbaran: Does it have to be a generated function? It's likely that constant propagation, inlining and optimization will make this fast anyway. And if not, `Diagonal{Float64}(I, n)` should probably be constructed at runtime (quoted return `:(Diagonal{Float64}(I, n))`) instead of compile time in case someone wants to modify it.

src/groups/group.jl
```diff
@@ -1,9 +1,9 @@
 @doc raw"""
     AbstractGroupOperation
 
-Abstract type for smooth binary operations $\circ$ on elements of a Lie group $G$:
+Abstract type for smooth binary operations $∘$ on elements of a Lie group $\mathcal{G}$:
```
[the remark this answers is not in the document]
sethaxen: It renders strangely when used in TeX as an operation (e.g. `x ∘ y`), but on its own as it is in these two strings, it renders correctly, so I'd like to use the symbol.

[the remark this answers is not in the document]
kellertuer: ah and on its own in markdown, too, but in math it looks fine (otherwise I'd recommended to just do `∘`, but then `$∘$` is fine.

src/groups/special_euclidean.jl
sethaxen: It doesn't need to be generated. For SE(3), there's no speed-up. It's just for higher dimensional SE(N) that you see a speed-up, but those are unlikely to be used.

Can you explain the benefit of returning quoted? How would one modify it?

mateuszbaran: Here is an example:
```julia
julia> @generated f(x) = Diagonal{Float64}(I, 3)
f (generic function with 1 method)

julia> using LinearAlgebra

julia> A = f(10)
3×3 Diagonal{Float64,Array{Float64,1}}:
 1.0   ⋅    ⋅ 
  ⋅   1.0   ⋅ 
  ⋅    ⋅   1.0

julia> A[1] = 10
10

julia> f(20)
3×3 Diagonal{Float64,Array{Float64,1}}:
 10.0   ⋅    ⋅ 
   ⋅   1.0   ⋅ 
   ⋅    ⋅   1.0

julia> @generated g(x) = :(Diagonal{Float64}(I, 3))
g (generic function with 1 method)

julia> B = g(10)
3×3 Diagonal{Float64,Array{Float64,1}}:
 1.0   ⋅    ⋅ 
  ⋅   1.0   ⋅ 
  ⋅    ⋅   1.0

julia> B[1] = 20
20

julia> g(99)
3×3 Diagonal{Float64,Array{Float64,1}}:
 1.0   ⋅    ⋅ 
  ⋅   1.0   ⋅ 
  ⋅    ⋅   1.0
```
The returned quoted expression is evaluated when the generated function is called, so it makes a difference when a mutable object is allocated.

sethaxen: Whoah, that is really good to know! You probably just saved me hours of future work hunting down bugs that would be caused by this. Thanks!

sethaxen:
Okay, all requested changes have been implemented. Unless there's anything else, I'd like to merge once the test suite has finished running.

src/groups/special_euclidean.jl
mateuszbaran: No problem :slightly_smiling_face: . Generated functions are quite tricky to write, so I prefer to keep them as simple as possible.

Issue #94, opened 2020-01-16
Run JuliaFormatter on source files


sethaxen:
I've been running [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) on new source files before committing. While it doesn't/can't enforce all of BlueStyle's recommendations, the formatting decisions it does make are compatible with BlueStyle and are quite nice. Running it ensures uniform style, even for things we miss. Also at some point in the future, we may be able to make format testing a part of our pipeline, like folks in Python can do with `black`.

 I think we'll just need one big PR that runs JuliaFormatter on all files. One formatting decision that this will change is the choice to switch from one import per line to multiple imports per line (JuliaFormatter prefers the latter, but this is actually a [BlueStyle convention](https://github.com/invenia/BlueStyle#module-imports); it makes diffs easier to read).

mateuszbaran:
OK, we can use JuliaFormatter.

sethaxen:
Okay, once #82 and #86 are merged, I'll submit a PR, since they make changes to quite a few existing files.

PR #97, opened 2020-01-18
More reformatting, additional conventions

sethaxen:
This PR runs JuliaFormatter on everything in `src/` (see #94 ). It works toward enforcing a few conventions:
- No dangling `=` signs. If a "one-line" function is too long to fit in a line,
  then make it multiline
- Always add a newline between things of different types (struct/method/const)
- Always add a newline between methods for different functions (including mutating/nonmutating variants)
- Prefer to have no newline between methods for the same function; when reasonable, merge the docstrings.
- The mutating method always goes after the nonmutating method
- Try to simplify the codebase by identifying repeating patterns and introducing helper functions to replace them
- For routine mutating calls like `y .= `, `copyto!`, `fill!`, `mut!`, or our own mutating functions, consistently assume that the returned value is the mutated object, as convention dictates.
- All `import`/`using`/`include` are in the main module file.

A few incorrect docstrings and one or two small bugs are also fixed.

### DIFF
--- a/docs/src/manifolds/rotations.md
+++ b/docs/src/manifolds/rotations.md
@@ -5,7 +5,8 @@ The manifold $\mathrm{SO}(n)$ of orthogonal matrices with determinant $+1$ in $\
 $\mathrm{SO}(n) = \bigl\{R \in \mathbb{R}^{n\times n} \big| RR^{\mathrm{T}} =
 R^{\mathrm{T}}R = \mathrm{I}_n, \det(R) = 1 \bigr\}$
 
-``\mathrm{SO}(n)`` is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
+The Lie group $\mathrm{SO}(n)$ is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
+See also [`SpecialOrthogonal`](@ref), which is this manifold equipped with the group operation.
 
 Tangent vectors are represented by elements of the corresponding Lie algebra, which is the tangent space at the identity element.
 This convention allows for more efficient operations on tangent vectors.
--- a/docs/src/manifolds/rotations.md
+++ b/docs/src/manifolds/rotations.md
@@ -5,11 +5,16 @@ The manifold $\mathrm{SO}(n)$ of orthogonal matrices with determinant $+1$ in $\
 $\mathrm{SO}(n) = \bigl\{R \in \mathbb{R}^{n\times n} \big| RR^{\mathrm{T}} =
 R^{\mathrm{T}}R = \mathrm{I}_n, \det(R) = 1 \bigr\}$
 
- $\mathrm{SO}(n)$ is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
+``\mathrm{SO}(n)`` is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
 
-Tangent vectors are represented by elements of the corresponding Lie algebra, which is the tangent space at the identity element. This convention allows for more efficient operations on tangent vectors. Tangent spaces at different points are different vector spaces.
+Tangent vectors are represented by elements of the corresponding Lie algebra, which is the tangent space at the identity element.
+This convention allows for more efficient operations on tangent vectors.
+Tangent spaces at different points are different vector spaces.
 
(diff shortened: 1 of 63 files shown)


mateuszbaran:
I like almost all of the changes here. I only don't like the way expressions like `condition || throw("something")` are divided between two lines. Specifically, that there is no indentation in the next line that starts with `throw`. It's not clear at the first glance that the second line is a continuation of them expression from the previous line. Unfortunately I don't see an option in JuliaFormatter that could change this. It's not that important though, I'm OK with not indenting the second line until JuliaFormatter adds an option for this.

src/manifolds/CholeskySpace.jl
```diff
@@ -31,13 +29,22 @@ The tolerance for the tests can be set using the `kwargs...`.
 """
 function check_manifold_point(M::CholeskySpace, x; kwargs...)
     if size(x) != representation_size(M)
-        return DomainError(size(x),"The point $(x) does not lie on $(M), since its size is not $(representation_size(M)).")
+        return DomainError(
+            size(x),
+            "The point $(x) does not lie on $(M), since its size is not $(representation_size(M)).",
```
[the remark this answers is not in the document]
sethaxen: That was all JuliaFormatter. 🙂

src/manifolds/Circle.jl
```diff
@@ -323,36 +326,50 @@ where [`log`](@ref) denotes the logarithmic map on `M`.
...
 end
 
-zero_tangent_vector(::Circle,x::Number) = zero(x)
-function zero_tangent_vector!(::Circle, v, x)
-    fill!(v, 0)
-    return v
-end
+zero_tangent_vector(::Circle, x::Number) = zero(x)
+
+zero_tangent_vector!(::Circle, v, x) = fill!(v, 0)
```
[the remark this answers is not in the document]
sethaxen: Yes, it is now unified everywhere (or at least it should be)

sethaxen:
> I like almost all of the changes here. I only don't like the way expressions like `condition || throw("something")` are divided between two lines. Specifically, that there is no indentation in the next line that starts with `throw`. It's not clear at the first glance that the second line is a continuation of them expression from the previous line. Unfortunately I don't see an option in JuliaFormatter that could change this. It's not that important though, I'm OK with not indenting the second line until JuliaFormatter adds an option for this.

Oh, I thought I had caught all of those, but I see now  that I missed a few ONB ones. I'll replace with an if statement.

Yes, this is an annoyance, and so is the wrapping a long single-liner at the equal sign. I'll probably file an issue to get options added to change that.

> Thanks for this thorough rework; are we now on a more-strict 80-characters limit? It seemed a little bit so. Despite that, adding this style as guide to contributions.md would be nice.

No, this is with JuliaFormatter's default, which is 92; that's also BlueStyle's limit. Some of our docstrings seem to be following an 80-character limit, which is probably worth changing. No formatter for those though. 😉 

What would you like to see in `contributions.md`?

kellertuer:
> What would you like to see in `contributions.md`?

It currently states that we follow CodeBlue, so maybe add, that it's CodeBlue and additional fine tuning by JuliaFomatter?

docs/src/manifolds/rotations.md
```diff
@@ -5,11 +5,16 @@ The manifold $\mathrm{SO}(n)$ of orthogonal matrices with determinant $+1$ in $\
 $\mathrm{SO}(n) = \bigl\{R \in \mathbb{R}^{n\times n} \big| RR^{\mathrm{T}} =
 R^{\mathrm{T}}R = \mathrm{I}_n, \det(R) = 1 \bigr\}$
 
- $\mathrm{SO}(n)$ is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
+``\mathrm{SO}(n)`` is a subgroup of the orthogonal group $\mathrm{O}(n)$ and also known as the special orthogonal group or the set of rotations group.
```
kellertuer: Does this do inline-math? Until now I always preferred `$...$` for inline math (and for a long time `\[...\]` for display) since it's compatible with LaTeX (even the nicer `\(...\)` works with MathJax. So if we change this, we should change it globally?

sethaxen: Yes, this does. I use it here because beginning a line with `$` makes Documenter think the equation should stand alone on its own line, which isn't what we want here. The double backtick syntax, which usually behaves just like `$`, does the right thing here. Maybe this is a Documenter bug, but this is a workaround. I also prefer the `$...$` in general.

sethaxen:
> Thanks for carefully reformatting and rephrasing the documentation

No problem, although this PR focuses mostly on code, and I only addressed documentation when I saw easy-to-fix issues along the way. In a future PR, I'd like to more carefully go through the documentation to unify notation and make the docs more simple and concise

> ...there were a few quite-german sentences (i.e. long ones) in there. Have you read Thomas Mann (The Buddenbrooks for example)? Maybe I read that too often ;)

Ha, I've not read the book, but my in-laws know if it.  I've noticed you _really_ like your "i.e." phrases. 😉 I tend toward long sentences too, but I've found simpler, more concise ones are usually easier to follow, especially for non-native English speakers, hence I prefer those for code documentation, which IMO shouldn't take as much work to parse as, say, a paper.

sethaxen:
So long as the test suite passes, I think this PR is ready to merge.

docs/src/manifolds/rotations.md
kellertuer: I would prefer then to rephrase this to “The manifold $\mathrm{SO}(n)$” (or the Lie group), since starting a sentence with a formula is not that nice anyways.

src/Manifolds.jl
```diff
@@ -33,6 +33,7 @@ import Base:
     transpose,
     zero
 import Distributions: _rand!, support
+import LightGraphs: AbstractSimpleGraph, is_directed
```
mateuszbaran: These could probably be imported with `using` instead? Similarly below in `import SimpleWeightedGraphs: AbstractSimpleWeightedGraph, get_weight`.

Issue #95, opened 2020-01-17
Respect input array type (static/non-static) unless unreasonable


sethaxen:
There are a few places in the codebase where we explicitly created `StaticArray`s or `Array`s regardless of the type of the input.

This isn't great. For one thing, it's always possible the dimension of a manifold is so large that a static array makes no sense. Also, for some of the Christoffel functions, even a 10-dimensional manifold can produce a length 1000 array, which isn't good even if the user did pass a static array.

I propose a few utility function of this form:

```julia
@generated maybesize(s::Size{S}) where {S} = prod(S) > 100 ? S : Size{S}()

_similar(x, args...) = similar(x, args...)
_similar(x, s::Size{S}) where {S} = similar(x, S...)
_similar(x::StaticArray, s::Size{S}) where {S} = similar(x, maybesize(s))
_similar(x, ::Type{T}, s::Size{S}) where {S,T} = similar(x, T, S...)
_similar(x::StaticArray, ::Type{T}, s::Size{S}) where {S,T} = similar(x, T, maybesize(s))

_reshape(x, args...) = reshape(x, args...)
_reshape(x, s::Size{S}) where {S} = reshape(x, S...)
_reshape(x::StaticArray, s::Size{S}) where {S} = reshape(x, s)
```

Unlike the usual `similar` and `reshape`, which if `Size` is passed will always return a sized array, these versions only return a sized array if the original array is sized and if the total length of the requested size is less than some hardcoded cut-off deemed by us to be the maximum size above which it is no longer worth using a sized array.

Whenever size information is known statically (e.g. using the `representation_size` or `manifold_dimension`), then one of these functions should be preferred over the original in base.

mateuszbaran:
This makes sense, I'm not fully satisfied with the way we currently handle static arrays.

One thing I'd change in you code is that generated function. Something like this:
```julia
@generated maybesize(s::Size{S}) where {S} = prod(S) > 100 ? S : :(s)
```
looks a bit nicer since the given size object would be reused instead of making a new one.

This change could be coordinated with another change I need to support nested power manifold. When a point is represented by an array of arrays, we should call similar on the innermost array instead of the outermost. The outer array should just be `map`ped and then the inner arrays would be filled. `similar` for `ProductRepr` already does that, it just needs to be extended to arrays of arrays. At the same time we could think about renaming this function as it's not really a standard `similar`.

mateuszbaran:
[...]

PR #98, opened 2020-01-18
Changes to PowerManifold

mateuszbaran:
I'm going to change a few things about `PowerManifold`, mainly add an option to represent points on a power manifold as nested arrays. It'll require some changes to allocation of new points in `ManifoldsBase.jl`.

As a start, I've changed the `representation_size` of `Circle` to `()`.

Ref #95


mateuszbaran:
Yes, and only `sym_rem` needed some tweaks. I'm often amazed by how easy it is to make such changes in our code. I'll document these changes when I'm finished with the new design of the power manifold and allocation.

`representation_size` equal to `()` means that point are represented by zero-dimensional arrays (for example `Ref(2.)`, `reshape([2.])`, `Scalar(2.0)`, `MArray(Scalar(2.0))`, zero-dimensional views, etc.). Using just numbers works with some methods but not others.

kellertuer:
For me the `sym_rem` still reads nearly the same, but I have to check what that difference now is (the difference is broadcast vs map?)

kellertuer:
...and for me, that some things are easier to do than we think, is to some extend a sign of well-thought design (without putting too much self-praise on us).

mateuszbaran:
Yes, I've just changed broadcast to `map` in `sym_rem`.

sethaxen:
Currently, with `Euclidean{Tuple{1}}`, we try to support both 1-element vectors and scalars as point types. Should we then limit `Euclidean{Tuple{1}}` to have vector elements and have `Euclidean{Tuple{}}` have scalar elements? This may be confusing to users. We would in effect be drawing a distinction between `ℝ` and `ℝ¹`.

kellertuer:
We are running into this dilemma anyways, also for the `Circle`, which is now by convention real numbers (and not vectors of length 1), which is exactely the difference in the `representation_size. returning `(1,)` or `()`. It's merely a technical complication, since mathematically we anyways identify numbers and vectors of length 1.

I would like to propose the following:

*  `Euclidean(1)` initialises the manifold to represent numbers (so internally its the `Tuple{}()`-type or `ℝ`.
* `Eculidean([1])` initialises the `ℝ¹` (returning `(1,)`as representation size).

This way the user get's what he expects in the usual case: `Euclidean(1)` is real numbers. They can however decide to explicitly have vectors of length 1.
We could even try to do something similar with the circle, though we have to invent new Field types for that I fear (or another parameter to dispatch on). So maybe we just leave the Circle as is. Basically `Torus(1)` should equal my proposed `Euclidean([1])` anyways right?

mateuszbaran:
One thing we should keep in mind is that if `representation_size` is equal to some tuple, then any mutable `AbstractArray` of that size should work as a point representation. Another thing to remember is that we weren't paying much attention to representing points using immutable objects, and representing them by just numbers can cause some problems. We need a distinction between `ℝ` and `ℝ¹`. Also, representation size `()` means representation using zero-dimensional scalars (but not just numbers). Representation using just numbers is actually the third option.

Maybe we could have one code for two of these cases, I don't know.

> Basically `Torus(1)` should equal my proposed `Euclidean([1])` anyways right?

Sounds right 🙂 .

sethaxen:
> I would like to propose the following:
> 
> * `Euclidean(1)` initialises the manifold to represent numbers (so internally its the `Tuple{}()`-type or `ℝ`.
> * `Eculidean([1])` initialises the `ℝ¹` (returning `(1,)`as representation size).
> 
> This way the user get's what he expects in the usual case: `Euclidean(1)` is real numbers. They can however decide to explicitly have vectors of length 1.

I think it would be cleaner for the user to enter `Euclidean()` to get `Euclidean{Tuple{}}` and `Euclidean(1)` to get `Euclidean{Tuple{}}` for several reasons. First, the use of the arguments is consistent across constructors (they are all packed into the `Tuple`). Second, the arguments in the constructor are the same as the splatted representation size. Third, `Euclidean` with no arguments gives `ℝ`, which is the only sensible way to interpret that. We will need to document that, as a convention, we represent `ℝ` and `ℝ¹` differently.

mateuszbaran:
> I think it would be cleaner for the user to enter `Euclidean()` to get `Euclidean{Tuple{}}` and `Euclidean(1)` to get `Euclidean{Tuple{}}` for several reasons.

I think you mean `Euclidean(1)` should give `Euclidean{Tuple{1}}`?

sethaxen:
> I think you mean `Euclidean(1)` should give `Euclidean{Tuple{1}}`?

Yes, sorry, I do mean that.

kellertuer:
Ah, I was a little confused of that mixup. Of course your proposal is more consistent, since any `Euclidean(n)` gets you a tuple-based manifold and `Euclidean()` and even be seen as a default ℝ, so I am also fine with this ordering.

mateuszbaran:
@kellertuer how do you want `Torus` and `GraphManifold` adapted to these changes to power manifolds that enable selection between multidimensional and nested representation? Currently both manifolds use the multidimensional representation but it could be configurable.

kellertuer:
Cool! I'd have to check the new code soon, then! For the `Torus`, I think both would be fine

* multidimensional as of now seems to have the disadvantage, that you always need a `1xn` vector so an `Array{T,2}` which is a little counter intuitive
* for nested I can nest with just a Float? Then I'd prefer that (i.e. row vector of floats)

For the `GraphManifold` I prefer nested.

mateuszbaran:
The code is more-or-less functionally complete. I just have to make more tests, docs and rename a couple of functions.

> * multidimensional as of now seems to have the disadvantage, that you always need a `1xn` vector so an `Array{T,2}` which is a little counter intuitive

No, multidimensional for torus is just a simple vector (`[1.0, 2.0, 0.0]`) while nested would be for example `[Ref(1.0), Ref(2.0), Ref(0.0)]`.

> * for nested I can nest with just a Float? Then I'd prefer that (i.e. row vector of floats)

No, in the nested representation you have at least two layers of array-like things. I think multidimensional is what you want here. Notice that I've changed the `representation_size` of circle to `()` so you don't need a two-dimensional array here. I'll write some documentation for that later.

> For the `GraphManifold` I prefer nested.

OK, I'll make nested the default for `GraphManifold`.

kellertuer:
Thanks for the clarification, then the `Torus` is multidimensional. And its even better, since that way we directly have an example for both the multidimensional and the nested Powermanifold.

mateuszbaran:
I think the PR is more or less complete. I just need to wait until `ManifoldsBase.jl` 0.4.0 is tagged to update Manifest file here and make sure CI passes. Do you think the PR looks OK?

kellertuer:
Since this is changes to the Power manifold – can you change the title of the doc page to read Power not Product, see https://julianlsolvers.github.io/Manifolds.jl/latest/manifolds/power.html ;) ?

docs/src/manifolds/power.md
```diff
@@ -1,4 +1,4 @@
-# Product manifold
+# Power manifold
```
kellertuer: Saw this one just, too, in the published docs.

mateuszbaran:
Already done :slightly_smiling_face: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/98/files#diff-93f9bdbf3611099d849ecfb3e13fc93eR1

src/manifolds/GraphManifold.jl
```diff
@@ -21,7 +21,7 @@ A type for a [`GraphManifold`](@ref) where the data is given on the vertices.
 struct VertexManifold <: GraphManifoldType end
 
 @doc doc"""
-    GraphManifold{G, M, T} <: AbstractPowerManifold{M}
+    GraphManifold{G, M, T} <: AbstractPowerManifold{M,,NestedPowerRepresentation}
```
kellertuer: Can we document here or in ? `graph.md` how one has to create such a nested data? That would be helpful to the user.

src/manifolds/GraphManifold.jl
```diff
@@ -210,28 +224,11 @@ function manifold_dimension(M::GraphManifold{<:AbstractGraph,<:Manifold,EdgeMani
...
-    representation_size(N::GraphManifold{G,M,EdgeManifold})
-
-returns the representation size of a point on the [`GraphManifold`](@ref) on the vertices of $G=(V,E),
-which is the representation size of `M` with $\lvert E \rvert$ added.
-"""
-function representation_size(M::GraphManifold{<:AbstractGraph,<:Manifold,EdgeManifold})
-    return (representation_size(M.manifold)..., ne(M.graph))
-end
-
+#=
```
kellertuer: what happened to the representation size – is it now just the one from `Power`?

src/manifolds/PowerManifold.jl
```diff
@@ -1,17 +1,45 @@
+"""
```
kellertuer: We should/could accompany the types by concrete examples how to generate such data, maybe here or in `power.md` or in the two occasions `GraphManifold` and `Torus` and link to those from here?

src/manifolds/PowerManifold.jl
```diff
@@ -116,6 +155,24 @@ function PrecomputedPowerOrthonormalBasis(
     return PrecomputedPowerOrthonormalBasis{typeof(bases),F}(bases)
 end
 
+const PowerManifoldMultidim = AbstractPowerManifold{<:Manifold, MultidimentionalArrayPowerRepresentation} where TSize
```
kellertuer: Maybe we could write this as `PowerManifoldMultidimensional`?

src/manifolds/Torus.jl
```diff
@@ -6,8 +6,8 @@ The n-dimensionsl torus is the $n$-dimensional product of the [´Circle`](@ref).
 The Circle is stored internally within `manifold`, such that all functions of
```
kellertuer: See comments above on documenting the `PowerManifold` examples.

test/utils.jl
```diff
@@ -12,6 +12,10 @@ using Test
 using LightGraphs
 using SimpleWeightedGraphs
 
+find_eps(x::Type{TN}) where TN<:Number = eps(real(TN))
```
kellertuer: is it a bad idea to just extend `eps` here?

docs/src/manifolds/power.md
mateuszbaran: So it looks correct now :slightly_smiling_face:

src/manifolds/GraphManifold.jl
mateuszbaran: Sure, I'll add an example to `graph.md`.

src/manifolds/GraphManifold.jl
mateuszbaran: There is none, essentially. If something isn't a flat array, it doesn't have representation size. We can think about returning `nothing` by default in such cases instead of throwing an error.

src/manifolds/PowerManifold.jl
mateuszbaran: OK, I'll add examples for graph manifold and torus and link it from here.

test/utils.jl
mateuszbaran: It could work but we wouldn't benefit from it. It's just a small utility for testing and it's more predictable if it's a separate function.

src/manifolds/Torus.jl
mateuszbaran: I've added a simple example. Would like to also have something else here?

kellertuer: Yes, looks nice, great.

mateuszbaran:
I hope these fails are because a wrong version of ManifoldsBase.jl is loaded. Locally tests pass for me.

mateuszbaran:
I think we can merge this now?

kellertuer:
Yes, code coverage looks great and the rest I already checked.

PR #102, opened 2020-01-23
Start testing Julia 1.4, stop testing Julia 1.2, and remove ODE support for Julia 1.0

sethaxen:
Julia 1.4.0-rc1 is now out. In addition to the nightly (which is now on Julia 1.5), we should start testing explicitly against 1.4. To keep our test suite roughly the same runtime, I've removed testing against Julia 1.2, since it is not a long term support release.

### DIFF
--- a/src/manifolds/MetricManifold.jl
+++ b/src/manifolds/MetricManifold.jl
@@ -528,7 +528,15 @@ function sharp!(M::N, v::TFVector, x, w::CoTFVector) where {N<:MetricManifold}
 end
 
 @doc doc"""
-    solve_exp_ode(M::MetricManifold, x, v, tspan; backend=:default, solver=AutoVern9(Rodas5()), kwargs...)
+    solve_exp_ode(
+        M::MetricManifold,
+        x,
+        v,
+        tspan;
+        backend = :default,
+        solver = AutoVern9(Rodas5()),
+        kwargs...,
+    )
 
 Approximate the exponential map on the manifold over the provided timespan
 assuming the Levi-Civita connection by solving the ordinary differential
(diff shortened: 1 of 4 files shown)


sethaxen:
Some of the builds are currently failing I believe because of the new version of OrdinaryDiffEq released today: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/releases/tag/v5.28.0

sethaxen:
Actually, the failing builds appear to be due to the release of DiffEqBase 6.13.0. Locally, pinning DiffEqBase to 6.12.5 for Julia 1.0 results in the tests passing. Specifically, I'm pretty sure it's something to do with this PR: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/426

I don't _think_ we're doing anything funny with our ODE integration, and this is just a minor release, so nothing _should_ have broken.

mateuszbaran:
Good idea, though I'd prefer to wait until there are no failures on 1.4 before merging this.

sethaxen:
> Good idea, though I'd prefer to wait until there are no failures on 1.4 before merging this.

I generally agree. I think for now though it makes sense to allow failures on 1.4 and if failures arise on our end, then fix them (e.g. in this PR before it is merged). It's always possible though that a failure will arise due to a dependency that isn't yet 1.4-compatible, and I don't think that should hold up future PRs.

mateuszbaran:
Sure, I didn't mean disallowing failures, just making sure that tests on 1.4 pass on at least a single commit. Otherwise these tests wouldn't provide much value for other PRs. We can have failures on 1.4 until it's actually released.

sethaxen:
It sounds like it's going to be tricky to resolve the failures on Julia 1.0. It would need to happen from the OrdinaryDiffEq side. See discussion here: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/426. Perhaps we should skip the OrdinaryDiffEq test on Julia 1.0?

mateuszbaran:
Sure, we can stop supporting OrdinaryDiffEq on Julia 1.0.

sethaxen:
Okay, I switched to testing `solve_exp_ode` only on 1.1 and greater and documented that 1.1 is required. I also added explicit tolerances to more of the metric tests. The test suite passes locally for me both on 1.0 and 1.4, so If the test suite passes for CI, I'd like to merge this.

mateuszbaran:
I wonder, which version of Julia is used for code coverage?

sethaxen:
I had assumed all versions are used, but then on another package that didn't seem to be the case, so I don't know. Sometimes I have no idea how it's counting. Sometimes it seems that Julia inlines a function and then it is never counted as being called. Codecov is a mystery to me.

src/manifolds/MetricManifold.jl
```diff
@@ -544,9 +552,16 @@ arguments that will be passed to `OrdinaryDiffEq.solve`.
...
+!!! note
+    This function only works for Julia 1.1 or greater, when
+    [OrdinaryDiffEq.jl](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl) is loaded with
+    ```julia
+    using OrdinaryDiffEq
+    ```
 """
 function solve_exp_ode(M, x, v, tspan; kwargs...)
-    error("solve_exp_ode not implemented on $(typeof(M)) for point $(typeof(x)), vector $(typeof(v)), and timespan $(typeof(tspan)). For a suitable default, enter `using OrdinaryDiffEq`.")
+    error("solve_exp_ode not implemented on $(typeof(M)) for point $(typeof(x)), vector $(typeof(v)), and timespan $(typeof(tspan)). For a suitable default, enter `using OrdinaryDiffEq` on Julia 1.1 or greater.")
```
kellertuer: Could you add a small test for this error message? We do that for several errors with an `EmptyManifold` for manifold-vunctions as well.

PR #105, opened 2020-01-27
Add show methods

sethaxen:
This PR adds convenient `show` methods for most types, extending the existing `show` methods previously added for `AbstractGroupManifold` and related types.

The general convention it follows is that show methods should primarily be informative about the object and secondarily should if possible be a valid constructor for the object.

Examples:

```julia
julia> Euclidean(3, 2)
Euclidean(3, 2; field = ℝ)

julia> Circle()
Circle(ℝ)

julia> Rotations(3) × Euclidean(3)
ProductManifold(Rotations(3), Euclidean(3; field = ℝ))

julia> Rotations(3) × Sphere(3) × Euclidean(3) × Stiefel(3, 2) × Grassmann(3,2)
ProductManifold(
    Rotations(3),
    Sphere(3),
    Euclidean(3; field = ℝ),
    Stiefel(3, 2, ℝ),
    Grassmann(3, 2, ℝ),
)

julia> Sphere(2)^(3, 2) # this is a new convenience syntax
PowerManifold(Sphere(2), 3, 2)

julia> B = ProjectedOrthonormalBasis(:svd)
ProjectedOrthonormalBasis(svd, ℝ)

julia> get_basis(Sphere(2), [0.0, 0.0, 1.0], B)
PrecomputedOrthonormalBasis with scalars in ℝ
Basis vectors:
3-element Array{Float64,1}:
 0.0
 1.0
 0.0
3-element Array{Float64,1}:
 1.0
 0.0
 0.0

 julia> ProductRepr(randn(4, 4), randn(3))
 ProductRepr{Tuple{Array{Float64,2},Array{Float64,1}}}
 Submanifold components:
 4×4 Array{Float64,2}:
  -0.128581   -1.1519     1.00885    0.747719
  -0.361213    0.125522   0.285183  -0.297439
   0.589948    0.625854  -1.4666    -1.7739  
  -0.0389574  -0.129482  -1.71002    0.522341
 3-element Array{Float64,1}:
  -0.9919280768212939 
   2.3109456533471118 
   0.28037969957815995
```

To be sufficiently informative, `show` methods for basis vectors and product representations do not give valid constructors.

`GraphManifold` currently doesn't have a `show` method since I haven't yet looked into a good way to display both its graph and manifold.


kellertuer:
I wasn't aware of the constructor-idea but despite that – great! Was something I also had in mind to do. Thanks! Will take a closer look, soon.

mateuszbaran:
Thanks for working on this. The new output of `show` looks definitely better.

kellertuer:
Uh and I just saw the `Sphere(2)^(2,3)` – very neat! Will directly include that into my talk Friday on Trust Regions (which is not yet based on Manifolds, but advertising this is great).

src/orthonormal_bases.jl
```diff
@@ -343,3 +343,39 @@ function get_basis(M::Manifold, x, B::ProjectedOrthonormalBasis{:gram_schmidt,
...
+    println(
+        io,
+        "DiagonalizingOrthonormalBasis with scalars in $(number_system(onb)) and 0 curvature in direction:",
+    )
+    show(io, mime, onb.v)
+end
+function show(io::IO, mime::MIME"text/plain", onb::PrecomputedOrthonormalBasis)
+    println(io, "PrecomputedOrthonormalBasis with scalars in $(number_system(onb))")
+    print(io, "Basis vectors:")
+    for v in onb.vectors
```
mateuszbaran: I think we may want to limit the number of vectors that are shown here. Maybe something like 5?

mateuszbaran:
```julia
julia> B = ProjectedOrthonormalBasis(:svd)
ProjectedOrthonormalBasis(svd, ℝ)
```
I think this should be printed as  `ProjectedOrthonormalBasis(:svd, ℝ)` so that it is printed how it can be constructed.

kellertuer:
For the graph manifold one could first `show` the graph then print whether ist data on edges or vertices and then with values in `show(M)`?

mateuszbaran:
> Uh and I just saw the `Sphere(2)^(2,3)` – very neat! Will directly include that into my talk Friday on Trust Regions (which is not yet based on Manifolds, but advertising this is great).

Cool! `Sphere(2)^(2,3)` might not be the reason people start using `Manifolds.jl` but it's neat indeed :wink: .

src/manifolds/ProductManifold.jl
```diff
@@ -578,6 +578,24 @@ function sharp!(M::ProductManifold, v::TFVector, x, w::CoTFVector)
     return v
 end
 
+function show(io::IO, M::ProductManifold)
+    length(M.manifolds) == 0 && return println(io, "ProductManifold()")
```
kellertuer: Does this happen? Well it's not explicitly excluded, though it seems not that reasonable.

src/manifolds/ProductManifold.jl
```diff
@@ -578,6 +578,24 @@ function sharp!(M::ProductManifold, v::TFVector, x, w::CoTFVector)
...
+    width = displaysize(io)[2]
+    strings = []
+    one_line_length = 17 + (length(M.manifolds) - 1) * 2
+    for m in M.manifolds
+        s = repr(m)
+        push!(strings, s)
+        one_line_length += length(s)
+    end
+    if one_line_length ≤ width
+        print(io, join(strings, ", "), ")")
```
kellertuer: Displaywise an `⨉` might also be nice, but it would not yield constructor with the `PowerManifold` upfront.

test/stiefel.jl
```diff
@@ -4,6 +4,7 @@ include("utils.jl")
     @testset "Real" begin
         M = Stiefel(3,2)
         @testset "Basics" begin
+            @test repr(M) == "Stiefel(3, 2, ℝ)"
```
kellertuer: I think `Stiefel{ℝ}(3, 2)` might be better readable (actually for any field-parametrised manifold) and is still a valid constructor?

src/manifolds/ProductManifold.jl
sethaxen: Well if it can happen (it currently can), then the case should exist so the `show` doesn't error. Even simple calls like `ProductManifold() |> manifold_dimension` error, so I think we should just error of one tries to construct with no submanifolds, and then we don't need this.

kellertuer: Sure, as long as we allow that to happen, it should have a show – I am also in favour of disallowing zero-element product manifolds.

src/manifolds/ProductManifold.jl
sethaxen: Yeah, for a small number of manifolds, that would look really nice, but for a large number, not so much. We could do something like 

```julia
Rotations(3)
 × Sphere(3)
 × Euclidean(3; field = ℝ)
 × Stiefel(3, 2, ℝ)
 × Grassmann(3, 2, ℝ)
```

or do dots just showing the first 2 and last 2 submanifolds like

```julia
Rotations(3) × Sphere(3) × … × Stiefel(3, 2, ℝ) × Grassmann(3, 2, ℝ)
```

Either way we should probably do something special to avoid printing all submanifolds for large product manfiolds.

test/stiefel.jl
sethaxen: I prefer the `Stiefel(3, 2, ℝ)` case because 1) it's probably the constructor that will be used more often and 2) it's fairly common for math notation of the field of manifolds to include it inside the parentheses (consider e.g. `SO(n, ℝ)` and `GL(n, ℝ)`)

sethaxen:
> > Uh and I just saw the `Sphere(2)^(2,3)` – very neat! Will directly include that into my talk Friday on Trust Regions (which is not yet based on Manifolds, but advertising this is great).
> 
> Cool! `Sphere(2)^(2,3)` might not be the reason people start using `Manifolds.jl` but it's neat indeed 😉 .

There's also
```julia
julia> Circle()^3
Torus(3)

julia> Circle()^(2, 3)
PowerManifold(Circle(ℝ), 2, 3)
```

The idea is that the power notation creates a manifold with a suitable power structure, not necessarily a `PowerManifold` (though both of these are `AbstractPowerManifold`). (in this sense, it is like `×` for `ProductManifold`). We also already have

```julia
julia> ℝ^2
Euclidean(2; field = ℝ)
```

And perhaps should have

```julia
julia> Euclidean(2)^(3, 4)
Euclidean(2, 3, 4; field = ℝ)
```

kellertuer:
I saw torus, which is neat, but even more the `Euclidean` special is even nicer.

test/stiefel.jl
kellertuer: I haven't worked with that much, so I trust your expertise. I felt like `Stiefel{ℝ}` to even be a kind of type (some real-valued Stiefel), so that felt nice in view, too. That that's not the most commonly used constructor is clear. So with your expertise, this is resolved.

src/orthonormal_bases.jl
sethaxen: Yes, I think for `ProductManifold`, `ProductGroup`, the product representations, and the bases, it would probably be best to truncate at some point. I'm thinking showing the first 2 and last 2 with dots in the middle if the number is beyond something like 5.

src/manifolds/ProductManifold.jl
kellertuer: For really large ones, I am not so sure, yes. That would be something like:
If this manifold and the last manifold do not fit into this line, but ` ... x` followed by the last manifold into the line. That would yield nearly your second line, with the exception that the ... would be before the last manifold not before the second to last.

src/orthonormal_bases.jl
mateuszbaran: Yes, that's a good approach.

src/manifolds/ProductManifold.jl
```diff
@@ -19,6 +19,7 @@ struct ProductManifold{TM<:Tuple} <: Manifold
 end
 
 ProductManifold(manifolds::Manifold...) = ProductManifold{typeof(manifolds)}(manifolds)
```
kellertuer: Just to be safe, it might be good to check that the `manifolds`-Tuple as a (strictly) positive length, since that would also lead to an zero-product product manifold

src/orthonormal_bases.jl
```diff
@@ -343,3 +343,66 @@ function get_basis(M::Manifold, x, B::ProjectedOrthonormalBasis{:gram_schmidt,
...
+function show(io::IO, mime::MIME"text/plain", onb::DiagonalizingOrthonormalBasis)
+    println(
+        io,
+        "DiagonalizingOrthonormalBasis with coordinates in $(number_system(onb)) and 0 curvature in direction:",
+    )
+    sk = sprint(show, "text/plain", onb.v, context = io, sizehint = 0)
+    sk = replace(sk, '\n' => "\n ")
+    print(io, sk)
+end
+function show(io::IO, mime::MIME"text/plain", onb::PrecomputedOrthonormalBasis)
```
kellertuer: Comparing this one with the next (Diagonalizing) I noticed that the next one prints an additional `:`  as well as `"Basis vectors:"´. Is that intended that way? Further the then-following lines are quite similar, they could also be put in a `_show_basis_vector_range_shortened` (or such) function that takes `io, vectors, limit=4,pre="",sym="")` ? Then one can even easily split this limit in the middle to obtain the current split idea.

sethaxen:
I made one change to the convention. Looking over the standard library, it seem that `show` methods with no mimetype are called by `print` and `repr` with no mimetype, and these are usually kept as one-line valid constructors.

On the other hand, when `show` with the "text/plain" mimetype is defined, it usually has a more complicated representation that is informative but not a constructor (e.g. `AbstractArray` or `SVD`). This is what is called in interactive contexts. I adopted that convention, so we now in the REPL we have:

```julia
julia> M = ProductManifold(Rotations(3), Euclidean(2, 3))
ProductManifold with 2 submanifolds:
 Rotations(3)
 Euclidean(2, 3; field = ℝ)

julia> print(M)
ProductManifold(Rotations(3), Euclidean(2, 3; field = ℝ))

julia> withenv("LINES" => 10, "COLUMNS" => 100) do
           repr("text/plain", ProductManifold([Euclidean(i) for i = 1:10]...)) |> println
       end
ProductManifold with 10 submanifolds:
 Euclidean(1; field = ℝ)
 Euclidean(2; field = ℝ)
 Euclidean(3; field = ℝ)
 ⋮
 Euclidean(9; field = ℝ)
 Euclidean(10; field = ℝ)

julia> ProductGroup(SpecialOrthogonal(3) × SpecialEuclidean(3))
ProductGroup with 2 subgroups:
 SpecialOrthogonal(3)
 SpecialEuclidean(3)

julia> ProductRepr(randn(3, 3), randn(2), randn(2), randn(2), randn(2))
ProductRepr with 5 submanifold components:
 Component 1 =
  3×3 Array{Float64,2}:
    0.458013  -0.434409   1.45428
   -0.759097   0.579221  -1.01714
    0.752913   0.719709   1.58326
 Component 2 =
  2-element Array{Float64,1}:
    0.4848805445027728
   -0.9835078976865191
 ⋮
 Component 4 =
  2-element Array{Float64,1}:
    1.447162575734385 
   -1.6870631057127565
 Component 5 =
  2-element Array{Float64,1}:
    1.0288503168025935
   -1.2198960580840423

julia> get_basis(Euclidean(2, 3), randn(2, 3), ArbitraryOrthonormalBasis())
PrecomputedOrthonormalBasis with coordinates in ℝ and 6 basis vectors:
 E1 =
  2×3 Array{Float64,2}:
   1.0  0.0  0.0
   0.0  0.0  0.0
 E2 =
  2×3 Array{Float64,2}:
   0.0  0.0  0.0
   1.0  0.0  0.0
 ⋮
 E5 =
  2×3 Array{Float64,2}:
   0.0  0.0  1.0
   0.0  0.0  0.0
 E6 =
  2×3 Array{Float64,2}:
   0.0  0.0  0.0
   0.0  0.0  1.0

julia> get_basis(Euclidean(2, 3), randn(2, 3), DiagonalizingOrthonormalBasis(randn(2, 3)))
PrecomputedDiagonalizingOrthonormalBasis with coordinates in ℝ and 6 basis vectors
Basis vectors:
 E1 =
  2×3 Array{Float64,2}:
   1.0  0.0  0.0
   0.0  0.0  0.0
 E2 =
  2×3 Array{Float64,2}:
   0.0  0.0  0.0
   1.0  0.0  0.0
 ⋮
 E5 =
  2×3 Array{Float64,2}:
   0.0  0.0  1.0
   0.0  0.0  0.0
 E6 =
  2×3 Array{Float64,2}:
   0.0  0.0  0.0
   0.0  0.0  1.0
Eigenvalues:
 6-element Array{Float64,1}:
  0.0
  0.0
  0.0
  0.0
  0.0
  0.0
```

test/orthonormal_bases.jl
```diff
@@ -94,3 +94,92 @@ struct NonBasis <: Manifolds.AbstractBasis{ℝ} end
...
 end
+
+@testset "Basis show methods" begin
+    @test sprint(show, ArbitraryOrthonormalBasis()) == "ArbitraryOrthonormalBasis(ℝ)"
+    @test sprint(show, ArbitraryOrthonormalBasis(ℂ)) == "ArbitraryOrthonormalBasis(ℂ)"
+    @test sprint(show, ProjectedOrthonormalBasis(:svd)) == "ProjectedOrthonormalBasis(:svd, ℝ)"
+    @test sprint(show, ProjectedOrthonormalBasis(:gram_schmidt, ℂ)) == "ProjectedOrthonormalBasis(:gram_schmidt, ℂ)"
+
+    @test sprint(show, "text/plain", DiagonalizingOrthonormalBasis(Float64[1, 2, 3])) == """
+    DiagonalizingOrthonormalBasis with coordinates in ℝ and 0 curvature in direction:
```
kellertuer: Zero curvature in direction is a little imprecise, but maybe the long explanation is too much for a `show`:
The DiagonalizingONB is an ONB in one tangent space equipped with eigenvalues of the curvature tensor that characterise how much a tangent vector changes when being transported along the geodesic given by the first tangent vector. That way the first tangent vector gets a zero, since its the covariant derivtative of the geodesic considered anyway.

The imprecision is now, that that does not mean zero curvature, the manifold is still (happily) curved as before, just the curvature tensor when restricted to this orthonormal Frame (=set of ONBs along the geodesic by parallel transport) has an eigenvalue of zero for that direction. And yes this is far too long for a print, so maybe just FYI.

src/manifolds/ProductManifold.jl
sethaxen: How would such a tuple be created? If one passes no args to `ProductManifold`, the below constructor is called, which throws the error. Only if one passes at least one arg, creating a strictly positive length tuple, would this constructor be used.

src/orthonormal_bases.jl
sethaxen: > Comparing this one with the next (Diagonalizing) I noticed that the next one prints an additional `:` as well as `"Basis vectors:"´. Is that intended that way? `

Yes, the second needs the `Basis vectors:` to clearly differentiate them from eigenvalues. If you think it would be better, I can also add `Basis vectors:` to the first to unify them. Good idea on the shared utility function.

src/manifolds/ProductManifold.jl
kellertuer: Ah, I did not try that, but you're right, that might not be possible, since Tuples are immutable, for Arrays this is different, for sure.

src/orthonormal_bases.jl
kellertuer: A, I missed the `Eigenvalues`, no than that difference is fine.

test/orthonormal_bases.jl
sethaxen: Thanks, that's helpful. I chose this wording as a synopsis of what the docs say, that is, it's the basis "...that diagonalizes the curvature tensor $R(u,v)w$ and where the direction `v` has curvature `0`.`"

I'm happy to update both the docs and the `show` method, with the latter remaining not much longer than it currently is. What would you propose?

kellertuer: the shortest way would be to replace curvature by eigenvalue?

sethaxen: Okay, I made a fix; let me know if it resolves this.

sethaxen:
@kellertuer I just added a `show` method for `GraphManifold`. If it wasn't what you had in mind, let me know.

src/manifolds/GraphManifold.jl
```diff
@@ -208,3 +206,22 @@ d_{\mathcal N} = \lvert E \rVert d_{\mathcal M}.
...
+    sg = replace(sg, '\n' => "\n$(pre)")
+    println(io, pre, sg)
+    println(io, "Manifold$(man_desc):")
+    sm = sprint(show, "text/plain", M.manifold, context = io, sizehint = 0)
+    sm = replace(sm, '\n' => "\n$(pre)")
+    print(io, pre, sm)
+    return nothing
+end
+
+function show(io::IO, mime::MIME"text/plain", M::EdgeGraphManifold)
```
kellertuer: I think the second type has to be `GraphManifold{G,TM,VertexManifold} where {G,TM}`, since the `VertexManifold` is the third parameter of the `GraphManifold`– same for the next.

kellertuer: Ah sorry, I missed the `const`s. >this is fine.

sethaxen:
With the most recent commits, this PR is feature-complete and should have tests for all additions (last I checked some were marked as not tested but definitely were).

We could consider in the future adding LaTeX show methods using [LaTeXStrings.jl](https://github.com/stevengj/LaTeXStrings.jl), but these would mostly only be useful in Jupyter notebooks and html docs, and they don't seem that useful in the end.

kellertuer:
I think we can postpone the LaTeXStrings for later; maybe they'll also provide a text/utf8-renderer for docstrings one, that would really be nice. I am fine with this PR when the test coverage is good. Thanks for working on this.

kellertuer:
I just noticed `show` for `VectorBundle` being unchecked and the new nice `Circle()^(2)` yielding a `Torus`. In general `get_basis`, `basis` and `get_vector` could get a little more attention, but not necessarily in this PR, I can do a few over in the docs when I anyways check all those.

sethaxen:
> I just noticed `show` for `VectorBundle` being unchecked

I see that, yes, I'll add a new `TestVectorSpaceType` so that we can test that fallback.

> and the new nice `Circle()^(2)` yielding a `Torus`.

There's actually an explicit test for this in `/test/torus.jl`, so I'm not certain why codecov doesn't count this.

> In general `get_basis`, `basis` and `get_vector` could get a little more attention, but not necessarily in this PR, I can do a few over in the docs when I anyways check all those.

Yes, I agree, and that that should be in a separate PR

Once the `VectorBundle` test is added and passes, I'll merge this.

PR #110, opened 2020-02-06
Taking Riemannian differentiation a bit more seriously

mateuszbaran:
I've started working on Riemannian differentiation support (mostly derivative, gradient and Jacobian for now). It's very much a work in progress but I'd like to get some comments if my approach makes sense to you.


mateuszbaran:
@kellertuer I should probably add a function here that returns the value of a hessian on a particular tangent vector, right? That seems to be what is needed for second-order optimization methods. This would take care of converting to and from the correct ONB for this operation.

kellertuer:
Yes, that would be nice, you can look at the Riemannian Trust Region either in Manopt or Manopt.jl how it'S done there as a first idea.

mateuszbaran:
`project` is supposed to perform the ambient -> Riemannian gradient conversion but I'm not sure how to use it for ambient -> Riemannian derivative conversion, especially for non-isometric embeddings. I'm not sure if it's even worth the trouble though, AD wouldn't give us much here anyway and finite differencing is quite easy in this case.

kellertuer:
Since the derivative should be seen as a linear functional, i.e. a Covector, the precise way (I think) would be to do a flat/sharp around the project.

mateuszbaran:
A slight problem here is that for me it's useful to still work with tangent vectors when calculating derivatives of curves. A simple finite differencing using `log` would give us a tangent vector, which then would be needlessly converted to and from the cotangent representation. Would it make more sense then to call the function that computes that tangent vector `differential` or `pushforward`? (or something else?)

kellertuer:
I think differential would be the right term. Not that I might also be interested in the adjoint of the differential at some point … but we can also discuss that later – for now I have a few explicitly implemented such adjoint differentials.

src/riemannian_diff.jl
```diff
@@ -9,97 +9,70 @@ abstract type AbstractRiemannianDiffBackend end
...
+hessian_vector_product(f, ft::RealField, p, X, backend::AbstractRiemannianDiffBackend)
 
-r_differential(f::AbstractCurve, p) = r_differential(f, p, rdiff_backend())
+differential(f, c::Curve, p) = differential(f, c, p, rdiff_backend())
 
-r_differential!(f::AbstractCurve, X, p) = r_differential!(f, X, p, rdiff_backend())
+differential!(f, c::Curve, X, p) = differential!(f, c, X, p, rdiff_backend())
 
-r_gradient(f::AbstractRealField, p) = r_gradient(f, p, rdiff_backend())
+gradient(f, rf::RealField, p) = gradient(f, rf, p, rdiff_backend())
```
kellertuer, line 65: I would still prefer seeing the manifold here, since there might be gradients that require information about the manifold or for conversion a basis in TpM, then it would be neat to have `M` here. Same for the differential and the hessian and the hessian vector product. For those two I think you mean that the hessian is `H(p)` while the product is `H(p)[X]`?

mateuszbaran: `rf` contains a manifold, and it can be easily extended to also contain other information. Maybe I should introduce `AbstractRealField` here though. This way it's straightforward to introduce additional information but putting it in `rf` instead of adding additional arguments (which will mess with AD backend information or, if keyword arguments, can't be dispatched over).

> For those two I think you mean that the hessian is `H(p)` while the product is `H(p)[X]`?

Yes, and more specifically, Hessian in my code is the matrix of Hessian coefficients in the given basis and Hessian-vector product is the tangent vector that's the value of Hessian on the given tangent vector (not its coefficients in a basis).

I'll have to think how to incorporate other representations of the Hessian but most likely it would be specified by `AbstractRiemannianDiffBackend`. And I'm going to separate `AbstractRiemannianDiffBackend` into different backends for gradients, hessians and differentials.

kellertuer: Thanks for the clarification, you'Re right, the manifold is withing `rf`, I was not careful enough trying to understand that.

For the hessian, we could nearly combine those right?
A `hessian(f,p)`, where `f` is a `AbstractRealField` has both manifolds, so one can take their `DefaultBases` as a matrix representation but it should maybe also be possible to specify both basis, i.e. `hessian(f,p,b1,b2)` - where `b1,b2` are `AbstractBase`s. For the hessian applied to a vector we have `hessian(f,p,X)`. What do you think?

mateuszbaran: > For the hessian, we could nearly combine those right?

Yes, it (`f` and `ft`) was even combined before. I'm not quite sure which version I prefer.

> A `hessian(f,p)`, where `f` is a `AbstractRealField` has both manifolds

What manifolds? I can see only one for Hessians.

> For the hessian applied to a vector we have `hessian(f,p,X)`

I would prefer to have different functions for coefficients of Hessians in bases and values of Hessians on tangent vectors.

kellertuer: Ah right, for the hessian it's only one.

I am not a fan of the currect hessian-vector-name, that's why I am looking for a new name – and I think it could be combined into one name, why should it not?

mateuszbaran: Making both variants methods of one function will cause ambiguity issues and, from my previous experience, I prefer to avoid method ambiguity rather than solving it.

I agree that `hessian_vector_product` is not a great name. Maybe the discussion from https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/45 would be relevant here?

kellertuer: Ah the ambiguities yes, that might be a problem.

Yes that reference is relevant here but from your current approach it solves my question therein – we just have to find a nice name :)

mateuszbaran:
This PR is getting too ambitious, I'm going to split things I need the most to a separate PR to merge them first. Specifically, just gradients and differentials of curves. Jacobians and Hessians will have to wait.

mateuszbaran:
I will close this PR. The only thing potentially still worth taking from this is a few methods for `RiemannianONBDiffBackend`.

Issue #21, opened 2019-06-28
Ideas for a Logo


kellertuer:
Now that we have a new name, maybe we could also think about a Logo. One of my ideas (that I had for Manopt.jl actually but then they were missing the opt-part, so I am working on something different there), is, to use the “three Julia dots” and their colors to represent a manifold each, but I am not a designer so I did not get that far. The three manifolds I had in mind are

* top left a sphere (well that's obvious) with geodesics
* top right a hyperbolic space (maybe also in the embedding, so the H2 in R3 ?) but I haven't found a good way to illustrate that
* bottom center the Euclidean space, a square with straight vertical and horizontal lines (so R2), one could also do a cube to illustrate R3.

What do you think? Do you have a good way to illustrate that? I can try to do that in TikZ, though there the sphere might be tricky, since tikz is mainly good for 2D graphics, not for 3D (Especially not for occlusion things).

mateuszbaran:
Julia project logos are usually quite simple and flat. Maybe we could use some nice tilings in Julia colors https://en.wikipedia.org/wiki/Uniform_tilings_in_hyperbolic_plane ?

kellertuer:
While they look nice, my only objection is, that that's only one manifold, and our package has now the plural s included.

sdaxen:
I like the idea of something flat and simple that conveys either the idea that this is a collection of manifolds or some representative aspect of generic manifolds that we provide.

kellertuer:
I just had a short idea for a logo yesterday night that I just finished in dark and bright mode (since I am using GitHub in dark mode these days and the docs also have dark mode). May main idea is: Keep the Julia dots and present just main things we have. Currently tangent vectors and geodesics. We could add (in blue or white/black) small neat features to the other dots; maybe a slightly bending distribution?

![logo_dark](https://user-images.githubusercontent.com/115566/73061553-bde4de00-3e9a-11ea-89fd-9619e384b492.png)
![logo_bright](https://user-images.githubusercontent.com/115566/73061558-bfaea180-3e9a-11ea-8776-3ffc1fee9821.png)

For compilation you need Tamil MN as font and XeteX (or disable the last draw) in the code

````latex

\documentclass[tikz]{standalone}
\usepackage{}
\usepackage{ifthen}
\usepackage{fontspec}
\setmainfont{Tamil MN}
\usetikzlibrary{arrows.meta}
\newboolean{darkmode}
\setboolean{darkmode}{false}
[...]

PR #113, opened 2020-02-09
Proposes a Logo

kellertuer:
This PR finishes the discussion on a logo as discussed in the chat and in #21 .

The logo is generated using the script in `assets/logo.jl` by @sethaxen – that I just reorganised a little bit, put a few comments and added a `plain` mode for the favicon.

## Logo
The logo is available in two flavours: bright and dark

which are also once saved without their backgrounds – and these are used in the docs by their naming scheme

## Favicon
The favicon was generated from the plain png

which in full size still has the patch dots. This yields the favicon (even with transparency)

generated with https://realfavicongenerator.net

## SVG
With Inkscape sag exports at least on Mac OS render strangely, but since there is actually no pixel image part in the logo, we could also go for an sag in the future.

Thanks for all interims discussions, I think we ended up with quite a nice logo that even can be simplified to a nice icon.

_edit_: In the preview you can already see both logos in action in both light and dark mode of the docs: https://julianlsolvers.github.io/Manifolds.jl/previews/PR113/


kellertuer:
Oh and I also found the fix, such that the dark logo in the generated PDF is actually on black grounds by adding a line to the preamble.

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
 # Power manifold
 
+A power manifold is based on a [`Manifold`](@ref) $\mathcal M$ to build a
+$\mathcal M^{n_1 \times n_2 \times \cdots \times n_m}$.
```
mateuszbaran: I think we agreed to write one sentence per line in Markdown files?

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+For the two special cases $m=1$ and $m=2$ we obtain
+one-dimensional data, i.e. vectors of points on $\mathcal M$ and matrices of images,
+respectively.
```
mateuszbaran: This sentence seems a bit hard to read, maybe it can be rephrased to something like
> In the case where $m=1$ we can represent a manifold-valued vector of data of length $n_1$, for example a time series.
> The case where $m=2$ is useful for representing manifold-valued matrices of data of size $n_1 \times n_2$, for example certain types of images.

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+
+## Example
+There are two ways to store the data: in a multidimensional array or in a nested array.
+
+Let's look at both for an example. Let $\mathcal M$ be `Sphere(2)` the 2-sphere. and we want
```
mateuszbaran:
```suggestion
Let's look at both for an example. Let $\mathcal M$ be `Sphere(2)` the 2-sphere and we want
```

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+to look at vectors of length 4.
+
+For the default, the [`ArrayPowerRepresentation`](@ref), we store the data in da multidimensional array,
```
mateuszbaran:
```suggestion
For the default, the [`ArrayPowerRepresentation`](@ref), we store the data in a multidimensional array,
```

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+        [1/sqrt(2.0), 1/sqrt(2.0), 0.0],
+        [1/sqrt(2.0), 0.0, 1/sqrt(2.0)],
+        [0.0, 1.0, 0.0]
+    ,dims=2)
+```
+
+which is a valid point i.e.
+
+```@example 1
+is_manifold_point(M,p)
```
mateuszbaran:
```suggestion
is_manifold_point(M, p)
```

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+```
+
+An advantage of this representation is, that it is quite efficient.
+A disadvantage might be both looking at the matrix and seeing the points.
```
mateuszbaran:
```suggestion
A disadvantage is not being able to easily identify parts of the multidimensional array that correspond to a single point on the base manifold.
```

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+For the default, the [`ArrayPowerRepresentation`](@ref), we store the data in da multidimensional array,
+
+```@example 1
+using Manifolds #hide
+M = PowerManifold(Sphere(2),4)
+p = cat([1.0, 0.0, 0.0],
+        [1/sqrt(2.0), 1/sqrt(2.0), 0.0],
+        [1/sqrt(2.0), 0.0, 1/sqrt(2.0)],
+        [0.0, 1.0, 0.0]
+    ,dims=2)
```
mateuszbaran: I would add here an example based on HybridArrays.jl. It's in the testset for `PowerManifold`, for example `HybridArray{Tuple{3,StaticArrays.Dynamic()}, Float64, 2}` for the case we have in this example.

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+There are two ways to store the data: in a multidimensional array or in a nested array.
+
+Let's look at both for an example. Let $\mathcal M$ be `Sphere(2)` the 2-sphere. and we want
+to look at vectors of length 4.
+
+For the default, the [`ArrayPowerRepresentation`](@ref), we store the data in da multidimensional array,
+
+```@example 1
+using Manifolds #hide
+M = PowerManifold(Sphere(2),4)
```
mateuszbaran:
```suggestion
M = PowerManifold(Sphere(2), 4)
```

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+    ,dims=2)
+```
+
+which is a valid point i.e.
+
+```@example 1
+is_manifold_point(M,p)
+```
+
+An advantage of this representation is, that it is quite efficient.
```
mateuszbaran:
```suggestion
An advantage of this representation is that it is quite efficient, especially when a `HybridArray` (from the `HybridArrays.jl` package) is used to represent a point on the power manifold.
```

docs/src/manifolds/power.md
```diff
@@ -1,5 +1,56 @@
...
+M = PowerManifold(Sphere(2), NestedPowerRepresentation(), 4)
+p = [ [1.0, 0.0, 0.0],
+      [1/sqrt(2.0), 1/sqrt(2.0), 0.0],
+      [1/sqrt(2.0), 0.0, 1/sqrt(2.0)],
+      [0.0, 1.0, 0.0]
+    ]
+```
+
+which is again a valid point so [`is_manifold_point`](@ref)`(M,p)` here also yields true.
+A disadvantage might be that with nested arrays one looses a little bit of performance.
```
mateuszbaran:
```suggestion
A disadvantage might be that with nested arrays one loses a little bit of performance.
```
I've just checked and a simple `log` is 3x faster for the manifold we have in this example for `HybridArray` than the nested representation. The difference will be even greater when `HybridArrays` can skip more allocations.

test/utils.jl
```diff
@@ -284,36 +284,38 @@ function test_manifold(M::Manifold, pts::AbstractVector;
...
-                @test project_tangent(M, x, bvectors[i]) ≈ bvectors[i]
+                @test norm(M, x, bvectors[i]) ≈ 1
+                for j in i+1:N
+                    @test real(inner(M, x, bvectors[i], bvectors[j])) ≈ 0 atol = sqrt(find_eps(x))
+                end
+            end
+            if isa(btype, ProjectedOrthonormalBasis)
+                # check projection idempotency
+                for i in 1:N
+                    @test project_tangent(M, x, bvectors[i]) ≈ bvectors[i]
```
mateuszbaran:
```suggestion
                    @test isapprox(M, x, project_tangent(M, x, bvectors[i]), bvectors[i])
```

src/Manifolds.jl
```diff
@@ -407,7 +415,8 @@ export allocate,
     zero_vector!,
     zero_tangent_vector,
     zero_tangent_vector!,
-    ×
+    ×,
+    uniform_distribution
```
mateuszbaran: It's not in the alphabetical order.

docs/src/manifolds/power.md
kellertuer: Yes, that was my german again peeking through. Your two sentences sound better.

docs/src/manifolds/power.md
kellertuer: I rephrased also the first one when adding your suggestion.

docs/src/manifolds/power.md
kellertuer: I added that to the example for completeness.

mateuszbaran: OK, thanks!

kellertuer:
Ah linking directly to files might not be best – well the first post captures the current status: Trying an all-plain approach as suggested in the chat. I will try an more-emphasised/bolder details next and then write a post to compare all 3 (dark and bright logos and screenshots from the docs) to compare.

kellertuer:
Here's my final comparison. One problem with making things more prominent (bolder) is, that the dotted lines, when set too bold, are more like a heap of dots, so that's the limit I took: here's a comparison of 3 version, where for 1 and 3 I would propose to use 2 as favicon, still:
(1) The original proposed by @sethaxen in his script
(2) a plain version as proposed in the chat
(3) a bolder version of 1 (tangent vector line with to 4->8, mesh line to 1->3, distribution dots scaled by 2.5, geodesics width to 12->20)

logo_1
logo_2
logo_3

all three logos are rendered the same way and exported the same way (transparent bg) in Inkskape. Similarly the docs-test is the same screenshot in the same browser of our index page.

docs1
docs2
docs3


Please discuss and decide; I am not sure whether I like 2 or 3; 3 is nicely bold and detailed, 2 is nicely plain.

sethaxen:
I don't like the bold colors of option (2). In particular, the blue on top the other colors. Yes, they're the Julia colors, but in the Julia logo, they stand separate. I do like the "pastel" colors in (1) and (3) from the increased transparency. Perhaps you could show a version of the bold one with that transparency for comparison? Also in the bold version, I think you could make the lines in the patches a little thicker and darker.

kellertuer:
Can you clarify which transparency you are referring for in (3)? It should have the same pastel colors as in (1). Making (3) with the transparency as in 2 makes it too strong I fear, I mean you already donÄt like the overlap of the colors there (and you're right it's little too strong in color contrast).

For the dotted lines, they are currently thickness `3` in (3), making them `4` makes them (visually to my eye) more like a heap of dots then lines. But I can try setting patch mesh-line thickness to 4 and making the dots denser.

sethaxen:
> Can you clarify which transparency you are referring for in (3)? It should have the same pastel colors as in (1). Making (3) with the transparency as in 2 makes it too strong I fear, I mean you already donÄt like the overlap of the colors there (and you're right it's little too strong in color contrast).

No, I'm saying make a new version of (2): on with colors more like (3) and dotted lines also more bold. The dots would ideally be spaced further apart in that case to avoid collision.
The dots are necessary to give a sense of perspective, but now they're so faint in (2) that it draws the attention to figure out if something is there. If they were just a little more apparent, then perhaps they wouldn't do that.

kellertuer:
Ah, then I confused your writing in the with those and the references, sorry. I changed the color of the lines – that are now black (white) with transparency themselves (50%) this way we can make them again as bold as desired (up to black). Further, the plain one now also has transparent geodesics to comply with your request.
I feel that on the back where the lines get quite close one nearly looses the impression of lines, though I put the dots just a little closer. Here are again: bright logo and logo in use:

logo

docs4

Now we of course can further play with transparency of the geodesics and boldness of the dots.

kellertuer:
Maybe another idea: avoid colliding colors – though that would mean doing geodesics in black… I like black, but I can see that the logo then get's a darker tough:

docs5

kellertuer:
I think this last one is actually my favourite – no overlapping colors, plain, still bright Julian colors...so here's the same image for dark style

docs6d

mateuszbaran:
There are so many designs that I'm getting lost. I'll be happy with whatever you choose 🙂 .

sethaxen:
I like this last one too. The only thing I'm not a huge fan of right now is the darkness of the black on the version on the white background. Perhaps a slightly lower opacity or a slightly thinner line would resolve that.

kellertuer:
Yes I had a similar feeling, it was a bit too dark. I reduced geodesic thickness (in both styles) from 40 to 35, which already looked better and reduced opacity of geodesic (similarly before already the mesh lines have reduced opacity in dark mode).

So here is the result – and I am happy that we agree on a logo (and maybe this is even its final form!)
docs7b
docs7d

sethaxen:
Looks great! I'm happy with it! Thanks for pushing this forward.

kellertuer:
Well thanks for starting the 3D idea. I will merge this and then adapt the readme to display the logo (some smaller version that I'll prepare now).

PR #114, opened 2020-02-10
Embedded manifolds and a global approach to transparent decorators

kellertuer:
I would like some feedback on my new approach on decorators. I haven't finished all rework for that but just sketched the idea in the docs and for a few functions in `Decorators.jl`.

I think this might replace the THTT with `is_decorator_manifold` and in order to dispatch on the abstract decorator level already this is even necessary. But I think it's worth it.

I got to this idea, because I noticed that I would have to programming quite a few passthroughs again for `EmbeddedManifold`, that we already have for group and metric (array is more special since it overwrites nearly all functions). I think it can also stay with manifolds for now – let's see how it works out, whether we want to even put it to base.

I will continue (maybe slowly) to implement more functions when I extend the embedded idea further.

See the `decorator.md` for the idea of the approach.

edit_: And this is of course (despite this rework/interface for decorators) an approach to #107 .

PR #115, opened 2020-02-13
Simplify the basis interface

kellertuer:
This is a sketch for now and aims to simplify the basis type. What this aims to do is the following

We have several types that mark _what type_ of a basis we want
* `ArbitraryBasis`
* `ArbitraryOrthonormalBasis`
* `DiagonalizingOrthonormalBasis`
all of this are `<: ArbitraryBasis`

and for each we previously had a Precomputed basis type. This PR aims to provide just one `CachedBasis <: ArbitraryBasis`, that is internally typed by an arbitrary basis itself and has just one field for stored data. It usually stores just the vectors. For the last type it also stores the direction and the eigenvalues in a seperate type.

I am not yet done with all rework, but I hope this avoid the double-types (T and PrecomputedT) for all basis types. This is not an urgent PR, but I will work on this, when I have time and the current other PRs are finished.


mateuszbaran:
Thanks, this approach should work I think. We could also rename `get_basis` to something like `cache_basis` to make it more clear what it does.

kellertuer:
I have to rework Bundle and Product and such, that's why I waited for some feedback – thanks.

I think you get a basis anyways – so `get_basis` is ok – if you pass a cached one, it just passes that back, so that's fine. I would however implement `get_coordinates` and `get_vector` also once for cached basis, where they do more `inner`s.
This should account for the case that the ad-hoc-basis generation that is now the default might be time-consuming (more time consuming than the memory consumption of caching).

mateuszbaran:
> I would however implement `get_coordinates` and `get_vector` also once for cached basis, where they do more `inner`s.

As far as I can see `get_coordinates` is currently only implemented twice (one variant for real-bases and one for the other (it doesn't account for things like complex bases of quaternionic manifolds but it doesn't seem particularly useful).

> This should account for the case that the ad-hoc-basis generation that is now the default might be time-consuming (more time consuming than the memory consumption of caching).

I don't see where the ad-hoc basis generation is the default. I was trying to make it so full basis generation doesn't happen unless the user specifically asks for it.

kellertuer:
> I don't see where the ad-hoc basis generation is the default. I was trying to make it so full basis generation doesn't happen unless the user specifically asks for it.

Oh that's what I meant with "ad-hoc", to avoid full basis generation. If you have a basis already stored, the user might just use that one. I will take care that that's the case and documented, then.

kellertuer:
I finished a few tasks today, the next thing – and that might require some internal tricks – is representing cached bases on product manifolds – I will try to understand the old approach and mimick that within the new structure.

but for the base manifolds everything should already work.

src/manifolds/PowerManifold.jl
```diff
@@ -151,18 +151,12 @@ The array `bases` stores bases corresponding to particular parts of the manifold
...
+struct PowerBasisData{TB<:AbstractArray{T} where {T}}
     bases::TB
 end
 
-function PrecomputedPowerOrthonormalBasis(
-    bases::AbstractArray{<:AbstractPrecomputedOrthonormalBasis},
-    F::AbstractNumbers = ℝ,
-) where {N}
-    return PrecomputedPowerOrthonormalBasis{typeof(bases),F}(bases)
+function PrecomputedPowerOrthonormalBasis(bases::AbstractArray{T}) where {T}
```
mateuszbaran: This function isn't needed now.

src/manifolds/PowerManifold.jl
```diff
@@ -151,18 +151,12 @@ The array `bases` stores bases corresponding to particular parts of the manifold
 
 The type parameter `F` denotes the [`AbstractNumbers`](@ref) that will be used as scalars.
 """
-struct PrecomputedPowerOrthonormalBasis{
-    TB<:AbstractArray{<:AbstractPrecomputedOrthonormalBasis},
-    F,
-} <: AbstractPrecomputedOrthonormalBasis{F}
+struct PowerBasisData{TB<:AbstractArray{T} where {T}}
```
mateuszbaran:
```suggestion
struct PowerBasisData{TB<:AbstractArray}
```
Both versions are equivalent here but this one is more clear.

kellertuer:
I am quite stuck on Product and Power to unify my cached approach. It seems I am missing something and I mainly do not understand how the old storage worked. I am missing `length` of my data every other time. Can you help me out here @mateuszbaran ? You know those two by heart.

My idea is to basically store a tuple of cached bases and return the old style representation of `get_vectors` – but I do not see how the previous precomputed storage worked nor how that was accessed. The internal storage might be very efficient but it's too clever for me to understand.

mateuszbaran:
What do you mean by missing the length of your data?

ONBs for the power manifold, for example, work by storing a basis for for each particular component. For example if you have an SPD-valued image, we have a separate basis at each point stored. That's what `PrecomputedPowerOrthonormalBasis` stores. Then `get_vector` and `get_coordinates` iterate over all point in such image and use the stored basis to convert between tangent vectors and their coordinates. Product manifolds work similarly, just with tuples of bases instead of arrays of them.

> My idea is to basically store a tuple of cached bases

Yes, that's what old style cached ONBs for product manifold do.

> and return the old style representation of `get_vectors`

What do you mean by this?

kellertuer:
Yes, the idea is clear, just the realisation/implementation – I can't get that to work. I tried to do the same approach as the original `Precomputed...` did (with a `.parts`) field but I am only running from one error message to another without understanding where they come from and especially not how to solve them. I seem to miss something in my understanding.

For example calling `get_vectors` seems sometimes to work currently? But the result then never passes any test. So it seems to return something different currently and I am not able to investigate this really. Actually I don't really know where to start :/. Maybe in the long run it would be nice to have a documentation of the internal functionality of both power and product manifolds. I tried to go line by line through the code but somehow, somewhere I seem to not understand what your code is doing, so I can't adapt it (to what I think is the same, also having a `.parts` field in its struct).

mateuszbaran:
Unfortunately both power and product manifolds are quite complicated, yes. I was focusing on performance and versatility so the code is not that simple. Some things like `zip_tuples` is working around Julia's performance problems.

`get_vectors` is probably a poor place to start. I haven't worked out a specialized implementation for the power manifold for example (the generic one should work though).

I'm not really sure where the problem may be, the rework should just be about replacing the struct with cached basis with another? I'll take a closer look what's wrong here during the weekend. So push whatever you feel like and I'll see what's wrong with it.

kellertuer:
Thanks for taking a look. Yes, I am not saying, you should've documented it directly; of course first it should work and be performant :) I am just stuck and don't know what's going wrong, since I mainly tried to follow your approaches, even storing in `.parts`. And I am mainly following the route: Replace all `PrecomputedX` by the CachedBasis with an internal data representation. 

And well – `get_vectors` is the first thing that fails – for the decorator that was a good way to go through the functions – just test by test. But I'll try again tomorrow, maybe after all afternoon trying to figure out what the code is doing and mainly getting either errors or very complicated data structures that didn't make much sense … it's better to take a rest.

kellertuer:
I got a small step further, but somehow sometimes I don't understand that the `CachedBasis` that I assume to have is actually only its internal data. Somewhere this is still confused (mainly also in my head I think).

kellertuer:
Taking a break was a good idea.

* [x] Product Manifold works for all 3 basis types.
* [ ] show methods do not yet work and we have to find a way to nicely plot those
* [ ] Power Manifold is missing.

src/bases.jl
```diff
@@ -68,96 +70,114 @@ Available methods:
...
 
+function _euclidean_basis_vector(p, i)
+    X = zero(p)
+    X[i] = 1
+    return X
+end
+
+_get_vectors(B::CachedBasis) = B.data
+_get_vectors(B::CachedBasis{BT,D,𝔽}) where {BT<:AbstractBasis, D<:DiagonalizingBasisData, 𝔽} = B.data.vectors
+_get_vectors(B::CachedBasis{BT,D,𝔽}) where {BT<:AbstractBasis, D<:ProductBasisData, 𝔽} = B.data.parts
```
mateuszbaran: As far as I can see this function is called by `get_vectors` which is supposed to return a `Vector` of tangent vectors. `data.parts` is a very different thing.

src/bases.jl
```diff
@@ -205,7 +234,7 @@ end
 Compute the basis vectors of the tangent space at a point on manifold `M`
 represented by `p`.
 
-Returned object derives from [`AbstractBasis`](@ref) and may have a field `.vectors`
+Returned object derives from [`AbstractBasis`](@ref) and may have a 𝔽 `.vectors`
```
mateuszbaran:
```suggestion
Returned object derives from [`AbstractBasis`](@ref) and may have a field `.vectors`
```
This refers to a field in a struct.

src/manifolds/PowerManifold.jl
```diff
@@ -188,7 +172,7 @@ end
 function basis(M::AbstractPowerManifold, p, B::AbstractBasis)
```
mateuszbaran: IIRC this method is was replaced by `get_basis` and should be deleted.

src/manifolds/PowerManifold.jl
```diff
@@ -338,28 +322,45 @@ function get_coordinates(
     M::AbstractPowerManifold,
     p,
     X,
-    B::PrecomputedPowerOrthonormalBasis,
-)
+    B::CachedBasis{<:AbstractBasis{ℝ},<:PowerBasisData,ℝ}
+) where {𝔽}
```
mateuszbaran: I can't see `𝔽` in types of arguments.

src/manifolds/ProductManifold.jl
```diff
@@ -248,8 +226,8 @@ function flat!(M::ProductManifold, ξ::CoTFVector, p, X::TFVector)
 end
 
 function get_basis(M::ProductManifold, p, B::AbstractBasis)
-    parts = map(t -> get_basis(t..., B), ziptuples(M.manifolds, submanifold_components(p)))
-    return PrecomputedProductOrthonormalBasis(parts)
+    parts = map(t -> get_basis(t..., B).parts, ziptuples(M.manifolds, submanifold_components(p)))
```
mateuszbaran: Why did you add this `.parts` thing here? We should store here either the entire basis or its `.data` field perhaps.

src/manifolds/VectorBundle.jl
```diff
@@ -287,18 +283,6 @@ function get_coordinates(M::VectorBundle, p, X, B::ArbitraryOrthonormalBasis) wh
     coord2 = get_coordinates(M.fiber, px, VXF, B)
     return vcat(coord1, coord2)
 end
-function get_coordinates(
```
mateuszbaran: Why did you remove this function?

src/bases.jl
kellertuer: The naming is not great, it's more like an internal extract.

src/manifolds/PowerManifold.jl
kellertuer: Oh that's just a leftover from a lot of trying.

src/manifolds/ProductManifold.jl
kellertuer: The parts themselves are `CachedBases` I think? I was trying to understand what it does and this yields consistent results, namely `get_basis` delivers a `CachedBasis` with clever internal storage that the other three functions work with. I hope.

src/manifolds/VectorBundle.jl
kellertuer: All that were previously `PrecomputedXBasis` based functions are now handled by the single `CachedBasis` function, which should be the one l. 168 or 171 in `bases.jl`. If I understood all implementations correctly (and the tests are still running, but I still have to check for coverage).

src/bases.jl
mateuszbaran: Well, here: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/115/files#diff-9dbb0d28119f714ea432396713884380R348 it's used to return vectors outside.

src/manifolds/VectorBundle.jl
mateuszbaran: These two methods in `bases.jl` are just generic implementations. We should still have specialized ones for product, power and vector bundle (they are much faster).

src/manifolds/ProductManifold.jl
mateuszbaran: > The parts themselves are `CachedBases` I think?

Yes, and `CachedBasis` has only one field called `data`. There is no `.parts` field. This should work without the `.parts` thing.

src/bases.jl
kellertuer: Thanks for noticing, then I have to check how that came into the code and what it actually should be.

src/manifolds/ProductManifold.jl
kellertuer: Interesting, must have also slipped in my slight confusion (see above).

src/bases.jl
kellertuer: The idea for the internal function is, that that can be used in show. But now I split them more carefully and it only returns vectors when its really array of vectors. I now just have to get the tests back up running. Problem stays: Someone has to redo the shows for cached data. I am not yet able to understand those tricky functions.

kellertuer:
While I had hoped I only renamed `hat` and `vee` in Rotations, the last commit sadly also broke all of `ForwardDiff` (including my shell giving up) – but after 2 hours just trying to fix bugs – I would maybe recommend not to rename hat and vee after all?

src/bases.jl
mateuszbaran: I'll take a look at the show methods for cached data in a moment.

mateuszbaran:
I've fixed these infinite recursion bugs in Rotations. There is just one test failing where the norm of a basis vector is not 1.

Do you plan to completely remove `vee` and `hat`? I think we could leave them.

kellertuer:
Yes, where the norm error comes from, I am not yet sure.

Thanks for noticing, I must have missed that recursion and it was hard to see.

Concerning naming, maybe @sethaxen might help. I feel, that `get_vector` and `get_coordinates` are more informal but maybe better to understand. We should definetly leave `hat` and `vee` as synonyms as they are now, maybe?

mateuszbaran:
If we leave `vee` and `hat` then we also should keep calling them in `Rotations.jl` instead of `get_vector` and `get_coordinates` I think.

BTW, I'll push `show` for cached product bases in a moment.

kellertuer:
Great, thanks for helping out. I think this is nearly done, we should

- [ ] check for the remaining norm (it's off by 10^-8 and I don't see why),
- [ ] decide which one will be the primary names (`hat`/`vee` or `get_vector`, `get_coordinates`
- [ ] last but not least update the docs accordingly.

kellertuer:
I found the bug with Rotations, it happened for `Float32` and adding a tolerance to said failing norm based on the `eps` did it. Now `Product` is again failing and I am not 100% sure why; maybe it's not yet onsistent in the sense that `CachedData`has an internal array of `CachedData`s? I think that's what some paces assume to have. But I again can't follow the 400+ character long `ProductRepr`s in the error messages.

mateuszbaran:
I've fixed most issues with product manifold bases. The remaining one is lack of `get_vector`/`get_coordinates` for reshaped array representation but I'd just disable this test for now. I'll have to revisit this representation in the future anyway (some memory layout changes are planned for Julia and it can impact performance here), for now using `ProductRepr` should be good enough.

kellertuer:
So its hopefully just documentation left to be revised. I'd be busy next week, but I'll try on my travels to read docs.

mateuszbaran:
Yes, documentation and making a decision about `vee`/`hat` vs `get_vector`/`get_coordinates`. I'd like to get @sethaxen's opinion about the second thing before the PR is merged.

sethaxen:
> Yes, documentation and making a decision about `vee`/`hat` vs `get_vector`/`get_coordinates`. I'd like to get @sethaxen's opinion about the second thing before the PR is merged.

I haven't had a chance to review this PR yet, but I can share my thoughts on `vee` and `hat`. First, I've only seen the `vee`/`hat` terminology used to describe the Lie algebra coordinates of Sym(n), SO(n), and SE(n). The specific basis vectors used for these are not necessarily consistent; eg for SO(n), the two different inner product definitions differ by a factor of 1/2. We've chosen the less common but more consistent one, where the basis associated with `vee`/`hat` is not orthonormal.

Consequently, I think the most sensible general meaning of `vee`/`hat` is some maps that vectorize a tangent vector and its inverse, so I'd prefer they be aliases of `get_coordinates`/ `get_vector` with something like `ArbitraryBasis`, that defaults to the same functions with `ArbitraryOrthonormalBasis`.

kellertuer:
Despite the renaming of `ArbitraryBasis` to `DefaultBasis` (and similarly for `ArbitraryOrthogonalBasis` and `ArbitraryOrthonormalBasis`) they are currently exactely these aliases. And with your reasoning that should then also stay that way, for people who expect these functions to work (or even feel that they want to reimplement only those).

src/bases.jl
```diff
@@ -342,26 +337,68 @@ function get_basis(M::Manifold, p, B::ProjectedOrthonormalBasis{:gram_schmidt,
...
+
+For array manifolds, this converts a vector representation of the tangent
+vector to an array representation. The [`vee`](@ref) map is the `hat` map's
+inverse.
+"""
+hat(M::Manifold, x, v) = get_vector(M, x, v, DefaultBasis())
+
+function hat!(M::Manifold, X, p, Xⁱ)
+    is_decorator_manifold(M) === Val(true) && return hat!(base_manifold(M), X, p, Xⁱ)
+    error("hat! operator not defined for manifold $(typeof(M)), array $(typeof(X)), point $(typeof(p)), and vector $(typeof(Xⁱ))")
```
sethaxen: I don't see a reason to not have this default:
```suggestion
    return copyto!(X, get_vector(M, p, Xⁱ, DefaultBasis()))
```

src/bases.jl
```diff
@@ -342,26 +337,68 @@ function get_basis(M::Manifold, p, B::ProjectedOrthonormalBasis{:gram_schmidt,
...
+
+````math
+∧ : X^i ↦ X^i e_i
+````
+
+For array manifolds, this converts a vector representation of the tangent
+vector to an array representation. The [`vee`](@ref) map is the `hat` map's
+inverse.
+"""
+hat(M::Manifold, x, v) = get_vector(M, x, v, DefaultBasis())
```
sethaxen:
```suggestion
hat(M::Manifold, p, Xⁱ) = get_vector(M, p, Xⁱ, DefaultBasis())
```

src/bases.jl
```diff
@@ -389,36 +426,75 @@ function show(io::IO, mime::MIME"text/plain", onb::DiagonalizingOrthonormalBasis
...
+
+````math
+\vee : X^i e_i ↦ X^i
+````
+
+For array manifolds, this converts an array representation of the tangent
+vector to a vector representation. The [`hat`](@ref) map is the `vee` map's
+inverse.
+"""
+vee(M::Manifold, x, v) = get_coordinates(M, x, v, DefaultBasis())
```
sethaxen:
```suggestion
vee(M::Manifold, p, X) = get_coordinates(M, p, X, DefaultBasis())
```

src/bases.jl
```diff
@@ -389,36 +426,75 @@ function show(io::IO, mime::MIME"text/plain", onb::DiagonalizingOrthonormalBasis
...
+
+function vee!(M::Manifold, Xⁱ, p, X)
+    is_decorator_manifold(M) === Val(true) && return vee!(base_manifold(M), Xⁱ, p, X)
+    error("vee! operator not defined for manifold $(typeof(M)), vector $(typeof(Xⁱ)), point $(typeof(p)), and array $(typeof(X))")
```
sethaxen:
```suggestion
    return copyto!(Xⁱ, get_coordinates(M, p, X, DefaultBasis()))
```

src/manifolds/Circle.jl
```diff
@@ -125,36 +125,36 @@ flat!(::Circle, ξ::CoTFVector, p, X::TFVector) = copyto!(ξ, X)
...
 
-get_coordinates(M::Circle{ℝ}, p, X, B::ArbitraryOrthonormalBasis) = X
+get_coordinates(M::Circle{ℝ}, p, X, B::DefaultOrthonormalBasis) = X
 function get_coordinates(M::Circle{ℝ}, p, X, B::DiagonalizingOrthonormalBasis)
     sbv = sign(B.frame_direction[])
     return X .* (sbv == 0 ? 1 : sbv)
 end
 """
-    get_coordinates(M::Circle{ℂ}, x, v, B::ArbitraryOrthonormalBasis)
+    get_coordinates(M::Circle{ℂ}, x, v, B::DefaultOrthonormalBasis)
```
sethaxen:
```suggestion
    get_coordinates(M::Circle{ℂ}, p, X, B::DefaultOrthonormalBasis)
```

src/bases.jl
```diff
@@ -169,12 +185,29 @@ Depending on the basis, `p` may not directly represent a point on the manifold.
...
-function get_vector(M::Manifold, p, X, B::AbstractPrecomputedOrthonormalBasis)
+function get_vector(M::ArrayManifold, p, X, B::AbstractBasis; kwargs...)
+    is_manifold_point(M, p, true; kwargs...)
+    size(X) == (manifold_dimension(M),) || error("Incorrect size of vector X")
+    return get_vector(M.manifold, p, X, B)
+end
+function get_vector(M::DefaultManifold, p, X, B::DefaultOrthonormalBasis)
+    return reshape(X, representation_size(M))
+end
+function get_vector(M::Manifold, x, v, B::DefaultBasis)
```
sethaxen: These methods need an update to our current notation.

src/bases.jl
```diff
@@ -68,96 +70,110 @@ Available methods:
...
+For the [`CachedBasis`](@ref) keep in mind that the reconstruction with [`get_vector`](@ref)
+requires either a dual basis or the cached basis to be selfdual, for example orthonormal
 
 See also: [`get_vector`](@ref), [`get_basis`](@ref)
 """
 function get_coordinates(M::Manifold, p, X, B::AbstractBasis)
     error("get_coordinates not implemented for manifold of type $(typeof(M)) a point of type $(typeof(p)), tangent vector of type $(typeof(X)) and basis of type $(typeof(B)).")
 end
-function get_coordinates(M::Manifold, p, X, B::AbstractPrecomputedOrthonormalBasis{ℝ})
+function get_coordinates(M::Manifold, x, v, B::DefaultBasis)
```
sethaxen: Outdated notation

src/manifolds/Rotations.jl
```diff
@@ -174,7 +174,7 @@ exp(::Rotations, ::Any...)
 exp!(M::Rotations, q, p, X) = copyto!(q, p * exp(X))
 function exp!(M::Rotations{2}, q, p, X)
     @assert size(q) == (2, 2)
-    θ = vee(M, p, X)[1]
+    θ = get_coordinates(M, p, X, DefaultBasis())[1]
```
sethaxen:
```suggestion
    θ = get_coordinates(M, p, X, DefaultOrthogonalBasis())[1]
```

src/manifolds/Rotations.jl
```diff
@@ -240,37 +240,40 @@ end
...
-end
-
-function get_vector(M::Rotations, p, X, B::ArbitraryOrthonormalBasis) where {N}
-    T = Base.promote_eltype(p, X)
-    return hat(M, p, X) ./ sqrt(T(2))
-end
-
 @doc raw"""
-    hat(M::Rotations, p, Xⁱ)
+    get_vector(M::Rotations, p, Xⁱ)
```
sethaxen: I'd prefer for `Rotations` that we use the more specific default basis `DefaultOrthogonalBasis`. That the basis is orthogonal is important for its use, and of course the behavior will be the same.
```suggestion
    get_vector(M::Rotations, p, Xⁱ, B:: DefaultOrthogonalBasis)
```

sethaxen: Here and below

src/manifolds/Rotations.jl
```diff
@@ -240,37 +240,40 @@ end
...
+    X = zeros(N,N)
+    return get_vector!(M, X, p, Xⁱ, B)
+end
+function get_vector(M::Rotations, p, X, B::DefaultOrthonormalBasis) where {N}
+    T = Base.promote_eltype(p, X)
+    return get_vector(M, p, X, DefaultBasis()) ./ sqrt(T(2))
+end
 
-function hat!(M::Rotations{2}, X, p, Xⁱ::Real)
+get_vector!(M::Rotations{2}, X, p, Xⁱ, B::DefaultBasis) = get_vector!(M, X, p, Xⁱ[1], B)
```
sethaxen: I don't know what the solution is, but it feels a bit odd seeing the mutating form of `get_vector!` used here when it isn't used in `bases.jl`.
Why for `get_vector` and `get_coordinates` do we not follow our usual pattern of the mutating function being wrapped in the non-mutating?

kellertuer: Yes, we could do that, indeed.

src/bases.jl
kellertuer: this would then default to `get_vector!` as soon as it's implemented?

src/bases.jl
```diff
@@ -169,12 +185,29 @@ Depending on the basis, `p` may not directly represent a point on the manifold.
 For example if a basis transported along a curve is used, `p` may be the coordinate
 along the curve.
 
+For the [`CachedBasis`](@ref) keep in mind that the reconstruction from [`get_coordinates`](@ref)
+requires either a dual basis or the cached basis to be selfdual, for example orthonormal
```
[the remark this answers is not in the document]
kellertuer: Yes that sounds reasonable, we should do that in the future.

src/manifolds/Rotations.jl
kellertuer: I will do the ones below later, since I am currently at the airport and have to first finish something on another branch then.

src/manifolds/Rotations.jl
kellertuer: I just did mutating versions of `get_coordinates` and `get_vector` and introduced our usual pattern. For the basis I have to think a little bit more about the corresponding allocate.

src/bases.jl
kellertuer: I actually implemented is as a fallback to the mutating one.

src/bases.jl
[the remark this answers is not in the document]
mateuszbaran: Hmm... I didn't think about reconstruction in non-orthonormal bases actually. Good point.

src/manifolds/Rotations.jl
mateuszbaran: Initially I wasn't particularly concerned about allocations here so I just didn't write the mutating variants. But just for consistency it is worth changing to the usual pattern.

kellertuer: Oh the tests still fail, I did not run all tests while travelling; might only come back to that on Thursday then.

kellertuer:
This branch currently only work with Base 0.4.1. We should maybe merge this first and then only the other branch to update to 0.5?

mateuszbaran:
OK, this seems to be almost complete so it is a good idea to merge this PR first.

mateuszbaran:
I've added some tests and exports for mutating `get_vector!` and `get_coordinates!`, though more will be needed (it should be a part of the generic testset).

kellertuer:
At least for Euclidean, Product and Power the switch to mutating has to be done and with `Euclidean` I am running into one of the errors that I don't understand and just randomly am able to resolve: No tho matching `MethodError: no method matching one(::Type{Any})` which is triggered somewhere in `allocate_result_type`. I had hoped that one is just not allowed to pass down a basis to `allocate` but that did not help.

mateuszbaran:
I've looked at that `InexactError` thing, there are a few ways it can be fixed:

1. Always requiring complex-valued array for representing points on a complex manifold.
2. Specialize `allocate_result` for  a few problematic cases like `get_vector`.
3. Always allocating complex-valued arrays when `allocate_result` is used on a complex manifold.

Option 1 looks less desirable than the other two and 3 looks slightly more appealing than 2 (but then we'll unnecessarily complexify some arrays). What do you think?

kellertuer:
Thanks for taking a look; I also think 1 would be too strong to impose, 2 would be closest to the current form (the `get_vector` always produced complex values), so I would go with that one and only if we notice this a few more times, then we should for consistency switch to 3.

mateuszbaran:
OK, I'll fix it using option 2 (most likely tomorrow).

mateuszbaran:
I've made a quick-ish fix for this problem. It needs more tests but the approach should work.

mateuszbaran:
I'm fixing the product manifold bug, it should be ready soon.

kellertuer:
Cool. I already removed the WIP since we are feature complete and it's only code coverage and bug fixing left.

kellertuer:
For basis I leave the code coverage for later; some `support`(`!`)s still don't have tests.

mateuszbaran:
Vector bundle a product manifold are fixed now, I'm trying to fix some group related bugs.

mateuszbaran:
Hmm... I'm not quite sure what should be done in these failing cases, and most likely a different solution will be needed once #114 is merged. Should we just disable the failing tests for now and fix them in #114?

BTW, this PR still needs appropriate `vee`, `vee!`, `hat`, `hat!`, `get_coordinates!` and `get_vector!` tests added to `test_manifold`.

kellertuer:
If we change that anyways in the other PR I think it's fine to finish these and adapt them over there.

Ah, yes, a similar block should be added, could you do that? We should also discuss how to maybe reduce test time again, we are hitting the time limit sometimes again.

mateuszbaran:
OK, I'll work on that tomorrow.

We'll have to take a look at reducing test times, right. Eventually we may have to split the package for this reason.

mateuszbaran:
I have added some tests for `vee` and `hat`, although I'm not finished yet and there seems to be much more group-related failures than I've initially anticipated.

Given that #114 is also getting close to being finished it may be a good idea to merge 114 first and then fix the failures here.

kellertuer:
Sorry that it's more complicated than I thought, I had hoped that `get_vectors` and `get_coordinates` would just be transparent wrt groups.

mateuszbaran:
The problem is, they are transparent if `Identity` is not involved and not transparent otherwise.

kellertuer:
Yes, because without group, `Identity` doesn't really exist, so with `Identity` you have to match that to something explicitly, that's where the choice of transparency comes from; but you're right that's not optimal. The best would be that in these cases some function, maybe `Identity(G)` returns an array representation if the identity.

mateuszbaran:
There is already the `identity` function which does that, although we still need #114 to nicely capture the cases where it's needed.

kellertuer:
Then its maybe better to really get the features of #114 here and check for the remaining cases that they all transparently work.

mateuszbaran:
Yes, that's what I think.

kellertuer:
The merge is nearly done, just that I noticed one thing: Currently the sphere fails, since the old fallback from `vee`/`hat` to `get_coordinates`/`get_vector` don't work; the defaults in `Manifold` are already defined in base (as non implemented) and `hat`/`vee` are still always transparent (since they were introduced in Base). We have to solve both; currently Sphere calls vee of Euclidean.

We should first decide, whether `get_bases`, `get_coordinates`, `get_vector` should also go over to base – I would say yes – then we can introduce the same default that's currently in Manifolds mani file.

mateuszbaran:
Yes, I'd just move the basis-related functions to Base. Most likely everything from `bases.jl` and `numbers.jl`. It's quite lightweight and the new interface looks really nice.

kellertuer:
After that fixing sphere is just setting the transparency of vee/hat correctly for non `Transparent` embedded manifolds

kellertuer:
I am not sure, whether we can get basis/vector/coordinates transparent because we run into ambiguous cases (manifold versus basis type). For now they are also only transparent for ArrayManifold (already implemented) and the default metric (not yet adapted).

Do you have an idea how to get them transparent?

I already isolated bases such that it can easily be transferred to ManifoldsBase, numbers can be transferred, too, once this runs fine.

mateuszbaran:
Hmm... would making each variant with different basis types transparent separately work? Somewhat similarly to `injectivity_radius`.

kellertuer:
That would be a fix, yes, even with explicitly making those that fall back to others not transparent. Will try that tomorrow.

kellertuer:
Hm, currently I don't really see how we can get basis/coordinates/vector transparent, I am running (for 1.5 hours) from one ambiguous case to the next without seeing an end. Any ideas welcome.

kellertuer:
Array turned out to be not so nice, product and power work, but I am now (again) fighting Metric & Group.

src/bases.jl
```diff
@@ -540,3 +489,138 @@ inverse.
...
+            dot_val = real(inner(M, p, bvectors[i], bvectors[j]))
+            if !isapprox(dot_val, 0; atol = eps(eltype(p)))
+                throw(ArgumentError("vectors number $i and $j are not orthonormal (inner product = $dot_val)"))
+            end
+        end
+    end
+    return B
+end
+# the following is not nice, can we do better when using decorators and a specific last part?
+get__coordinates(M::ArrayManifold, p, X, B::AbstractBasis; kwargs...) = _get__coordinates(M,p,X,B, kwargs...)
```
mateuszbaran: Why does this have two underscores between `get` and `coordinates`?

kellertuer: The error only occurred for `get_vector`, I wanted to adapt it for `coordinates`, too, and then copy/paste/replace did this. Thanks for noticing.

kellertuer:
Hm, I can get group to sometimes work, when array does not and vice versa. Allocation/similar with `identity` is something that really bothers me, I run into this problem far too often recently.

src/bases.jl
```diff
@@ -0,0 +1,595 @@
...
+    println(io, "\nand $(nv) basis vector$(nv == 1 ? "" : "s").")
+    print(io, "Basis vectors:")
+    _show_basis_vector_range_noheader(io, vectors; max_vectors = 4, pre = "  ", sym = " E")
+    println(io, "\nEigenvalues:")
+    sk = sprint(show, "text/plain", B.data.eigenvalues, context = io, sizehint = 0)
+    sk = replace(sk, '\n' => "\n ")
+    print(io, ' ', sk)
+end
+
+vee(M::Manifold, p, X) = get_coordinates(M, p, X, DefaultBasis())
```
mateuszbaran: Wouldn't it cause problems that `vee` has a different default in `ManifoldsBase.jl`?

mateuszbaran:
Thanks for working on it :+1: . I think we either need to completely decouple `vee`/`hat` from `get_vector`/`get_coordinates` or make a new PR to ManifoldsBase.jl that couples them there.

src/bases.jl
kellertuer: Currently I am getting an overwrite warning, but this will replace the one in Base once tests here are running; then We can put bases and numbers over to base.

mateuszbaran: I hope there are no failures caused by that warning, I have always resolved overwrite warnings before fixing bugs (I'm not sure though if it's strictly necessary).

kellertuer: I am testing with the overwritten thing, I am quite sure that'll work. Most things even do work already.

kellertuer:
I nearly decoupled that, vee/hat are just frontends that map to the others with OrthogonalBasis(). It's just identity that I worked on for an hour now without getting it fixed. That's really a little messy with Identity. Was one of my most time-intensive part last time, too. And I'll have to stop for today.

mateuszbaran:
I think `Identity` turns out to be more trouble than its worth.

kellertuer:
I see that it causes a lot of trouble, so what was it introduced for? Where does it really have a useful reason?

mateuszbaran:
I'm not sure. It was originally a part of #25 by @sethaxen and I kept it while working on #68 . It wasn't that bad back then. I guess it's there as an optimization opportunity? I'm not particularly attached to `Identity`.

kellertuer:
I think it might be beneficial for the Lie algebra and it might even be that once the remaining bugs (I think its 2, one for vee and the analogous for hat) it might be nice to have. It's just really not easy to go through the different dispatches and look where we currently are too transparent of do not dispatch on the mutating one early enough. Somewhere there lies the problem.

kellertuer:
So I nearly made it. The bug in `SpecialEuclidean` as well as special orthogonal was the following

Originally `vee` is just passed down transparently, and just `allocate` takes care that the right dimension is allocated for a coordinate in the identity. Since I wasn't sure, I disallowed that. Turns out, that `vee!` in Line 681 in `Rotations` in the end handles the corresponding cases; note that `p` can be `Any` and is not used. The last point is why it works. Not my favourite solution, when Identity just comes from the decorator basically.

I also commented out the definitions of my `hat`/`vee` in ManifoldsBase and that just removed the warnings, so I will do that once this is working fine.

For the remaining problem: With the new naming `vee` is not as transparent as before, goes over to `get_coordinate` that goes over to `get_coordinate!` and that is not implemented in `Rotations`. So here we have to decide what we should do with this case.

*  make vee transparent (carefully) to (for example) git line 679 in `Rotations`.
* implement that line as `get_coordinates` with the `DefaultOrthogonalBasis`
* find any other nice solution?

Currently 3 tests fail, but all with the same error.

mateuszbaran:
Which commit did you think about where you refer to these specific lines? It doesn't look like it's the last commit. `get_coordinates!` is nicely implemented there in `Rotations.jl` for the default orthogonal basis.

kellertuer:
Thanks for noticing, then I was imprecise in my observations, all these inheritance, transparency, default values and dispatches get you quite dizzy after an hour of jumpfing from file to file. If I remember the error message correctly (I could've copied it), `get_coordinates!` was still on `GroupManifold` somewhere, and it came all the way from `vee` via `get_coordinates`, so somewhere we have to transparently pass through group instead of dispatching further I think? It's really a large knot of chaos with vee and get_coordinates both being there nearly doing the same (same for hat and get_vector), just that the short ones are working on OrthogonalBasis only.

The lines refer to the last commit but the implementations of vee/vee! that were reached before the decorator was there (I think).

mateuszbaran:
"Large knot of chaos" doesn't really make me confident we have a well-designed code there. Maybe making `vee` just an alias for `get_coordinates` with `VeeOrthogonalBasis` would make things easier? Then we'd just have `get_coordinates` to deal with. `VeeOrthogonalBasis` would just by default be `DefaultOrthogonalBasis`.

kellertuer:
I think that would make it easier. Like we carefully did with transparency and mutating/nonmutating, it would be good to have just vee->vee!, vee->get_coordinates(...,::DefaultOrthogonalBassis) and in the manifolds just implement coordinates then (we could still document specific things for vee? I'm not sure). Currently it's not completely clear how far transparency “acts” until one transitions from vee!->get_coordinates! (or similarly for the non-mutating). That's hard to track.

mateuszbaran:
I'd just jump from vee to get_coordinates immediately.
```julia
vee(M::Manifold, p, X) = get_coordinates(M, p, X, VeeOrthogonalBasis())
vee!(M::Manifold, Y, p, X) = get_coordinates!(M, Y, p, X, VeeOrthogonalBasis())
hat(M::Manifold, p, X) = get_vector(M, p, X, VeeOrthogonalBasis())
hat!(M::Manifold, Y, p, X) = get_vector!(M, Y, p, X, VeeOrthogonalBasis())
```
And these would be the only methods of these four functions. No need to think about propagating through `vee` and `hat`.

mateuszbaran:
`VeeOrthogonalBasis` would use `DefaultOrthogonalBasis` by default but it would still be overridable.

kellertuer:
Yes, that sounds good.

kellertuer:
I hoped (after I wanted to spend an hour on the remaining stuff and it turned out to be 5 until now) that this would fix it, but we are back to 
* `groups/special_euclidean.jl` has 23 (again) tests failing that is a missing fallback, that I had introduced but lost in renaming vee?! I don't know.
* `groups/metric.j` has the same error twice.

The main problem is that the following lines (similarly for the mutating and coordinates)

````Julia
function get_vector(M::GM, e::Identity{GroupManifold{NT,GT}}, X, B::VeeOrthogonalBasis) where {GM<:AbstractGroupManifold,NT<:Manifold,GT}
    error("get_vector not applicable for group manifold $(M), and identity $(e), $(X), since the identity does not match.")
end
````

are necessary (I think) to avoid that an identity that does not fit, can be used (tested by `group_general.jl`, 26-31).
Still the error messages in above files read
````
  get_coordinates not applicable for group manifold SpecialOrthogonal(2), and identity Identity(SpecialOrthogonal(2)), [0.0 -1.0; 1.0 0.0], since the identity does not match.
````
Despite(!) the fact that for the case that `GM` and the identities group manifold are the same there is the case
````Julia
function get_coordinates(M::GT, e::Identity{GT}, X, B::VeeOrthogonalBasis) where {GT <: AbstractGroupManifold}
    return get_coordinates(decorated_manifold(M), e, X, B)
end
````
which should have been met as far as I read the error message above, where it did not take this one.

So I am not seeing the details anymore, but I am also tired. If you want to try, to do, I need a break from this. Identity might be nice somewhere, it's quite uncomfortable in these dispatches.

**Comment by mateuszbaran** [comment]
I've started fixing things by storing the representation of identity inside the `Identity` objects which makes making generic fallbacks easy. It causes, though, some weird issues for the special euclidean group, I'm not sure how could it have ever worked.

**Comment by kellertuer** [comment]
Originally it was just passed down to `Rotations`, where the `vee` is independently of `p` implemented and hence also works for non-arrays; just the allocation for `vee!` is adapted accordingly.

**Comment by mateuszbaran** [comment]
Well, the first thing I noticed is that `identity!(G::SemidirectProductGroup, q, p)` gets a matrix in `q` and `Identity` in `p` in one of the tests, and then calls `submanifold_components` on that matrix, which obviously fails.

**Comment by mateuszbaran** [comment]
OK, I've figured it out.

**Comment by mateuszbaran** [comment]
It looks like now it's just a matter of fixing propagation of `get_vector` etc. which should be much easier since `Identity` can just be unwrapped by default. SpecialEuclidean reports only 3 errors now.

**Comment by mateuszbaran** [comment]
I've fixed a few things, storing the point in `Identity` seems like a better option generally. I'm just not sure how to adapt some group array manifold tests (for example the one I've marked as broken).

File: test/groups/array_manifold.jl [hunk 22]
```diff
@@ -11,15 +13,15 @@
     p, q = [exp(M, eg, hat(M, eg, ωi)) for ωi in ω]
     X = hat(M, eg, [-1.0, 2.0, 0.5])
 
-    e = Identity(AG, p)
-    @test Manifolds.array_value(e) === e
-    @test Manifolds.array_point(e) === e
+    e = make_identity(AG, p)
+    @test Manifolds.array_value(e).p == Manifolds.array_value(e.p)
+    @test_broken Manifolds.array_point(e).p == Manifolds.array_point(e.p)
```
kellertuer: This might just be tricky because `array_point(e)` has to “unwrap” both `e.p` and `e` itself and create a new identity from that? Does it do that?
Is within `e` the `p` also stored as `ArrayPoint`? Otherwise the right would not make sense.

mateuszbaran: `array_value` unwraps both `e.p` and `e` itself right now. `array_point` at the moment doesn't wrap `Identity`.

kellertuer: ah, I see, then we should maybe add that double, wrap?

kellertuer:
We should maybe also add back `Identity(G)` to generate the default (`Float64`-based) identity? Something like `Identity(G::SpecialOrthogonalGroup{N}) here {N} = Identity(G, Matrix{Float64}(I,N,N)` (sketched, not testes)?

test/groups/array_manifold.jl
kellertuer: Could just (roughly) be `array_point(e::Identity) = Identity(e.group, array_point(e.p))` ? Or even wrap also `e`? maybe wrapping the inner would be enough?

kellertuer:
Status: We are back to `get_coordinates` and `get_vector` fails due to identity (the now nicer one) that somehow don't find their functions (and one gets the “not implemented” fallback errors) left are

- [ ] 23 errors in `groups/special_euclidean.jl`
- [ ] 2 errors in `groups/metric.jl`

and I think all are related to `Rotations` where the functions perfectly act transparent (from SE/SO to Rotations). If one comments out these Manifold/Identity fallbacks, either allocate or similar are not able to handle identity.

mateuszbaran:
> We should maybe also add back `Identity(G)` to generate the default (`Float64`-based) identity? Something like `Identity(G::SpecialOrthogonalGroup{N}) here {N} = Identity(G, Matrix{Float64}(I,N,N)` (sketched, not testes)?

Maybe. I'm not sure if it's actually necessary but it looks doable (although I'm not sure if `SpecialEuclidean` actually uniquely determines the representation of identity).

> Status: We are back to `get_coordinates` and `get_vector` fails due to identity (the now nicer one) that somehow don't find their functions (and one gets the “not implemented” fallback errors) left are
> 
> * [ ]  23 errors in `groups/special_euclidean.jl`
> * [ ]  2 errors in `groups/metric.jl`
> 
> and I think all are related to `Rotations` where the functions perfectly act transparent (from SE/SO to Rotations). If one comments out these Manifold/Identity fallbacks, either allocate or similar are not able to handle identity.

I'll take a look at it later.

mateuszbaran:
I've spent some time fixing problems and `SpecialOrthogonal` looks like it has incorrect transparency indicators, as it fails even if there is no `Identity` involved. I'll leave it for now.

mateuszbaran:
All these forwarding and dispatch for get_coordinates and get_vector look way too convoluted now,  more than for any other function, and I don't see why some of this complexity is needed.

kellertuer:
I think it's the first function with some actual dispatch involved? Maybe it got a little out of hand and can be simplified, I just don't really see, how. And in all that convoluted code I think it can have happened that some transparency is surely wrong (and it maybe should be transparent then?)

kellertuer:
So I noticed that we do not stick to the scheme as we do otheriwise and made a new approach with the non-mutating being less transparent. This reduces the `special_euclidean` down to 3 problems with product, the 2 in metrics are the same and `ProductManifold` still does not like something. Still, I feel this approach is cleaner. I haven't checked your new macro but added the constant to get Circle running.

mateuszbaran:
One of the remaining problems is caused by `get_vector!` not being transparent for `SpecialEuclidean` since `submanifold_components` has a special method for it.

mateuszbaran:
This is a bit too much I think:  https://github.com/JuliaNLSolvers/Manifolds.jl/blob/master/src/groups/special_euclidean.jl#L43-L63 . Special casing of `submanifold_components` with respect to manifold type effectively makes all functions that use it intransparent.

kellertuer:
I don't completely understand what you mean nor what you propose. Is that too complicated in implementation? Should we remove that and internally “just“ use the two `view` calls?

For the `VectorBundle` there is mainly the problem that the allocate is not yet provided for `get_coordinates`/`get_vector`.

mateuszbaran:
> I don't completely understand what you mean nor what you propose. Is that too complicated in implementation? Should we remove that and internally “just“ use the two `view` calls?

I don't really know how to fix the problem with SpecialEuclidean. We probably either need to get rid of the special case I've linked to or maybe switch dispatch of `get_vector!` to `:parent` on SpecialEuclidean? I didn't try all options yet.

> For the `VectorBundle` there is mainly the problem that the allocate is not yet provided for `get_coordinates`/`get_vector`.

I'm looking into it right now.

mateuszbaran:
I've just checked that we have 254 method ambiguities that are not triggered probably just because our tests are not exhaustive enough. Some are very interesting, for example:
```
inverse_translate_diff!(G::AbstractDecoratorManifold, ::Val{:intransparent}, Y, p, q, X)
in Manifolds at /home/mateusz/.julia/dev/ManifoldsBase/src/DecoratorManifold.jl:250, 
inverse_translate_diff!(M::ArrayManifold, Y, p, q, X, conv::ActionDirection)
in Manifolds at /home/mateusz/repos/Manifolds.jl/src/groups/array_manifold.jl:140
```
Most of them are caused by similar `::Val` vs `::ActionDirection` confusion. The call that would trigger it makes no sense though. There are also some related to bases that look actually worth fixing. You can generate the full list using `Test.detect_ambiguities(Manifolds)`.

kellertuer:
I didn't know that one. They mostly seem to be related to optional arguments in transparent functions. How can we best solve that? Putting the `Val` last could solve it maybe? If we could also find one that passes down and then only dispatches (for Basis for example); that would be awesome, though I have no clue how to do that.

mateuszbaran:
IIRC we had `Val` last at some point and it messed up with default arguments. Probably making it last without a default value would work best, although ambiguities related to `Val` are not the biggest problem right now.

mateuszbaran:
It turns out that the problem with SpecialEuclidean was at the beginning of the call chain, not at the end. All SpecialEuclidean tests passed for me locally.

kellertuer:
The only remaining one seems to be in the very last test, that for non-default metrics on should not be allowed to call the `DefaultOrthogonalBasis` or the orthonormal one. Making `get_basis` transparent could be a fix?

mateuszbaran:
I've fixed that problem with `get_basis` on a metric manifold.

src/Manifolds.jl
```diff
@@ -110,6 +111,7 @@ using ManifoldsBase: QRRetraction, PolarRetraction, ProjectionRetraction
 using ManifoldsBase: AbstractInverseRetractionMethod, LogarithmicInverseRetraction
 using ManifoldsBase: QRInverseRetraction, PolarInverseRetraction, ProjectionInverseRetraction
 using ManifoldsBase: AbstractVectorTransportMethod, ParallelTransport, ProjectionTransport
+using ManifoldsBase: ℝ, ℂ, ℍ
```
kellertuer: Mabe you can add `real_dimension` as an import here, too?

src/manifolds/Euclidean.jl
```diff
@@ -25,6 +25,8 @@ function Euclidean(n::Vararg{Int,I}; field::AbstractNumbers = ℝ) where {I}
     return Euclidean{Tuple{n...},field}()
 end
 
+^(𝔽::AbstractNumbers, n) = Euclidean(n...; field = 𝔽)
```
kellertuer: This is new, right? Neat idea!

src/Manifolds.jl
mateuszbaran: ManifoldsBase.jl exports `real_dimension` anyway and as far as I can see it's not extended with new methods in Manifolds.jl.

src/manifolds/Euclidean.jl
mateuszbaran: We had this before, just in `numbers.jl` for some reason.

kellertuer: Ah I looked for it bun then I must have missed it, looks really nice, especially handy if you do a demo (and know how to type that utf8 letter fast, I'll have to check though).

src/Manifolds.jl
kellertuer: Sorry then I was imprecise, I meant to add it to this `using` (not import) because I think we use it in manifolds.

mateuszbaran: We already have `using ManifoldsBase` so the symbol `real_dimension` is visible anyway.

mateuszbaran:
I think instead of adding a `Val{:parent}` parameter to functions we may prefer to use different function names (for example `exp`, `exp_transparent`, `exp_parent` and `exp_intransparent`). I'll experiment with that.

kellertuer:
Wouldn't that be close to the cases where we started implementing the `::Val{:value}` cases and we decided that's not nice? For the user it would be best, if he/she only has to care about `exp` with the normal interface.

mateuszbaran:
I think the `::Val` variants are internal only anyway (so the user will only see them in stacktraces), and modifying the name is currently my best idea for dealing with ambiguities.

src/manifolds/Euclidean.jl
mateuszbaran: In REPL, VSCode and Juno, you can insert them by typing `\bbR`, `\bbC` and `\bbH`.

kellertuer:
I think it's less flexible (i.e. harder to extend to another case), but since that will also only happen in a macro, it should be fine. And I am sure it will avoid ambiguitites (I should check for those more often).

mateuszbaran:
It would be somewhat hard to extend to another case without modifying ManifoldsBase.jl itself, yes. In my opinion, though, it's a small price to pay.

I'll add checks for the number of ambiguities to the test suite so that it'll be impossible to forget about checking them :slightly_smiling_face: .

src/product_representations.jl
```diff
@@ -310,7 +310,7 @@ function allocate(
     return ProductArray(ShapeSpec, allocate(x.data, T), x.reshapers)
 end
 allocate(p::ProductArray, ::Type{T}, s::Size{S}) where {S,T} = Vector{T}(undef, S)
-allocate(p::ProductArray, ::Type{T}, s::Integer) where {T} = Vector{T}(undef, s)
+allocate(p::ProductArray, ::Type{T}, s::Int64) where {T} = Vector{T}(undef, s)
```
kellertuer: Is it necessary to be that specific? What was the problem with `Integer`?

mateuszbaran: There was an ambiguity because `allocate` has `Int64` in ManifoldsBase in one place, I'll fix this later. Right now I'm transitioning decorators in ManifoldsBase to the new parent/transparent/intransparent dispatch based on different functions.

mateuszbaran:
`injectivity_radius` is another major place with ambiguities, primarily because the (manifold, point) version collides with the (manifolds, retraction method) version. We can either put disambiguations like I just did for the sphere everywhere or split this into two functions (one at point and another one global for the entire manifold). Which option do you prefer?

kellertuer:
Can we find a ways to kind of make the second argument optional? Like (sketch)

````julia
injectivity_radius(M::Sphere,p,m::AbstractRetractionMethod) = injectivity_radius(M,m)
````

or would that leave ambiguitites? For sphere that would reduce the last two lines to one, well not much, but I think it's nicer?
I would prefer the way you did for sphere; it's not that many lines (and with above idea if it works a few less) and shows more precisely all possibilities in one small block of code.

**Comment by mateuszbaran** [comment]
Your method would unfortunately be ambiguous with this one: https://github.com/JuliaNLSolvers/ManifoldsBase.jl/blob/54b6c86272bac4eb177238dd3076c7149741016b/src/ManifoldsBase.jl#L366 .

I'll keep it one function then.

**Comment by kellertuer** [comment]
That's sad, because yes also that case is reasonable. Then there is no way around doing all cases.

**Comment by mateuszbaran** [comment]
There is yet another ambiguity about `injectivity_radius`: this one https://github.com/JuliaNLSolvers/ManifoldsBase.jl/blob/54b6c86272bac4eb177238dd3076c7149741016b/src/ManifoldsBase.jl#L363 conflicts with things like `injectivity_radius(::Sphere, ::Any)`: the second argument type is not specific enough. So we need another method `injectivity_radius(::Sphere, ::AbstractRetractionMethod)` that calls `injectivity_radius(::Manifold, ::AbstractRetractionMethod)`. There would be no such problem if it were two functions.

**Comment by mateuszbaran** [comment]
If I haven't overlooked anything, all remaining ambiguities are related to either `injectivity_radius` or `Identity` (mostly its interactions with bases). I'm not sure how much do we want to get rid of the `Identity`-related ones... it doesn't look nice.

**Comment by kellertuer** [comment]
We can also do an issue for that asking for clarification what we want to use identity for before continuing to patch ambiguitites, instead of extending this PR into too many directions.

**Comment by mateuszbaran** [comment]
Yes, I also think we can solve the issues with `Identity` in another PR. I'd like to solve `injectivity_radius` though, it doesn't look complicated.

With `injectivity_radius` solved I think we can release ManifoldsBase 0.6 and polish this PR (it looks essentially complete to me).

**Comment by kellertuer** [comment]
The merge was quite easy. As soon as we hve done `Base` (where only test coverage is missing I think) I can work on test coverage here (for example dispatches).

**Comment by mateuszbaran** [comment]
Nice! Changes in base are almost complete, there is only coverage to be improved and the issue with `injectivity_radius`. For the second thing, would you prefer adding a method `injectivity_radius(::M, ::AbstractRetractionMethod)` for each manifold `M` or splitting it into two functions?

**Comment by kellertuer** [comment]
I don't see a good name for a splitted version; so I would maybe prefer to add these functions to all manifolds.

**Comment by kellertuer** [comment]
Can we somehow trigger the two warnings in the ProjectedBasis?  I am not yet sure when these two will happen anyways.

**Comment by mateuszbaran** [comment]
The "not linearly independent" warning is already triggered sometimes, just not for the `DefaultManifold` (it would be for any manifold of dimension lower than the ambient space).

The "input vector has length 0" as an extremely low change of ever happening but arguably we may want to change the condition from `nrmΞₙ == 0` to something like `abs(nrmΞₙ) < eps` for some very small epsilon (maybe on the order of `eps(T)^2` for numeric type `T`).

The "only found K vectors" error can easily happen when projection does not result in the correct number of linearly independent vectors (for example it projects all vectors to the zero vector).

**Comment by kellertuer** [comment]
Yes I meant the norm-zero and the less vectors warnings that are not covered. In theory, of course that can happen, but I am not that sure how to create a test case.

**Comment by mateuszbaran** [comment]
They can be triggered by a special manifold with `norm` returning 0 for some vectors that aren't exactly length 0 (for example sphere with a modification where `norm` returns exactly 0 when it's calculated to be less then some epsilon).

**Comment by kellertuer** [comment]
Will try that tomorrow. Should they stay warnings or even get errors? Neither with norm zero nor with too less vectors we have a basis.

**Comment by mateuszbaran** [comment]
The "length 0" warning doesn't necessarily mean that we won't finally get a basis, it's just that the projection turned out to be exactly 0 instead of approximately 0 (in the second case it would be discarded at the "not linearly independent" warning). The "only K vectors found" warning actually means we don't have a basis, so it should be an error (perhaps with a keyword argument that allows for returning incomplete sets of vectors),

**Comment by mateuszbaran** [comment]
The CI failures are probably caused by this: https://github.com/JuliaMath/DoubleFloats.jl/pull/101 .

**Comment by mateuszbaran** [comment]
Closing and reopening triggers a new CI build.

**Comment by kellertuer** [comment]
Here, we should once check that all new get_vector/get_coordinates are covered and look for optimising testing? Despite that this is also ready for review/merge.

**Comment by mateuszbaran** [comment]
Yes, the main thing this is lacking is more tests of `get_coordinates` and `get_vector`.

I don't really have any easy ideas for optimizing testing. Testing time is most likely dominated by compilation times (and we have a lot of methods to compile). One idea is to split the package (for example statistics-related things and group-related things could be separate), another one is setting up a virtual machine on one of our computers and run the full test suite only there while Travis and AppVeyor perform less extensive tests. Splitting would likely be better short-term.

**Comment by kellertuer** [comment]
I would prefer splitting and yes, while our 3-level idea is too complicated maybe, splitting groups would be a good idea. So what do we do about this branch then if we can't finish tests that properly?

**Comment by mateuszbaran** [comment]
For this PR, we can re-run tests once or twice hoping that the next time the time limit won't be exceeded. We can also disable some test for CI and run them only locally.

By the way, splitting groups away may be a little tricky because together with them we need to have there corresponding manifolds (Euclidean, Rotations, Product, Circle, maybe Power).

**Comment by kellertuer** [comment]
I have an idea for splitting groups but let's finish a few things first ;)

**Comment by kellertuer** [comment]
Interestingly the merge has messed up the DiffEqBase problem again, but I think I have taken over the Manifest and Project (despite ManifoldsBase now being on 0.6.0) from the Grassmann approach. @mateuszbaran can you check that again? I don't see a difference, but the tests are failing agin.

**Comment by mateuszbaran** [comment]
I think it's not the merge, it's the change in the registry: https://github.com/JuliaRegistries/General/pull/11221/files

**Comment by kellertuer** [comment]
I saw that the merged branch had Comap 2.2, with that, the test (just started) loaded DiffEqBase 6.22.2 again (and noch 6.23), which hopefully makes it work again.

**Comment by kellertuer** [comment]
I hopefully did the easier ones and opened an issue for #125 – so now just tests for the new basis scheme with

- [ ] `GroupManifold`
- [ ] `VectorBundle`
- [ ] `PowerManifold`
- [ ] `ProductManifold`

For these I especially have to check how far certain bases can be used there.

**Comment by mateuszbaran** [comment]
Hmm.. there are still some random CI failures here and there, for example in Circle, and some related to group manifold which do look particularly easy.

**Comment by kellertuer** [comment]
Failures or missing lines? The Mac OS tests currently stall and stop working.
The one line in circle is actually called but optimised away for example.

**Comment by mateuszbaran** [comment]
Sorry, I meant uncovered lines. When I opened the codecov report there were a few more uncovered lines in some methods of get_vector or get_coordinates. Maybe it was an old report. Anyway I'm trying to improve coverage a bit and will push a commit if things go well.

**Comment by kellertuer** [comment]
I just finished group manifolds (especially identity and) `get_vector`/`get_coordinates`. I think this should be fine (up to Mac OS tests now not even starting, but there's already a thread on the travis discourse on that).

File: src/manifolds/Sphere.jl [hunk 26]
```diff
@@ -26,6 +26,8 @@ struct Sphere{N} <: AbstractEmbeddedManifold{DefaultIsometricEmbeddingType} end
 
 Sphere(n::Int) = Sphere{n}()
 
+base_manifold(M::Sphere) = M
```
mateuszbaran: Why do we have this?

kellertuer: That ends the (otherwise infinite) loop of locking for base manifolds... Oh, maybe here this line is not necessary because it has no `Val{N}`.

mateuszbaran: This doesn't seem to be needed.

kellertuer: You might be right. Let's remove it.

mateuszbaran: There is already one definition for all `AbstractEmbeddedManifold`s so that's why it should work.

test/groups/array_manifold.jl
mateuszbaran: I've changed `array_point(e::Identity)` to your definition. Sorry for the delay, I forgot about it.

kellertuer:
So let's wait for tests and then we can merge.

PR #120, opened 2020-03-09
Introduce the Generalized Stiefel manifold

kellertuer:
This PR is to introduce the generalized (complex) Stiefel manifold. It further introduces a `project_point!` for Stiefel. Most of this PR was implemented by @manuelweisser – and now just testing is missing. I already adapted and merged the current master. 

- [x] Implementation
- [x] Documentation
- [x] Testing


src/manifolds/GeneralizedStiefel.jl
```diff
@@ -0,0 +1,223 @@
...
+Compute the inner product for two tangent vectors `X`, `Y` from the tangent space of `p`
+on the [`GeneralizedStiefel`](@ref) manifold `M`. The formula reads
+
+````math
+(X, Y)_p = \operatorname{trace}(v^{\mathrm{H}}Bw),
+````
+i.e. the metric induced by the scalar product `B` from the embedding, restricted to the
+tangent space.
+"""
+inner(M::GeneralizedStiefel, p, X, Y) = dot(X, M.B * Y)
```
mateuszbaran: I think it doesn't look like an isometric embedding when we have a custom `inner`?

```julia
struct GeneralizedStiefel{n,k,F,TB<:AbstractMatrix} <: AbstractEmbeddedManifold{AbstractIsometricEmbeddingType}
```

kellertuer: Good that you noticed, that was too  fast copied from Stiefel where this is indeed the case (and this PR also covers transferring Stiefel to exactely this).

.travis.yml
```diff
@@ -16,6 +16,7 @@ branches:
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/ # build tagged releases
 jobs:
   allow_failures:
+    - julia: 1.4
```
mateuszbaran:
```suggestion
```

appveyor.yml
```diff
@@ -12,6 +12,7 @@ platform:
 # (tests will run but not make your overall status red)
 matrix:
   allow_failures:
+  - julia_version: 1.4
```
mateuszbaran:
```suggestion
```

docs/make.jl
```diff
@@ -15,6 +15,7 @@ makedocs(
                 "Circle" => "manifolds/circle.md",
                 "Euclidean" => "manifolds/euclidean.md",
                 "Fixed-rank matrices" => "manifolds/fixedrankmatrices.md",
+                "Generalized Stiefel" => "manifolds/grassmann.md",
```
mateuszbaran:
```suggestion
                "Generalized Stiefel" => "manifolds/generalizedstiefel.md",
```

src/manifolds/EmbeddedManifold.jl
```diff
@@ -127,6 +127,16 @@ function default_decorator_dispatch(M::EmbeddedManifold)
     return default_embedding_dispatch(M)
 end
 
+@doc doc"""
+    default_embedding_dispatch(M)
```
mateuszbaran:
```suggestion
    default_embedding_dispatch(M::AbstractEmbeddedManifold)
```

src/manifolds/EmbeddedManifold.jl
```diff
@@ -127,6 +127,16 @@ function default_decorator_dispatch(M::EmbeddedManifold)
...
+
+A function to indicate that an [`AbstractEmbeddedManifold`](ref) is the default
```
mateuszbaran:
```suggestion
This method indicates that an [`AbstractEmbeddedManifold`](ref) is the default
```

src/manifolds/EmbeddedManifold.jl
```diff
@@ -127,6 +127,16 @@ function default_decorator_dispatch(M::EmbeddedManifold)
...
+and hence acts completely transparent and passes all functions transparently onwards.
```
mateuszbaran:
```suggestion
and hence acts completely transparently and passes all functions transparently onwards.
```

src/manifolds/GeneralizedStiefel.jl
```diff
@@ -0,0 +1,242 @@
+@doc doc"""
+    GeneralizedStiefel{n,k,T} <: AbstractEmbeddedManifold{AbstractIsometricEmbeddingType}
```
mateuszbaran:
```suggestion
    GeneralizedStiefel{n,k,T} <: AbstractEmbeddedManifold{AbstractEmbeddingType}
```
We could show that it's an isometric embedding in a metric manifold (Euclidean with metric given by the metric B)?

By the way, I feel like we should always give a concrete type here (that's what I meant earlier by `DefaultEmbeddingType` and `DefaultIsometricEmbeddingType` and why I was confused as to what it meant before the name was changed).

src/manifolds/GeneralizedStiefel.jl
```diff
@@ -0,0 +1,242 @@
...
+````
+"""
+manifold_dimension(::GeneralizedStiefel{n,k,ℝ}) where {n,k} = n * k - div(k * (k + 1), 2)
+manifold_dimension(::GeneralizedStiefel{n,k,ℂ}) where {n,k} = 2 * n * k - k * k
+manifold_dimension(::GeneralizedStiefel{n,k,ℍ}) where {n,k} = 4 * n * k - k * (2k - 1)
+
+@doc doc"""
+    project_point(M::GeneralizedStiefel,p)
+
+Projects `p` from the embedding onto the [`GeneralizedStiefel`](@ref) `M`, i.e. compute `q`
```
mateuszbaran:
```suggestion
Project `p` from the embedding onto the [`GeneralizedStiefel`](@ref) `M`, i.e. compute `q`
```

src/manifolds/GeneralizedStiefel.jl
```diff
@@ -0,0 +1,242 @@
...
+
+Project `X` onto the tangent space of `p` to the [`GeneralizedStiefel`](@ref) manifold `M`.
+The formula reads
+
+````math
+\operatorname{proj}_{\operatorname{St}(n,k)}(p,X) = X - p\operatorname{Sym}(p^{\mathrm{H}}BX),
+````
+
+where $\operatorname{Sym}(y)$ is the symmetrization of $y$, e.g. by
+$\operatorname{Sym}(y) = \frac{y^{\mathrm{H}}+y}{2}$.
```
mateuszbaran: The implementation does something different though (IIRC it just takes the upper triangular part of the matrix).

test/generalized_stiefel.jl
```diff
@@ -0,0 +1,89 @@
...
+        M = Stiefel(3,2,ℂ)
+        @testset "Basics" begin
+            @test repr(M) == "Stiefel(3, 2, ℂ)"
+            @test representation_size(M) == (3,2)
+            @test manifold_dimension(M) == 8
+            @test !is_manifold_point(M, [1., 0., 0., 0.])
+            @test !is_tangent_vector(M, [1.0 0.0; 0.0 1.0; 0.0 0.0], [0., 0., 1., 0.])
+            x = [1.0 0.0; 0.0 1.0; 0.0 0.0]
+            @test_throws DomainError is_manifold_point(M, [:a :b; :c :d; :e :f],true)
+            @test_throws DomainError is_tangent_vector(M, x, [:a :b; :c :d; :e :f], true)
```
mateuszbaran: Could you add some tests here with actually complex-valued matrices?

src/manifolds/GeneralizedStiefel.jl
kellertuer: Right, I will change this line.

test/generalized_stiefel.jl
kellertuer: Yes, I hope that works too, but that definetlyshould be tested. This is just a first version that does run properly and the next push will increase code coverage and cover the complex case.

kellertuer:
Strangely most `default_metric_dispatch` definitions are currently not called, I have to check how I might have deactivated that (that's where this PR looses code coverage); I had hoped that I only forward all `AbstractEmbeddedManifolds` to their corresponding transparent dispatches.

mateuszbaran:
I don't really understand that either, maybe CI made a mistake here? Sometimes in such cases I'm just adding a `print` statement to the method that is reported as not called to make sure it's a real problem.

kellertuer:
Well I can state that it does not print so it's not called ;) I have to check where I introduced that since I thought I'd only made the `AbstractEmbeddedManifold` case more secure.

mateuszbaran:
Probably some optimization pass gets rid of these functions (through inlining and constant propagation) before the intermediate representation used by CI is generated. I have no idea though why it was correctly identified before but not in this PR.

kellertuer:
Just group is going down in code covergae because I did not care about hat/vee/get_vectors/get_coordinates to not interfere with the other PR. I did however fix a few other things to increace code coverage in general.

src/manifolds/GeneralizedStiefel.jl
```diff
@@ -0,0 +1,245 @@
...
+The manifold is named after
+[Eduard L. Stiefel](https://en.wikipedia.org/wiki/Eduard_Stiefel) (1909–1978).
+
+# Constructor
+    GeneralizedStiefel(n, k, B=I_k, F=ℝ)
+
+Generate the (real-valued) Generalized Stiefel manifold of $n\times k$ dimensional
+orthonormal matrices with scalar product `B`.
+"""
+struct GeneralizedStiefel{n,k,TB<:AbstractMatrix,F} <: AbstractEmbeddedManifold{AbstractEmbeddingType}
```
mateuszbaran:
```suggestion
struct GeneralizedStiefel{n,k,TB<:AbstractMatrix,F} <: AbstractEmbeddedManifold{DefaultEmbeddingType}
```
I think it makes sense to have a concrete type here. `DefaultEmbeddingType` would just be a simple struct for default embeddings that are not transparent isometric.

src/manifolds/GeneralizedStiefel.jl
```diff
@@ -0,0 +1,245 @@
...
+````
+
+where $\operatorname{Sym}(y)$ is the symmetrization of $y$, e.g. by
+$\operatorname{Sym}(y) = \frac{y^{\mathrm{H}}+y}{2}$.
+"""
+project_tangent(::GeneralizedStiefel, ::Any...)
+
+function  project_tangent!(M::GeneralizedStiefel, Y, p, X)
+    A = p'*M.B'*X
+    copyto!(Y, X - p*Hermitian(0.5*(A+A')))
```
mateuszbaran:
```suggestion
    copyto!(Y, X - p*Hermitian((A+A')/2))
```

src/manifolds/Stiefel.jl
```diff
@@ -32,10 +31,13 @@ The manifold is named after
 
 Generate the (real-valued) Stiefel manifold of $n × k$ dimensional orthonormal matrices.
 """
-struct Stiefel{n,k,𝔽} <: Manifold end
+struct Stiefel{n,k,𝔽} <: AbstractEmbeddedManifold{AbstractIsometricEmbeddingType} end
```
mateuszbaran: I'd prefer a concrete type here too.

mateuszbaran:
I think this looks OK, apart from these minor issues. Do you want to merge it before of after #115 ?

src/manifolds/GeneralizedStiefel.jl
kellertuer: Yes that sounds reasonable. Will change that.

kellertuer:
I don't have a strong preference, the advantage of merging this first, would be thst this is done. I will add the basis later, when I thought about that. I would  help merging this to base, but I was carefull to hopefully not interfere too much.
If we wait for base, this might be open for a while.

src/manifolds/GeneralizedStiefel.jl
kellertuer: But it might take some time, I get errors in some dispatches after introducing that type (and a corresponding default iso type).

mateuszbaran:
OK, let's merge this first then.

kellertuer:
Well first of all I messed up the code by introducing the concrete types and I can't see what happened, want to take a look?

mateuszbaran:
Sure, I'll take a look once I finish the current batch of changes to bases.

kellertuer:
So if one does not confuse the two new types (`DefaultEmbeddingType`, `DefaultIsometricEmbeddingType`) it seems to work. Good idea.

mateuszbaran:
Yes, also the non-generalized Stiefel should be isometric, right?

mateuszbaran:
I'll commit the fix in a moment.

kellertuer:
Ah, yes, thanks for noticing, got the two new types directly mixed up in search/replace.

kellertuer:
So as soon as the tests are passing I'll merge this.

PR #124, opened 2020-03-20
URL changes related to changing Github ORG

mateuszbaran:


Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 Arpack = "7d9fca2a-8960-54d3-9f78-7d1dccf2cb97"
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
```
kellertuer: Is this still necessary now that the other branch just works fine, where I think we did not fix Base? Or you you want to keep it just to be safe?

mateuszbaran: We did pin DiffEqBase to version 6.20 in the basis PR as well. The issue with DiffEqBase is not solved yet (it was solved for a short while and then broken again).

kellertuer: Then let's stay at this until the figured it out.

Issue #121, opened 2020-03-10
Naming Scheme for project() and embed()


kellertuer:
I noticed that I introduced a slight naming inconsistency with `EmbeddedManifolds`. Currently there are

* `project_point(M,p)` (and `project_point!(M, q, p)`) for points
* `project_tangent(M,p,X)` (and `project_tangent!(M, Y, p,X)`) for tangents

but the opposite direction is currently
* `embed(M,p)` (and `embed!(M, q, p)`) for points
* `embed(M,p,X)` (and `embed!(M, Y, p, X)`) for tangents.

How should we unify this? I think one could simplify the first to just a `project`(`!`)? The only argument I see against that is that `project(M,p,X)` for tangents and `project!(M,q,p)` for points are quite similar in writing and arguments but do quite different things.

What do you think?

mateuszbaran:
This inconsistency doesn't bother me that much. We can drop `_point` and `_tangent` if you want, ambiguity is not that much of an issue since we have different naming conventions for points and tangent vectors. Also `isapprox` follows the same pattern.

kellertuer:
Right, `isapprox` follows the same scheme, then I would prefer to drop `_point` and `_tangent` to be consistent with `embed`. Yes we are not completely consistent, but here we can easily introduce a consistency. I will do this when the current two PRs are done.

PR #127, opened 2020-03-23
Move EmbeddedManifold and adapt Bases

kellertuer:
This PR moves EmbeddedManifold to base and fixes the changes necessary for the new basis interpretation, for example

- [x] complex Euclidean is still confused which basis to use (a real valued 3-dim one with complex coefficients or a 6-dim complex one with possibly real coefficients).

It also solves #121.


kellertuer:
This should work once the Bugfix in JuliaManifolds/ManifoldsBase.jl#29 is merged.

PR #130, opened 2020-03-27
Add CompatHelper as GitHub Action

sethaxen:


kellertuer:
We could do the same with `ManifoldsBase`.

sethaxen:
> We could do the same with `ManifoldsBase`.

Does it need it? ATM its only compat entry has no upper bound to update.

kellertuer:
As far as I read, it would be good, if it gets an upper bound; but for compat helper you'Re right: we don't need compat helper (yet) for just one entry

mateuszbaran:
Technically, the only dependency of `ManifoldsBase` is Julia (and two of its stdlibs). I think we can manually update it when Julia 2.0 is released 🙂 .

PR #131, opened 2020-03-27
Bump StatsBase and Distributions compats

sethaxen:


mateuszbaran:
Before merging this we should make sure that CI actually uses these versions on at least Julia 1.4.

PR #138, opened 2020-03-29
CompatHelper should use the oldest supported (by us) Julia version

mateuszbaran:
Maybe this will work?

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: julia-actions/setup-julia@latest
         with:
-          version: 1.3
+          version: 1.0
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
       - name: CompatHelper.main()
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: julia-actions/setup-julia@latest
         with:
-          version: 1.0
+          version: 1.1
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
       - name: CompatHelper.main()


sethaxen:
What does this do for us?

kellertuer:
Oh, the problem might have vanished with your manual update. It tried twice to do a PR to update Arpack and SpecialFunctions, when that yielded errors due to our Manifest. Maybe this PR is currently obsolete since with #140 this was solved.

In general the problem with CompatHelper seems to be, that it sometimes is not able to get Julia 1.0 constraints right, since CompatHelper itself is only running on/with Julia 1.1 onwards.

dkarrasch:
But CompatHelper is not supposed to get any constraints right. It simply recognizes when packages (that you depend on) are released with a version number that is not covered by your current Project.toml. Then it makes a PR that adds this new version, so you can run CI tests on it. If that newly added version is incompatible with, say, Julia v1.0, that should be written in its own Project.toml, so that Pkg will use a compatible version there. That's what all the Project.toml business is good for.

I think this PR is obsolete, because it simply determines on which Julia version the CompatHelper is running. It does not determine w.r.t. which Julia versions compatibility is supposed to be guaranteed. And in fact, you do want to use more recent package versions on Julia v1.4 than on v1.0 (think of OrdinaryDiffEq.jl, for instance).

I haven't closely followed the progress here for a while, but if you have issues with versions of packages that you depend on, I believe it's because those packages have messed up their registry/Project.toml compat bounds.

kellertuer:
Actuallly @sethaxen resolved the issues yesterday. The origin of this PR came from #132 which we closed since it was not compatible with our Manifest at that time and then it was repeated by CompatHelper the next day (#136). With now a resolved Project.toml and no necessity for a checked in Manifest, this PR is obsolete.

mateuszbaran:
The problem I was trying to solve here was that CompatHelper opened two separate PRs for the same version of Arpack and SpecialFunctions. This seems to be resolved now by #140 .

One of the issue was that Pkg on newer versions of Julia was picking too new versions of packages (not compatible with Julia 1.0). I had to generate Manifold.toml files with Julia 1.0 to get versions compatible with 1.0. I was hoping that setting CompatHelper to use Julia 1.1 will make it use Pkg from 1.1 that, at least once, generated Manifest.toml compatible with Julia 1.0 as well. 

> But CompatHelper is not supposed to get any constraints right. It simply recognizes when packages (that you depend on) are released with a version number that is not covered by your current Project.toml. Then it makes a PR that adds this new version, so you can run CI tests on it. If that newly added version is incompatible with, say, Julia v1.0, that should be written in its own Project.toml, so that Pkg will use a compatible version there. That's what all the Project.toml business is good for.
> 
> I think this PR is obsolete, because it simply determines on which Julia version the CompatHelper is running. It does not determine w.r.t. which Julia versions compatibility is supposed to be guaranteed. And in fact, you do want to use more recent package versions on Julia v1.4 than on v1.0 (think of OrdinaryDiffEq.jl, for instance).

CompatHelper also tries to update Manifold.toml and newer versions of Julia can't generate Manifest file compatible with Julia 1.0. It wouldn't be that much of a problem if CompatHelper opened one PR per release of a package we depend on but it looks like that wasn't the case.

PR #139, opened 2020-03-29
generalize the sphere manifold to the complex case and to unit norm matrices

kellertuer:
This PR introduces a generalisation of the `Sphere` manifold to two further cases

* (Frobenius) unit norm matrices
* complex vectors
* complex unit norm matrices

while keeping the old interface, that `Sphere(n)` introduces the `(n+1,)`-dimensional unit vectors, for matrices the interface of `(n,m)´ unit matrices are generated by `Sphere(m,n)`.

I think this could even be generalized (easily?) to Frobenius unit norm arrays, though with the change to a tuple as first parameter, it might be a little complicated to keep backwards-compatibility to the `Sphere(n)`.


src/manifolds/Sphere.jl
```diff
@@ -1,30 +1,43 @@
...
+
+    Sphere(n,m[, field=ℝ])
+
+Generate $𝕊^{n,m} ⊂ ℝ^{n,m}$.
 """
-struct Sphere{N} <: AbstractEmbeddedManifold{DefaultIsometricEmbeddingType} end
+struct Sphere{n,m,𝔽} <: AbstractEmbeddedManifold{DefaultIsometricEmbeddingType} end
 
-Sphere(n::Int) = Sphere{n}()
+Sphere(n::Int, field::AbstractNumbers=ℝ) = Sphere{n+1,1,field}()
```
mateuszbaran: I think it would be more consistent to use a tuple here instead of two size arguments. We did it similarly for Euclidean. By the way, we could even have multi-index spheres then.

kellertuer: Yes, I already discussed with Seth shortly yesterday, that this could also cover unit norm tensors. Maybe its better (for consistency) to introduce a different name for the general case, since otherwise the argument of the constructor is inconsistent (n - dim of the manifold for vectors, (n1,n2,n3,...) - dim of embedding otherwise; dim is then their product -1). So I will do a `GeneralizedSphere` for the tensors and use the subtype `Sphere` just for convenience then speaking of vectors. That's the plan.

mateuszbaran: That's a good idea.

kellertuer:
Locally the tests are running, but there's two questions left before I start working on the docs

* its now a `const` since I could not get it working with subtyping – but the `? Sphere` still works with that
* I noticed that some `show` methods print default values, is that intended? We could also just print `Euclidean(3)` instead of `Euclidean(3; field=ℝ)` and I even think we are not completely consistent on this
* I would like to keep both manifolds, though they have different names on the same `sphere.md` docs page – what do you think? The sphere alone would just be the const I fear.

Still to do

- [ ] write test cases for the complex sphere
- [ ] write test cases for the generalized sphere
- [ ] write test cases for the complex generalized sphere
- [ ] adapt docs.

sethaxen:
> * I noticed that some `show` methods print default values, is that intended? We could also just print `Euclidean(3)` instead of `Euclidean(3; field=ℝ)` and I even think we are not completely consistent on this -> I noticed that mainly my Sphere was off, so I added that to the Sphere. We should still check that all others not also print the field in case they have different possible fields.

The reason I did this was because we then wouldn't need to worry about making sure every constructor with defaults had a specialized `show` method. Also we wouldn't need to worry about synchronizing. I don't have a strong objection to specializing, it's just more work.

> * its now a `const` since I could not get it working with subtyping – but the `? Sphere` still works with that

What were the problems with subtyping? The downside to the const approach is that `Sphere(2)` makes `Sphere{3}`, which will definitely end up confusing users.

kellertuer:
Well, I can only get an `invalid subtyping error` or not getting the numbers right since you would need
```julia
struct Sphere{N,F} <: GeneralizedSphere{Tuple{N+1},F} where {N,F}
```
The `N+1` yields an error that the `+` is not defined, since `N` is a type, so yes, currently `Sphere(2)` is `Sphere{3}`. So I haven't found (even with a lot of trying) a sub typing that works. If you have an idea how to do that, I'd happily try that, I also don't like the internal type that much (with the offset), but I don't see the solution currently.

Concerning the field I think I did not do that for all (lie GeneralizedStuefel and such) I will make sure, that it is unified within this PR then.

sethaxen:
> Well, I can only get an `invalid subtyping error` or not getting the numbers right since you would need
> 
> ```julia
> struct Sphere{N,F} <: GeneralizedSphere{Tuple{N+1},F} where {N,F} end
> ```
> 
> The `N+1` yields an error that the `+` is not defined, since `N` is a type, so yes, currently `Sphere(2)` is `Sphere{3}`. So I haven't found (even with a lot of trying) a sub typing that works. If you have an idea how to do that, I'd happily try that, I also don't like the internal type that much (with the offset), but I don't see the solution currently.

I think the usual way to do this is with an auxiliary type parameter near the end of the signature and have the most explicit constructor enforce the link between the auxiliary type and the other types. e.g. if `GeneralizedSphere` is an abstract type

```julia
struct Sphere{N,F,M} <: GeneralizedSphere{Tuple{M},F}
    function Sphere{N,F,M}() where {N,F,M}
        M == N + 1 || error()
        return new{N,F,M}()
    end
    Sphere{N,F}() where {N,F} = Sphere{N,F,N+1}()
end
```

Then define other convenient constructors like `Sphere(n)`.

> Concerning the field I think I did not do that for all (lie GeneralizedStuefel and such) I will make sure, that it is unified within this PR then.

Sounds good!

kellertuer:
Ah, not it is not abstract, but I can of course do an `AbstractSphere` as a common abstract type for both the generalized and the classical sphere. That even solves the problems I had when I wanted to start documentations, since I can use the abstract case for documentations.

sethaxen:
I wonder if there's a better name than `GeneralizedSphere`. The thing is "generalized" is so vague. In many cases I think would would assume generalized means more fields are allowed. But one could also generalize the sphere by giving it a radius, etc. I thought the same thing with `GeneralizedGrassmann` and `GeneralizedStiefel`. For those, it seemed a name like `ScaledGrassmann` and `ScaledStiefel` would be more descriptive since they are embedded in a scaled (really, scaled and rotated) Euclidean space, and other generalizations are possible (e.g. multidimensional arrays like you have here, such as in http://www.mathem.pub.ro/proc/bsgp-04/K04-TREN.PDF).

kellertuer:
That's a good point. I merely adapted the `GeneralizedStiefel` from both manopt.org and the common literature. But you're right, even there, the generalized Stiefel is not consistently used.

For the sphere I think this naming is easy – `TensorSphere`! As we discussed `HighDimensionalArraySphere` just does not read that nice.

For `GeneralizedStiefel`/-`Grassmann` I find `Scaled`- a good solution, but we should point out in the documentation then, that in the literature these are usually called generalized. Higher Dimensional cases (if we do them) are then (consistently to above sphere) called `TensorStiefel`. Renaming these might be a little beyond this PR and should be done in a new one.

 What do you think?

sethaxen:
> That's a good point. I merely adapted the `GeneralizedStiefel` from both manopt.org and the common literature. But you're right, even there, the generalized Stiefel is not consistently used.
> 
> For the sphere I think this naming is easy – `TensorSphere`! As we discussed `HighDimensionalArraySphere` just does not read that nice.

I agree, something like `MultidimensionalArraySphere` is just too long. My issue with `TensorSphere` is that it uses a rather informal definition of tensor, a tensor being more than a multidimensional array. What about just `ArraySphere`? In Julia, vectors, matrices, and multidimensional arrays are all just `Array`, so it accurately describes what its points are and is not long.

> For `GeneralizedStiefel`/-`Grassmann` I find `Scaled`- a good solution, but we should point out in the documentation then, that in the literature these are usually called generalized. Higher Dimensional cases (if we do them) are then (consistently to above sphere) called `TensorStiefel`.

I'm not as familiar with the literature on these as you are. I'd never actually come across generalized Stiefel and Grassmann in the literature, but when I went looking for them, I found different generalizations. If these are indeed the standard terms, then we should probably keep them. I do though like the clarity of `Scaled`. I'm also planning to add a `ScaledEuclideanMetric`, which would have the same inner product they use.

> Renaming these might be a little beyond this PR and should be done in a new one.

I agree, I just felt it was a good time to bring it up. 🙂

kellertuer:
I don't like `ArraySphere`, since we have `ArrayManifold` in `ManifoldsBase`, that means something different, namely that testing is activated; so we need something different.

For `Stiefel` and `Grassmann` I will carefully check, but I read about one different generalized Grassmann. I will think about that and open an issue on that then.

mateuszbaran:
Would this definition of unit norm multi-index array actually correspond to some (basis-independent) notion of a unit-norm tensor?

And, since we are talking about spheres, there is also the Hilbert sphere which I'm going to add to FunManifolds.jl at some point, although it requires a separate implementation of basically every function (implementing it as just a high-dimensional sphere leads to a relatively poor approximation).

kellertuer:
At least for the `HilbertSphere` its good to have an `AbstractSphere`, so that's good.
Yes, If you interpret a multidimensional array as a linear map or a tenor, then the spheres introduced here are sets of unit norm tensors.

mateuszbaran:
OK, then calling it a `TensorSphere` makes sense.

kellertuer:
I took `TensorSphere` for now, but it can still be easily renamed (just a global search&replace). I also introduced the `AbstractSphere` and that even avoids the additional parameter @sethaxen proposed, and a lot of function can even be dispatched on the abstract level; Oblique also got nicer (we  could even extend in to complex oblique actually).

I also added some tests for the functions that are not already covered by the general case.

So I just have to extend the docs, which I think should still stay on one page, but I will look into that tomorrow evening.

kellertuer:
Ah, I see the problem with `TensorSphere`, since we here have unit norm linear maps and missing the multi part...so we need a different name.

mateuszbaran:
What do you mean by missing the multi part?

sethaxen:
> I don't like `ArraySphere`, since we have `ArrayManifold` in `ManifoldsBase`, that means something different, namely that testing is activated; so we need something different.

Is this the only reason you don't like `ArraySphere`? Because I also think `ArrayManifold` is not a good name. An `ArrayManifold` has really nothing to do with arrays. You can pass it a point that does not even wrap an array, and it will work just fine. Even as you say here, it's all about testing, so it would probably make more sense to be named `TestedManifold`, `ValidatedManifold`, `CheckedManifold`, or something similar.

kellertuer:
> What do you mean by missing the multi part?

That was the try of a short explanation, why Tensor is wrong.

> Is this the only reason you don't like ArraySphere?

I am not yet sure, it imples that its unit norm arrays? I am not yet 100% convinced its a good name

But you are right, the original idea calling the current `ArrayManifold` what it is called might be misleading; the name stems from the idea, that the arguments are more properly typed arrays and thinking of that it might even be false for manifolds with own types...

So yes renaming it is a good idea and since we validate your names sound intriguing. But _if_ we change the name, I have a small issue with the names proposed: It's not the manifold that is validates, its the operations on there. But currently I am not yet sure how to reflect that in a name.
Maybe its more secure?

sethaxen:
> I am not yet sure, it imples that its unit norm arrays? I am not yet 100% convinced its a good name

But isn't that exactly what this is? A manifold with points `p` such that `norm(p::Array)==1`. `UnitNormManifold` would be perhaps an even more explicit name that releases the constraint that points be arrays, but it doesn't make explicit that this is a generalization of the `Sphere`. Or have I misunderstood what the multidimensional generalization is defined as.

> Edit: Maybe `ValidationManifold` – since you can validate your results and this makes that distinct from the manifold being validated?

I like `ValidationManifold`. I guess the only downside to it I see is that the checking methods it uses internally are called `check_` not `validate_`. But `CheckingManifold` sounds weird. `ValidationManifold` is nice and only slightly inconsistent.

kellertuer:
Ok
* I'll rename the generalisation to `ArraySphere`
* rename the `ArrayManifold` to `ValidationManifold`, once the other PR over at base is finished (let's not do too many things concurrently)

src/manifolds/Sphere.jl
```diff
@@ -1,81 +1,146 @@
+"""
+    AbstractSphere{𝔽} <: AbstractEmbeddedManifold{DefaultIsometricEmbeddingType}
+
+An abstract type to represent a unit sphere that is represented isometrically in the embedding.
+"""
+abstract type AbstractSphere{𝔽} <:
+    AbstractEmbeddedManifold{DefaultIsometricEmbeddingType} where {𝔽<:AbstractNumbers} end
+
+@doc raw"""
+    Sphere{n,𝔽} = AbstractSphere{𝔽}
```
sethaxen, line 10:
```suggestion
    Sphere{n,𝔽} <: AbstractSphere{𝔽}
```

src/manifolds/Sphere.jl
```diff
@@ -274,93 +342,98 @@ end
...
-    uniform_distribution(M::Sphere, p)
+    uniform_distribution(M::AbstractSphere, p)
 
-Uniform distribution on given [`Sphere`](@ref) `M`. Generated points will be of
+Uniform distribution on given [`AbstractSphere`](@ref) `M`. Generated points will be of
 similar type as `p`.
 """
-function uniform_distribution(M::Sphere, p)
+function uniform_distribution(M::AbstractSphere, p)
     d = Distributions.MvNormal(zero(p), 1.0)
```
sethaxen, line 414: Hmm, this won't generalize beyond the 1D case. You could also extend to the 2D case using `MatrixNormal` as in #142, but to go beyond that with `ProjectedPointDistribution` you'll need something else. I don't think Distributions supports multidimensional array distributions.

sethaxen: You could always take the stand normal implementation in #142 and generalize it for multidimensional arrays.

src/manifolds/Sphere.jl
```diff
@@ -274,93 +342,98 @@ end
...
     return ProjectedPointDistribution(M, d, project!, p)
 end
 
 @doc doc"""
-    vector_transport_to(M, p, X, q, ::ParallelTransport)
+    vector_transport_to(M::AbstractSphere, p, X, q, ::ParallelTransport)
 
-Compute the paralllel transport of the tangent vector `X` at `p` to `q`,
-provided, the [`geodesic`](@ref) between `p` and `q` is unique
+Compute the paralllel transport on the [`Sphere`](@ref) of the tangent vector `X` at `p`
```
sethaxen, line 421:
```suggestion
Compute the parallel transport on the [`Sphere`](@ref) of the tangent vector `X` at `p`
```

src/manifolds/Sphere.jl
kellertuer: Then we should for this PR maybe just restrict it to the `Sphere` again? Does it work for the complex case?

Once #142 is done we can generalise it to the `ArraySphere`, too.

sethaxen: That's the other thing. I don't think Distributions.jl supports any complex distributions currently. These gaps are getting annoying enough that it's probably time to start implementing some of #57, at least the `Uniform` distribution for starters. I'd say limit to real `Sphere` for now, and once I've made a PR for `Uniform`, we can switch it over.

kellertuer: Thanks for noticing, I will restrict it.

src/manifolds/Sphere.jl
```diff
@@ -1,81 +1,146 @@
...
+An abstract type to represent a unit sphere that is represented isometrically in the embedding.
+"""
+abstract type AbstractSphere{𝔽} <:
+    AbstractEmbeddedManifold{DefaultIsometricEmbeddingType} where {𝔽<:AbstractNumbers} end
+
+@doc raw"""
+    Sphere{n,𝔽} <: AbstractSphere{𝔽}
+
+The (unit) sphere manifold $𝕊^{n}$ is the set of all unit norm vectors in $𝔽^{N}$, $𝔽 elements. The sphere is
+represented in the embedding, and currently supports both vectors and matrices, i.e.
```
mateuszbaran, line 13: I think representing points on a `Sphere` using matrices isn't really advisable (although it should more or less work in practice).

src/manifolds/Sphere.jl
```diff
@@ -1,81 +1,146 @@
...
+
+````math
+𝕊^{n} := \bigl\{ p \in 𝔽^{n+1}\ \big|\ \lVert p \rVert = 1 \bigr\}
+````
+
+where $𝔽\in\{ℝ,ℂ\}$. Note that compared to the [`ArraySphere`](@ref), here the
+argument of the manifold is the dimension of the manifold, i.e. $𝕊^{n} ⊂ 𝔽^{n+1}$, $n\in ℕ$.
```
mateuszbaran, line 20:
```suggestion
argument `n` of the manifold is the dimension of the manifold, i.e. $𝕊^{n} ⊂ 𝔽^{n+1}$, $n\in ℕ$.
```

src/manifolds/Sphere.jl
```diff
@@ -1,81 +1,146 @@
...
+
+The tangent space at point p is given by
```
mateuszbaran, line 22:
```suggestion
The tangent space at point $p$ is given by
```

src/manifolds/Sphere.jl
```diff
@@ -84,59 +149,61 @@ function check_tangent_vector(
...
     V = nullspace(A)
-    κ = ones(N)
+    κ = ones(n)
     if !iszero(B.frame_direction)
         # if we have a nonzero direction for the geodesic, add it and it gets curvature zero from the tensor
         V = cat(B.frame_direction / norm(M, p, B.frame_direction), V; dims = 2)
         κ[1] = 0 # no curvature along the geodesic direction, if x!=y
     end
-    Ξ = [V[:, i] for i = 1:N]
+    Ξ = [V[:, i] for i = 1:manifold_dimension(M)]
```
mateuszbaran, line 206:
```suggestion
    Ξ = [V[:, i] for i = 1:n]
```
This can still be just `n` like before.

src/manifolds/Sphere.jl
```diff
@@ -147,112 +214,113 @@ Represent the tangent vector `X` at point `p` from the [`Sphere`](@ref) `M` in
 an orthonormal basis by rotating the vector `X` using the rotation matrix
 $2\frac{q q^\mathrm{T}}{q^\mathrm{T} q} - I$ where $q = p + (1, 0, …, 0)$.
 """
-function get_coordinates(M::Sphere{N}, p, X, B::DefaultOrthonormalBasis) where {N}
-    if isapprox(p[1], 1)
+function get_coordinates(M::Sphere{n,ℝ}, p, X, B::DefaultOrthonormalBasis) where {n}
+    if isapprox(abs(p[1]), 1)
```
mateuszbaran, line 218: Why did you add `abs` here? I think there real error here is in the `else` section (`xp1` should probably be `p .+ ntuple(i -> ifelse(i == 1, -1, 0), n+1)`). Similarly for `get_vector`

Ref: https://github.com/JuliaManifolds/Manifolds.jl/pull/86#discussion_r364019155

src/manifolds/Sphere.jl
```diff
@@ -147,112 +214,113 @@ Represent the tangent vector `X` at point `p` from the [`Sphere`](@ref) `M` in
...
         return X[2:end]
     else
-        xp1 = p .+ ntuple(i -> ifelse(i == 1, 1, 0), N + 1)
-        return (2*xp1*dot(xp1, X)/dot(xp1, xp1)-X)[2:end]
+        xp1 = p .+ ntuple(i -> ifelse(i == 1, 1, 0), n+1)
+        return (2*xp1*real(dot(xp1, X))/real(dot(xp1, xp1))-X)[2:end]
```
mateuszbaran, line 222:
```suggestion
        return (2*xp1*dot(xp1, X)/dot(xp1, xp1)-X)[2:end]
```
This real does nothing since the method is only for the real sphere anyway.

test/sphere.jl
```diff
@@ -85,4 +85,26 @@ include("utils.jl")
...
     end
+
+    @testset "Complex Sphere" begin
+        M = Sphere(2,ℂ)
+        @test repr(M) == "Sphere(2; field = ℂ)"
+        @test typeof(get_embedding(M)) === Euclidean{Tuple{3},ℂ}
+        @test representation_size(M) == (3,)
+        p = [1.0, 1.0im, 1.0]
+        q = project(M,p)
+        @test is_manifold_point(M,q)
```
mateuszbaran, line 96: I'd suggest adding a test for tangent vector projection on the complex sphere.

mateuszbaran: Or, even better, call `test_manifold` for complex sphere and tensor sphere.

src/manifolds/Sphere.jl
kellertuer: Oh, that's a leftover from previous docs, I will omit the matrices part.

src/manifolds/Sphere.jl
kellertuer: I added the `abs` when thinking of the complex case, but forgot to remove it when leaving this to the real case. For the real case the tuple is fine.

The basis for the complex case is something I first have to think about.

src/manifolds/Sphere.jl
kellertuer: yes this is also a leftover from my try to get a complex sphere basis, that I missed to remove, thanks for noticing.

test/sphere.jl
kellertuer: I am not so sure, we are again at 51 minutes currently for Julia 1.0, which already kills travis from time to time, running a whole `test_manifold` might be quite long (ok and I was lazy to think of 3 points on both of the manifolds).

mateuszbaran: Then maybe we can have just a test for tangent vector projection here.

kellertuer: Ok, I'll hack in a few numbers.

PR #140, opened 2020-03-29
Remove version constraints

sethaxen:
Out of sheer morbid curiosity, remove the version constraints on SpecialFunctions and Arpack introduced in #92, as well as the Manifest.


kellertuer:
So at least on travis it seems to work, so AppVeyor should just do strange things.

What I think is really nice, is, that we don't need an explicit Manifest anymore; could you adapt the `Project.toml`. within `docs/`, too?

sethaxen:
Yeah, looks like everything builds just fine now. I removed the docs manifest as well and made its `Project.toml` consistent with that of the project. Since this releases a lot of constraints, I incremented the version number so we can do another patch release if all tests pass.

mateuszbaran:
Great, thanks for checking if the manifest file is still needed.

PR #142, opened 2020-03-31
Add uniform distribution for Stiefel/Grassmann, fixes #141

### LINKED ISSUE #141: Add uniform_distribution for Stiefel?
Hi! I'm really excited about this package. I have the following snippet floating around in my code for generating a random point uniform on the Stiefel:

```Julia
struct UnifSt{T<:Integer} <: Sampleable{Matrixvariate, Continuous}
    k::T    # inner dimension
    m::T    # outer dimension
    UnifSt{T}(k::T,m::T) where {T<:Integer} = k > m ? error("k > m") : new(k,m)
end
UnifSt(k::T,m::T) where {T<:Integer} = UnifSt{T}(k,m)

Base.size(d::UnifSt) = (d.m,d.k)
Base.size(d::UnifSt, i) = i::Integer <= 2 ? size(d)[i] : 1
function Distributions._rand!(rng::AbstractRNG, d::UnifSt, A::AbstractMatrix)
    randn!(rng,A)
    F = svd!(A)
    A .= F.U*F.Vt
end
```

and I'd be happy to replace it by using `Manifolds.jl`. ;)

Looking through the `uniform_distribution` code for the sphere, I think I could try adding the Stiefel analogue - it seems it would be largely the same.

Is this something already in the works? I'd be happy to make a pull request! :)

dahong67:
Fixes #141 

Added `uniform_distribution` for `Stiefel` following the example for `Sphere`. In order to fit within the `ProjectedPointDistribution` framework, created a thin wrapper around `randn` called (for now) `RandnMatrix`. Happy to iterate on this, not sure if it's the best way!

Some todos:
+ [x] ~~(maybe) simplify implementation of `project!(M::Stiefel, q, p)` to eliminate (potentially unnecessary) eigendecomposition of U'U = I (up to numerical errors).~~ (new pr: #153)
+ [x] ~~complex-valued case? (currently not handled)~~ (new issue: #152)
+ [x] Grassmann (should be easy once we're happy with the Stiefel)
+ [x] ~~(maybe) tackle generalized Stiefel/Grassmann~~ (new issue: #152)
+ [x] tests (will likely follow whatever is done for `Sphere`)

### DIFF
--- a/src/manifolds/Grassmann.jl
+++ b/src/manifolds/Grassmann.jl
@@ -384,6 +384,7 @@ show(io::IO, ::Grassmann{n,k,𝔽}) where {n,k,𝔽} = print(io, "Grassmann($(n)
     uniform_distribution(M::Grassmann{n,k,ℝ}, p)
 
 Uniform distribution on given (real-valued) [`Grassmann`](@ref) `M`.
+Specifically, this is the normalized Haar measure on `M`.
 Generated points will be of similar type as `p`.
 
 The implementation is based on Section 2.5.1 in [^Chikuse2003];
--- a/src/manifolds/Grassmann.jl
+++ b/src/manifolds/Grassmann.jl
@@ -395,7 +395,13 @@ see also Theorem 2.2.2(iii) in [^Chikuse2003].
     > doi: [10.1007/978-0-387-21540-2](https://doi.org/10.1007/978-0-387-21540-2).
 """
 function uniform_distribution(M::Grassmann{n,k,ℝ}, p) where {n,k}
-    return uniform_distribution(Stiefel(n, k, ℝ), p)
+    μ = Distributions.Zeros(n, k)
+    σ = one(eltype(p))
+    Σ1 = Distributions.PDMats.ScalMat(n, σ)
+    Σ2 = Distributions.PDMats.ScalMat(k, σ)
+    d = MatrixNormal(μ, Σ1, Σ2)
+
+    return ProjectedPointDistribution(M, d, (M, q, p) -> (q .= svd(p).U), p)
 end
 
 @doc raw"""
--- a/src/manifolds/Grassmann.jl
+++ b/src/manifolds/Grassmann.jl
@@ -389,3 +389,19 @@ which is given by a zero matrix the same size as `p`.
 zero_tangent_vector(::Grassmann, ::Any...)
 
 zero_tangent_vector!(::Grassmann, X, p) = fill!(X, 0)
+
+"""
+    uniform_distribution(M::Grassmann{n,k,ℝ}, p)
+
+Uniform distribution on given (real-valued) [`Grassmann`](@ref) `M`.
+Generated points will be of similar type as `p`.
+
+The implementation is based on Section 2.5.1 in [^Chikuse2003];
+see also Theorem 2.2.2(iii) in [^Chikuse2003].
+
+[^Chikuse2003]:
+    > Y. Chikuse: "Statistics on Special Manifolds", Springer New York, 2003,
+    > doi: [10.1007/978-0-387-21540-2](https://doi.org/10.1007/978-0-387-21540-2).
+"""
+uniform_distribution(M::Grassmann{n,k,ℝ}, p) where {n,k} =
+    uniform_distribution(Stiefel(n, k, ℝ), p)
--- a/src/manifolds/Grassmann.jl
+++ b/src/manifolds/Grassmann.jl
@@ -380,16 +380,6 @@ end
 
 show(io::IO, ::Grassmann{n,k,𝔽}) where {n,k,𝔽} = print(io, "Grassmann($(n), $(k), $(𝔽))")
 
-@doc raw"""
-    zero_tangent_vector(M::Grassmann, p)
-
-Return the zero tangent vector from the tangent space at `p` on the [`Grassmann`](@ref) `M`,
-which is given by a zero matrix the same size as `p`.
-"""
-zero_tangent_vector(::Grassmann, ::Any...)
-
-zero_tangent_vector!(::Grassmann, X, p) = fill!(X, 0)
-
 """
     uniform_distribution(M::Grassmann{n,k,ℝ}, p)
 
@@ -403,5 +393,16 @@ see also Theorem 2.2.2(iii) in [^Chikuse2003].
     > Y. Chikuse: "Statistics on Special Manifolds", Springer New York, 2003,
     > doi: [10.1007/978-0-387-21540-2](https://doi.org/10.1007/978-0-387-21540-2).
 """
-uniform_distribution(M::Grassmann{n,k,ℝ}, p) where {n,k} =
-    uniform_distribution(Stiefel(n, k, ℝ), p)
+function uniform_distribution(M::Grassmann{n,k,ℝ}, p) where {n,k}
+    return uniform_distribution(Stiefel(n, k, ℝ), p)
+end
+
+@doc raw"""
+    zero_tangent_vector(M::Grassmann, p)
+
+Return the zero tangent vector from the tangent space at `p` on the [`Grassmann`](@ref) `M`,
+which is given by a zero matrix the same size as `p`.
+"""
+zero_tangent_vector(::Grassmann, ::Any...)
+
+zero_tangent_vector!(::Grassmann, X, p) = fill!(X, 0)
--- a/src/manifolds/Stiefel.jl
+++ b/src/manifolds/Stiefel.jl
@@ -35,7 +35,11 @@ struct Stiefel{n,k,𝔽} <: AbstractEmbeddedManifold{DefaultIsometricEmbeddingTy
 
 Stiefel(n::Int, k::Int, field::AbstractNumbers = ℝ) = Stiefel{n,k,field}()
 
-function allocation_promotion_function(M::Stiefel{n,k,ℂ}, f, args::Tuple) where {n,k}
+function allocation_promotion_function(
+    M::Stiefel{n,k,ℂ},
+    f,
+    args::Tuple,
+) where {n,k}
     return complex
 end
 
@@ -47,8 +51,7 @@ Check whether `p` is a valid point on the [`Stiefel`](@ref) `M`=$\operatorname{S
 complex conjugate transpose. The settings for approximately can be set with `kwargs...`.
 """
 function check_manifold_point(M::Stiefel{n,k,𝔽}, p; kwargs...) where {n,k,𝔽}
-    mpv =
-        invoke(check_manifold_point, Tuple{supertype(typeof(M)),typeof(p)}, M, p; kwargs...)
+    mpv = invoke(check_manifold_point, Tuple{supertype(typeof(M)), typeof(p)}, M, p; kwargs...)
     mpv === nothing || return mpv
     c = p' * p
     if !isapprox(c, one(c); kwargs...)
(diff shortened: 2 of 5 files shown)


dahong67:
Did a quick benchmark to see if `RandnMatrix` actually yields any improvement over using `MatrixNormal` from `Distributions.jl`. If not much improvement, might be simpler to just use `MatrixNormal` instead.

```julia
julia> using Random, Distributions, Manifolds

julia> using BenchmarkTools, Random, Distributions, Manifolds

julia> A = zeros(500,200);

julia> d1 = Manifolds.RandnMatrix(size(A)...)
Manifolds.RandnMatrix(n=500, m=200)

julia> cov1, cov2 = Distributions.PDMats.ScalMat.(size(A),one(eltype(A)),one(eltype(A)))
(PDMats.ScalMat{Float64}(500, 1.0, 1.0), PDMats.ScalMat{Float64}(200, 1.0, 1.0))

julia> d2 = MatrixNormal(zero(A),cov1,cov2)
MatrixNormal{Float64,Array{Float64,2},PDMats.ScalMat{Float64},PDMats.ScalMat{Float64}}(
M: [0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0]
U: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
V: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)

julia> rng = Random.GLOBAL_RNG
Random._GLOBAL_RNG()

julia> Distributions._rand!(rng,d1,A); Distributions._rand!(rng,d2,A);

julia> @btime Distributions._rand!($rng,$d1,$A);
  776.034 μs (0 allocations: 0 bytes)

julia> @btime Distributions._rand!($rng,$d2,$A);
  1.119 ms (18 allocations: 3.06 MiB)
```

Seems a bit slower and allocates (likely for intermediates).

kellertuer:
Concerning the complex case, currently your dispatch still would be called; you could use `Stiefel{m,n,ℝ}` as type to ensure it only gets called for the real-valued case.

src/manifolds/Stiefel.jl
```diff
@@ -318,3 +318,26 @@ i.e. `(n,k)`, which is the matrix dimensions.
...
+"""
+function uniform_distribution(M::Stiefel, p)
+    d = RandnMatrix(size(p)...)
+    return ProjectedPointDistribution(M, d, project!, p)
+end
+
+struct RandnMatrix <: ContinuousMatrixDistribution
+    n::Integer
+    m::Integer
+end
```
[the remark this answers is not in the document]
sethaxen: But we might ultimately want to switch over to `MatrixGaussian` when it is merged: https://github.com/JuliaStats/Distributions.jl/pull/1091. In a later PR, of course.

src/manifolds/Stiefel.jl
```diff
@@ -318,3 +318,26 @@ i.e. `(n,k)`, which is the matrix dimensions.
...
+Base.size(d::RandnMatrix) = (d.n,d.m)
+Base.size(d::RandnMatrix, i) = i::Integer <= 2 ? size(d)[i] : 1
+Distributions._rand!(rng::AbstractRNG, d::RandnMatrix, A::AbstractMatrix) = randn!(rng,A)
+Distributions._logpdf(d::RandnMatrix, x::AbstractArray) = sum(x) do xij
+    dij = Distributions.Normal(zero(xij),one(xij))
+    return Distributions.logpdf(dij,xij)
+end
```
sethaxen: Because this is just a multivariate normal variable reshaped into a matrix you should be able to do this:
```suggestion
function Distributions._logpdf(d::RandnMatrix, x::AbstractArray)
    dim = prod(size(d))
    σ = one(eltype(x))
    dvec = MvNormal(dim, σ)
    return Distributions._logpdf(dvec, vec(x))
end
```
This is nice because the `MvNormal(::Integer, ::Real)` signature makes a `ZeroMeanIsoNormal`, which leverages all the efficiencies of having a zero mean and scalar covariance.

src/manifolds/Stiefel.jl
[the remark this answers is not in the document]
sethaxen: I think also it's customary for distributions to store an eltype, so it probably be good to add that to the type.

sethaxen:
Hmm, now that I think of it, even though it's somewhat less efficient right now, perhaps we should just wrap `MatrixNormal`. We can get some speed-ups with `Distributions.Zeros(A)`. This is because with I think with https://github.com/JuliaStats/Distributions.jl/pull/1091, we'll get essentially the same thing.

src/manifolds/Stiefel.jl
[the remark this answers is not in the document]
dahong67: Generally agreed with your later suggestion to use `MatrixNormal` for now, but will commit this suggestion too so so that we also have the improved alternative handy. :)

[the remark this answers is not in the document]
dahong67: Agreed on both, I'm committing this suggestion and will add eltype in a follow-up commit.

src/manifolds/Stiefel.jl
dahong67: Good point, incorporated!

dahong67:
I agree, seems like `Distributions.jl` is the appropriate place for something like this anyway. :)

I'm incorporating your earlier improvements first though so we have an alternative handy. After that I'll add a commit replacing it with `MatrixNormal`.

dahong67:
Okay, done! New pull request is simpler/nicer since it doesn't involve creating a new distribution for the standard normal matrix. The final `MatrixStandardNormal` implementation can be found in commit bb6d9ae in which it was removed.

If this is looking good, I propose to:
+ add tests similar to those in `test/sphere.jl`
+ work on Grassmann analogue (seems we'll need a new `project!` method for that)
+ open a separate pull request for simplification of `project!(M::Stiefel, q, p)`
+ leave complex-valued Stiefel/Grassmann and generalized Stiefel/Grassman for future work (would be happy to open an issue to document the missing functionality) ;)

@kellertuer and @sethaxen, how does that sound to you?

kellertuer:
Sounds good. I haven't looked for a `project!` for onto Grassmann, so if you have a reference there, we can definitely add that. Your proposal sounds good :)

dahong67:
Agreed, I too need to think a bit more about a `project!` for Grassmann. Maybe tomorrow - will share what I learn!

src/manifolds/Stiefel.jl
```diff
@@ -318,3 +318,25 @@ i.e. `(n,k)`, which is the matrix dimensions.
...
+
+[^Chikuse2003]:
+    > Y. Chikuse: "Statistics on Special Manifolds", Springer New York, 2003,
+    > doi: [10.1007/978-0-387-21540-2](https://doi.org/10.1007/978-0-387-21540-2).
+"""
+function uniform_distribution(M::Stiefel{n,k,ℝ}, p) where {n,k}
+    μ = Distributions.Zeros(p)
+    σ = one(eltype(p))
+    Σ1 = Distributions.PDMats.ScalMat(size(p,1),σ)
+    Σ2 = Distributions.PDMats.ScalMat(size(p,2),σ)
```
sethaxen, line 338: I would use the explicit dimensions here, so an error will get thrown by Distributions if `p` has the wrong size. Also for `Zeros` above:
```suggestion
    Σ1 = Distributions.PDMats.ScalMat(n,σ)
    Σ2 = Distributions.PDMats.ScalMat(k,σ)
```

dahong67: Sounds good and agreed, incorporating now!

src/manifolds/Grassmann.jl
```diff
@@ -380,6 +380,23 @@ end
...
+
+The implementation is based on Section 2.5.1 in [^Chikuse2003];
+see also Theorem 2.2.2(iii) in [^Chikuse2003].
+
+[^Chikuse2003]:
+    > Y. Chikuse: "Statistics on Special Manifolds", Springer New York, 2003,
+    > doi: [10.1007/978-0-387-21540-2](https://doi.org/10.1007/978-0-387-21540-2).
+"""
+function uniform_distribution(M::Grassmann{n,k,ℝ}, p) where {n,k}
+    return uniform_distribution(Stiefel(n, k, ℝ), p)
```
mateuszbaran, line 397: If you want to re-use the uniform distribution implementation for `Stiefel`, you should make a helper function that is called from both methods of `uniform_distribution`. Otherwise the distribution returned for `Grassmann` will have wrong manifold set.

dahong67: Good point! I've implemented a separate `uniform_distribution` for `Grassmann`. The code overlaps but I think this is probably the conceptually simpler approach. :)

dahong67:
Fun to get back to this! (I'm working on this during my breaks.) I've incorporated the comments from @sethaxen and @mateuszbaran above (thanks again for all the feedback/guidance!), added tests and added a `Grassmann` version. New issue #152 discusses the complex-valued and generalized cases, and new pr #153 implemented the simplified `project!` method for `Stiefel`.

Some things to highlight:
+ After thinking a bit about `project!` for `Grassmann` and discussing briefly with a friend (@matthew-kvalheim), I'm still not entirely sure of what it should be in general. So I've instead used an anonymous function for `uniform_distribution`.
+ I haven't thought through whether the `uniform_distribution` for `Grassmann` is the Hausdorff measure so left that off the docstring for now - thoughts?
+ `JuliaFormatter` also made changes to other parts of the code (i.e., parts I did not touch). I didn't commit those since that seemed beyond the scope of this pr.
+ I tried running (local) tests with `TEST_FLOAT32`, `TEST_DOUBLE64` and `TEST_STATIC_SIZED` set to `true`, but other parts seemed to fail. Is this expected? Tests pass locally for me otherwise.

How is it looking now? Anything further needed to merge? :)

mateuszbaran:
> * I tried running (local) tests with `TEST_FLOAT32`, `TEST_DOUBLE64` and `TEST_STATIC_SIZED` set to `true`, but other parts seemed to fail. Is this expected? Tests pass locally for me otherwise.

That's not exactly expected. It was fixed as a part of #150 but it wasn't merged yet and should be in a separate PR (since I don't know when/if #150 is going to be merged).

sethaxen:
> That's not exactly expected. It was fixed as a part of #150 but it wasn't merged yet and should be in a separate PR (since I don't know when/if #150 is going to be merged).

Yes, the fixes for StaticArrays testing part of #150 should definitely be split out. Do we still want the CI jobs for it too? Also, I didn't do anything in #150 on the alternate float tests.

mateuszbaran:
> > That's not exactly expected. It was fixed as a part of #150 but it wasn't merged yet and should be in a separate PR (since I don't know when/if #150 is going to be merged).
> 
> Yes, the fixes for StaticArrays testing part of #150 should definitely be split out. Do we still want the CI jobs for it too? Also, I didn't do anything in #150 on the alternate float tests.

I'm not sure about splitting the CI jobs. I'm currently splitting these fixes to a separate PR.

dahong67:
Exciting to see this merged (and even already tagged!), thank you all for your patient guidance! :)

mateuszbaran:
Thank you for your contribution :slightly_smiling_face:

Issue #144, opened 2020-04-01
Speed up CI


sethaxen:
There are a few more things we can do to speed up CI and avoid timeouts:

1. Split the tests into groups that are run in parallel.
Choose an environment variable that if set can filter the tests to be run down to a group, defaulting to all. Then we set that variable in the build matrix in CI so the groups run in parallel.

2. Switch to another CI provider
e.g. GitHub Actions will let us run 20 Linux jobs in parallel, with 6 hour time limits. There's the org https://github.com/julia-actions, which has a number of tools to make GitHub Actions easy with Julia projects.

GitHub Actions is a little less user friendly for CI than Travis, and either way, it will benefit from test groups, so my thinking is to start with (1) above.

kellertuer:
I think this issue is finished, well maybe up to the point whether to deactivate AppVeyor and/or Travis?

mateuszbaran:
I think #150 is supposed to resolve this.

sethaxen:
GitHub Actions runs the test suite pretty fast, so I only saw about a 10 minute speed-up in CI with #150. But because the groups spawn so many jobs, it also fills up the 20-job queue pretty quickly, so I think it's no that useful yet, but probably will be later. I think we could deactivate Travis and Appveyor now and just revert that PR if we have problems with GH actions.

kellertuer:
Main question is whether we want to keep this issue open and want to do this or whether we are fine for now with the current hour of testing

sethaxen:
Might also be worth looking at some of the speed-ups used for the ChainRules test suite:
- https://github.com/JuliaDiff/ChainRules.jl/pull/499
- https://github.com/JuliaDiff/ChainRules.jl/pull/500

kellertuer:
...if only I would understand _anything_ from those changes.  I have absolutely no clue what both PRs are doing :(

sethaxen:
> ...if only I would understand _anything_ from those changes. I have absolutely no clue what both PRs are doing :(

You're in good company: https://github.com/JuliaDiff/ChainRules.jl/pull/499#issuecomment-898254561

kellertuer:
Excellent, with Frames on my side – I am fine I think.

I saw one thing that might do the one thing I describe above. But the timings for some manifolds are also still running, I will report :)

sethaxen:
[...]

PR #146, opened 2020-04-02
Split tests into groups for CI

sethaxen:
As proposed in #144, this PR splits tests into groups for faster runtimes. It's currently 3 groups with roughly equal runtimes. To reduce the number of jobs, I discontinued testing against 1.3, but we can always turn that back on.

I also saw that there was a huge runtime difference (5-6 minutes) between the statistics tests for Julia 1.0 and 1.4, coming entirely from `median` on the sphere running for the maximum number of iterations. Reducing the tolerances fixed that.


mateuszbaran:
Thanks, that's a good idea. I also think we can discontinue testing on 1.3. Maybe we could have another job for testing static arrays on Julia 1.4? We already have TEST_STATIC_SIZED.

sethaxen:
> Despite Mac OS being slow as so often in tests, this is really neat!

Yeah, we'll see how fast it is in the end. 38 different jobs is quite a lot. As long as the total suite runs significantly faster than the 35-40 minutes that we saw with GitHub Actions before splitting, then I think it's worth it. 

I added a new workflow for running CI on statically sized matrices. But this also repeats a lot of computation; ideally we would test statically sized or not statically sized, but not both. It's currently failing. Could someone take a look at that error message?

mateuszbaran:
I've checked these error messages and my commit should fix that problem.

sethaxen:
It seems like the changes in this PR really help with Travis but for GitHub Actions don't make much of a difference. Runtime is still 30-40 minutes for GA. But since for GA, fast fail mode is the default (if one job fails, it kills all jobs), splitting into groups means we fail faster, which is nice for not filling up the queue. So with that, I think this PR is ready to merge.

kellertuer:
The test suite looks really nice, but why did code coverage reduce by 16%?

sethaxen:
I don't know. It could be the usual codecov nonsense. If you look through the logs, the jobs that ran the tests it says it doesn't see were all run, and code coverage was uploaded. The commit it's comparing to was one that didn't run Travis, so maybe that messed with it somehow. We could merge and see after the suite runs on master whether that fixes the codecov issues. If not, we could revert and troubleshoot.

kellertuer:
Hm, ok, let's hope its just codecov nonsense then. I just noticed, because I invested quite some time writing tests for edge cases.

PR #147, opened 2020-04-02
Add GitHub Actions CI

sethaxen:
Testing out using GitHub Actions as CI, as proposed in #144.


mateuszbaran:
Do we need to trigger it on both `push` and `pull_request`?

kellertuer:
For documenter you might want to generate a set of keys – and do the GitHub Actions have a variable environment? You should then add the key as described here https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions-1 – and since this is a different config file, a separate key should work, so we don't have to use the travis one twice. Or wa can anyways decide to do just one deploy (the one here?).

sethaxen:
35 minutes to run the entire CI is pretty nice! Should be even lower with #146 

> Do we need to trigger it on both `push` and `pull_request`?

No, I'm going to remove `push`. 

> For documenter you might want to generate a set of keys – and do the GitHub Actions have a variable environment? You should then add the key as described here https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions-1 – and since this is a different config file, a separate key should work, so we don't have to use the travis one twice. Or wa can anyways decide to do just one deploy (the one here?).

At the moment, everything seems to work for the docs build. It didn't deploy, but it's not yet configured to deploy the preview build on PRs. Yes, I think ultimately, we'll want to just deploy once. My thinking is that if we like GitHub Actions for CI, then we can remove Travis and Appveyor, since it works on all 3 OSes.

I've heard that codecov doesn't currently work if the branch isn't from the same repo. If we choose to merge this when completed, I'll open a PR from a fork to test this.

sethaxen:
Okay, this seems to me to be ready for review and possible merging (once tests pass). Perhaps we can run this along with Travis and Appveyor for a little while and see if we like it enough to switch entirely to it.

One thing I noticed that's pretty cool is that when Documenter runs, a "documenter/deploy".
job appears. Clicking "Details" next to that run opens up the preview docs URL!

kellertuer:
The preview URL is _pretty_ cool, I always had to remember the URL style manually.

I think we can run them in parallel for a while, we just should do the deploy on only one, so remove it from travis maybe?

sethaxen:
Good idea. Done,

kellertuer:
The `[skip ci]` does nor yet seem to work? Can you do that label-based approach?

sethaxen:
Ah, it's not supposed to work. [skip ci] tells both Travis and Appveyor not
to run on that commit. I was just trying to avoid using up the queue with
frequent pushing. I might be able to do something similar with GitHub
actions, but I'd rather save that for another PR.

PR #149, opened 2020-04-04
Revert "Split tests into groups for CI"

sethaxen:
Reverts JuliaManifolds/Manifolds.jl#146


kellertuer:
I don't know why it missed out on all function signatures, but it did,
this might make Travis fail again in the long run (pun intended), but if Github Actions work fine, we can also use those, they seem to be really nice.

So I would prefer this version. Thanks for this PR.

sethaxen:
Yeah, this is just to revert until I can figure out what's wrong with the other PR. I have a theory. But you're right, this works for now, so we can keep moving forward. I'll merge.

PR #150, opened 2020-04-04
Split tests into groups for CI

sethaxen:
This is a copy of #146, which was merged and then reverted.


kellertuer:
Should we maybe close this, until we require something like that again? Currently with GitHub Actions it might not be that necessary.

sethaxen:
Sounds good to me

Issue #103, opened 2020-01-25
Add the simplex


sethaxen:
We should at least have the unit simplex, potentially also the general simplex

kellertuer:
You mean the Fisher-Rao metric on the interior of the probability simplex like https://ipa.math.uni-heidelberg.de/dokuwiki/Papers/Astroem2016d.pdf section 2.1? I have to check a few more references, but yes that could and should definitely be done.

sethaxen:
No, I wasn't proposing a particular choice of metric, per se. But I was proposing the unit/standard/probability simplex (see e.g. https://en.wikipedia.org/wiki/Simplex#The_standard_simplex). I have not had a chance to read the linked paper.

This particular manifold arises frequently in statistics, since points on the unit simplex are weight vectors that sum to 1, such as those in mixture models. I don't know how the general simplex is used.

kellertuer:
If we implement this we should at least have a metric; but I also haven't had time to look at this closer – and would also first like to look at embeddings (and finish the logo and finish the abstract).

kellertuer:
I read a little bit in the paper mentioned above, it's a metric for the relative interior of the simplex, so the boundary is excluded, but despite that they have nice formulae (though not that nicely organized/presented) for exp/log/inner/distance...

kellertuer:
I don't like the choice here https://www.manopt.org/reference/manopt/manifolds/multinomial/multinomialfactory.html that much, they implement a projection onto tangent spaces (does the previuos one also have) but a fake one onto the real interior again and the metric is the one from the embedding, which I think is strange (also no exp/log). So I think it will boil down to the interior of the simplex, not the simplex.

mateuszbaran:
"Simplex" may not be the greatest name for the discrete Fisher-Rao metric on the space of discrete probability distributions. And it's indeed only for the interior of the simplex, the definition of this metric assumes positive probabilities everywhere. Anyway, probably the simplest way to deal with the Fisher-Rao metric is by applying the square root and using a (scaled) sphere geometry... at least that's what I was doing in a very similar case.

mateuszbaran:
Technically the boundary of a simplex transforms OK to the boundary of the non-negative orthant of a sphere so there may be a unique extension of the metric to the boundary (assuming we want to deal with manifolds with boundaries) but I didn't think about it too much.

kellertuer:
[...]

PR #151, opened 2020-04-07
Introduce the manifold of the probability simplex

kellertuer:
This PR works on #103 and adds the probability simplex.
Documentation is already added, the tests are in preparation, currently either log or exp are still wrong.

I might also introduce the multinomial matrices manifold, since this as much work as `Oblique`, so just about 15 lines of code – which would then cover the manifold described [here](https://www.manopt.org/reference/manopt/manifolds/multinomial/multinomialfactory.html) .


src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,263 @@
...
+operations $X_p^2$ and $\sqrt{p}$.
+"""
+exp(::ProbabilitySimplex, ::Any...)
+
+function exp!(::ProbabilitySimplex, q, p, X)
+    Xp = X./sqrt.(p)
+    Xp_n = Xp./norm(Xp)
+    q .= 0.5*(p+Xp_n.^2) + 0.5*(p-Xp_n.^2).*cos(norm(Xp)) + Xp_n.*sqrt.(p).*sin(norm(Xp))
+    return q
+end
```
sethaxen, line 150: By my local tests, this seems to produce the same output as mapping to the sphere, following the geodesic, and then mapping back. But doing that is quite a bit more efficient. Here's an example that's basically that but avoids the final squaring in-place.
```suggestion
function exp!(::ProbabilitySimplex, q, p, X)
    s = sqrt.(p)
    Xs = X ./ s ./ 2
    θ = norm(Xs)
    q .= (cos(θ) .* s .+ usinc(θ) .* Xs).^2
    return q
end
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,263 @@
...
+@doc raw"""
+    inner(M::ProbabilitySimplex,p,X,Y)
+
+Compute the inner product of two tangent vectors `X`, `Y` from the tangent space $T_p𝒮$ at
+`p`. The formula reads
+````math
+g_p(X,Y) = \sum_{i=1}^{n+1}\frac{X_iY_i}{p}
+````
+"""
+inner(::ProbabilitySimplex, p, X, Y) = sum( (X.*Y)./p )
```
sethaxen, line 161: Like with `distance` above, this can be made more efficient by looping instead of broadcasting.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,263 @@
+@doc raw"""
+    ProbabilitySimplex{n} <: AbstractEmbeddedManifold{ℝ,DefaultEmbeddingType}
+
+The (relative interior of) the probability simplex is the set
+````math
+𝒮^{n} := \biggl\{ p \in ℝ^{n+1}\ \big|\ p_i > 0 \text{ for all } i=1,…,n+1,
```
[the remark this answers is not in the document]
kellertuer: I will add all that then, I don't have a strong preference towards the mathcal S, so yes, we can definitely use `Δ^n`

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,263 @@
...
+@doc raw"""
+    distance(M,p,q)
+
+Compute the sistance of two points on the [`ProbabilitySimplex`](@ref) `M`.
+The formula reads
+````math
+d_{𝒮}(p,q) = 2\arccos \biggl( \sum_{i=1}^{n+1} \sqrt{p_i q_i} \biggr)
+````
+"""
+distance(::ProbabilitySimplex,p,q) = 2*acos(sum(sqrt.(p.*q)))
```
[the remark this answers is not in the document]
kellertuer: Thanks for the suggestion. Does the last line (without `*`) work?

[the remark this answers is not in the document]
sethaxen: Yeah, but you can add it back in for clarity.

sethaxen:
Are you using the Fisher-Rao metric (equivalently the metric on the Sphere of radius 2) or the metric on the Sphere of radius 1? I know the paper uses the latter; I'm not 100% certain how the geodesic on the sphere changes when the radius is doubled.

kellertuer:
Oh, then I misread the paper, because as far as I saw it was the 2-sphere that I wanted to refer tom ,thanks for noticing. For clarity I wanted to go for the Fisher Rao metric (and also introduce that type), so I have to check again.

sethaxen:
Yeah, they actually call it the 2-sphere, which was confusing at first, because they also give it a different number of dimensions, so I'm pretty sure they use 2-sphere in a nonstandard way to mean sphere of radius 2.

kellertuer:
Ah yes, I checked, they use the sphere of radius 2, so it is Fisher Rao.

kellertuer:
Something is still wrong, I will check the formulae carefully tonight, but with the numbers in the test cases I still get
````julia
julia> norm(M,p,Y)
0.681385143869247

julia> distance(M,exp(M,p,Y),p)
0.6340174486112579
````
so either inner (and hence norm) or exp/distance contain a bug.

**Comment by sethaxen** [comment]
> Edit: Do you know a better reference with explicit formulae directly for Fisher-Rao then?

Do you mean formulae for exponential map, etc or for the entries of the metric matrix? For the former, this paper handles it just fine I think.

**Comment by kellertuer** [comment]
Yes, for the exponential map, but now that we found out it'S both the 2-sphere, I just have to find the bug that I programmed, i.e. were I misinterpreted their implicit broadcast they do in their paper throughout.

**Comment by sethaxen** [comment]
Ah, no, it looks like you just happened to pick an unfortunate base point and tangent vector: 
test
The point corresponds to `exp(M,p,Y)`, and it occurs right after a turn at the boundary, hence the discrepancy. Multiply `Y` by 0.9, and the test will pass.

And just for fun, continuing the geodesic:
test2

**Comment by kellertuer** [comment]
Thanks for noticing, so if we could find, an `injectivity_radius` (that actually depends on `p` would be helpful I think.

**Comment by sethaxen** [comment]
Yes I agree, although we should probably clarify what we mean in this context by injectivity radius. Technically, for much of the geodesic above, the exponential map is injective, but I think extending the exponential map like this is a trick since it technically passes through the boundary where it is discontinuous. So the injectivity radius should probably point to the nearest boundary, which would be as simple as finding the length of the shortest geodesic that hits the orthant at a right angle.

File: src/manifolds/ProbabilitySimplex.jl [hunk 5]
```diff
@@ -0,0 +1,294 @@
...
+\log_pq = \frac{d_{Δ^n}(p,q)}{\sqrt{1-⟨\sqrt{p},\sqrt{q}⟩}}(\sqrt{pq} - ⟨\sqrt{p},\sqrt{q}⟩p),
+````
+
+where $pq$ and $\sqrt{p}$ is meant elementwise.
+"""
+log(::ProbabilitySimplex, ::Any...)
+
+function log!(M::ProbabilitySimplex, X, p, q)
+    s = dot(sqrt.(p),sqrt.(q))
+    X .= (distance(M,p,q)/(1-s^2)) .* ( sqrt.(p.*q) .- s.*p )
```
sethaxen, line 226: The denominator is missing a `sqrt` from the paper, also here is a version that avoids duplicate calculations
```suggestion
    z = sqrt.(p .* q)
    s = sum(z)
    X .= 2acos(s) / sqrt(1 - s^2) .* (z .- s .* p)
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,294 @@
...
+````math
+\operatorname{retr}_p X = \frac{p\mathrm{e}^X}{⟨p,\mathrm{e}^X⟩},
+````
+
+where multiplication, exponentiation and division are meant elementwise.
+"""
+retract(::ProbabilitySimplex, ::Any, ::Any, ::SoftmaxRetraction)
+
+function retract!(::ProbabilitySimplex, q, p, X, ::SoftmaxRetraction)
+    q .= p.*exp.(X)./(dot(p,exp.(X)))
```
sethaxen, line 278: Haven't tested this but should be more efficient

```suggestion
    s = zero(eltype(q))
    @inbounds for i in eachindex(q, p, X)
        q[i] = p[i] * exp(X[i])
        s += q[i]
    end
    q ./= s
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,294 @@
...
+
+````math
+\operatorname{proj}_{Δ^n}(p,Y) = p\bigl(Y - ⟨p,Y⟩\mathbb{1}),
+````
+
+where multiplication is meant elementwise and $\mathbb{1}$ is the vector of ones.
+"""
+project(::ProbabilitySimplex, ::Any, ::Any)
+
+project!(::ProbabilitySimplex, X, p, Y) = X .= (p.*( Y .- dot(p,Y)))
```
sethaxen, line 254:
```suggestion
function project!(::ProbabilitySimplex, X, p, Y)
    s = zero(eltype(X))
    @inbounds for i in eachindex(X, p, Y)
        X[i] = p[i] * Y[i]
        s += X[i]
    end
    X .-= s .* p
    return X
end
```

sethaxen:
We can also set `GeodesicInterpolation` as the default method for `mean`, since as the paper notes we can essentially compute mean on the Sphere. The technique on the Sphere of radius 2 would be valid in a geodesic ball of radius π, but the maximum geodesic radius of the simplex is π/2, so the technique applies everywhere.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,294 @@
...
+function distance(::ProbabilitySimplex, p, q)
+    sumsqrt = zero(Base.promote_eltype(p, q))
+    @inbounds for i in eachindex(p, q)
+        sumsqrt += sqrt(p[i] * q[i])
+    end
+    return 2*acos(sumsqrt)
+end
+
+embed!(::ProbabilitySimplex, q, p) = copyto!(q, p)
+embed!(::ProbabilitySimplex, Y, p, X) = copyto!(Y, X)
```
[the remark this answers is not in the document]
mateuszbaran: This, I think, depends on whether we want an isometric embedding or not. This one is not isometric so scaling is not necessary (although I guess we may prefer an isometric embedding as a default for `ProbabilitySimplex`).

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,294 @@
...
+Compute a first order approximation by projection. The formula reads
+````math
+\operatorname{retr}^{-1}_p q = \bigl( I - \frac{1}{n}\mathbb{1}\mathbb{1}^\mathrm{T}\bigr)(\log(q)-\log(p))
+````
+where $\mathbb{1}$ is the column vector containing ones and $\log$ is applied elementwise.
+"""
+inverse_retract(::ProbabilitySimplex, ::Any, ::Any, ::SoftmaxInverseRetraction)
+
+function inverse_retract!(::ProbabilitySimplex{n}, X, p, q, ::SoftmaxInverseRetraction) where {n}
+    X .= (one(zeros(n+1,n+1)) - 1/n.*ones(n+1,n+1) ) * (log.(q) - log.(p))
```
mateuszbaran, line 206:
```suggestion
    logdiff = log.(q) .- log.(p)
    meanlogdiff = mean(logdiff)
    copyto!(X, logdiff)
    X .-= meanlogdiff
```
This looks like it gives the same result but is much faster.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: I've checked your version for 3-element vectors and it's about two times faster.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: Oh, it can actually be even faster.

src/manifolds/ProbabilitySimplex.jl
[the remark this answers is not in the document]
kellertuer: Currently it's the `DefaultEmbedding`, that does not enforce (or use) any property. If we scale correctly we could go to `IsometricEmbedding´ and not implement inner – but use the embedding fallback for that. I feel that here our inner is faster than the default.

I currently use the embedding only within the checks, since I can just pass `p` or `X` to the embedding to check its size and that it's real valued (since that's all that `Euclidean` checks)

So the Isometric one might be nice to have, but we have to overwrite inner anyways, so there wouldn't be much use in terms of lines of code.

I personally have to preference, since the checks I did would be invariant to the scale

src/manifolds/ProbabilitySimplex.jl
kellertuer: Cool! Thanks for checking; Yes, my version built whole matrices, that was fore sure not efficient.

src/manifolds/ProbabilitySimplex.jl
kellertuer: Ah, I already noticed that the log (numerically) might be wrong, thanks for checking.

kellertuer:
Despite an ambiguity with `retract!` (which might be due to the field parameter we introduced to the manifold), there seems to be mainly a `NaN` error lot for `inverse_retract!` and `log!` for the case that both points are equal. I will fix those in the evening. Thanks for all the speedup suggestions;
avoid the matrix was something I already saw, but the other allocations...nice!

kellertuer:
From the Current tests just the
* [ ] projection onto the tangent does not yet work, maybe due to scaling.
* [ ] The mean is sill missing
* [ ] we have to decide on the embedding.
* [ ] if that exists we could provide an injectivity radius.

mateuszbaran:
This should work for the injectivity radius:
```julia
function injectivity_radius(M, p)
    i = argmin(p)
    q = copy(p)
    q[i] = 0
    q ./= sum(q)
    return distance(M, p, q)
end
```
It follows the Seth's idea of projecting to the nearest boundary and calculating the distance. It seems to work. I could make it allocation-free by the way but this implementation makes it clearer what it's doing.

I don't really have a strong opinion about the embedding.

kellertuer:
Is that the exact injectivity radius (i.e. the maximum one)? Otherwise there might be a point a little closer that is already on the boundary and would yield a smaller radius – that would not be best, I would prefer a “safe” injectivity radius in the sense that it's a pessimistic approximation but within that ball you definetly have exp/log defined.

Concerning the embedding, I think, the isometric one might be neccesary in order to make our `project`s work, since otherwise (since the embedded p and the projected one differ in length due to nonisometric necessity of scaling) the tests we do for project (and the idea that project on the tangent should be the identity) fail.

mateuszbaran:
> Is that the exact injectivity radius (i.e. the maximum one)? Otherwise there might be a point a little closer that is already on the boundary and would yield a smaller radius – that would not be best, I would prefer a “safe” injectivity radius in the sense that it's a pessimistic approximation but within that ball you definetly have exp/log defined.

Yes, that should be pessimistic (up to numerical errors).

> Concerning the embedding, I think, the isometric one might be neccesary in order to make our `project`s work, since otherwise (since the embedded p and the projected one differ in length due to nonisometric necessity of scaling) the tests we do for project (and the idea that project on the tangent should be the identity) fail.

I think I've just fixed the tangent space projection anyway. That 4 in my formula comes from the fact that Fisher-Rao maps to a scaled sphere

mateuszbaran:
Although I'm not exactly sure it's the right projection yet.

kellertuer:
The tests run fine and your completely correct with the scaling in the inner product (to obtain orthogonality).

mateuszbaran:
I am? I was just starting to doubt my math there. I'm again thinking my coefficient near `sqrt.(p)` may not be right.

kellertuer:
You are currently implementing (2.9a) of https://arxiv.org/pdf/1810.06970.pdf (with the slight caveat that you also update/care about the scaling)

mateuszbaran:
But the scaling (by 4) should probably vanish actually. If I use the isometric embedding, I should also scale `sqrt.(p)` by 2 (which I'm not doing) and that would cancel with the 4 I've added. If I use the isometric-up-to-scaling embedding to the unit sphere then there would be no 4 anywhere.

kellertuer:
Oh indeed, the 4 is not required? Now I am surprised since I do not see what we did differently, but also without 4 it works. Now we just fight ambiguities with injectivity radius again (I tried and failed to resolve those)

mateuszbaran:
OK, this time I'm fairly sure my projection is right at least at the point with equal probabilities. I've added a test for it. I'm slightly confused how I should actually transform vectors at other points for projection.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -190,10 +190,14 @@ function injectivity_radius(M::ProbabilitySimplex, p)
     q ./= sum(q)
     return distance(M, p, q)
 end
+function injectivity_radius(M::ProbabilitySimplex)
+    error("The injectivity radius on $(M) depends on the point and has no closed form globally")
```
mateuszbaran, line 194: Isn't the global radius supposed to be the infimum of injectivity radii over all points of a manifold? Then it would just be 0 here.

kellertuer: Yes, we could do that, might kick a few algorithms off, so I thought maybe an error is nicer, but – well we could also take 0.

mateuszbaran: Yes, right, if an algorithm depends on the global injectivity radius and it's 0, the algorithm would likely fail anyway. If we go with throwing an error, we should adjust documentation for `injectivity_radius` explaining that it throws an error when it's 0.

kellertuer: Oh I already fixed it to return zero for now, since that's how its documented. Any algorithm depending on it being positive should check then, you're right.

mateuszbaran:
We could also add projection of points (at least for points with all positive coordinates).

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,352 @@
...
+    )
+
+Compute the Riemannian [`mean`](@ref mean(M::Manifold, args...)) of `x` using
+[`GeodesicInterpolationWithinRadius`](@ref), where the minimum of the injectivity radii
+of the points in `x` is used as the value `a`.
+"""
+mean(::ProbabilitySimplex, ::Any...)
+
+function mean!(M::ProbabilitySimplex, p, x::AbstractVector, w::AbstractVector; kwargs...)
+    a = a = min(injectivity_radius.(Ref(M),x)...)
```
sethaxen: As I mentioned in my comment, we don't need a radius here: https://github.com/JuliaManifolds/Manifolds.jl/pull/151#issuecomment-610866602

kellertuer:
But then with an error for negative/zero entries, since I don't want to “epsilon-relax-project”, since that is not a projection.

mateuszbaran:
Yes, sure, throwing an error when a coordinate is non-positive is reasonable.

src/manifolds/ProbabilitySimplex.jl
kellertuer: Ok.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,352 @@
...
+
+compute the injectivity radius on the [`ProbabilitySimplex`](@ref) `M` at the point `p`,
+i.e. the distanceradius to a point near/on the boundary, that could be reached by following the
+geodesic.
+"""
+function injectivity_radius(M::ProbabilitySimplex, p)
+    i = argmin(p)
+    q = copy(p)
+    q[i] = 0
+    q ./= sum(q)
```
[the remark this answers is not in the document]
mateuszbaran: The `project!` I've written also checks that coordinates are positive which wouldn't work here.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -307,12 +307,20 @@ project `Y` from the embedding onto the tangent space at `p` on
...
-\operatorname{proj}_{Δ^n}(p,Y) = pY -⟨p,Y⟩p.
+\operatorname{proj}_{Δ^n}(p,Y) = Y - ⟨p,Y⟩p.
 ````
 """
 project(::ProbabilitySimplex, ::Any, ::Any)
 
-function project!(::ProbabilitySimplex{n}, X, p, Y) where {n}
+function project!(::ProbabilitySimplex, q, p)
+    if any(x -> x <= 0, p)
+        throw(DomainError(p, "All coordinates of the projected point must be positive"))
```
kellertuer, line 317:
```suggestion
        throw(DomainError(p, "All coordinates of point from the embedding, that should be projected, must be positive, otherwise the projection is not well defined."))
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,371 @@
...
+i.e. the distanceradius to a point near/on the boundary, that could be reached by following the
+geodesic.
+"""
+function injectivity_radius(M::ProbabilitySimplex, p)
+    i = argmin(p)
+    q = copy(p)
+    q[i] = 0
+    q ./= sum(q)
+    return distance(M, p, q)
+end
```
sethaxen, line 192: Combining with `distance` and then simplifying produces this much simpler one. Also by not calling `distance`, we don't imply that the boundary point is in the `ProbabilitySimplex`: 
```suggestion
function injectivity_radius(::ProbabilitySimplex{n}, p) where {n}
    i = argmin(p)
    s = sum(p) - p[i]
    return 2 * acos(sqrt(s))
end
```

test/probability_simplex.jl
```diff
@@ -54,13 +54,5 @@ include("utils.jl")
         X2 = X .+ 10
         Y = project(M, p, X2)
         @test isapprox(M, p, X, Y)
-
-        X = log(M, q, p)
```
mateuszbaran: Why did you remove these tests?

kellertuer: I hope not, but I had a strange merge, which I hope I aborted and restarted :/

docs/make.jl
```diff
@@ -20,7 +23,9 @@ makedocs(
                 "Grassmann" => "manifolds/grassmann.md",
                 "Hyperbolic space" => "manifolds/hyperbolic.md",
                 "Lorentzian manifold" => "manifolds/lorentz.md",
+                "Multinoial matrices" => "manifolds/multinomial.md",
```
mateuszbaran, line 26:
```suggestion
                "Multinomial matrices" => "manifolds/multinomial.md",
```

kellertuer: More interestingly, the docs page is still empty...I was too fast I fear, still checking.

test/probability_simplex.jl
kellertuer: It's restored.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,371 @@
...
+
+Compute the Riemannian [`mean`](@ref mean(M::Manifold, args...)) of `x` using
+[`GeodesicInterpolationWithinRadius`](@ref), where the minimum of the injectivity radii
+of the points in `x` is used as the value `a`.
+"""
+mean(::ProbabilitySimplex, ::Any...)
+
+function mean!(M::ProbabilitySimplex, p, x::AbstractVector, w::AbstractVector; kwargs...)
+    a = minimum(injectivity_radius.(Ref(M), x))
+    return mean!(M, p, x, w, GeodesicInterpolationWithinRadius(a); kwargs...)
```
sethaxen, line 300: We don't need the radius check here, a noted in my earlier comment.

```suggestion
    return mean!(M, p, x, w, GeodesicInterpolation(); kwargs...)
```

kellertuer: Oh, I also had thought I did that earlier, I had some strange behaviour of my sourcetree where I seem to have lost a few lines before committing them. Will change that and update the docstring.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+This set is also called the unit simplex or standard simplex.
+
+The tangent space is given by
+````math
+T_pΔ^n = \biggl\{ X ∈ ℝ^{n+1}\ \big|\ ⟨\mathbb{1},X⟩ = \sum_{i=1}^{n+1} X_i = 0 \biggr\}
+````
+
+We employ this set and its tangent spaces with the [`FisherRaoMetric`](@ref) exploiting
+the isometry to the $n$-sphere of radius 2, i.e. a scaled version of the [`Sphere`](@ref).
+The corresponding diffoemorphism $\varphi\colon \mathbb Δ^n \to \mathcal N$,
```
sethaxen, line 20:
```suggestion
The corresponding diffeomorphism $\varphi: \mathbb Δ^n → \mathcal N$,
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+where $\mathcal N \subset 2𝕊^n$ is given by %\varphi(p) = 2\sqrt{p}$.
```
sethaxen, line 21:
```suggestion
where $\mathcal N \subset 2𝕊^n$ is given by $\varphi(p) = 2\sqrt{p}$.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+
+This implementation follows the notation in [^ÅströmPetraSchmitzerSchnörr2017]
```
sethaxen, line 23:
```suggestion
This implementation follows the notation in [^ÅströmPetraSchmitzerSchnörr2017].
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+
+This set is also called the unit simplex or standard simplex.
+
+The tangent space is given by
+````math
+T_pΔ^n = \biggl\{ X ∈ ℝ^{n+1}\ \big|\ ⟨\mathbb{1},X⟩ = \sum_{i=1}^{n+1} X_i = 0 \biggr\}
+````
+
+We employ this set and its tangent spaces with the [`FisherRaoMetric`](@ref) exploiting
+the isometry to the $n$-sphere of radius 2, i.e. a scaled version of the [`Sphere`](@ref).
```
sethaxen, line 19: These are technically 2 different approaches that happily coincide here. So we can say something like "The manifold is implemented assuming the Fisher-Rao metric for the multinomial distribution, which is equivalent to the induced metric from isometrically embedding the probability simplex in the $n$-sphere of radius 2."

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+    > arxiv: [1603.05285](https://arxiv.org/abs/1603.05285).
+"""
+struct ProbabilitySimplex{n} <: AbstractEmbeddedManifold{ℝ,DefaultEmbeddingType} end
+
+ProbabilitySimplex(n::Int) = ProbabilitySimplex{n}()
+
+"""
+    SoftmaxRetraction <: AbstractRetractionMethod
+
+Describes a retraction that is based on the Softmax function.
```
sethaxen, line 38: It's not usually capitalized
```suggestion
Describes a retraction that is based on the softmax function.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+"""
+struct SoftmaxRetraction <: AbstractRetractionMethod end
+
+"""
+    SoftmaxInverseRetraction <: AbstractInverseRetractionMethod
+
+Describes an inverse retraction that is based on the Softmax function.
```
sethaxen, line 45:
```suggestion
Describes an inverse retraction that is based on the softmax function.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+"""
+struct SoftmaxInverseRetraction <: AbstractInverseRetractionMethod end
+
+"""
+    FisherRaoMetric <: Metric
+
+The Fisher-Rao metric or Fisher information metric is a particular Riemannian metric which
```
sethaxen, line 52: We may want to move this to `MetricManifold.jl`, since it's not unique to the simplex.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,369 @@
...
+can be defined on a smooth statistical manifold, i.e., a smooth manifold whose points are
+probability measures defined on a common probability space.
+
+It is currently used in the discrete case for the [`ProbabilitySimplex`](@ref).
```
sethaxen, line 56: I wouldn't say this because it's too narrow. We can say instead "See for example the [`ProbabilitySimplex`](@ref)"

src/manifolds/Multinomial.jl
```diff
@@ -0,0 +1,93 @@
...
+    return check_tangent_vector(
+        PowerManifold(M.manifold, m),
+        p,
+        X;
+        check_base_point = check_base_point,
+        kwargs...,
+    )
+end
+
+get_iterator(M::MultinomialMatrices{n,m}) where {n,m} = 1:m
```
sethaxen, line 87: Would it make more sense to use `Base.OneTo(m)`? Maybe the compiler can do some optimizations if it knows the first element is 1.

kellertuer:
Will run the formatter at the end, is there a way to just run it on a few files? Last time I tried, I had to manually just check in the files I worked on.

src/manifolds/ProbabilitySimplex.jl
kellertuer: Maybe as soon as a second manifold uses this?

sethaxen:
> Will run the formatter at the end, is there a way to just run it on a few files? Last time I tried, I had to manually just check in the files I worked on.

Yeah, you can just pass those files to `format`. But ideally, every PR runs JuliaFormatter before merging so it only changes the added code. I'll eventually add a Linter to CI so it yells at us if we didn't run it.

src/manifolds/Multinomial.jl
kellertuer: I don't see an argument against this, so yes, I'll do that.

kellertuer:
Then I have to check the docs again how to best run it as one of the last commits (maybe also the one where we bump version).

src/manifolds/Multinomial.jl
mateuszbaran: I doubt it would make any noticeable difference in this case. The loops where it's used are relatively expensive anyway.

kellertuer: But it also does no harm, right?

mateuszbaran: Currently my number 1 idea for speeding up `PowerManifold` is marking more methods as `@inline` to inline them into these loops in `PowerManifold` so that views that I make there could be stack-allocated.

kellertuer: Sounds nice. For now I changed this line to @sethaxen|s suggestion, since I think it's at least not worse that `1:m`, and if we finde even better ways, we can change that further.

mateuszbaran: Yes, of course, there is definitely no harm in this.

src/manifolds/Multinomial.jl
```diff
@@ -0,0 +1,97 @@
...
+This yields exactly the same metric as
+considering the product metric of the probablity vectors, i.e. [`PowerManifold`](@ref) of the
+$(n-1)$-dimensional [`ProbabilitySimplex`](@ref).
+
+The [`ProbabilitySimplex`](@ref) is stored internally within `M.manifold`, such that all functions of
+[`AbstractPowerManifold`](@ref) can be used directly.
+
+# Constructor
+
+    MultinomialMatrices(n,m)
```
mateuszbaran, line 21:
```suggestion
    MultinomialMatrices(n, m)
```

src/manifolds/Multinomial.jl
```diff
@@ -0,0 +1,97 @@
...
+end
+
+MultinomialMatrices(n::Int, m::Int) =
+    MultinomialMatrices{n,m,n - 1}(ProbabilitySimplex(n - 1))
+
+^(M::ProbabilitySimplex{N}, m::Int) where {N} =
+    MultinomialMatrices{manifold_dimension(M) + 1,m,N}(M)
+
+@doc raw"""
+    check_manifold_point(M::MultinomialMatrices,p)
```
mateuszbaran, line 38:
```suggestion
    check_manifold_point(M::MultinomialMatrices, p)
```

kellertuer: Interesting, I even ran formatter, but thanks for noticing.

mateuszbaran: I think the formatter doesn't format docstrings.

kellertuer: Ah, thanks for noting that, I even got a little lazy, but will carefully still check docs then.

src/manifolds/Multinomial.jl
```diff
@@ -0,0 +1,97 @@
...
+MultinomialMatrices(n::Int, m::Int) =
+    MultinomialMatrices{n,m,n - 1}(ProbabilitySimplex(n - 1))
+
+^(M::ProbabilitySimplex{N}, m::Int) where {N} =
+    MultinomialMatrices{manifold_dimension(M) + 1,m,N}(M)
+
+@doc raw"""
+    check_manifold_point(M::MultinomialMatrices, p)
+
+Checks whether `p` is a valid point on the [`MultinomialMatrices`](@ref)`{m,n}` `M`, i.e. is a matrix
```
sethaxen, line 40:
```suggestion
Checks whether `p` is a valid point on the [`MultinomialMatrices`](@ref)`(m,n)` `M`, i.e. is a matrix
```

src/manifolds/Multinomial.jl
```diff
@@ -0,0 +1,97 @@
...
+    end
+    return check_manifold_point(PowerManifold(M.manifold, m), p; kwargs...)
+end
+@doc raw"""
+    check_tangent_vector(M::MultinomialMatrices p, X; check_base_point = true, kwargs...)
+
+Checks whether `X` is a valid tangent vector to `p` on the [`MultinomialMatrices`](@ref) `M`.
+This means, that `p` is valid, that `X` is of correct dimension and columnswise
+a tangent vector to the columns of `p` on the [`ProbabilitySimplex`](@ref).
+The optional parameter `check_base_point` indicates, whether to call [`check_manifold_point`](@ref)  for `p`.
```
sethaxen, line 60:
```suggestion
The optional parameter `check_base_point` indicates, whether to call
[`check_manifold_point`](@ref check_manifold_point(::MultinomialMatrices, ::Any))  for `p`.
```

src/manifolds/Multinomial.jl
```diff
@@ -0,0 +1,97 @@
...
+end
+
+get_iterator(M::MultinomialMatrices{n,m}) where {n,m} = Base.OneTo(m)
+
+@generated manifold_dimension(::MultinomialMatrices{n,m}) where {n,m} = (n - 1) * m
+
+@generated representation_size(::MultinomialMatrices{n,m}) where {n,m} = (n, m)
+
+show(io::IO, ::MultinomialMatrices{n,m}) where {n,m} =
+    print(io, "MultinomialMatrices($(n),$(m))")
```
sethaxen, line 97:
```suggestion
function show(io::IO, ::MultinomialMatrices{n,m}) where {n,m}
    print(io, "MultinomialMatrices($(n),$(m))")
end
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+    end
+    return nothing
+end
+
+"""
+    check_tangent_vector(M::ProbabilitySimplex, p, X; check_base_point = true, kwargs... )
+
+Check whether `X` is a tangent vector to `p` on the [`ProbabilitySimplex`](@ref) `M`, i.e.
+after [`check_manifold_point`](@ref)`(M,p)`, `X` has to be of same dimension as `p`
+and its elements have to sum to one.
```
sethaxen, line 97:
```suggestion
after [`check_manifold_point`](@ref check_manifold_point(::ProbabilitySimplex, ::Any))`(M,p)`,
`X` has to be of same dimension as `p` and its elements have to sum to one.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+The optional parameter `check_base_point` indicates, whether to call
+[`check_manifold_point`](@ref)  for `p` or not.
```
sethaxen, line 99:
```suggestion
`check_manifold_point`  for `p` or not.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+    end,
+)
+
+@doc raw"""
+    inner(M::ProbabilitySimplex,p,X,Y)
+
+Compute the inner product of two tangent vectors `X`, `Y` from the tangent space $T_pΔ^n$ at
+`p`. The formula reads
+````math
+g_p(X,Y) = \sum_{i=1}^{n+1}\frac{X_iY_i}{p}
```
sethaxen, line 214:
```suggestion
g_p(X,Y) = \sum_{i=1}^{n+1}\frac{X_iY_i}{p_i}
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+    end
+    return d
+end
+
+@doc raw"""
+    inverse_retract(M::ProbabilitySimplex, p, q, ::SoftmaxInverseRetraction)
+
+Compute a first order approximation by projection. The formula reads
+````math
+\operatorname{retr}^{-1}_p q = \bigl( I - \frac{1}{n}\mathbb{1}\mathbb{1}^\mathrm{T}\bigr)(\log(q)-\log(p))
```
sethaxen, line 230: 11^T is just too weird:
```suggestion
\operatorname{retr}^{-1}_p q = \bigl( I_{n+1} - \frac{1}{n}\mathbb{1}^{n+1,n+1} \bigr)(\log(q)-\log(p))
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+````
+where $\mathbb{1}$ is the column vector containing ones and $\log$ is applied elementwise.
```
sethaxen, line 232:
```suggestion
where $\mathbb{1}^{m,n}$ is the size `(m,n)` matrix containing ones, and $\log$ is applied elementwise.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+Returns the manifold dimension of the probability simplex in $ℝ^{n+1}$, i.e.
+````math
+    \dim_{Δ^n} = n.
+````
+"""
+manifold_dimension(::ProbabilitySimplex{n}) where {n} = n
+
+@doc raw"""
+    mean(
+        S::ProbabilitySimplex,
```
sethaxen, line 286:
```suggestion
        M::ProbabilitySimplex,
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -0,0 +1,373 @@
...
+        x::AbstractVector,
+        [w::AbstractWeights,]
+        method = GeodesicInterpolationWithinRadius();
```
sethaxen, line 289: So I _still_ think that we can just use `GeodesicInterpolation` here, as the entire simplex is included in the corresponding radius for the sphere of radius 2, unless there's a good reason not to?

sethaxen:
Oh, and maybe also increment the version number for a patch release?

src/manifolds/ProbabilitySimplex.jl
kellertuer: Oh, I misread your last request on that, will change this to the right form then

kellertuer:
bump is one of the last things I usually do but yes, will do that now, when working through your remarks.

src/manifolds/ProbabilitySimplex.jl
sethaxen: Thanks! Yeah, just plain `GeodesicInterpolation()` everywhere should do it.

PR #153, opened 2020-04-08
Simplify project! for Stiefel

dahong67:
Do simplifications of `project!` method discussed in #141 and mentioned in #142.

Key idea is that `s.U' * s.U` in the previous implementation should (up to numerical issues) be approximately `I` in which case the remaining code simplifies.

### DIFF
--- a/src/manifolds/Stiefel.jl
+++ b/src/manifolds/Stiefel.jl
@@ -235,7 +235,7 @@ project(::Stiefel, ::Any, ::Any)
 
 function project!(M::Stiefel, q, p)
     s = svd(p)
-    q .= s.U * s.V'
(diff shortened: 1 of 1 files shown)


sethaxen:
Better yet, now master should just work, so merge master in, and if it passes, I'll merge the PR.

sethaxen:
Oh nevermind, I see that the CI is running correctly after reopening, it's just that the checks also show the old failed CI because those have different names. Okay, we'll just see if it passes.

dahong67:
Wow, you are quite efficient. Thanks for resolving the testing stuff and merging! :)

kellertuer:
Thanks for helping and contributing :)

PR #154, opened 2020-04-08
Run JuliaFormatter, add JuliaFormatter workflow

sethaxen:
This essentially adds a linter to CI.

### DIFF
--- a/src/groups/group.jl
+++ b/src/groups/group.jl
@@ -235,9 +235,16 @@ function get_vector(M::Manifold, e::Identity, X, B::VeeOrthogonalBasis)
     return get_vector(M, e.p, X, B)
 end
 for MT in GROUP_MANIFOLD_BASIS_DISAMBIGUATION
-    eval(quote
-        @invoke_maker 1 Manifold get_vector(M::$MT, e::Identity, X, B::VeeOrthogonalBasis)
-    end)
+    eval(
+        quote
+            @invoke_maker 1 Manifold get_vector(
+                M::$MT,
+                e::Identity,
+                X,
+                B::VeeOrthogonalBasis,
+            )
+        end,
+    )
 end
 function get_vector!(M::AbstractGroupManifold, Y, e::Identity, X, B::VeeOrthogonalBasis)
     M != e.group && error("On $(M) the identity $(e) does not match to perform get_vector!")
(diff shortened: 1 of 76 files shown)


mateuszbaran:
What does it actually do? It just fails the test if formatting makes any changes?

sethaxen:
> What does it actually do? It just fails the test if formatting makes any changes?

That's right, yeah. And it runs it on the entire repo, not just the source code. It's not included in the CI workflow, so its status won't impact our badges; it's mostly there as a reminder to us before we merge PRs, but I enabled it for tags and master as for the CI.

sethaxen:
I'm also going to use this PR to troubleshoot why nightly failures are not tolerated.

sethaxen:
The current failure is due to JuliaFormatter not being idempotent on our `@invoke_maker` calls. Relevant issue: https://github.com/domluna/JuliaFormatter.jl/issues/218

src/groups/group.jl
```diff
@@ -225,25 +230,29 @@ function get_vector(M::AbstractGroupManifold, e::Identity, X, B::VeeOrthogonalBa
     return get_vector(decorated_manifold(M), e.p, X, B)
 end
 function get_vector(M::Manifold, e::Identity, X, B::VeeOrthogonalBasis)
-    M != e.group.manifold && error("On $(M) the identity $(e) does not match to perform get_vector.")
+    M != e.group.manifold &&
+    error("On $(M) the identity $(e) does not match to perform get_vector.")
```
mateuszbaran, line 234: This is one thing the formatter does I don't like: breaking `condition && something()` lines without indenting. Maybe we should just replace them with `if` in cases where they are too long?

sethaxen: Ah, yes, I think we've agreed before that way's better. Ideally in the future with the workflow, we'll all run JuliaFormatter before merging, and large runs like this where it's easy to miss these won't be necessary. If you open an issue on JuliaFormatter, they might agree and change the default.

sethaxen: I created an issue for this: https://github.com/domluna/JuliaFormatter.jl/issues/222

mateuszbaran: Thanks!

PR #155, opened 2020-04-08
Allow nightly failures on GitHub Actions

sethaxen:
By splitting into its own job

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI-nightly
 on:
   push:
     branches: [master]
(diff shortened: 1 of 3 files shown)


sethaxen:
Failures due to bugs introduced in FiniteDifferences v0.9.3. See https://github.com/JuliaDiff/FiniteDifferences.jl/issues/67

sethaxen:
One downside to GitHub actions is when the job is allowed to fail, then if it does fail, it is still marked as succeeding. One alternative would be to split the nightly tests into its own workflow, which would prevent it from marking the build as failing on our badge but would still mark the checks for a commit a failing if the nightly failed. Thoughts?

kellertuer:
I think the own-workflow approach would be best.

PR #160, opened 2020-04-09
Power/product metrics wrap other metrics

### LINKED ISSUE #125: Should PowerMetric store a metric inside?
I just wanted to write a small test for `det_local_metric` on a `MetricManifold{PowerManifold,PowerMetric}` , seee https://github.com/JuliaManifolds/Manifolds.jl/blob/161014340e55ef5cb1e9df00a28655209a18128a/src/manifolds/PowerManifold.jl#L240 and encountered a problem: The current version can't work, since the inner calls are never `MetricManifold`s.

In order for them to be, one has to create a `MatricManifold` of the powers internal `Manifold` and an according `Metric`.  But at that point we only have the `PowerMetric`. So I would like to propose to store a `Metric` inside the `PowerMetric` for exactely cases like this. What do you think? Like for the one manifold, one metric should be enough.

mateuszbaran:
This fixes #125.


src/manifolds/PowerManifold.jl
```diff
@@ -82,13 +82,17 @@ function PowerManifold(
 end
 
 @doc raw"""
-    PowerMetric <: Metric
+    PowerMetric(metric::metric)
```
kellertuer, line 85:
```suggestion
    PowerMetric{<:Metric} <: Metric
```

src/manifolds/PowerManifold.jl
```diff
@@ -82,13 +82,17 @@ function PowerManifold(
...
 
 Represent the [`Metric`](@ref) on an [`AbstractPowerManifold`](@ref), i.e. the inner
 product on the tangent space is the sum of the inner product of each elements
 tangent space of the power manifold.
+
+It is constructed based on the given `metric` on the powered space.
```
kellertuer, line 91: Could you the explicit constructor signature here?

src/manifolds/ProductManifold.jl
```diff
@@ -42,12 +42,22 @@ end
 
 const PRODUCT_BASIS_LIST_CACHED = [CachedBasis]
 
-"""
+@doc raw"""
     ProductMetric <: Metric
```
kellertuer, line 46:
```suggestion
    ProductMetric{<:(Tuple{N,Metric} where N)} <: Metric
```

src/manifolds/PowerManifold.jl
mateuszbaran: OK, done.

src/manifolds/ProductManifold.jl
```diff
@@ -42,12 +42,22 @@ end
 
 const PRODUCT_BASIS_LIST_CACHED = [CachedBasis]
 
-"""
-    ProductMetric <: Metric
+@doc raw"""
+    ProductMetric{<:(Tuple{N,Metric} where N)} <: Metric
+
+A type to represent the product of `metrics` for a [`ProductManifold`](@ref).
```
sethaxen, line 48: `metrics` here isn't a variable or type in this docstring, so it doesn't need the ticks
```suggestion
A type to represent the product of [`Metric`](@ref)s for a [`ProductManifold`](@ref).
```

src/manifolds/ProductManifold.jl
```diff
@@ -42,12 +42,22 @@ end
 
 const PRODUCT_BASIS_LIST_CACHED = [CachedBasis]
 
-"""
-    ProductMetric <: Metric
+@doc raw"""
+    ProductMetric{<:(Tuple{N,Metric} where N)} <: Metric
```
sethaxen, line 46: The parens aren't necessary

```suggestion
    ProductMetric{<:Tuple{N,Metric} where N} <: Metric
```

sethaxen:
Oh, also this is technically a breaking change, so we'll want to increment the version number accordingly after this. I'd like to get the distribution stuff in its own patch release, so I'll do that first.

mateuszbaran:
Sure, I can definitely wait with this and call this 0.2.0. By the way, now that I've done this, extending this power/product metrics to all metric-dependent functions (exp, log, inner etc.) looks like something that should also be done.

sethaxen:
> Sure, I can definitely wait with this and call this 0.2.0. By the way, now that I've done this, extending this power/product metrics to all metric-dependent functions (exp, log, inner etc.) looks like something that should also be done.

Hmm, is there a way we can use the default metric functionality to make this happen with just a few lines of code?

mateuszbaran:
> > Sure, I can definitely wait with this and call this 0.2.0. By the way, now that I've done this, extending this power/product metrics to all metric-dependent functions (exp, log, inner etc.) looks like something that should also be done.
> 
> Hmm, is there a way we can use the default metric functionality to make this happen with just a few lines of code?

As far as I'm aware, not without some changes to the default metric functionality itself. The more I think about it, the less I like this re-wrapping that needs to be done here. And it doesn't even really add new functionality, it's just another way to achieve something that is and will be possible anyway.

sethaxen:
> As far as I'm aware, not without some changes to the default metric functionality itself. The more I think about it, the less I like this re-wrapping that needs to be done here. And it doesn't even really add new functionality, it's just another way to achieve something that is and will be possible anyway.

I guess the main purpose is to enable one to equip a `ProductManifold` with a different product metric than the one that results from each of the (maybe default) metrics of the submanifold. That is, to avoid enforcing that if one wants to do this, that they need to equip each submanifold with a metric before taking the cartesian product. I like removing the order constraint, but since product metrics only make sense when they map one-to-one to each submanifold, it may after all be cleaner to enforce the order.

mateuszbaran:
Honestly, it feels like too much hassle for a minor convenience. I think I'll close the PR unless you or Ronny want to work on it.

sethaxen:
Sounds good to me.

kellertuer:
Maybe the question is: Do we need both orders (Product wraps metric and Metric wraps product) or is one enough? Both `ProductMetric` and `PowerMetric` aim to do the second approach (and I think that requires storing the metric internally).

Is it worth it? I don't know. In the beginning I thought, since you would store the metric just once (instead of within every product factor, for power you actually really only store it once within the  (metric) manifold). So maybe the second way (MetricManifold(Power)) is not the way we want to go, but just provide (Power(MetricManifold. I would be fine with that, though that would close this and #125 with “No, we don't do that”. I would be fine with that decision, too.

mateuszbaran:
OK, so then I'll close both this and #125 .

kellertuer:
Yes. I think, we can even remove `PowerMetric` and `ProductMetric` then?

mateuszbaran:
All these two metrics do is allowing people to wrap power and product manifolds in `MetricManifold`. I don't know how useful it is when it doesn't change the metric at all.

kellertuer:
Thats exactly my conclusion, I don't think its useful, even because you can't change the metric in the current form.

mateuszbaran:
OK, so we can remove it then.

sethaxen:
Last I checked some metric functions only worked for metric manifolds. If
that's still the case, then these metrics would still be necessary.

kellertuer:
No, you would put the metric on the one (for power) or the element-manifolds (for product) and then it works again, I think? But just to be sure we can also leave them in for ow and check more precisely wether they are useful (if you can't change the metric).

PR #161, opened 2020-04-09
Add more blue style consistency, more docs sections

sethaxen:
These options bring the style more in-line with Blue Style and document the format conventions.

This PR does not yet but will also bring us in-line with the using/import style conventions, which we currently do not follow: https://github.com/invenia/BlueStyle#module-imports.


kellertuer:
We could add a contribute section that link to the contributing md in the readme? And we could also add an authors section to the docs maybe? We could mention core team and contributors?

sethaxen:
> We could add a contribute section that link to the contributing md in the readme?

Better yet, we can load the contributing readme as a docs page. See the [preview docs](https://juliamanifolds.github.io/Manifolds.jl/previews/PR161/generated/contributing.html).

> And we could also add an authors section to the docs maybe? We could mention core team and contributors?

Are you thinking in `index.html` or as its own page? Maybe something like

## Core Developers:
- Seth Axen
- Mateusz Baran
- Ronny Bergmann

See all contributors [on GitHub](https://github.com/JuliaManifolds/Manifolds.jl/graphs/contributors).

kellertuer:
I think the index should welcome users and get them started and not directly mention our names – maybe literature (for starters), but not developers. I was more thinking of a `about.md` that is linked from the menu either after Home (so second entry) or quite late, maybe before Notations.

kellertuer:
Could we extend the docs action just a lilttle bit to get the link to preview directly like here https://github.com/climate-machine/CLIMA/pull/953/files#diff-654965a0f240efc62d1413a634170e97R28-R34 ?

src/Manifolds.jl
```diff
@@ -42,6 +9,7 @@ import ManifoldsBase:
     array_value,
     base_manifold,
     check_manifold_point,
+    check_manifold_point__transparent,
```
kellertuer, line 12: why do we have to explicitly import transparent functions?

src/Manifolds.jl
```diff
@@ -99,60 +64,68 @@ using HybridArrays
 using LinearAlgebra
 using LightGraphs
```
kellertuer, line 65: I thought that just `using Package` would bring all its public functions to local scope, so why do we do a using and then import specific ones separately again?

src/groups/group.jl
```diff
@@ -957,25 +976,25 @@ struct MultiplicationOperation <: AbstractGroupOperation end
 
 const MultiplicationGroup = AbstractGroupManifold{𝔽,MultiplicationOperation} where {𝔽}
 
-*(e::Identity{G}) where {G<:MultiplicationGroup} = e
-*(::Identity{G}, p) where {G<:MultiplicationGroup} = p
-*(p, ::Identity{G}) where {G<:MultiplicationGroup} = p
-*(e::E, ::E) where {G<:MultiplicationGroup,E<:Identity{G}} = e
-*(::Identity{<:MultiplicationGroup}, e::Identity{<:AdditionGroup}) = e
+Base.:*(e::Identity{G}) where {G<:MultiplicationGroup} = e
```
kellertuer, line 979: I wasn't aware of the `:` notation – and: We now prepend the package for every foreign function definition that we extend? Because then we might miss `Base.exp` and `Base.log` (mainly when defining more docstrings, since we implement `exp!` and `log!`)?

sethaxen: I don't follow. `exp!` and `log!` are unique to our package. We don't need to use `Base.exp` and `Base.log` when defining docstrings, as that doesn't extend the function. And if we accidentally implement an `exp` (without the prepend), the test suite raises an error asking us if we meant to import `Base.exp`. Is there another scenario where this could break things?

src/Manifolds.jl
sethaxen: I think you mean `using`? Yeah, that's a carry-over from the older code. Having now looked at LightGraphs, I see that `AbstractGraph` is exported, so we don't need to "use" it.

src/groups/group.jl
kellertuer: Ah, I wasn't aware, that this does not hold for docstrings.

src/Manifolds.jl
sethaxen: I don't know. These were imported in `MetricManifold.jl`, so I moved them here. If they're not necessary, I could remove them (one is explicitly extended).

kellertuer: So `exp!__intransparent` is imported but not extended, so that can be left out, but I don't recall why we did that with `inner`, but it doesn't look nice. I have to check whether that's necessary.

sethaxen:
> Currently a lot of functions gained their package prepended (norm for example) but other didn't (exp/log) and neither did the functions from `ManifoldsBase`. So what's the reasoning or rule w would like to follow?

BlueStyle recommends never importing functions for overloading. Instead, you prepend. This prevents accidentally overloading something, and it also makes it easier when reading the code to know where a function came from. See https://github.com/invenia/BlueStyle#module-imports.

I modified the BlueStyle standard here. Since we "own" ManifoldsBase, and its functions are meant to be exported here, we import them and then treat them internally as if they had been originally written in our package, then re-export them. That will make it easier to transition functions to ManifoldsBase.

Where aren't `exp` and `log` extended?

sethaxen:
> Could we extend the docs action just a lilttle bit to get the link to preview directly like here https://github.com/climate-machine/CLIMA/pull/953/files#diff-654965a0f240efc62d1413a634170e97R28-R34 ?

Nice idea! Certainly, I'll add it.

src/Manifolds.jl
kellertuer: and for `check_manifold_point__` I don't see where that's used?

src/Manifolds.jl
kellertuer: We could check for ManifoldsBase, too – since we use the full package, I think most specific function/type uses can be omitted?

kellertuer:
That sounds like a reasonable code style to follow, ok.

src/Manifolds.jl
sethaxen: Ah, I see what the issue is. `@decorator_transparent_fallback` creates a method with a name like `exp!__intransparent`, but since that particular function is also created by `@decorator_transparent_fallback` in ManifoldsBase, it needs to be imported. There must be a more hygienic  way to do this that doesn't require the user to know the internals of the macro in ManifoldsBase and whether the macro has been called there.

sethaxen: That's beyond the scope of this issue, so I'll leave the imports in.

kellertuer: Yes, leave it imported for now then. I didn't notice this when @mateuszbaran moved to changing the dispatch-per-symbol to different functions, but we should look for a – as you said – more hygenic way to do this and leave the internals “hidden”.

It might be that just for this one case its required and it might also be solved by moving MetricManifold to base (which I would actually favour, including moving basic metric types) – as well as maybe moving power and product... but thats all beyond this PR for sure ;)

kellertuer:
Can we maybe even do comment only once? Like codecov does?

sethaxen:
You mean build the docs only once, or only comment once? With the current implementation, it should on every commit:
1. Build the docs
2. Deploy the docs
3. If building the docs was successful and on a PR and has not commented before, post a comment with the link

As we discussed earlier, I plan to deactivate preview doc deployment by default on a future PR, using a label to activate it.

kellertuer:
The not-commented-before condition was, what I did not yet see and feared not being implemented – then – great!

Yes switchting that to not always do docs would also be nice, but not urgent.

sethaxen:
Any objections to merging this PR?

kellertuer:
Oh, we actually don't need the comment, Frederik just commented on Slack, that the last status links to the preview docs already... so we can revert the comment update.

PR #165, opened 2020-04-10
Fixing StaticArray tests

mateuszbaran:


mateuszbaran:
I'm not sure why one job failed, how to re-run it or why there seems to be no code coverage running here.

sethaxen:
> I'm not sure why one job failed
After googling it, it seems to be something that sometimes happens on os x. Should be rare, but we'll monitor and file a support ticket if it happens a lot

> how to re-run it (I see only re-run all option)
Yeah, looks like you can only re-run all

> or why there seems to be no code coverage running here.

Might be connected to the stalled job. I'd just re-run and then wait.

PR #168, opened 2020-04-14
get_component and set_component!

### LINKED ISSUE #167: Easier access to power and product manifold points/tangents
While currently porting Manopt, I noticed, that it might be nice to have easier access methods. As an example for something like

````julia
M = Stiefel(3,2)^(5,4)
````

and having `p` on the manifold, it would be nice to have something similar to `p[I]`, that we would do for arrays – especially independent of the internal representation (array or nested).

What do you think about overwriting `getindex` and similarly `setindex`?

I was thinking about something like
````julia
a = p[M,2,3]
p[M,3,3] = a
```` 
would be possible? For `PowerManifold`, I think that should be doable, for `Product` accessing more than one element might yield a slight problem with the reshaper?

* Is that too much of an abuse of the index-access notation?
* is that doable witth Julia? (I tried shortly but got lost in technical details).

What I would especially like, if such an access is independent of the structure (nested/array) for ease of use.

### PR DESCRIPTION
**Pull request by mateuszbaran**
Fixes #167.

### CODE REVIEW

File: src/manifolds/PowerManifold.jl [hunk 1]
```diff
@@ -725,6 +745,17 @@ function retract!(M::AbstractPowerManifold, q, p, X, method::PowerRetraction)
     return q
 end
 
+"""
+    set_component!(M::AbstractPowerManifold, q, p, idx...)
+
+Set the component of a point `q` on an [`AbstractPowerManifold`](@ref) `M` at index `idx`
+to `p`.
```
kellertuer, line 752:
```suggestion
to `p`, which itself is a point on the [`Manifold`](@ref) the power manifold is build on.
```

src/manifolds/ProductManifold.jl
```diff
@@ -871,6 +880,15 @@ function representation_size(M::ProductManifold)
     return (mapreduce(m -> prod(representation_size(m)), +, M.manifolds),)
 end
 
+"""
+    set_component!(M::ProductManifold, q, p, i)
+
+Set the `i`th component of a point `q` on a [`ProductManifold`](@ref) `M` to `p`.
```
kellertuer, line 886:
```suggestion
Set the `i`th component of a point `q` on a [`ProductManifold`](@ref) `M` to `p`, where `p` is a point on the [`Manifold`](@ref) this factor of the product manifold consists of.
```

src/product_representations.jl
```diff
@@ -207,8 +209,8 @@ submanifold_component(p, i::Integer) = submanifold_component(p, Val(i))
 Get the projected components of `p` on the submanifolds of `M`.
```
kellertuer, line 209:
```suggestion
Get the projected components of `p` on the submanifolds of `M`. The components are returned in a Tuple.
```

mateuszbaran:
I don't really understand what is going on with this code coverage. Anyway, it indeed does not appear as a check below comments in a PR (luckily we still have a comment with a report).

PR #169, opened 2020-04-15
A small PR to extend vector transport to product & power

kellertuer:
Which is very similar to how retract works on there. I still have to check for code coverage.


kellertuer:
Once this is done, the code of Manopt is ported and should work (still have to finish docs).

mateuszbaran:
Looks OK (except there are no tests for the new functionality right now).

> Once this is done, the code of Manopt is ported and should work (still have to finish docs).

That's great! 🎉

kellertuer:
I ran formatter and checked all files (actually nearly all changes?) in and...the check still says nearly all files are not formatted?

src/manifolds/Grassmann.jl
```diff
@@ -416,6 +416,20 @@ function uniform_distribution(M::Grassmann{n,k,ℝ}, p) where {n,k}
...
+@doc raw"""
+    vector_transport_to(M::Grassmann,p,X,q,::ProjectionTransport)
+
+compute the projection based transport on the [`Grassmann`](@ref) `M` by
+interpreting `X` from the tangent space at `p` as a point in the embedding and
+projecting it onto the tangent space at q.
+"""
+vector_transport_to(::Grassmann, ::Any, ::Any, ::Any, ::ProjectionTransport)
+
+function vector_transport_to!(M::Grassmann, Y, p, X, q, ::ProjectionTransport)
```
mateuszbaran, line 428: The `vector_transport_to!` method from ManifoldsBase should already cover this case I think?

kellertuer: Oh, maybe it does, then I was maybe just confused by the non-existing default. I'll check (we have a test for this now so removing it has to still pass).

kellertuer: Removing it yields a not-implemented error. I am not sure its implemented (but from the anti patterns post we should remove those own-errors anyways somewhen).

So for now we need that.

mateuszbaran: We do have a default implementation: https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/0f3c5933ab565fc0297c6e339e1bae5717f05e0c/src/ManifoldsBase.jl#L875 , just the `EmbeddedManifold` is `intransparent` for this transport method (I think it should be `parent`).

kellertuer: ah, maybe it should be transparent. But I would leave this version here as is until we have fixed manifolds base, after an update there, this still works but can be simplified.

mateuszbaran: By the way, I'm not sure if projection transport actually does the right thing for Lie groups. Let's take, for example, the circle group on a real plane with unit element (1, 0). Projection at (0, 1) performs the rotation by -pi/2. Meanwhile, projection transport should be just the identity function.

mateuszbaran: OK, sure, we can keep it for now.

kellertuer: For the circle group and sphere for far apart points – yes we don't want to have projection ... that's why I am maybe against a default.

kellertuer:
I am not sure where this one loses 2.14%, that's just a hiccup again I think.

mateuszbaran:
Yes, most likely a hiccup. It may fix itself in a while, like it did for my last PR.

src/manifolds/Grassmann.jl
mateuszbaran: I'm not sure we are on the same page here. The circle group was just an example that would apply to all Lie groups where we represent tangent vectors in the Lie algebra. I think we just can't really have the same function for Euclidean to Riemannian gradient conversion and projection vector transport. Or at least we need a projection type argument or something.

kellertuer: Yes, in that we are on the same page. For me that yields, that we should not do a generic ProjectionTransport implementation (the one we currently have), since it breaks for the Circle (without a parameter that ensures a valid definition). For that we would need a more specific projection transport type, yes.

mateuszbaran: OK. it also breaks for `Rotations` which leaves it with no valid vector transport at the moment. I'll open an issue.

PR #174, opened 2020-04-19
Modified constructor of PowerManifold

### LINKED ISSUE #170: What should ^ do on a PowerManifold?
Maybe it should just make a "flat" `PowerManifold` just with additional indices. Similarly for `PowerManifold` constructor.

mateuszbaran:
Fixes #170.

### DIFF
--- a/src/manifolds/PowerManifold.jl
+++ b/src/manifolds/PowerManifold.jl
@@ -64,21 +64,39 @@ By default, the [`ArrayPowerRepresentation`](@ref) of points
 and tangent vectors is used, although a different one, for example
 [`NestedPowerRepresentation`](@ref), can be given as the second argument to the
 constructor.
+When `M` is a `PowerManifold` (not any [`AbstractPowerManifold`](@ref)) itself, given
+dimensions will be appended to the dimensions already present, for example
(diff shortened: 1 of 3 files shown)


src/manifolds/PowerManifold.jl
```diff
@@ -64,22 +64,40 @@ By default, the [`ArrayPowerRepresentation`](@ref) of points
 and tangent vectors is used, although a different one, for example
 [`NestedPowerRepresentation`](@ref), can be given as the second argument to the
 constructor.
+When `M` is a `PowerManifold` (not any [`AbstractPowerManifold`](@ref)) itself, given
+dimensions will be appended to the dimensions already present, for example
+`PowerManifold(PowerManifold(Sphere(2), 2), 3)` is equivalent to
+`PowerManifold(Sphere(2), 2, 3)`. This feature preserves the representation of the inner
```
kellertuer, line 70: I like this completeness, not only in the docs but also, that it's possible to be overwritten.

mateuszbaran: Yes, the ability to override it comes for free thanks to `AbstractPowerManifold` :slightly_smiling_face: .

PR #177, opened 2020-04-28
Fixes for vector bundle of product and power manifolds

mateuszbaran:
Fix for a problem reported on Slack and a few related issues I've encountered while fixing that bug.


src/manifolds/PowerManifold.jl
```diff
@@ -839,6 +839,21 @@ function vector_bundle_transport(fiber::VectorSpaceType, M::PowerManifold)
     return PowerVectorTransport(ParallelTransport())
 end
 
+function vector_transport_direction(M::AbstractPowerManifold, p, X, d)
+    return vector_transport_direction(M, p, X, d, PowerVectorTransport(ParallelTransport()))
+end
+
+function vector_transport_direction!(M::AbstractPowerManifold, Y, p, X, d)
```
kellertuer, line 846: The `PowerVectorTransport` was used to avoid ambiguities, right?

mateuszbaran: Yes, exactly, there were many ambiguities otherwise.

kellertuer: Ok.

PR #179, opened 2020-04-29
Provide get/setindex for Power and Product manifold.

kellertuer:
This is still work in progress, but the `ProductManifold` should already work.

For the `PowerManifold` I would like to have arbitrary indexing (multidimensional), too, but I can't resolve the ambuguity: Since any tuple is valid, the ambiguity is with the original implementation, treating the manifold as an element of the tuple.


mateuszbaran:
Thanks for starting this. This `setindex!` and `getindex` need stricter type bounds to reduce ambiguities, I'll try that later.

mateuszbaran:
I've improved this code a bit. There are only 4 new ambiguities that look relevant and as far as I can see any code that could trigger them can't be correct anyway (correct as in doing something remotely reasonable).

These are somewhat unusual definitions and in their defense I can say that 1) there is no type piracy and 2) the important pieces of design were posted on Slack in dev-good-practice and no one has found any problems with it.

kellertuer:
This looks really neat now, thanks for helping.

PR #183, opened 2020-05-01
Extrinsic mean estimator

### LINKED ISSUE #77: Add the extrinsic mean estimator
It's just the mean in the ambient space and point projection. Not a particularly good estimator but it is fast and simple.

mateuszbaran:
I've made a simple implementation of the extrinsic mean estimator. Resolves #77 .

### DIFF
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -425,7 +425,6 @@ function Statistics.mean!(
     x::AbstractVector,
     w::AbstractVector,
     ::ExtrinsicEstimation;
-    p0 = x[1],
     kwargs...,
 )
     embedded_x = map(p -> embed(M, p), x)
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -19,6 +19,16 @@ Method for estimation using the cyclic proximal point technique.
 """
 struct CyclicProximalPointEstimation <: AbstractEstimationMethod end
 
+"""
+    ExtrinsicEstimation <: AbstractEstimationMethod
+
+Method for estimation in the ambient space and projecting to the manifold.
+
+For [`mean`](@ref) estimation, [`GeodesicInterpolation`](@ref) is used for mean estimation
+in the ambient space.
+"""
+struct ExtrinsicEstimation <: AbstractEstimationMethod end
+
 _unit_weights(n::Int) = StatsBase.UnitWeights{Float64}(n)
 
 @doc raw"""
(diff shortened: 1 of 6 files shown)


src/statistics.jl
```diff
@@ -409,6 +419,21 @@ end
     kwargs...,
 )
 
+function Statistics.mean!(
+    M::Manifold,
```
kellertuer, line 423:
```suggestion
    M:: AbstractEmbeddedManifold,
```
I think this only works for embedded manifolds, so we have to take that into account for this type.

mateuszbaran: I think we also want it to work for wrapped embedded manifolds (when the outermost thing is not `<:AbstractEmbeddedManifold`).

mateuszbaran:
Yes, sure, extrinsic median would also make sense.

src/statistics.jl
kellertuer: should we then maybe restrict it to abstract decorators? Or do you think that is too restrictive?

mateuszbaran: There doesn't seem to be any good reason for restricting it, the primary reason for restricting argument types should be polymorphic dispatch: https://white.ucc.asn.au/2020/04/19/Julia-Antipatterns.html and here there are no methods we could potentially be concerned about.

mateuszbaran:
Hmm... after thinking about it a bit, what would be the point of using the extrinsic median?

kellertuer:
> Hmm... after thinking about it a bit, what would be the point of using the extrinsic median?

I would ask the same for the mean ;) For both, if possible, I would prefer the intrinsic one, but the extrinsic one might be cheaper.

src/statistics.jl
kellertuer: ok, then let's leave this as is.

mateuszbaran:
The extrinsic mean can be much cheaper than the intrinsic one so there is at least one good reason. I'm not sure how much performance could be gained in the case of median, there doesn't seem to be a fast way to estimate the median in the Euclidean space anyway. Anyway, it looks like something for another PR.

kellertuer:
Sure we can do it in another PR.

The median is still reasonably cheap in the Euclidean space (sorting once basically) compared to manifolds (nonsmooth optimization, so not even a gradient descent applicable).

mateuszbaran:
Sorting only works in R^1, in R^d things are a bit more complicated: https://arxiv.org/abs/1606.05225 .

kellertuer:
Ah, you're right, still the multivariate Euclidean case should still be more efficient than the manifold-case; all those methods look better than doing our current method on manifolds (and I am not aware of a better one in general).

mateuszbaran:
Yes, these methods for R^d are very likely faster. The problem is I don't think any of our current dependencies implements any of these methods for multivariate Euclidean.

kellertuer:
Ah, I warnt aware of that, still; the proximal maps involved on Euclidean space are faster (just a plus) compared to the spherical ones, for example. So conceptually we could still do that,
and add a more clever on by dependency or from Julia Base, when it is available.

So conceptually I would still opt for an extrinsic median, but we can do that in a different PR and then this is fine.

mateuszbaran:
OK, I'll then open an issue for extrinsic median and cover that one last line that I've changed.

PR #185, opened 2020-05-07
Introduces the manifold of centered matrices

r-dornig:
Transfers the manifold of centered matrices from the Matlab package manopt ( see https://www.manopt.org/reference/manopt/manifolds/euclidean/centeredmatrixfactory.html) to Julia.


r-dornig:
I added a test tolerance for manifold points and it seems to fail for fixed rank matrices on Julia 1.4. But it works locally on my Julia 1.4.

.github/workflows/ci.yml
```diff
@@ -22,6 +22,4 @@ jobs:
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
       - uses: julia-actions/julia-uploadcodecov@latest
```
kellertuer, line 24:
```suggestion
      - uses: codecov/codecov-action@v1
```
sorry we have to also exchange the action as it seems.

PR #186, opened 2020-05-11
Don't upload to codecov on Julia nightly tests

mateuszbaran:


kellertuer:
Im I am not able to tag lines 10-12 in said yaml we are working on here but I would like to replace
```
    name: Julia nightly - ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
```
with
```
    name: Julia nightly - ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    continue-on-error: ${{ matrix.experimental }}
    strategy:
        fail-fast: false
```
and fail-fast is on the same indentation as matrix after that. That might yield a green check for all nighties, but all nightlies are always done and they do not change the global green check, see https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

kellertuer:
Ah, despite going bogus as so often, this seems to work for codecov.

mateuszbaran:
I've made the change you suggested. I've also tried enabling pushing to codecov only on a single configuration but this doesn't seem to work? We'll see.

kellertuer:
my proposal seems not to work, since currently nightly-ci is empty. Since your version pushed coverage, let's revert that test to your proposal of just not pushing coverage..

mateuszbaran:
OK, so do we merge this as-is?

kellertuer:
Hoping, that the code coverage is the usual hiccup, I'd say yes, let's merge.

PR #187, opened 2020-05-12
New vector transports

mateuszbaran:
Makes changes to support https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/40 .

TODO:

- [x] Remove Manifest after ManifoldsBase 0.9 is tagged.
- [x] Fix all tests.
- [x] Check documentation.
- [x] Look which manifolds could have special `mid_point` methods.
- [x] Add vector transport tests.

By the way, currently mean on the real circle is calculated using `sum` instead of `mean`. It doesn't look correct but apparently there are some tests.


kellertuer:
Note that the real circle's mean is a sum modulo 2π, that is the correct formula for the mean on the real circle.

mateuszbaran:
Could you link to some explanation of it? It would, sort of, work on the complex circle ( https://en.wikipedia.org/wiki/Mean_of_circular_quantities#Mean_of_angles ) but why would mean of angles 0.5 and 0.5 be 1?

kellertuer:
Maybe you're right and there's a `1/n` missing. I will look for a reference.

kellertuer:
I'm fine with the current state after a short check on code coverage. Maybe the spds (affine metric) can be optimized in mid point, one can check manopt for that (or pymanoth), they have named it `pairmean`.

mateuszbaran:
I didn't find much there, I think new method for `mid_point` can be added in another PR. Can I merge this now and tag 0.3?

kellertuer:
Nearly, just, why do we los 0.1% in coverage?

mateuszbaran:
Coverage does something weird again. I will add a few tests and it may help.

PR #188, opened 2020-05-13
fixes the mean for the real-valued circle.

kellertuer:


mateuszbaran:
Thanks! I've added mean on the complex circle and did some minor changes, does it look OK to you?

mateuszbaran:
I also don't know why mean on the real circle worked on `Array`s instead of `AbstractVectors` like other means.

kellertuer:
I noticed that for complex numbers, when the sum is zero, the mean is not defined.
When I tried to find a reference for that I found that a Matlab package for example does not care about that, but I also found the variance.

But then I noticed that the Statistics interface for variance is different from ours: `var` has the mean as a keyword, `varm` – so when I write the tests I might adapt that here as well (I think we even have some warnings on that for a while). @sethaxen would that be ok with you to adapt the interfaces to the interface as described here https://juliastats.org/StatsBase.jl/stable/scalarstats/#Statistics.var i.e. with the mean being a keyword? The circle does not need it anyways (actually `mean_and_var` might even be the more efficient way to implement this).

sethaxen:
> But then I noticed that the Statistics interface for variance is different from ours: `var` has the mean as a keyword, `varm`

Yeah, it does look like our `var` is more like their `varm` when it has mean as an arg. I feel like there was a reason I did this, but I can't think of a good one now, and we should definitely try to be consistent with the interfaces of the packages that define these methods.

> – so when I write the tests I might adapt that here as well (I think we even have some warnings on that for a while).

It would be better to handle this in its own PR, so that if we need to revert it at any point, we don't need to disentangle it from the other changes in this PR.

> @sethaxen would that be ok with you to adapt the interfaces to the interface as described here https://juliastats.org/StatsBase.jl/stable/scalarstats/#Statistics.var i.e. with the mean being a keyword?

I won't have time to work on this soon, but I agree that it would be a good thing to do and would endorse anyone else doing it. 🙂

kellertuer:
Don't worry I'll start that then next as soon as this PR is done, and it only needs a few tests for now (var here uses the old interface currently).

kellertuer:
I removed my variance from here too, since I will do that then in the next PR on reworking the stats functions. This now also covers tests for the new catched cases and hence it should be done.

PR #190, opened 2020-05-16
Proposes a tutorial on how to implement a manifold

kellertuer:
My aim is to maybe slowly extend this and write more tutorials in forthcoming PRs.

### DIFF
--- a/docs/src/tutorials/manifold.md
+++ b/docs/src/tutorials/manifold.md
@@ -68,7 +68,7 @@ Since the latter might be something we want to [dispatch](https://en.wikipedia.o
 
 ```@example manifold-tutorial
 """
-    MySphere{N} <: Manifold{ℝ}
+    MySphere{N} <: Manifold{ManifoldsBase.ℝ}
 
 Define an `n`-sphere of radius `r`. Construct by `MySphere(radius,n)
 """
--- a/docs/src/tutorials/manifold.md
+++ b/docs/src/tutorials/manifold.md
@@ -24,6 +24,7 @@ After that, we will
 * [implement](@ref manifold-tutorial-fn) two functions, the exponential map and the manifold dimension.
 
 ## [Technical Preliminaries](@id manifold-tutorial-prel)
+
 There is only two small  technical things we require.
 First of all our [`Manifold`](@ref)`{𝔽}` has a parameter `𝔽`.
 This parameter indicates the [`number_system`](@ref) the manifold is based on, e.g. real-valued (`ℝ`) or complex-valued (`ℂ`) data.
@@ -53,6 +54,7 @@ Long story short: if possible, implement the mutating vresion [`exp!`](@ref exp!
 Many functions that build upon basic functions employ the mutating variant, too, to avoid reallocations.
 
 ## [Startup](@id manifold-tutorial-startup)
+
 As a start, let's load `ManifoldsBase.jl` and import the functions we consider throughout this tutorial.
 For implementing a manifold, loading the interface should suffice for quite some time.
 
@@ -62,6 +64,7 @@ import ManifoldsBase: check_manifold_point, check_tangent_vector, manifold_dimen
 ```
 
 ## [Goal](@id manifold-tutorial-task)
+
 As an example, let's implement the sphere, but with a radius $r$. Since this radius is a property inherent to the manifold.
 The second information, we want to store is the dimension of the sphere, for example whether it's the 1-sphere, i.e. the circle, represented by vectors $p\in\mathbb R^2$ or the 2-sphere in $\mathbb R^3$.
 Since the latter might be something we want to [dispatch](https://en.wikipedia.org/wiki/Multiple_dispatch) on, we model it as a parameter of the type. Hence we define
--- a/docs/src/tutorials/manifold.md
+++ b/docs/src/tutorials/manifold.md
@@ -13,7 +13,8 @@ This tutorial demonstrates how to easily set your own manifold up within `Manifo
 
 If you looked around a little and saw the [interface](../interface.md), the amount of functions and possibilities, it might seem that a manifold might take some time to implement.
 This tutorial demonstrates that you can get your first own manifold quite fast and you only have to implement the functions you actually need. For this tutorial it would be helpful if you take a look at our [notation](../notation.md).
-For this tutorial it is helpful, if you heard of the exponential map, tangent vectors and the dimension of a manifold.
+This tutorial assumes that you heard of the exponential map, tangent vectors and the dimension of a manifold. If not, please read for example [[do Carmo, 1992](#doCarmo1992)],
+Chapter 3, first.
 
 We start with two technical preliminaries. If you want to start directly, you can [skip](@ref manifold-tutorial-task) this paragraph and revisit it for two of the implementation details.
 
@@ -65,9 +66,15 @@ import ManifoldsBase: check_manifold_point, check_tangent_vector, manifold_dimen
 
 ## [Goal](@id manifold-tutorial-task)
 
-As an example, let's implement the sphere, but with a radius $r$. Since this radius is a property inherent to the manifold.
+As an example, let's implement the sphere, but with a radius $r$.
+Since this radius is a property inherent to the manifold, it will become a field of the manifold.
 The second information, we want to store is the dimension of the sphere, for example whether it's the 1-sphere, i.e. the circle, represented by vectors $p\in\mathbb R^2$ or the 2-sphere in $\mathbb R^3$.
-Since the latter might be something we want to [dispatch](https://en.wikipedia.org/wiki/Multiple_dispatch) on, we model it as a parameter of the type. Hence we define
+Since the latter might be something we want to [dispatch](https://en.wikipedia.org/wiki/Multiple_dispatch) on, we model it as a parameter of the type.
+
+In general the `struct` of a manifold should provide information about the manifold, which are inherent to the manifold or has to be available without a specific point or tangent vector present.
+This is -- most prominently -- a way to determine the manifold dimension.
+
+For our example we define
 
 ```@example manifold-tutorial
 """
@@ -84,7 +91,6 @@ nothing #hide
 ```
 
 Here, the last line just provides a nicer print of a variable of that type
-The manifold should only contain information that it might require independently of its points or tangents.
 Now we can already initialize our manifold that we will use later, the $2$-sphere of radius $1.5$.
 
 ```@example manifold-tutorial
@@ -196,3 +202,17 @@ Sometimes a default implementation is provided; for example if you implemented [
 With [`log!`](@ref log!(M::Manifold, X, p, q)) and [`inner`](@ref inner(M::Manifold, p, X, Y)) you get the [`distance`](@ref distance(M::Manifold, p, q)), and so.
 
 In summary with just these few functions you can already explore the first things on your own manifold. Whenever a function from `Manifolds.jl` requires another function to be specifically implemented, you get a reasonable error message.
+
+## Literature
+
+```@raw html
+<ul>
+<li id="doCarmo1992">
+    [<a>doCarmo, 1992</a>]
+    M. P. do Carmo,
+    <emph>Riemannian Geometry</emph>,
+    Birkhäuser Boston, 1992,
+    ISBN: 0-8176-3490-8.
+</li>
+</ul>
+```
--- a/docs/src/tutorials/manifold.md
+++ b/docs/src/tutorials/manifold.md
@@ -75,7 +75,7 @@ Define an `n`-sphere of radius `r`. Construct by `MySphere(radius,n)
 struct MySphere{N} <: Manifold{ℝ} where {N}
     radius::Float64
 end
-MySphere(radius,n) = MySphere{n}(radius)
+MySphere(radius, n) = MySphere{n}(radius)
 Base.show(io::IO, M::MySphere{n}) where {n} = print(io, "MySphere($(M.radius),$n)")
 nothing #hide
 ```
--- a/docs/src/tutorials/manifold.md
+++ b/docs/src/tutorials/manifold.md
@@ -23,7 +23,7 @@ After that, we will
 * [implement](@ref manifold-tutorial-checks) two tests, so that points and tangent vectors can be checked for validity, for example also within [`ValidationManifold`](@ref),
 * [implement](@ref manifold-tutorial-fn) two functions, the exponential map and the manifold dimension.
 
-## [Technical Preliminaries](@id manifold-tutorial-prel)
+## [Technical preliminaries](@id manifold-tutorial-prel)
 
 There is only two small  technical things we require.
 First of all our [`Manifold`](@ref)`{𝔽}` has a parameter `𝔽`.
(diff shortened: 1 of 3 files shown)


docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+If you already have allocated memory to the variable that should store the result, it is better to perform the computations directly in that memory and avoid reallocations. For example
+
+```julia
+q = similar(p)
+exp!(M, q, p, X)
+```
+
+calls [`exp!`](@ref exp!(M::Manifold, q, p, X)), which modifies its input `q` and returns the resulting point in there.
+Actually these two lines are the default implementation for [`exp`](@ref exp(M::Manifold, p, X)).
+Note that for a unified interface, the manifold `M` is _always_ the first parameter, and the variable the result will be stored to in the mutating variants is _always_ the second parameter.
```
mateuszbaran, line 51: There will be one exception to this rule, specifically when a function takes another function as an argument, and it would make sense to write that function using the do-notation. Then `M` would be second and the mutating variable would be third.

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+For implementing a manifold, loading the interface should suffice for quite some time.
+
+```@example manifold-tutorial
+using ManifoldsBase, LinearAlgebra, Test
+import ManifoldsBase: check_manifold_point, check_tangent_vector, manifold_dimension, exp!
+```
+
+## [Goal](@id manifold-tutorial-task)
+
+As an example, let's implement the sphere, but with a radius $r$. Since this radius is a property inherent to the manifold.
```
mateuszbaran, line 68: The second sentence in this line is missing something.

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+struct MySphere{N} <: Manifold{ManifoldsBase.ℝ} where {N}
+    radius::Float64
+end
+MySphere(radius, n) = MySphere{n}(radius)
+Base.show(io::IO, M::MySphere{n}) where {n} = print(io, "MySphere($(M.radius),$n)")
+nothing #hide
+```
+
+Here, the last line just provides a nicer print of a variable of that type
+The manifold should only contain information that it might require independently of its points or tangents.
```
mateuszbaran, line 87: Might be worth adding that it should always be possible to determine the dimension of a manifold based on the manifold object only.

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+Now we can already initialize our Manifold, we use later, the $2$-sphere of radius $1.5$.
+
+```@example manifold-tutorial
+S = MySphere(1.5,2)
```
mateuszbaran, line 91:
```suggestion
S = MySphere(1.5, 2)
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+    radius::Float64
+end
+MySphere(radius, n) = MySphere{n}(radius)
+Base.show(io::IO, M::MySphere{n}) where {n} = print(io, "MySphere($(M.radius),$n)")
+nothing #hide
+```
+
+Here, the last line just provides a nicer print of a variable of that type
+The manifold should only contain information that it might require independently of its points or tangents.
+Now we can already initialize our Manifold, we use later, the $2$-sphere of radius $1.5$.
```
mateuszbaran, line 88:
```suggestion
Now we can already initialize our manifold that we will use later, the $2$-sphere of radius $1.5$.
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+```@example manifold-tutorial
+S = MySphere(1.5,2)
+```
+
+## [Checking points and tangents](@id manifold-tutorial-checks)
+
+If we have now a point, represented as an array, we would first like to check, that it is a valid point on the manifold.
+For this one can use the easy interface [`is_manifold_point`](@ref is_manifold_point(M::Manifold, p; kwargs...)). This internally uses [`check_manifold_point`](@ref check_manifold_point(M, p; kwargs...)).
+This is what we want to implement.
+We have to return the error, if `p` is not on `M` and `nothing` otherwise.
```
mateuszbaran, line 99:
```suggestion
We have to return the error if `p` is not on `M` and `nothing` otherwise.
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+
+We have to check two things: that a point `p` is a vector with `N+1` entries and it's norm is the desired radius.
+To spare a few lines, we can use [lazy evaluation](https://en.wikipedia.org/wiki/Lazy_evaluation) instead of `if` statements.
```
mateuszbaran, line 102:
```suggestion
To spare a few lines, we can use [short-circuit evaluation](https://docs.julialang.org/en/v1/manual/control-flow/#Short-Circuit-Evaluation-1) instead of `if` statements.
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+
+An implementation of the mutation version, see the [technical note](@ref manifold-tutorial-prel), reads
+
+```@example manifold-tutorial
+function exp!(M::MySphere{N}, q, p, X) where {N}
+    nX = norm(X)
+    if nX == 0
+        q .= p
+    else
+        q.= cos(nX/M.radius)*p + M.radius*sin(nX/M.radius) .* (X./nX)
```
mateuszbaran, line 171:
```suggestion
        q .= cos(nX/M.radius)*p + M.radius*sin(nX/M.radius) .* (X./nX)
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,198 @@
...
+* [model](@ref manifold-tutorial-task) the manifold
+* [implement](@ref manifold-tutorial-checks) two tests, so that points and tangent vectors can be checked for validity, for example also within [`ValidationManifold`](@ref),
+* [implement](@ref manifold-tutorial-fn) two functions, the exponential map and the manifold dimension.
+
+## [Technical preliminaries](@id manifold-tutorial-prel)
+
+There is only two small  technical things we require.
+First of all our [`Manifold`](@ref)`{𝔽}` has a parameter `𝔽`.
+This parameter indicates the [`number_system`](@ref) the manifold is based on, e.g. real-valued (`ℝ`) or complex-valued (`ℂ`) data.
+This can be used to simultaneously implement both a real-valued and a complex-valued type of a manifold.
```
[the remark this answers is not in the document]
kellertuer: If you have `Circle{R}` and `Circle{C}` that's easier to do a common implementation for some functions compared to introducing a common abstract type.

That might not have been, how we _got_ to that parametrization, but it's a good reason anyways.

docs/src/tutorials/manifold.md
kellertuer: But currently we don't have that.

docs/src/tutorials/manifold.md
kellertuer: Maybe I phrase that even further, when introducing the struct, which thing the struct should be able to do.

docs/src/tutorials/manifold.md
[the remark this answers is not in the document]
mateuszbaran: Well, you can have a number system parameter in manifold types without having it as a parameter of `Manifold`. So a number system parameter of, let's say, `Symmetric` is there to avoid adding another abstract type while a number system parameter of the `Manifold` type it derives from is there to have `number_of_coordinates` for free.

`Circle` may not be the best example here because it overwrites `number_of_coordinates`.

docs/src/tutorials/manifold.md
mateuszbaran: Yes, we don't have that yet on the master branch but it's already a part of #110 . We can of course modify tutorial in that PR then.

kellertuer: Maybe changing it, when necessary might be best. And it stays a mostly or usually, still.

docs/src/tutorials/manifold.md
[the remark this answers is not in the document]
kellertuer: Having it as a parameter was a choice, yes. Still I think the `number_of_coordinates` is not be the main reasons for our manifolds (it was to pull it up to Manifold, yes).

In general our manifolds have this parameter for easy dispatch per field type. So what do you propose to write in the tutorial?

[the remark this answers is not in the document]
mateuszbaran: In a basic tutorial like this I think it would be enough to say that for real manifolds users need to put `ManifoldsBase.ℝ` there and not worry about it, and for complex manifolds tell the reader that it has important interactions with bases.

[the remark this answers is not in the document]
kellertuer: So I could just leave out line 31? That would be fine with me.

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,218 @@
...
+* [model](@ref manifold-tutorial-task) the manifold
+* [implement](@ref manifold-tutorial-checks) two tests, so that points and tangent vectors can be checked for validity, for example also within [`ValidationManifold`](@ref),
+* [implement](@ref manifold-tutorial-fn) two functions, the exponential map and the manifold dimension.
+
+## [Technical preliminaries](@id manifold-tutorial-prel)
+
+There are only two small technical things we need to explain at this point.
+First of all our [`Manifold`](@ref)`{𝔽}` has a parameter `𝔽`.
+This parameter indicates the [`number_system`](@ref) the manifold is based on, e.g. real-valued (`ℝ`) or complex-valued (`ℂ`) data.
+This can be used to simultaneously implement both a real-valued and a complex-valued type of a manifold.
```
mateuszbaran, line 32:
```suggestion
This parameter indicates the [`number_system`](@ref) the manifold is based on, for example `ℝ` for real manifolds. It is important primarily for defining bases of tangent spaces.
See [`Symmetric`](@ref) as an example of defining both a real-valued and a complex-valued symmetric manifolds using one type.
```
Maybe something like this?

docs/src/tutorials/manifold.md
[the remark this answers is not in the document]
mateuszbaran: I have added a suggested change below.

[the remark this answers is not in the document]
kellertuer: Thanks for the careful review.

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,218 @@
...
+In general the `struct` of a manifold should provide information about the manifold, which are inherent to the manifold or has to be available without a specific point or tangent vector present.
+This is -- most prominently -- a way to determine the manifold dimension.
+
+For our example we define
+
+```@example manifold-tutorial
+"""
+    MySphere{N} <: Manifold{ℝ}
+
+Define an `n`-sphere of radius `r`. Construct by `MySphere(radius,n)
```
mateuszbaran, line 83:
```suggestion
Define an `n`-sphere of radius `r`. Construct by `MySphere(radius,n)`
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,218 @@
...
+[isapprox(p,-q); is_manifold_point(S,q)]
+```
+
+## [Conclusion](@id manifold-tutorial-outlook)
+
+You can now just continue implementing further functions from the [interface](../interface.md),
+but with just [`exp!`](@ref exp!(M::Manifold, q, p, X)) you for example already have
+
+* [`geodesic`](@ref geodesic(M::Manifold, p, X)) the (not necessarily shortest) geodesic emanating from `p` in direction `X`.
+* the [`ExponentialRetraction`](@ref), that the [`retract`](@ref retract(M::Manifold, p, X))ion uses by default.
```
mateuszbaran, line 197:
```suggestion
* the [`ExponentialRetraction`](@ref), that the [`retract`](@ref retract(M::Manifold, p, X)) function uses by default.
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,218 @@
...
+
+For the [`shortest_geodesic`](@ref shortest_geodesic(M::Manifold, p, q)) the implementation of a logarithm [`log`](@ref ManifoldsBase.log(M::Manifold, p, q)), again better a [`log!`](@ref log!(M::Manifold, X, p, q)) is necessary.
+
+Sometimes a default implementation is provided; for example if you implemented [`inner`](@ref inner(M::Manifold, p, X, Y)), the [`norm`](@ref norm(M, p, X)) is defined. You should overwrite it, if you can provide a more efficient version, gut for a start, the default should suffice.
```
mateuszbaran, line 201:
```suggestion
Sometimes a default implementation is provided; for example if you implemented [`inner`](@ref inner(M::Manifold, p, X, Y)), the [`norm`](@ref norm(M, p, X)) is defined. You should overwrite it, if you can provide a more efficient version. For a start the default should suffice.
```

docs/src/tutorials/manifold.md
```diff
@@ -0,0 +1,218 @@
...
+## [Technical preliminaries](@id manifold-tutorial-prel)
+
+There are only two small technical things we need to explain at this point.
+First of all our [`Manifold`](@ref)`{𝔽}` has a parameter `𝔽`.
+This parameter indicates the [`number_system`](@ref) the manifold is based on, for example `ℝ` for real manifolds. It is important primarily for defining bases of tangent spaces.
+See [`Symmetric`](@ref) as an example of defining both a real-valued and a complex-valued symmetric manifolds using one type.
+
+Second, a main design decision of `Manifold.jl` is that most functions are implemented as mutating functions, i.e. as in-place-computations. There usually exists a non-mutating version that falls back to allocating memory and calling the mutating one. This means you only have to implement the mutating version, _unless_ there is a good reason to provide a special case for the non-mutating one, i.e. because in that case you know a far better performing implementation.
+
+Let's look at an example. The exponential map $\exp_p\colon T_p\mathcal M \to \mathcal M$ that maps a tangent vector $X\in T_p\mathcal M$ from the tangent space at $p\in \mathcal M$ to the manifold.
```
mateuszbaran, line 36:
```suggestion
Let's look at an example. The exponential map $\exp_p\colon T_p\mathcal M \to \mathcal M$ that maps a tangent vector $X\in T_p\mathcal M$ from the tangent space at $p\in \mathcal M$ to the manifold. You don't have to implement it for your manifold but it is very useful.
```

kellertuer: What is the aim of this new sentence? It's anyways meant as an example, so I don't feel its necessary to state that it's a useful function (and sometimes you even can't implement it – in closed form).

mateuszbaran: Primarily I wouldn't want to leave the reader with an impression that the exponential map is necessary. Maybe it's poorly worded but at the moment it's somewhat unclear which functions are necessary for which functionality. Although that may be an idea for a different part of documentation. What do you think?

kellertuer: You're right, that's a good point, but I'd like to add a paragraph in the introduction then; will think about one. I think here, also the place is not best, since its in the technical details section.

mateuszbaran: Good idea, a paragraph in the introduction would be better 👍 .

kellertuer:
Thanks, I'll merge this so it's on dev and with the next feature it'll be on the main docs then.

PR #192, opened 2020-05-27
adds a small security check to sphere log.

kellertuer:


src/manifolds/Sphere.jl
```diff
@@ -290,8 +290,9 @@ log(::AbstractSphere, ::Any...)
 
 function log!(M::AbstractSphere, X, p, q)
     cosθ = real(dot(p, q))
+    cosθ = abs(cosθ) > 1 ? sign(cosθ) * one(cosθ) : cosθ
```
sethaxen, line 293: Since now you have upper and lower bounds together, we could just use `clamp`:

```suggestion
    cosθ = clamp(real(dot(p, q)), -1, 1)
```

src/manifolds/Sphere.jl
```diff
@@ -290,8 +290,9 @@ log(::AbstractSphere, ::Any...)
...
     if cosθ ≈ -1 # appr. opposing points, return deterministic choice from set-valued log
-        fill!(X, 0)
+        fill!(X, 0.0)
```
sethaxen, line 295: This change should not be necessary, since 0 will be promoted as necessary.

kellertuer: ah, ok, thanks, I wasn't sure (or confused in between).

src/manifolds/Sphere.jl
kellertuer: > What does this PR change?

It adds the check for -1/clamp for the acos (and/or approx)? At least I ran once into the situation, where the approx did not trigger (let's say for the extreme case of `-1.1`) and I ended up in the second case. So for security I would like to clamp upfront.

I also had checks in the second case, that for `norm(p-q)` small or  `X` small one directly returns the zero-vector, too, but I could not get that stably to run and it does not seem necessary, so only the clamp stayed.

kellertuer: (i am not sure why it was added here when I said, quote in reply and it added it here?)

sethaxen: > It adds the check for -1/clamp for the acos (and/or approx)? At least I ran once into the situation, where the approx did not trigger (let's say for the extreme case of `-1.1`) and I ended up in the second case. So for security I would like to clamp upfront.

Ah, good point! Okay, this is fine then.

PR #193, opened 2020-06-20
Bug in check_manifold_point for FixedRankMatrices

GillesBareilles:
Hi,
I think there's a small bug in the `check_manifold_point` method for `FixedRankMatrices`. I think these simple changes are enough to fit the manifold definition exactly

The error can be retrieved by doing
```julia
M = FixedRankMatrices{5, 6, 3, ℝ}()
x = SVDMPoint(rand(5, 6), 3)
check_manifold_point(M, x)
```

### DIFF
--- a/test/fixed_rank.jl
+++ b/test/fixed_rank.jl
@@ -2,8 +2,10 @@ include("utils.jl")
 
 @testset "fixed Rank" begin
     M = FixedRankMatrices(3, 2, 2)
+    M2 = FixedRankMatrices(3, 2, 1)
     Mc = FixedRankMatrices(3, 2, 2, ℂ)
     x = SVDMPoint([1.0 0.0; 0.0 1.0; 0.0 0.0])
+    x2 = SVDMPoint([1.0 0.0; 0.0 1.0; 0.0 0.0],1)
     v = UMVTVector([0.0 0.0; 0.0 0.0; 1.0 1.0], [1.0 0.0; 0.0 1.0], zeros(2, 2))
     @test repr(M) == "FixedRankMatrices(3, 2, 2, ℝ)"
     @test repr(Mc) == "FixedRankMatrices(3, 2, 2, ℂ)"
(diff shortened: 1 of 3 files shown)


kellertuer:
You'Re right, thanks for noticing :)
Could you add a small test to check this specific case in `test/fixed_rank_matrices`?
I see that I missed that case, since I just wrote a test for `n=k`.

GillesBareilles:
My pleasure :)
Would these changes do it ?

test/fixed_rank.jl
```diff
@@ -2,8 +2,10 @@ include("utils.jl")
 
 @testset "fixed Rank" begin
     M = FixedRankMatrices(3, 2, 2)
+    M2 = FixedRankMatrices(3, 2, 1)
     Mc = FixedRankMatrices(3, 2, 2, ℂ)
     x = SVDMPoint([1.0 0.0; 0.0 1.0; 0.0 0.0])
+    x2 = SVDMPoint([1.0 0.0; 0.0 1.0; 0.0 0.0],1)
```
kellertuer, line 8:
```suggestion
    x2 = SVDMPoint([1.0 0.0; 0.0 1.0; 0.0 0.0], 1)
```

GillesBareilles:
Oh right, I forgot about the formatting (nice thing its integrated to github!).
There is still one test not passing, although it seems to concern the Grassman manifold (and only on one platform?)

kellertuer:
Yes, we had still to check that nightlies are allowed to fail without making the global test indicator indicate a fail. They are already allowed to fail without canceling other tests (if a non-nightly test fails, all other tests are stopped).

If you want, you can already bump the version number to `0.3.3`, otherwise I will do that after merging before registering the new (bug fixed) version.

GillesBareilles:
Ok, thanks for the explanation!

So to change the version number, I guess I should edit the corresponding entry of `Project.toml`. Is there something else to be done?

kellertuer:
Yes its just the entry in the `Project.toml`.

GillesBareilles:
Thanks for the help!

PR #194, opened 2020-06-23
Modify CI nightly fail policy

kellertuer:
set nightly ci runs to be allowed to fail without making the complete workflow fail.


kellertuer:
I am a little surprised that this still yields a failing global indicator, since this

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

states otherwise. I will try to find a way, such that in the list they may fail while the global icon should stay green (for nightlies).

kellertuer:
Strangely this does not seem to work as I hoped for – so should we close this until we find a way for the nightly not to affect the “global green light”?

mateuszbaran:
It doesn't work currently so we can either keep this PR open to remind us that it should be resolved or close it and open another PR to try a new solution. Your call :slightly_smiling_face: .

sethaxen:
Everything should already be set up how it should be, as in, the nightly CI and the regular CI are separate actions, and the badge is set to only show the status of regular CI. Indeed, I'm pretty sure that in the past, that's what it did. But it looks like now GitHub uses the name CI to [search the actions](https://github.com/JuliaManifolds/Manifolds.jl/actions?query=workflow%3ACI+branch%3Amaster) and returns the status of all hits, which now includes CI-nightly. The solution is to change the name of one of the workflows so searching the name of one doesn't fetch the other, then update the badge accordingly if necessary.

sethaxen:
The problem with `continue_on_error` is that currently it marks whole tests as pass, and you need to actually open the log to find out there's a failure.

kellertuer:
Well currently if nightly fails, our global icon is fail, see the last commit on master. `continue-on-error` just means that with one test (in nightly) failing, the other nightly tests are still started. This is for example not the case for the non-nightly tests.

sethaxen:
And I believe the reason it's not currently working is that you need to designate specific step with `continue_on_error`, not whole jobs.

sethaxen:
Yes, because the badge is set to return the results of a search right now. The search is for actions with the name CI, and CI-nightly contains that name. Hence changing the name of the CI-nightly workflow (e.g. to just "nightly") will fix this.

kellertuer:
Ah, interesting, something that I learned today :) 

Though I also meant the “badge” / icon after a commit or PR, but that will always show a fail if it failed, right? So I might adopt this PR to just change the Badge then.

kellertuer:
This is now just a minor change, namely for the badge – but I think it can be merged.

sethaxen:
Is it the filename of the workflow or the name of the workflow that is used for the badge?

kellertuer:
I renamed the nightly workflow to read `nightly.yml` so the badge uses still the `ci`-selector as before.

sethaxen:
Which motivates my question. Is the `CI` selector hitting the nightly CI because the filename is `ci-nightly.yml` or because the name of the workflow _in_ the file is `CI-nightly`?

kellertuer:
Ah, thanks for noticing, then renaming the workflow (just to be safe) should also be done, you're right, since I don't know which of the two it checks

sethaxen:
I think so, yeah. Once that change is made, I agree it looks good to merge.

PR #195, opened 2020-07-10
Gradients and differentials

mateuszbaran:
I have extracted gradients and differentials from PR #110 so that this part can be finished more quickly.


mateuszbaran:
So, this is just the first part but it works and is usable for calculation of gradients and differentials (of curves). I guess the performance could be improved in some places but it should be good enough for now.

Some things defined here should eventually be moved to ManifoldsBase.jl.

src/differentiation.jl
```diff
@@ -0,0 +1,139 @@
...
+"""
+abstract type AbstractDiffBackend end
+
+struct NoneDiffBackend <: AbstractDiffBackend end
+
+
+"""
+    _derivative(f, t[, backend::AbstractDiffBackend])
+
+Compute the derivative of a callable `f` at time `t` computed using the given `backend`,
```
kellertuer, line 16: We could add the formulae here, i.e. for a function $f\colon I \to \mathcal M$, where $I$ is some interval containing $t$, this computes $Df(t)$ and store that in `X`? And I think it is not (yet) the covariant one since this is the file on differentiation, i.e. in the embedding?

src/riemannian_diff.jl
```diff
@@ -0,0 +1,232 @@
...
+differential(::Any, ::Curve, ::Real, ::AbstractRiemannianDiffBackend)
+
+
+"""
+    gradient(f::Any, rf::RealField, p, backend::AbstractRiemannianDiffBackend = rgradient_backend())
+
+Compute the Riemannian gradient of a real field of type `rf` represented by function `f`
+at point `p` using the given backend.
+"""
+gradient(::Any, ::RealField, ::Any, ::AbstractRiemannianDiffBackend)
```
kellertuer, line 25: Does this also require the manifold? I would've expected that. Ah Maybe not in general for any field, but for a real field (which would be my main application I think) using

`gradient(M,f,p,backend)`

would be my favourite.

src/riemannian_diff.jl
```diff
@@ -0,0 +1,232 @@
...
+    ) <: AbstractRiemannianDiffBackend
+
+Riemannian differentiation based on differentiation in an [`AbstractOrthonormalBasis`](@ref)
+`basis` using specified `retraction`, `inverse_retraction` and using backend `diff_backend`.
+"""
+struct RiemannianONBDiffBackend{
+    TADBackend<:AbstractDiffBackend,
+    TRetr<:AbstractRetractionMethod,
+    TInvRetr<:AbstractInverseRetractionMethod,
+    TBasis<:Union{
```
kellertuer, line 58: Why store both? The cached basis “knows” its internal ONB?

src/riemannian_diff.jl
```diff
@@ -0,0 +1,232 @@
...
+
+[^Absil2008]:
+    > Absil, P. A., et al. Optimization Algorithms on Matrix Manifolds. 2008.
+"""
+struct RiemannianProjectionGradientBackend{TADBackend<:AbstractDiffBackend} <:
+       AbstractRiemannianDiffBackend
+    diff_backend::TADBackend
+end
+
+function gradient(f, ft::RealField, p, backend::RiemannianProjectionGradientBackend)
```
kellertuer, line 221: I wanted to ask about this approach already in the last variant, but yes, this is quite nice and general.

src/differentiation.jl
mateuszbaran: Yes, this is actually just a unified way to call plain Euclidean differentiation methods (AD, finite differences), not Riemannian AD. I will add a comment about it to the documentation.

src/riemannian_diff.jl
mateuszbaran: The manifold is currently stored in `RealField`. Anyway, maybe you are right and I'm trying too hard to make `gradient` compatible with the do-notation.

By the way, do you think `jacobian(M, N, f, p, backend)` looks OK?

kellertuer: We could go for the `do` scheme, but even then the gradient could be `gradient(f, M, p)` (though not that nice compared to our other signatures).

I would've expected your `AbstractMaps` for the Jacobian more, but I think your signature looks great.

src/riemannian_diff.jl
mateuszbaran: What do you mean by both? It stores either a cached or a non-cached basis. It's a union to only let orthonormal cached bases.

kellertuer: Ah, ok. Then I misread the code and its fine.

src/riemannian_diff.jl
mateuszbaran: OK, before committing to a design, one last thing. Let's say I want to compute the differential of a curve on the 2-sphere. This would be `differential(Sphere(2), f, p)` and all information about discretization would be stored in `f`, right?

kellertuer: Yes, I would say so. For my cases f would not be discretised even, so if it would be, f should have that information.

mateuszbaran: I've made the changes. Now there seems to be no point in having types for maps so I think I'll remove them.

test/differentiation.jl
```diff
@@ -161,42 +161,38 @@ end
...
     s2 = Sphere(2)
-    s2c = Manifolds.Curve(s2)
     p = [0.0, 0.0, 1.0]
     q = [1.0, 0.0, 0.0]
     c1(t) = geodesic(s2, q, p, t)
-    @test domain(s2c) === ℝ
 
     Xval = [-sqrt(2) / 2, 0.0, sqrt(2) / 2]
-    @test isapprox(s2, c1(π / 4), differential(c1, s2c, π / 4), Xval)
+    @test isapprox(s2, c1(π / 4), differential(s2, c1, π / 4), Xval)
```
kellertuer, line 169: I would expect a differential to be a linear functional, i.e. for real-valued functions a co-vector? This does not necessarily contradict to this line, just for clarification.

mateuszbaran: It was already discussed in the previous PR: https://github.com/JuliaManifolds/Manifolds.jl/pull/110#issuecomment-607219033 . BTW, differential of a curve $c$ at $t$ is indeed a linear functional, but as far as I know $\mathbb{R} -> T_{c(t)} M$, so it doesn't look like a covector.

Anyway, it's a good idea to clarify it in documentation.

kellertuer: Ah yes, we already discussed that.

The differential just at a point c(t) is a covector :) not the complete differential. If w look at

$D_t c(t)$ this is a map from T_{c(t)}M to R, so $D_tc(t)[X]$ to be precise and its linear. So its a linear function on T_{c(t)}M or in other words an element from T*_{c(t)}M, a covector. And that is what you compute here.

As long as the vector you return is untyped, that's completely fine, so its really something for the docs, and if we decide to type something we should take that into account.

mateuszbaran: > _edit_: ops. I confused the curve with a function on the manifold, the above argument is the other way around and your right it should returrn tangents here.

OK, so I'll just edit the documentation.

mateuszbaran: Does the documentation look OK now?

kellertuer: I am not sure about $df_t(t)$, I would'Ve thought its $df_t(t)$ (take the derivative with respect to the variable t and then plug in the value of t afterwards), but the rest looks very good.

mateuszbaran: In $df_{t_1}(t_2)$,  $t_1$ is a point in $\mathbb{R}$ but $t_2$ is a vector from $T_{t_1} \mathbb{R}$, so using the same value in both places doesn't look right. Anyway, you can check it for a curve on the real plane: let's say we have f(t) = (1-t, t^2). The differential at $t_1$ is df_{t_1}(u) = (-u, 2 u t_1) and df_{t_1}(1) is exactly the right tangent vector.

kellertuer: Hm, then I confuse something again in your notation. For me writing $df_t$ mint something different.

mateuszbaran: I was trying to follow the notation from https://juliamanifolds.github.io/Manifolds.jl/stable/notation.html . How would you write it then?

mateuszbaran: > Ah I see, I would write $D_tf$ to indicate taking the derivative with respect to t, D_tf(t_1)$ to indicate the (wholesaled) derivative at t_1 and D_tf(t_1[t_2]$ for your note. But for sure we can keep your notation, I was just confused,

Hmm, maybe we should change the primary notation for differentials.

kellertuer: I see, yes, then your notation was correct wrt that list and I was mainly confused. The one problem I see, is, that sometimes it is useful to distinguish which variable one takes the differential of $d\varphi_p$ would for me indicate that. but I would still write $d\varphi_p(p_1)[X]$ (one could leave our $_p(p_1)$ and just write $_{p_1}$ if it is clear because we have only one variable) but then I would still write square brackets for the direction maybe?

mateuszbaran: OK, makes sense.

mateuszbaran: Could you expand a bit on the distinguishing variables part? At the moment we only define differentials of functions between two manifolds and it always is a map between the respective tangent spaces. Do you mean it as using different charts, taking the differential for only a subspace of the tangent space, or something else?

kellertuer: Imagine the power (or product) manifold and you only take the differential with respect to the first or second argument. So for example having a function $f(p,q)$ and you look at $df_p(p_0,q_0)[X]$ where $X$ is from the tangent space of the p component at $(p_0,q_0)$. That's where you really want to keep the variable $p$ of the differential.

mateuszbaran: I've changed it, does it look OK now?

kellertuer: Yes, that looks quite good, for your one-variable-function I think you can leave out the $(t)$ since it can not (or no longer) be confused with the direction $[1]$, so $df_t[1]$ would be fine for me, too.

PR #197, opened 2020-07-12
Printing fixes for Julia 1.6

### LINKED ISSUE #196: Fix printing tests on Julia master
Julia master recently changed how some types are printed so some printing tests need to be updated. See for example https://github.com/JuliaManifolds/Manifolds.jl/pull/195/checks?check_run_id=858788815 .

mateuszbaran:
Should fix #196 .

### DIFF
--- a/src/manifolds/Stiefel.jl
+++ b/src/manifolds/Stiefel.jl
@@ -192,9 +192,9 @@ function inverse_retract!(::Stiefel{n,k}, X, p, q, ::QRInverseRetraction) where 
     A = p' * q
     R = zeros(typeof(one(eltype(p)) * one(eltype(q))), k, k)
     for i in 1:k
-        b = zeros(i)
+        b = zeros(eltype(R), i)
         b[i] = 1
-        b[1:(end - 1)] = real(-transpose(R[1:(i - 1), 1:(i - 1)]) * A[i, 1:(i - 1)])
+        b[1:(end - 1)] = -transpose(R[1:(i - 1), 1:(i - 1)]) * A[i, 1:(i - 1)]
         R[1:i, i] = A[1:i, 1:i] \ b
     end
     return copyto!(X, q * R - p)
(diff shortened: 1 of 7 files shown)


src/manifolds/Stiefel.jl
```diff
@@ -194,7 +194,7 @@ function inverse_retract!(::Stiefel{n,k}, X, p, q, ::QRInverseRetraction) where
     for i in 1:k
         b = zeros(i)
         b[i] = 1
-        b[1:(end - 1)] = -transpose(R[1:(i - 1), 1:(i - 1)]) * A[i, 1:(i - 1)]
+        b[1:(end - 1)] = real(-transpose(R[1:(i - 1), 1:(i - 1)]) * A[i, 1:(i - 1)])
```
mateuszbaran, line 197: @kellertuer This conversion of complex -> real was exact and thus non-throwing on Julia 1.0-1.4 but nightly has a small inexactness. Is it better to throw away the complex part or make `b` have the same eltype as `R`?

kellertuer: I think I would prefer the `eltype`, since for a complex case, I am not 100% sure this will stay just real.

test/fixed_rank.jl
```diff
@@ -9,36 +9,70 @@ include("utils.jl")
...
-      1.0  1.0
-    M factor:
-     2×2 Array{Float64,2}:
-      1.0  0.0
-      0.0  1.0
-    Vt factor:
-     2×2 Array{Float64,2}:
-      0.0  0.0
-      0.0  0.0"""
+    if VERSION >= v"1.6.0-DEV.430"
```
kellertuer, line 12: This got quite long, could we maybe generate this string, such tthat we check that from "U factor:" onwards the lines are queal to the SVD print? Maybe that would be a better test.

mateuszbaran: I'll try that.

mateuszbaran: I tried that and there are subtle differences, I'm not sure it's worth figuring out how to generate that from SVD print.

kellertuer: Ok, if this works, we can leave it for now, if I find time I can still try my advanced idea (which would get us independent of changes wrt to SVD printings).

PR #199, opened 2020-07-14
Generic manifold tests reusable by other packages

mateuszbaran:
I've moved `test_manifold` into a module in `Manifolds` so that I can re-use that function in FunManifolds.jl (and, also, so that other people could easily test their manifolds using it).

I've also changed two things, namely `zero_vector!` isn't tested for non-mutable manifold representations and the expected type for `manifold_dimension` is configurable (as I don't think it's appropriate to give a finite dimension to functional manifolds). Although maybe it should return the dimension of the finite approximation that is used, I'm not sure yet.


mateuszbaran:
Maybe `test_manifold` should have its own documentation page?

src/test.jl
```diff
@@ -0,0 +1,612 @@
+
```
kellertuer, line 1: Maybe we could call this file `ManifoldTests.jl`?

src/test.jl
```diff
@@ -0,0 +1,612 @@
...
+- `test_musical_isomorphisms = false` : test musical isomorphisms
+- `test_mutating_rand = false` : test the mutating random function for points on manifolds
+- `test_project_tangent = false` : test projections on tangent spaces
+- `test_representation_size = true` : test repersentation size of points/tvectprs
+- `test_tangent_vector_broadcasting = true` : test boradcasting operators on TangentSpace
+- `test_vector_transport = false` : test vector transport
+- `test_vector_spaces = true` : test Vector bundle of this manifold
+
+"""
+function test_manifold(
```
kellertuer, line 60: We should use this opportunity to check that all keywords are well documented, useful and complete. For example I recently added the `is_tangent_atol_multiplier` when working on a manifold with a student.

src/test.jl
mateuszbaran: Makes sense since it defines a separate module.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.3.3"
+version = "0.3.4"
```
kellertuer, line 4:
```suggestion
version = "0.3.5"
```

src/test.jl
mateuszbaran: I've added brief descriptions of all keyword arguments and sorted them alphabetically.

mateuszbaran:
I needed to define the `ManifoldTests` module outside of the lazily loaded code fragment because that code could be re-run multiple times which leads to module redefinition and isn't great. Alternatively I could perhaps check if that module is already defined.

kellertuer:
This last step I do not yet fully understand, I thought submodules would just be done like said here https://docs.julialang.org/en/v1/manual/modules/index.html#Relative-and-absolute-module-paths-1 ? Maybe can you provide more explanation or a link?

mateuszbaran:
Which part of the last step? I don't really like how I had to resolve this and there is some wiggle room to improve it (but not much). One thing that looks impossible is defining modules within a `@requires` block so `ManifoldTests` needs to be defined outside. Within the `@requires` block, however, we can populate that module in any way we want (remembering that we are doing it outside of that module). `@eval` is one option but I could also add a method to `test_manifold` without it, I just don't want to pollute the `Manifolds` module with new imports just for testing and what I've done is the simplest solution I've found.

I don't actually have any link for that. I was trying to find a solution online but it was easier to figure something out myself.

kellertuer:
Hm, I see, I did not think of the eval. And what if we have a `test_manifold` that calls the `diff_tests` which are within `ManifoldTest` defined within a require?

mateuszbaran:
I see, this would make even more sense. It would be possible to run tests not related to ForwardDiff or ReverseDiff without loading these packages.

mateuszbaran:
I've made these changes, is there anything else I should improve?

kellertuer:
I just have a general question left, since I do not see directly, whether that works: If I have a package that for whatever reason uses `ManifoldsBase` to define a manifold, can I just do `use Manifolds.ManifoldsTest` to load the test function and run it? Or is the workflow then different?

mateuszbaran:
Yes, that would work. The package itself could just depend on `ManifoldsBase`, and only the test suite would have to load `using Manifolds: ManifoldTests`, for example.

PR #201, opened 2020-07-23
Elliptope manifold

kellertuer:
This PR introduces the Elliptope manifold.
Still to do:
- [x] tests


kellertuer:
Locally the tests already run fine, here, the elliptope tests also run fine; however, due to the ReverseDiff bug not all other tests do currently run, so I can not yet look for code coverage.

Despite that this should be finished.

mateuszbaran:
Nice. I think we can for now mark the graph manifold in the nested array representation as not compatible with ReverseDiff, at least until https://github.com/JuliaDiff/ReverseDiff.jl/issues/140 is solved.

kellertuer:
Yes sure, we can just deactivate reverse_diff for the graph manifold for now; that's easier than fixing the version of reverse diff

kellertuer:
Now I am a little bit confused. Setting `test_reverse_diff = false` for the graph manifold still ran reverse-diff stuff thereon that (of course) failed. Can you see why?

src/manifolds/Elliptope.jl
```diff
@@ -0,0 +1,203 @@
+@doc raw"""
+    Elliptope{N,K} <: AbstractEmbeddedManifold{ℝ,DefaultIsometricEmbeddingType}
+
+The Elliptope manifold, also known as the set of correlation matricesmanifold of symmetric positive definite matrices, i.e.
```
mateuszbaran, line 4: There is something wrong with this sentence.

src/manifolds/Elliptope.jl
```diff
@@ -0,0 +1,203 @@
...
+\bigr\}.
+\end{aligned}
+````
+
+And this manifold is working solely on the matrices $q$. Note that this $q$ is not unique,
+indeed for any orthogonal matrix $A$ we have $(qA)(qA)^{\mathrm{T}} = qq^{\mathrm{T}} = p$,
+so the manifold implemented here is the quotient manifold. The unit diagonal translates to
+unit norm columns of $q$.
+
+The tangent space at $p$, $T_p\mathcal E(n,k)$ also represented matrices $Y\in ℝ^{n × k}$ reads as
```
mateuszbaran, line 21:
```suggestion
The tangent space at $p$, denoted $T_p\mathcal E(n,k)$, is also represented by matrices $Y\in ℝ^{n × k}$:
```
If you want you can write it differently but the original sentence doesn't look right.

src/tests/tests_general.jl
```diff
@@ -31,7 +31,8 @@ that lie on it (contained in `pts`).
 - `is_point_atol_multiplier = 0`: determines atol of `is_manifold_point` checks.
 - `is_tangent_atol_multiplier = 0`: determines atol of `is_tangent_vector` checks.
 - `mid_point12 = shortest_geodesic(M, pts[1], pts[2], 0.5)`: if not `nothing`, then check
-    that `mid_point(M, pts[1], pts[2])` is approximately equal to `mid_point12`.
+    that `mid_point(M, pts[1], pts[2])` is approximately equal to `mid_point12`. This is
+    automatically set to `nothing` then `text_exp_log` is set to false.
```
mateuszbaran, line 35:
```suggestion
    by default set to `nothing` if `text_exp_log` is set to false.
```

src/tests/tests_general.jl
```diff
@@ -31,7 +31,8 @@ that lie on it (contained in `pts`).
 - `is_point_atol_multiplier = 0`: determines atol of `is_manifold_point` checks.
 - `is_tangent_atol_multiplier = 0`: determines atol of `is_tangent_vector` checks.
 - `mid_point12 = shortest_geodesic(M, pts[1], pts[2], 0.5)`: if not `nothing`, then check
```
mateuszbaran, line 33:
```suggestion
- ` mid_point12 = test_exp_log ? shortest_geodesic(M, pts[1], pts[2], 0.5) : nothing`: if not `nothing`, then check
```

test/elliptope.jl
```diff
@@ -0,0 +1,53 @@
...
+    @test is_tangent_vector(
+        M,
+        q2,
+        vector_transport_to(M, q, Y, q2, ProjectionTransport());
+        atol = 10^-15,
+    )
+
+    types = [Matrix{Float64}]
+    TEST_FLOAT32 && push!(types, Matrix{Float32})
+    TEST_STATIC_SIZED && push!(types, MMatrix{3,3,Float64,9})
```
mateuszbaran, line 34: Are you sure it's the correct matrix size?

kellertuer: Well, yes I am sure, but only that it is the wrong matrix size ;) Sorry was a copy&paste error.

PR #203, opened 2020-07-25
Spectrahedron manifold

kellertuer:
This manifold is only slightly different from the Elliptope (#201), namely it is not unit diagonal but unit trace symmetric positive semidefinite matrices of rank k.


src/manifolds/Spectrahedron.jl
```diff
@@ -0,0 +1,208 @@
...
+@doc raw"""
+    check_tangent_vector(M::Spectrahedron, q, Y; check_base_point = true, kwargs...)
+
+Check whether $X = qY^{\mathrm{T}} + Yq^{\mathrm{T}}$ is a tangent vector to
+$p=qq^{\mathrm{T}}$ on the [`Spectrahedron`](@ref) `M`,
+i.e. atfer [`check_manifold_point`](@ref) of `q`, `Y` has to be of same dimension as `q`
+and a $X$ has to be a symmetric matrix with trace.
+The optional parameter `check_base_point` indicates, whether to call [`check_manifold_point`](@ref)  for `q`.
+The tolerance for the base point check and zero diagonal can be set using the `kwargs...`.
+Note that symmetric of $X$ holds by construction an is not explicitly checked.
```
mateuszbaran, line 88:
```suggestion
Note that symmetry of $X$ holds by construction and is not explicitly checked.
```

test/spectrahedron.jl
```diff
@@ -0,0 +1,57 @@
...
+    @test is_tangent_vector(
+        M,
+        q2,
+        vector_transport_to(M, q, Y, q2, ProjectionTransport());
+        atol = 10^-15,
+    )
+
+    types = [Matrix{Float64}]
+    TEST_FLOAT32 && push!(types, Matrix{Float32})
+    TEST_STATIC_SIZED && push!(types, MMatrix{3,3,Float64,9})
```
mateuszbaran, line 37: This is probably not the right size of matrix.

Issue #200, opened 2020-07-16
Cleaning up the documentation menu


mateuszbaran:
The menu on the left hand side of our documentation is quite chaotic at the moment, we should make it more organized.

kellertuer:
Maybe we could first look for a good organisation here?
I think it would be nice to have About and Contributim further up. What do you propose for the structure in order to be more organised?

mateuszbaran:
> Maybe we could first look for a good organisation here?

Yes, that's what I've opened this issue for :slightly_smiling_face: .

> I think it would be nice to have About and Contributim further up. What do you propose for the structure in order to be more organised?

I don't know, many documentation sites don't even have such things in the menu. Documenter itself has Contributing but it's at the very bottom.

Among other things I feel like Statistics and Distributions shouldn't be at the top level and things under the Library item are somewhat arbitrarily selected.

kellertuer:
Well what about the following

* Home
* ManifoldsBase
* How To
  * Tutorials... I hope to do a few soon...
* Manifolds
  * ..long list...
* $Name
  * Distributions
  * Statistics
  * ... further features, e.g. AD?
* Misc.
  * About
  * Contribute
  * Notation
* Library

I am not that much in favour of the current Library items. An Internal functions within Misc would be good. But what about the `Public` section? AD should go into $Name

and for $Name we need a good name - Features on Manifolds?

mateuszbaran:
Look good. The public section could maybe be called "Utilities" and go to $Name? We could remove the "Library" section and move all items from there to $Name as well. Maybe @sethaxen would have some suggestions?

kellertuer:
Calling "Public" "Utilities" is a good idea I think.

sethaxen:
So I prefer `How to` be named something like `Examples`, because in principle sample scripts without much annotation could be in there in addition to more elaborate tutorials.

I can't think of a better name for $Name than Features on Manifolds.

kellertuer:
I am fine with `Examples`, my idea with `How To` or `Get Started` was to get a reader pointed to the page with the tutorials in the sense that this is a point to directly get started. But `Examples` is fine with me.

Features on Manifolds (or just Features, since everything is on/with manifolds?) is a good idea.

kellertuer:
We are currently missing `_jacobian` in the docs, see https://github.com/JuliaManifolds/Manifolds.jl/runs/917792384#step:4:151 – should I remove that line when revising the docs anyways?

[...]

PR #204, opened 2020-07-25
Symmetric positive semidefinite, fixed rank manifold

kellertuer:
A third manifold from the same regime, where even an exp/log is available (and maybe a second metric).


kellertuer:
After this one I might tackle the menu issue (#200), so I will only update the version with that next PR on the menu. I will also check a few different other menus during this week to check further how others are doing menus.

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
```diff
@@ -0,0 +1,280 @@
...
+````math
+    \exp_q Y = q+Y.
+````
+!!! note
+
+    Since the manifold is represented in the embedding and is a quotient manifold,
+    the exponential and logarithmic map are a bijection only with respect to the equivalence
+    classes. Computing
+    ````math
+        q_2 = exp_p(log_pq)
```
mateuszbaran, line 152:
```suggestion
        q_2 = \exp_p(\log_pq)
```

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
```diff
@@ -0,0 +1,280 @@
...
+Compute the logarithmic map on the [`SymmetricPositiveSemidefiniteFixedRank`](@ref) manifold
+by minimizing $\lVert p - qY\rVert$ with respect to $Y$.
+
+!!! note
+
+    Since the manifold is represented in the embedding and is a quotient manifold,
+    the exponential and logarithmic map are a bijection only with respect to the equivalence
+    classes. Computing
+    ````math
+        q_2 = exp_p(log_pq)
```
mateuszbaran, line 190:
```suggestion
        q_2 = \exp_p(\log_pq)
```

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
```diff
@@ -0,0 +1,280 @@
+@doc raw"""
+    SymmetricPositiveSemidefiniteFixedRank{n,k,𝔽} <: AbstractEmbeddedManifold{𝔽,DefaultIsometricEmbeddingType}
+
+The [`Manifold`](@ref) $ \operatorname{SPS}_k(n)$ consisting of the real- or complex-valued
+symmetric positive semidefinite matrices of size $n × n$ and rank $k$, i.e. the set
+
+````math
+\operatorname{SPS}_k(n) = \bigl\{
```
mateuszbaran, line 8:
```suggestion
\operatorname{SPS}_k(n,𝔽) = \bigl\{
```
What do you think about adding the field here?

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
```diff
@@ -0,0 +1,280 @@
...
+
+@doc raw"""
+    manifold_dimension(M::SymmetricPositiveSemidefiniteFixedRank{n,k,𝔽})
+
+Return the dimension of the [`SymmetricPositiveSemidefiniteFixedRank`](@ref) matrix `M` over the number system
+`𝔽`, i.e.
+
+````math
+\begin{aligned}
+\dim \mathrm{Sym}(n,ℝ) &= kn - \frac{k(k-1)}{2},\\
```
mateuszbaran, line 210:
```suggestion
\dim \operatorname{SPS}_k(n,ℝ) &= kn - \frac{k(k-1)}{2},\\
```

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
```diff
@@ -0,0 +1,280 @@
...
+\dim \mathrm{Sym}(n,ℂ) &= 2kn - k^2,
```
mateuszbaran, line 211:
```suggestion
\dim \operatorname{SPS}_(n,ℂ) &= 2kn - k^2,
```

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
kellertuer: looks a little too crowded for a math symbol to me; I think until now we never or nearly never did that?

kellertuer:
> Does this metric on SPS_k(n) correspond to one of our metrics of SPD when k=n?

No. for $k=n$ the $Y$ represented here would for $X=YY^T$ elements on the manifold, so it would maybe be the Cholesky factor of X? and with the rank constraint, you are right it would be SPD, but the metric is still the Euclidean one on Y, which is not the same as we do on SPD (the linear-affine one), and I think its also none of the other two (log-Cholesky has the log in between for example, log Euclidean, too) – so no; none of the metrics we do on SPD would come out if you take k=n here, I think.

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
mateuszbaran: It's sort of needed in formulas for manifold dimension below. Anyway, the meaning looks clear even if the field is omitted here.

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
```diff
@@ -0,0 +1,280 @@
...
+@doc raw"""
+    distance(M::SymmetricPositiveSemidefiniteFixedRank, p, q)
+
+Compute the distance between two points `p`, `q` on the
+[`SymmetricPositiveSemidefiniteFixedRank`](@ref), which is the Frobenius norm of $Y$ which
+minimizes $\lVert p - qY\rVert$ with respect to $Y$.
+"""
+distance(M::SymmetricPositiveSemidefiniteFixedRank, p, q) = norm(M, p, log(M, p, q))
+
+embed!(::SymmetricPositiveSemidefiniteFixedRank, q, p) = copyto!(q, p)
```
mateuszbaran, line 134: Shouldn't embed perform the decomposition of `p` (n by n matrix) and write only the (n by k) part to `q`?

mateuszbaran:
I think it would be nice to write about this case in the documentation for SPD for example.

By the way, I think I see what the quotient manifold would be useful for here. By embedding this in quotient of n by k matrices by right actions of k by k rotations we would have `exp`, `log`, `distance` and `isapprox` for free. And we would have another manifold.

kellertuer:
I am not yet sure what that would help on the SPDs since the metzrix is a completely different one; sure, the construction of the Log-Cholesky metric is on some sense similar (as a factorization, though there its unique).

But concerning the Quotient manifold – exactly, I first saw it for `isapprox` and (here easier) `exp` (which might have to use an embed - + - project in general), but you're right with the right action we would get log and hence also distance in one generic approach.

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
kellertuer: No. This is currently modeled as embedding this manifold into `n \times k` matrices to “steal” their metric, so it is not the classical embedding one would first think of ( doing qq^T, but maybe the p here is misleading). So it is an embedding of the parametrizarion (q), wince (if k is far less then n) you actually never want to store the large `n \times n` matrix. So that's also one point where I noticed, since embedding might be misleading, naming it quotient would fit better.

mateuszbaran: OK, I see.

src/manifolds/SymmetricPositiveSemidefiniteFixedRank.jl
kellertuer: I clarified, that we might add the field here and there. We should maybe globally unify where and how we add the field to a namifold symbol in mathematical formulae.

mateuszbaran: Thanks, great :slightly_smiling_face: . Unifying this notation makes sense.

PR #206, opened 2020-07-28
Restructure Documentation menu

kellertuer:
Starts work on #200.

I made one change that w should discuss: To unify urls I moved the `generated/contributions.md` to `misc/contributions.md` though it is still generated/copied. This unifies URLs. What do you think? Move it back to generated or keep it in misc?

Second, Documenter claims that the docstring for `Manifolds.test_manifold`, but actually I can not see how that gets there, since we have `Manifolds.ManifoldTests.test_manifold` which is also properly documented therein. I am not sure how this starts.


docs/make.jl
```diff
@@ -61,16 +61,18 @@ makedocs(
...
-            "Public" => "lib/public.md",
-            "Internals" => "lib/internals.md",
-            "Differentiation" => "lib/differentiation.md",
-            "Testing" => "lib/testing.md",
+        "Features on Manifolds" => [
+            "Differentiation" => "features/differentiation.md",
+            "Distributions" => "features/distributions.md",
+            "Statistics" => "features/statistics.md",
+            "Utilities" => "features/utilities.md",
+            "Testing" => "features/testing.md",
```
mateuszbaran, line 69: Maybe we could order these alphabetically? That is put "Testing" before "Utilities".

kellertuer: Of course, I was just a little too fast.

docs/src/misc/contributing.md
```diff
@@ -0,0 +1,93 @@
+# Contributing to `Manifolds.jl`
```
mateuszbaran, line 1: This file should be in `.gitignore` so that it's not pushed here.

mateuszbaran: I don't have anything against generating this `.md` file into `misc`.

mateuszbaran:
I don't know what the issue with `test_manifold` is but I like the new structure of the menu 👍 .

PR #207, opened 2020-07-28
Changes power and product manifold to return `CompositeException`s.

### LINKED ISSUE #47: manifold_point_error can't return multiple independent issues
As discussed in #46 , the current approach to error checking allows reporting only a single error at a time. For some manifolds (for example product and power manifolds) multiple independent issues may arise at the same time. We should have a way to report all of them in `manifold_point_error`.

Ref #24 .

kellertuer:
This should fix #47. Tests are still running fine, but that also means we never actually tested which errors are returned.


src/manifolds/ProductManifold.jl
```diff
@@ -129,21 +129,21 @@ end
...
-        err === nothing || return err
+function check_manifold_point(
+    M::ProductManifold,
+    p::Union{ProductRepr,ProductArray};
+    kwargs...,
+)
+    ts = ziptuples(M.manifolds, submanifold_components(M, p))
+    e = [check_manifold_point(t...; kwargs...) for t in ts]
+    errors = filter((x) -> !isnothing(x), e)
+    if length(errors) > 0
```
mateuszbaran, line 145: You can also do `count(x -> !isnothing(x), e)` instead of filtering and calling `length`.

kellertuer: But what is I also only want to return the array of errors? That's mainly what I am filtering for.

mateuszbaran: Yes, you are right, but this made me realize -- will there be sufficient information in the error to tell which components the errors correspond to? I get the filtering now but maybe we should zip `e` with component indices before filtering?

kellertuer: I am not 100% sure it would work (with CompositeException) but it might, and would be helpful, yes.

kellertuer: I thought a little bit about this, I don't feel so god about collecting something else then errors in the CompositeException, though the type it accepts seems to alsow for that. Do you think it might be enough that our error messages usually repeat the data that is wrong?

kellertuer: So yes it does work, but the composite error prints quite strange then (so if it does not contain an array of errors).

kellertuer: Finally I checked whether one can “edit” an error message, and no, they are immutable; since we do not know which errors appear, w can not that easily prepend something like `"On Comp #1:` before the error.

mateuszbaran: Right, that printing is a problem. Ideally I guess we would have a separate exception type that encapsulates errors for specific components and stores component index, and has custom printing. Anyway, what you already did here is definitely an improvement over the current behavior so we can ignore these indices.

PR #208, opened 2020-08-03
More Multinomial manifolds: Double Stochastic and Symmetric Multinomial Manifold

kellertuer:
This is some work on the two remaining multinomial manifolds that I am aware of, the doubly stochastic multinomial and the symmetric multinomial manifold.

- [x] test doubly stochastic
- [x] implement symmetric multinomial
- [x] test symmetric multinomial


kellertuer:
This week seems to be the week of all tests failing. SimpleGraph? I don't know why that now fails... does not locally.

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,286 @@
...
+       AbstractEmbeddedManifold{ℝ,DefaultEmbeddingType} where {N} end
+
+function MultinomialDoubleStochasticMatrices(n::Int)
+    return MultinomialDoubleStochasticMatrices{n}()
+end
+
+@doc raw"""
+    check_manifold_point(M::MultinomialDoubleStochasticMatrices, p)
+
+Checks whether `p` is a valid point on the [`MultinomialDoubleStochasticMatrices`](@ref)`(m,n)` `M`,
```
mateuszbaran, line 53:
```suggestion
Checks whether `p` is a valid point on the [`MultinomialDoubleStochasticMatrices`](@ref)`(n)` `M`,
```

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,286 @@
...
+check_manifold_point(::MultinomialDoubleStochasticMatrices, ::Any)
+function check_manifold_point(
+    M::MultinomialDoubleStochasticMatrices{n},
+    p;
+    kwargs...,
+) where {n}
+    mpv =
+        invoke(check_manifold_point, Tuple{supertype(typeof(M)),typeof(p)}, M, p; kwargs...)
+    mpv === nothing || return mpv
+    c = sum(p, dims = 1) # due to symmetry we only have to check the cols
```
mateuszbaran, line 65: What symmetry do you mean here?

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,286 @@
...
+        Tuple{supertype(typeof(M)),typeof(p),typeof(X)},
+        M,
+        p,
+        X;
+        check_base_point = false, # already checked above
+        kwargs...,
+    )
+    mpv === nothing || return mpv
+    c = sum(X, dims = 1) # due to symmetry, we only have to check columns
+    if !isapprox(norm(c), 0.0; kwargs...)
```
mateuszbaran, line 119: Shouldn't you also check that `X` is symmetric?

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,286 @@
...
+        M::MultinomialDoubleStochasticMatrices,
+        p;
+        maxiter = 100,
+        tolerance = eps(eltype(p))
+    )
+
+project a matrix `p` with positive entries applying Sinkhorn's algorithm.
+"""
+function project(M::MultinomialDoubleStochasticMatrices, p; kwargs...)
+    q = allocate_result(M, project, p)
```
mateuszbaran, line 211: Do we actually need to implement this method?

src/manifolds/MultinomialSymmetric.jl
```diff
@@ -0,0 +1,252 @@
+@doc raw"""
+    MultinomialSymmetric{n} <: AbstractEmbeddedManifold{ℝ, DefaultEmbeddingType}
+
+The multinomial symmetric matrices manifold consists of all matrices with unit sum columns`m` column vectors, where each column is of length
```
mateuszbaran, line 4:
```suggestion
The multinomial symmetric matrices manifold consists of all symmetric `n×n` matrices with positive entries such that each column sums to one, and where each column is of length
```

src/manifolds/MultinomialSymmetric.jl
```diff
@@ -0,0 +1,252 @@
...
+`n` and unit norm, i.e.
```
mateuszbaran, line 5: Are you sure about the unit norm requirement?

src/manifolds/MultinomialSymmetric.jl
```diff
@@ -0,0 +1,252 @@
...
+end
+
+embed!(::MultinomialSymmetric, q, p) = copyto!(q, p)
+embed!(::MultinomialSymmetric, Y, ::Any, X) = copyto!(Y, X)
+
+@doc raw"""
+    inner(M::MultinomialSymmetric{n}, p, X, Y) where {n}
+
+Compute the inner product on the tangent space at $p$, which is the elementwise
+inner product similar to the [`Hyperbolic`](@ref) space, i.e.
```
mateuszbaran, line 128: To me it looks more like the metric of `ProbabilitySimplex` or `MultinomialMatrices`.

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,286 @@
...
+        return DomainError(
+            r,
+            "The matrix $(X) is not a tangent vector to $(p) on $(M), since its rows do not sum up to zero.",
+        )
+    end
+    return nothing
+end
+
+function decorated_manifold(::MultinomialDoubleStochasticMatrices{N}) where {N}
+    return Euclidean(N, N; field = ℝ)
```
mateuszbaran, line 136: It could also be `MultinomialMatrices`. Do we have any general guidelines about handling such hierarchies of embeddings? Do we want to put something more specific here first?

docs/make.jl
```diff
@@ -34,6 +34,9 @@ makedocs(
                 "Hyperbolic space" => "manifolds/hyperbolic.md",
                 "Lorentzian manifold" => "manifolds/lorentz.md",
                 "Multinomial matrices" => "manifolds/multinomial.md",
+                "Multinomial doubly stochastic matrices" =>
```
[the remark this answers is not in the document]
kellertuer: Ah yes, it was first called symmetric doubly (until I noticed that its not symmetric), that's why it's consistently one-off. Will fix that, of course.

src/manifolds/MultinomialDoublyStochastic.jl
kellertuer: see first amswer, I'll revise that comment.

src/manifolds/MultinomialDoublyStochastic.jl
kellertuer: I wanted to pass down the `kwargs`, since usually we don't do that for project.

src/manifolds/MultinomialSymmetric.jl
kellertuer: unit one norm to be precise, I'll rephrase.

src/manifolds/MultinomialSymmetric.jl
kellertuer: It's also `Hyperbolic(1)^(n x n)` :)

src/manifolds/MultinomialDoublyStochastic.jl
mateuszbaran: Right, it makes sense. Could you add a comment here in the code with this explanation?

src/manifolds/MultinomialSymmetric.jl
mateuszbaran: For n=1 this says `<X, Y>_p = X[1]*X[1]/p[1]` while `Hyperbolic(1)` has `<X, Y>_p = -X[2]Y[2] + X[1]Y[1]`. I don't see much similarity here. Am I missing something?

src/manifolds/MultinomialDoublyStochastic.jl
kellertuer: Indeed, I have to check; probably one has to have one check less (Multinomial checks for stochastic columns), one can reuse inner, sounds good.

In general I would give the rule of thumb: Use an embedding that gives you the most possible reuse of functions.

src/manifolds/MultinomialDoublyStochastic.jl
kellertuer: Of course.

src/manifolds/MultinomialSymmetric.jl
kellertuer: Oh! I mixed that up. It's `SymmetricPositiveDefinite(1,1)` (not `Hyperbolic`), i.e. element wise positive numbers. But also here we could embed into `MultinomialMatrices` to avoid implementing it again here.

src/manifolds/MultinomialDoublyStochastic.jl
kellertuer: Now I do that again – but omit something else ;)

src/manifolds/MultinomialSymmetric.jl
kellertuer: This is resolved by switching the embedding type.

src/manifolds/MultinomialDoublyStochastic.jl
kellertuer: I noticed that it changed a few things
* while column unit sum is not checked here, we only check rows
* similarly for the symmetric case, switching to `MultinomialMatrices` avoids positivity check and columns check (and hence row by symmetry) which are all done in the embedding, but symmetric is checked now therein (using the check from `SymmetricMatrices(n)` explicitly to also reuse code there).

kellertuer: I further notices that with an abstract type (by symmetry both are doubly stochastic) one could even avoid reimplementing projection of a point, so thanks for that hint.

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,258 @@
+@doc raw"""
+    AbstractMultinomialDoublyStochastic{N} <: AbstractEmbeddedManifold{ℝ, DefaultIsometricEmbeddingType}
+
+A comon type for manifolds that are doubly stochastic, for example by direct constraint
```
mateuszbaran, line 4:
```suggestion
A common type for manifolds that are doubly stochastic, for example by direct constraint
```

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -0,0 +1,258 @@
...
+````
+where $⊙$ denotes the Hadamard or elementwise product and $\mathbb{1}_n$ is the vector of length $n$ containing ones.
+The two vectors $α,β ∈ ℝ^{n×n}$ are computed as a solution (typically using the left pseudo inverse) of
+````math
+    \begin{pmatrix} I_n & p\\p^{\mathrm{T}} & I_n \end{pmatrix}
+    \begin{pmatrix} α\\ β\end{pmatrix}
+    =
+    \begin{pmatrix} Y\mathbf{1}\\Y^{\mathrm{T}}\mathbf{1}\end{pmatrix},
+````
+where $I_n$ is teh $n×n$ unit matrix and $\mathbf{1}_n$ is the vector of length $n$ containing ones.
```
mateuszbaran, line 161:
```suggestion
where $I_n$ is the $n×n$ unit matrix and $\mathbf{1}_n$ is the vector of length $n$ containing ones.
```

PR #209, opened 2020-08-04
A little rework do reactivate Julia 1.0 tests

kellertuer:
I could not yet fix all introduced errors, but 1.0 tests are running again (with some subarray error that I do not yet understand).


kellertuer:
This is really strange currently, here also, the Julia 1.0 test gives _very_ different errors from mine locally, my local ones ran until the circle at least, here they don't even properly start. I thought the Project.toml was aiming at such ways to be run on different machines? I am a little lost today with tests.

PR #210, opened 2020-08-06
Proper errors on Power and Product

kellertuer:
To continue #207 this PR introduces two special error types for composite manifolds like power and product: the `ComponentManifoldError` consists of an error and an index, where this error occurred,
the `CompositeManifoldError` collects a set of these if there is more than one error.

Maybe a separate docs page would be still nice.


docs/src/features/utilities.md
```diff
@@ -14,3 +14,13 @@ Manifolds.StaticReshaper
 Manifolds.ArrayReshaper
 Manifolds.make_reshape
 ```
+
+# Specific error messages
```
mateuszbaran, line 18: I would call this "Specific exception types".

test/groups/group_operation_action.jl
```diff
@@ -14,6 +14,9 @@ include("group_utils.jl")
     @test repr(A_left) == "GroupOperationAction($(repr(G)), LeftAction())"
     @test repr(A_right) == "GroupOperationAction($(repr(G)), RightAction())"
 
+    @test switch_direction(LeftAction()) == RightAction()
+    switch_direction(RightAction()) == LeftAction()
```
mateuszbaran, line 18:
```suggestion
    @test switch_direction(RightAction()) == LeftAction()
```

kellertuer: Ah, yes. I am wondering by the way, why the group test coverage is so low, the tests here seem quite exhaustive, but then some do not seem to “hit” their lines.

mateuszbaran: I don't know, I didn't check that in a long time. There are some real gaps in coverage there IIRC.

kellertuer: Me neither, but at first glance, most things should be covered but aren't – so we might check somewhen, to get coverage a little further.

PR #211, opened 2020-08-06
Connections

mateuszbaran:
I have split the connections part from #202 . This should be much easier to finish in a reasonable amount of time. The key functionality is already here, I will just add more tests and examples.


src/connections.jl
```diff
@@ -0,0 +1,173 @@
...
+end
+
+"""
+    HessianOperator{TC<:AbstractAffineConnection}
+
+Defines the Hessian operator for the given connection. Definition follows [^Absil2008],
+Section 5.5.
+
+[^Absil2008]:
+    > Absil, P. A., et al. Optimization Algorithms on Matrix Manifolds. 2008.
```
kellertuer, line 108:
```suggestion
    >Absil, P.-A., Mahony, R., Sepulchre, R.:
    >Optimization Algorithms on Matrix Manifolds,
    > Princeton University Press, 2008,
    > doi: [10.1515/9781400830244](https://doi.org/10.1515/9781400830244>,
    > [open access](http://press.princeton.edu/chapters/absil/).
```

and it would maybe be – for completeness - a nice idea to repeat some part of the definition?

src/connections.jl
```diff
@@ -0,0 +1,173 @@
...
+    apply_operator!(
+        F::HessianOperator,
+        p,
+        X,
+        f,
+        diff_backend::AbstractRiemannianDiffBackend,
+        grad_backend::AbstractRiemannianDiffBackend,
+    )
+
+Apply the Hessian operator `F` at point `p` to function `f` in the direction `f` using
```
kellertuer, line 124: f is a function and a direction at the same time?

src/connections.jl
```diff
@@ -0,0 +1,173 @@
...
+function apply_operator(F::AbstractAffineConnection, p, X, Y)
+    return apply_operator(F, p, X, Y, rdifferential_backend())
+end
+
+
+function apply_operator!(F::AbstractAffineConnection, Z, p, X, Y)
+    return apply_operator!(F, Z, p, X, Y, rdifferential_backend())
+end
+
+"""
```
kellertuer, line 28: I think previously we always collected types atop?
And – is this meant to work similar to a metric? Ah, no, it has `M` inside, so we do not need another decorator. Looking forward to see one or two examples how this will work :)

src/connections.jl
mateuszbaran: I was just too sloppy with that docstring, I'll fix it.

src/connections.jl
mateuszbaran: Right, I'll move the type to the top.

mateuszbaran:
Type stability of some functions is less than ideal right now.

kellertuer:
I shortly looked at the docs

* currently we have 3 times the Absil reference in a footnote – would it make sense to get that to one string? 
* could we also add a formula for the connection?
* It would be good to link to Hessian and connection in the apply_operator functions.

mateuszbaran:
> * currently we have 3 times the Absil reference in a footnote – would it make sense to get that to one string?

I don't really see any problem with multiple footnotes. At least docstrings are more complete this way.

> * could we also add a formula for the connection?

Which formula do you mean?

> * It would be good to link to Hessian and connection in the apply_operator functions.

OK, I'll add that.

kellertuer:
> > * currently we have 3 times the Absil reference in a footnote – would it make sense to get that to one string?
> 
> I don't really see any problem with multiple footnotes. At least docstrings are more complete this way.

ok. I still hope for a proper BibTeX support in Documenter, that would solve these.

> 
> > * could we also add a formula for the connection?
> 
> Which formula do you mean?

Maybe the formal definition (p. 94, Absil et al, first Formula) and the usual notation with \nabla_\eta \xi ? There we should carefully look for notation, since covectors have small greek letters at our point, so maybe $X(p)$, $Y(p)$ or something?

src/connections.jl
```diff
@@ -68,8 +68,9 @@ base_manifold(F::LeviCivitaConnection) = F.manifold
 @doc raw"""
     apply_operator(F::LeviCivitaConnection, p, X, Y)
 
-Compute the value of the Levi-Civita connection at point `p`, in the direction pointed by
-tangent vector `X` at `p`, of the vector field on `F.manifold` defined by a function `Y`.
+Compute the value of the Levi-Civita connection of type [`LeviCivitaConnection`](@ref) at
+point `p`, in the direction pointed by tangent vector `X` at `p`, of the vector field on
+`F.manifold` defined by a function `Y`.
 The formula reads $(\nabla_X \mathit{Y})_p$.
```
kellertuer, line 74: Why the `mathit` here? We should also be careful for the Levi Cevita, since it usually applies to vector fields not to vectors, then your notation might be confusing here.

mateuszbaran: Right, that `mathit` is not necessary. Connections look pretty well defined for `X` being just a tangent vector but `Y` really needs to be a vector field (and it is in the tests). Do you have a better notation for vector fields?

kellertuer: We basically only have capital greek letters left, right? If we write upfront $X,Y \in \mathcal{X}(\mathcal M)$ it might help and we might use X,Y for both vectors and vector fields, but all this should also be added to the notations.md.

mateuszbaran:
> > > * currently we have 3 times the Absil reference in a footnote – would it make sense to get that to one string?
> > 
> > 
> > I don't really see any problem with multiple footnotes. At least docstrings are more complete this way.
> 
> ok. I still hope for a proper BibTeX support in Documenter, that would solve these.

Yes, that would be the best solution.

> > Which formula do you mean?
> 
> Maybe the formal definition (p. 94, Absil et al, first Formula) and the usual notation with \nabla_\eta \xi ? There we should carefully look for notation, since covectors have small greek letters at our point, so maybe $X(p)$, $Y(p)$ or something?

I think what we have here slightly deviates from that mathematical definition. Using Absil's notation, we have fixed point `x` and so `\eta` is also just a tangent vector. We just take the value of the corresponding vector field at point `x` and values at other points are irrelevant. `\xi` needs to actually be a vector field. In out notation we could call it `Y(p)` or `Y_p`.

src/connections.jl
mateuszbaran: Mostly, we already use `\Xi` for a set of tangent vectors. I have written a separate response below regarding notation.

kellertuer: Ok, then lets use X and Y and phrase is carefully in the doc text.

mateuszbaran:
Riemannian Hessian currently returns wrong results for some reason. I compared it to the output of the formula with the Weingarten map and it's different.

mateuszbaran:
Ah, I think I get it now. Our gradient is doesn't reproduce the expected first-order behavior so taking a connection on it doesn't have any chance to reproduce the Hessian. I guess I'll just have to let it be that way and add explicit Weingarten maps for going from Euclidean to Riemannian Hessian, just like in Pymanopt.

mateuszbaran:
This approach to connections is now obsolete, see #381 and #387 for some discussions.

PR #212, opened 2020-08-27
Projection on power manifolds

mateuszbaran:
Just two new methods :slightly_smiling_face: .


kellertuer (approved):
Nice. One could also add docs to the corresponding `project` functions, such that this is also reflected in the docs.

PR #213, opened 2020-08-31
Moving group-related tests to ManifoldTests

mateuszbaran:
This will be useful in https://github.com/JuliaManifolds/FunManifolds.jl/pull/10 .


kellertuer:
Seems reasonable, maybe formatter run is missing?

mateuszbaran:
Yes it is, there is also a weird failure on nightly.

mateuszbaran:
It looks like there was another change to printing recently: https://github.com/JuliaLang/julia/pull/37085 .

kellertuer:
Good to know – if we encapsulate those tests further, as I started for one of them, this would automatically be correct (though I don't completely remember where I did that ;)).

PR #214, opened 2020-09-03
Embed for Euclidean

mateuszbaran:
`Euclidean` already has `project`, so it could also have `embed`.


kellertuer (approved):
For sure. This is nice to have.

Issue #158, opened 2020-04-09
Add overloads for Base.in


sethaxen:
The following overloads would be nice for syntactic convenience:

```julia
Base.in(M::Manifold, p) = is_manifold_point(M, p)
Base.in(TS::TangentSpaceAtPoint, X) = is_tangent_vector(TS.fiber.manifold, TS.point, X)
```

e.g. a user could write `[1, 0, 0] ∈ Sphere(2)` or `[0.5, 0.5, 0.5] ∈ ProbabilitySimplex(2)`

kellertuer:
Should be documented and implemented within `ManifoldsBase` – but it's a very neat idea!

mateuszbaran:
Nice idea! The second method, though, can't currently be implemented in `ManifoldsBase` (it doesn't have `TangentSpaceAtPoint`, although it perhaps should be in `ManifoldsBase`).

sethaxen:
> Nice idea! The second method, though, can't currently be implemented in `ManifoldsBase` (it doesn't have `TangentSpaceAtPoint`, although it perhaps should be in `ManifoldsBase`).

Perhaps. Though that would require fibers and a few other things being migrated over, right? This method is more for completeness, `TangentSpaceAtPoint` is a bit long to write out, and one might as well call `is_tangent_vector`. Where I would see it being useful is if one wanted to check many vectors against a tangent space, and then they would create one of these and then use `in` many times for brevity.

mateuszbaran:
It wouldn't be that much to migrate. Just `VectorBundleFibers`, `VectorSpaceAtPoint` and a few constants and methods. I think this should be done at some point anyway (although it could wait until we see how vector bundles should interact with bases).

sivapvarma:
I would like to take this up. I like the idea behind the project and it would be cool to have all of manopt functionality in a Julia package.

What is the current consensus on what should be moved and add to `ManifoldsBase`.

kellertuer:
There's https://github.com/JuliaManifolds/Manopt.jl which moved to using ManifoldsBase recently. It will cover Swarm-Optim and LRBFGS soon(ish), too.

sivapvarma:
My idea was to start by adding the overloads seth mentioned in this issue to ManifoldsBase.

Since the discussion above points to others things that should be moved like `VectorBundleFibers`, `VectorSpaceAtPoint`, I want to see if there is a clear consensus on what should be moved from here to ManifoldsBase

[...]

PR #215, opened 2020-09-07
Implement Base.in for manifold point and tangent vector tests

kellertuer:
Solves #158 and beautifies Tangent Space at point prints.

By the way: Is there a reason, we do not use `TangentSpace(M,p)` for the tangent space at point?


src/manifolds/VectorBundle.jl
```diff
@@ -472,6 +472,11 @@ function get_vectors(M::VectorBundleFibers, p, B::CachedBasis)
 end
 Base.@propagate_inbounds Base.getindex(x::FVector, i) = getindex(x.data, i)
 
+function Base.getproperty(TpM::TangentSpaceAtPoint, sym::Symbol)
```
mateuszbaran, line 475: Extending `getproperty` can cause serious type instability. Theoretically this method could be inlined and `sym` constant-propagated but it may not happen in practice.

test/vector_bundle.jl
```diff
@@ -114,28 +114,11 @@ struct TestVectorSpaceType <: VectorSpaceType end
...
-             3-element Array{Float64,1}:
-              1.0
-              0.0
-              0.0"""
-        end
-
+        t_xs = sprint(show, "text/plain", t_x)
+        sp = sprint(show, "text/plain", x)
+        sp = replace(sp, '\n' => "\n ")
+        t_xs_test = "Tangent space to the manifold $(M) at point:\n $(sp)"
```
mateuszbaran, line 120: Thanks, that's a much nicer test 🙂 .

src/Manifolds.jl
```diff
@@ -165,9 +165,29 @@ include("groups/rotation_action.jl")
 
 include("groups/special_euclidean.jl")
 
-
 include("tests/ManifoldTests.jl")
 
+@doc raw"""
+    p ∈ M
+    Base.in(p,M; kwargs...)
```
mateuszbaran, line 172: I would put `Base.in` in the first line of the docstring, similarly to the docstring in `Base`.

mateuszbaran:
> By the way: Is there a reason, we do not use `TangentSpace(M,p)` for the tangent space at point?

I think that's just a design choice. We need to have the `VectorSpaceAtPoint` type anyway and `TangentSpaceAtPoint` follows the same naming convention as other types in `VectorBundle.jl`. Anyway `TangentSpace(M,p)` could also construct an object of type `TangentSpaceAtPoint`.

src/Manifolds.jl
```diff
@@ -368,6 +390,7 @@ export ×,
     retract!,
     riemann_tensor,
     set_component!,
+    setproperty!,
```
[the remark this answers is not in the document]
kellertuer: Thanks, good to know.

src/manifolds/VectorBundle.jl
kellertuer: My hope was to have the `TangentSpaceAtPoint` with easy accessors for point and manifold (`TpM.fiber.manifold` feel weird here).

src/Manifolds.jl
```diff
@@ -51,7 +51,7 @@ import ManifoldsBase:
     vector_transport_to!,
     zero_tangent_vector,
     zero_tangent_vector!
-import Base: isapprox, length, isempty, showerror
+import Base: in, getproperty, isapprox, isempty, length, setproperty!, showerror
```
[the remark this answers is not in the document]
kellertuer: Thanks, I had that one in mind at that point but then noticed, that the tangent space is immutable...

src/manifolds/VectorBundle.jl
kellertuer: What would you propose to do to avoid the `getproperty` but still get a nice interface?

mateuszbaran: We can always extend `base_manifold` to `TangentSpaceAtPoint` to return `TpM.fiber.manifold`, I don't have a better idea.

kellertuer: good idea.

src/Manifolds.jl
```diff
@@ -333,6 +353,7 @@ export ×,
     is_manifold_point,
     is_tangent_vector,
     isapprox,
+    in,
```
mateuszbaran, line 356: We don't need to export `in` either I think.

src/manifolds/VectorBundle.jl
```diff
@@ -472,6 +479,8 @@ function get_vectors(M::VectorBundleFibers, p, B::CachedBasis)
 end
 Base.@propagate_inbounds Base.getindex(x::FVector, i) = getindex(x.data, i)
 
+base_manifold(TpM::TangentSpaceAtPoint) = TpM.fiber.manifold
```
mateuszbaran, line 482:
```suggestion
base_manifold(VpM::VectorSpaceAtPoint) = VpM.fiber.manifold
```

src/manifolds/VectorBundle.jl
```diff
@@ -89,10 +89,17 @@ const TangentSpaceAtPoint{M} =
...
+space at `p` on the [`Manifold`](@ref) `M`.
 """
 TangentSpaceAtPoint(M::Manifold, p) = VectorSpaceAtPoint(TangentBundleFibers(M), p)
 
+"""
+    TangentSpace(M::Manifold, p)
+
+Return a [`TangentSpaceAtPoint`](@ref) representing tangent space at `p` on the [`Manifold`](@ref) `M`.
+"""
+TangentSpace(M::Manifold, p) = VectorSpaceAtPoint(TangentBundleFibers(M), p)
```
mateuszbaran, line 101: This line actually isn't covered by tests I think.

mateuszbaran (approved):
After adding that one last test this should be ready to be merged 👍 .

src/manifolds/VectorBundle.jl
kellertuer: This contradicts line 211 in the same file, where base_manifold of the vectorslaceatpoint is set to the fiber; what is now the right definition?

kellertuer: Ah, I see, both do basically the same, just written differently. We keep the one that already existed.

kellertuer:
Ah we

- [ ] still need a test for VectorSpaceAtPoint (since the tangent vector now has its own)

src/manifolds/VectorBundle.jl
mateuszbaran: OK, cool, so we don't even need this method.

test/vector_bundle.jl
```diff
@@ -144,6 +129,13 @@ struct TestVectorSpaceType <: VectorSpaceType end
         @test ct_x.fiber.fiber == CotangentSpace
         @test t_x.point == x
         @test ct_x.point == x
+        # generic vector space at
+        fiber = VectorBundleFibers(TestVectorSpaceType(), M)
+        v_x = VectorSpaceAtPoint(fiber, x)
+        v_xs = sprint(show, "text/plain", v_x)
+        fiber_s = sprint(show, "text/plain", fiber)
+        v_xs_test = "VectorSpaceAtPoint{VectorBundleFibers{TestVectorSpaceType,Sphere{2,ℝ}},Array{Float64,1}}\nFiber:\n $(fiber_s)\nBase point:\n $(sp)"
+        @test v_xs == v_xs_test
```
mateuszbaran, line 138: This will fail on nigtly because `Array{Float64,1}` is printed there as `Vector{Float64}`

kellertuer: I will try to replace that part with a `$(typeof(x))`, that should wrap that I think?

mateuszbaran (approved):

PR #216, opened 2020-09-07
Embeddings for Euclidean into Euclidean

kellertuer:
A Euclidean manifold can be embedded into Euclidean if dimensions fit.

For this PR two things are missing still
- [x] dispatch still returns transparent and the implementation is not met
- [x] test cases have to be implemented


src/manifolds/Euclidean.jl
```diff
@@ -132,6 +132,39 @@ embed(::Euclidean, p)
...
+        throw(DomainError("Invalid embedding, since Euclidean dimension ($(n)) is longer than embedding dimension $(m)."))
+    any(n[1:lm] .> m) &&
+        throw(DomainError("Invalid embedding, since Euclidean dimension ($(n)) has entry larger than embedding dimensions ($(m))."))
+    # put p into q
+    q[[1:ind_n for ind_n in n]..., ones(lm - ln)...] .= p
+    # set remaining entries to zero
+    q[
+        [n[i] == m[i] ? (1:n[i]) : ((n[i] + 1):m[i]) for i in 1:lm]...,
+        [1:n[i] for i in (lm + 1):ln]...,
+    ] .= zero(eltype(p))
```
mateuszbaran, line 164: I don't think it will work for cases like embedding 2x2 Euclidean into 3x3 Euclidean. How about `fill!`ing entire `q` with `zero`s first and then copying the relevant part?

kellertuer: yes, filling is easier, and yes I missed some parts in setting zeros here.

kellertuer:
Found my bug with the transparency (missed a type parameter in the Embedded manifold) and wrote all necessary tests.

src/manifolds/Euclidean.jl
```diff
@@ -132,6 +132,39 @@ embed(::Euclidean, p)
...
+The remainder is filles with zeros for sure.
+"""
+function embed(M::EmbeddedManifold{𝔽,Euclidean{n,𝔽},Euclidean{m,𝔽},ET}, p) where {n,m,𝔽,ET}
+    q = allocate(p, representation_size(M.embedding))
+    embed!(M, q, p)
+    return q
+end
+
+function embed!(
+    M::EmbeddedManifold{𝔽,Euclidean{nL,𝔽},Euclidean{mL,𝔽},ET},
```
mateuszbaran, line 149: I think we don't have to enforce the two `Euclidean` spaces being over the same number system? Invalid conversions will fail during assignment anyway.

src/manifolds/Euclidean.jl
```diff
@@ -132,6 +132,39 @@ embed(::Euclidean, p)
...
+    m = size(q)
+    lm = length(m)
+    (length(n) > length(m)) &&
+        throw(DomainError("Invalid embedding, since Euclidean dimension ($(n)) is longer than embedding dimension $(m)."))
+    any(n .> m[1:ln]) &&
+        throw(DomainError("Invalid embedding, since Euclidean dimension ($(n)) has entry larger than embedding dimensions ($(m))."))
+    # put p into q
+    fill!(q, 0.0)
+    # fill „top left edge“ of q with p.
+    q[[1:ind_n for ind_n in n]..., ones(Int, lm - ln)...] .= p
```
mateuszbaran, line 164:
```suggestion
    q[n_ind -> map(Base.OneTo(ind_n), n)..., ntuple(_ -> 1, lm - ln)...] .= p
```
I haven't tested it but it should be fine (and type-stable).

src/manifolds/Euclidean.jl
kellertuer: Yes, we sure can do that. And embedding from R to C should then just work, that's neat.

src/manifolds/Euclidean.jl
kellertuer: Thanks, type (in)stability is something I do not always see.

kellertuer: Hm, I can't even compile this? It says `syntax: "..." expression outside`, so what are we missing here?

kellertuer: Ah, you shuffled it a little bit, it's `q[map(n_ind -> Base.OneTo(ind_n), n)..., ntuple(_ -> 1, lm - ln)...]`.

mateuszbaran: Right, yes, thanks for fixing it. One thing to remember about splatting (the ... thing) is to only splat tuples when you care about performance.

kellertuer: Now that works, but with different fields is still a little tricky, just leaving out the inner Fs did not do the trick.

mateuszbaran: ... and this time I've mixed `n_ind` with `ind_n`.

kellertuer: ;) Long day for me already, should've seen that. But I also got the different fields to work.

src/manifolds/Euclidean.jl
```diff
@@ -132,6 +132,39 @@ embed(::Euclidean, p)
...
+    q = allocate(p, representation_size(M.embedding))
+    embed!(M, q, p)
+    return q
+end
+
+function embed!(
+    ::EmbeddedManifold{𝔽,Euclidean{nL,𝔽a},Euclidean{mL,𝔽b},ET},
+    q,
+    p,
+) where {nL,mL,𝔽,𝔽a,𝔽b,ET}
```
mateuszbaran, line 152: I think 𝔽 and 𝔽b should be the same actually.

src/manifolds/Euclidean.jl
mateuszbaran: Nice :slightly_smiling_face:

src/manifolds/Euclidean.jl
kellertuer: They _were before_, but above you. asked whether the number fields have to be the same. No. You can embed R in C for sure, that's why I changed that.

What do you prefer, your last remark or this ;) ?

kellertuer: Sorry, I was too fast here, Fa can be different, but Fb and F should be the same, for sure. Will change that

mateuszbaran: When embedding, for example, R^2 in C^3, you have `EmbeddedManifold{ℂ,Euclidean{Tuple{2},ℝ},Euclidean{Tuple{3},ℂ},ET}`, right? 𝔽a and 𝔽b are different but 𝔽 and 𝔽b are not. Or did I mix something up again?

mateuszbaran: OK, so we have this one clear :slightly_smiling_face: .

test/euclidean.jl
```diff
@@ -147,4 +147,31 @@ include("utils.jl")
         @test ℍ^2 === Euclidean(2; field = ℍ)
         @test ℍ^(2, 3) === Euclidean(2, 3; field = ℍ)
     end
+
+    @testset "Embeddings into larger Euclidean Manifolds" begin
```
mateuszbaran, line 151: I think it would be good to have a simple test for embedding a real Euclidean space into a complex Euclidean space.

kellertuer: Good hint, I mixed up the fields (the embedded manifold is still of field type as the smaller one, not as the embedding). I added a test for that.

mateuszbaran: Right, I mixed it up as well when I suggested this change. Thanks for fixing it :slightly_smiling_face: .

mateuszbaran (approved):

src/manifolds/Euclidean.jl
```diff
@@ -132,6 +132,42 @@ embed(::Euclidean, p)
 
 embed!(::Euclidean, q, p) = copyto!(q, p)
 
+@doc raw"""
+    embed(::EmbeddedManifold{𝔽,Euclidean{𝔽},Euclidean{𝔽},ET}, p)
```
mateuszbaran, line 136:
```suggestion
    embed(::EmbeddedManifold{𝔽,Euclidean{n,𝔽},Euclidean{m,𝔽2},ET}, p)
```
One last thing I hope.

src/manifolds/Euclidean.jl
```diff
@@ -132,6 +132,42 @@ embed(::Euclidean, p)
...
+
+A [`Euclidean`](@ref) `M` manifold can be embedded into a [`Euclidean`](@ref) `N`
+if the length of the array dimension and representation size is elementwise less or equal.
+The remainder is filles with zeros for sure.
+"""
+function embed(
+    M::EmbeddedManifold{𝔽,Euclidean{n,𝔽},Euclidean{m,𝔽2},ET},
```
mateuszbaran, line 143: Actually, shouldn't this be defined for all embedded manifolds, not just this one? After all it's just the generic allocate-then-mutate function...

mateuszbaran: Wait, it's already in ManifoldsBase: https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/7471d948b4e5a0485cff823313233fb2397c485b/src/ManifoldsBase.jl#L260

src/manifolds/Euclidean.jl
kellertuer: Thanks, I missed to fix the docs.

src/manifolds/Euclidean.jl
kellertuer: Well, nearly. The one in base allocates an object of size of `M.manifold` somehow, not of `M.embedding`, that's what this one fixes.

mateuszbaran: Oh, I think it's a bug in ManifoldsBase then (and should be fixed there).

kellertuer: Should we leave this here until its fixed?

mateuszbaran: I can quickly fix it in base.

mateuszbaran:
Hmm... and since we are at it embedding tangent vectors would also be nice :slightly_smiling_face: . Let me just fix that problem in ManifoldsBase first.

mateuszbaran:
... and corresponding `project` methods would also be nice.

kellertuer:
Wel already have a few `project(M,p,X)` and `project!(M,Y,p,X)`, so having the corresponding `embed`s  would be reasonable.

src/manifolds/Euclidean.jl
mateuszbaran: Embedding should work with ManifoldsBase 0.9.2 without this method 🙂

kellertuer: Right, I changed that.

kellertuer:
Hm, locally it works, on the CI there is a document/@doc error that I can't reproduce nor understand. Can you take a look?

kellertuer:
I fear your version `0.9.2` might have broken a few manifolds?

kellertuer:
I created JuliaManifolds/ManifoldsBase.jl#49 – since the error introduces was merely me using a loophole in the documentation/usage of `allocate`.

kellertuer:
Hm, a short test revealed, the `embed` and `project` I yesterday deleted in `Base` are actually indeed required for this here to work, otherwise I can't get `EmbeddedManifold` to ever allocate the correct return value for the `O2` manifold in the euclidean tests (line 172).

mateuszbaran:
> I might just reimplement embed and project and make some tests for those explicitly in base.

This is probably the best solution.

mateuszbaran:
I think we need to remove now all the unused `embed!` methods.

mateuszbaran:
Coverage highlights them.

kellertuer:
Well, are they replaced by the method from base, so really unneeded? I think the `AbstractEmbeddedManifold`now provide the default, since we said, that the abstract case always assumes that the points are represented in the embedding; this yields that the generic function already does the job, so we really don't need the special embed-but-just-copy-functions anymore?

mateuszbaran:
Well, now there seems to be no need to ever `embed!` for `AbstractEmbeddedManifold` I think. Notice that there is no generic embed in base for `AbstractEmbeddedManifold`.

kellertuer:
Cool. No there is no need for a generic `embed`, since we have that on the `Manifold` level already for a default embedding.

I have to check some teaching but will remove the noch obsolete `embed!`s afterwards.

kellertuer:
Hm, that's strange. When I delete line 87 in Centered Matrices, which codecov reports as not covered, then the test fails since no method is found.

mateuszbaran:
The centered manifold tests fail just because there is an explicit test for `embed`. If you remove it nothing fails.

> Actually for the `AbstractEmbeddedManifold` I don't see a generic `embed!` in ManifoldsBase; we sure could add one, but currently there is none. Actually I think we even missed an `embed(M, p, X)` for the `EmbeddedManifold` (and its projection counterpart).

Right, I guess there will be another embedding PR to ManifoldsBase :slightly_smiling_face: .

kellertuer:
Hm, but why do I want to remove the test? `embed` and `embed!` should both work for `CenteredMatrices` for sure, or not?

mateuszbaran:
Hmm... actually, in what circumstances would using `embed` on for example `CenteredMatrices` be a good idea? The idea of using representation of the embedding suggests that all operations should actually use it. `project!` for `AbstractEmbeddedManifold` is also supposed to return a valid point or tangent vector in the embedding, right?

kellertuer:
Yes, neither `embed!` not `project!` should change array dimensions, I think we agreed on that recently.

mateuszbaran (approved):
I'm also fine with merging it as-is and leaving this issue for later  🙂.

PR #217, opened 2020-09-09
The manifold of positive numbers (and according powers)

kellertuer:
While basically `SymetricPositiveDefinite(1)` also is the manifold of positive numbers, this manifold is the specific manifold of positiv real numbers. The according powre manifolds are then vectors, matrices, arrays of positiv numbers with the corresponding power metric.


kellertuer:
Interestingly I have code coverage for all lines, but codecov does not like to mark all...not yet sure why.

src/manifolds/PositiveNumbers.jl
```diff
@@ -0,0 +1,224 @@
+@doc raw"""
+    PositiveNumbers <: Manifold{ℝ}
+
+The hyperbolic manifold of positive numbers $H^1$ is a the hyperbolic manifold either
+represented by just positive numbers or the [Poiincaré half-plane model](https://en.wikipedia.org/wiki/Poincaré_half-plane_model).
```
mateuszbaran, line 5:
```suggestion
represented by just positive numbers or the [Poincaré half-plane model](https://en.wikipedia.org/wiki/Poincaré_half-plane_model).
```

docs/src/manifolds/positivenumbers.md
```diff
@@ -0,0 +1,11 @@
+# Positive Numbers
+
+Additionally to the [`PositiveNumbers`](@ref), there are also short forms for its
```
mateuszbaran, line 3:
```suggestion
The manifold [`PositiveNumbers`](@ref) represents positive numbers with hyperbolic geometry. Additionally, there are also short forms for its
```

src/Manifolds.jl
```diff
@@ -219,6 +220,10 @@ export Euclidean,
     MultinomialMatrices,
     MultinomialSymmetric,
     Oblique,
+    PositiveNumbers,
+    PositiveMatrices,
+    PositiveVectors,
+    PositiveArrays,
```
mateuszbaran, line 226: This is not in alphabetical order.

src/manifolds/PositiveNumbers.jl
```diff
@@ -0,0 +1,224 @@
...
+@doc raw"""
+    check_manifold_point(M::PositiveNumbers, p)
+
+Check whether `p` is a point on the [`PositiveNumbers`](@ref) `M`, i.e. $p>0$.
+"""
+function check_manifold_point(M::PositiveNumbers, p; kwargs...)
+    if p <= 0
+        return DomainError(
+            p,
+            "The point $(p) does not lie on $(M), since its is nonpositive.",
```
mateuszbaran, line 50:
```suggestion
            "The point $(p) does not lie on $(M), since it is nonpositive.",
```

src/manifolds/PositiveNumbers.jl
```diff
@@ -0,0 +1,224 @@
...
+Return the dimension of the [`PositiveNumbers`](@ref) `M`,
+i.e. of the 1-dimensional hyperbolic space,
+
+````math
+\dim(H^1) = 1
+````
+"""
+manifold_dimension(::PositiveNumbers) = 1
+
+mid_point(M::PositiveNumbers, p1, p2) = exp(M, p1, 0.5 * log(M, p1, p2))
```
mateuszbaran, line 161:
```suggestion
mid_point(M::PositiveNumbers, p1, p2) = exp(M, p1, log(M, p1, p2) / 2)
```
This won't force type promotion to at least `Float64`.

src/manifolds/PositiveNumbers.jl
```diff
@@ -0,0 +1,224 @@
...
+    X::Real,
+    q::Real,
+    ::ParallelTransport,
+)
+    return X * q / p
+end
+
+function vector_transport_direction(
+    M::PositiveNumbers,
+    p::Number,
```
mateuszbaran, line 215: Why do you use sometimes `Number` and sometimes `Real`? I think using `Real` everywhere should be fine.

test/positive_numbers.jl
```diff
@@ -0,0 +1,45 @@
...
+        @test isinf(injectivity_radius(M, ExponentialRetraction()))
+        @test project(M, 1.5, 1.0) == 1.0
+        @test zero_tangent_vector(M, 1.0) == 0.0
+    end
+    types = [Float64]
+    TEST_FLOAT32 && push!(types, Float32)
+    for T in types
+        @testset "Type $T" begin
+            pts = convert.(Ref(T), [1.0, 4.0, 2.0])
+            test_manifold(
```
mateuszbaran, line 31: I think you should also call `test_manifold` on the `PositiveVectors` manifold.

mateuszbaran:
Codecov sometimes reports wrong, so they could be false positives.

src/manifolds/PositiveNumbers.jl
kellertuer: oh sorry, this was left over, I first though on C one could do the half-plane model, but that would be a bad name to do that here. So I have to delete that line anyways.

src/manifolds/PositiveNumbers.jl
kellertuer: You'Re right, I will unify that

test/positive_numbers.jl
kellertuer: No, it's 3 points on `PositiveNumbers()` not one point on the `PositiveVectors(3)`.

mateuszbaran: Yes, that's right, what I had in mind was adding another call to `test_manifold` below that tests the manifold `PositiveVectors(3)`.

mateuszbaran: Just to make sure nothing breaks there.

kellertuer: It's merely an alias for a PowerManifold, so that should not be necessary,

mateuszbaran: I know but it's easy to write functions on `PositiveNumbers` in a way that's not compatible with `PowerManifold`. I would even say, it most likely won't work as is since it generally assumes that points are represented by array-like objects (including points on submanifolds). With Circle/Torus we actually allowed points on Circle to be represented by one-element arrays.

kellertuer: Ok, you might be right, I'll do such a test tomorrow and maybe add also the array variants if necessary.

kellertuer: had a minute - there it is.

src/manifolds/PositiveNumbers.jl
```diff
@@ -0,0 +1,241 @@
...
+@doc raw"""
+    distance(M::PositiveNumbers, p, q)
+
+Compute the distance on the [`PositiveNumbers`](@ref) `M`, which is
+
+````math
+d(p,q) = \Bigl\lvert \log \frac{p}{q} \Bigr\rvert = \lvert \log p - \log q\rvert.
+````
+"""
+distacne(::PositiveNumbers, ::Any, ::Any)
```
mateuszbaran, line 82:
```suggestion
distance(::PositiveNumbers, ::Any, ::Any)
```

src/manifolds/PositiveNumbers.jl
```diff
@@ -0,0 +1,241 @@
...
+
+distance(::PositiveNumbers, p::Real, q::Real) = abs(log(p / q))
+distacne(::PositiveNumbers, p, q) = sum(abs.(log.(p ./ q)))
```
mateuszbaran, line 85:
```suggestion
distance(::PositiveNumbers, p, q) = sum(abs.(log.(p ./ q)))
```

PR #219, opened 2020-09-10
Poincaré models for Hyperbolic space

kellertuer:
I started testing how different formats for the same manifold work, i.e. besides the arrays, which are assumed to be form the hyperbolic model implicitly there is now
* the Hyperboloid model
* the Poincaré ball model
* the Poincaré half plane model (maybe better to name it half space)?

Where all have types for their `MPoints` and `TVectors`. All concerning converts have been implemented. Still todo 

- [x] exp/log/inner for the plane model
- [x] check which models do not directly work, especially those that originally came from the embedding (which only holds for the first model)
- [x] check whether we have to extend the `allocate_result` functions
- [x] add tests. 
- [x] convert tangent vectors


src/manifolds/Hyperbolic.jl
```diff
@@ -31,12 +31,223 @@ struct Hyperbolic{N} <: AbstractEmbeddedManifold{ℝ,DefaultIsometricEmbeddingTy
...
+end
+
+@doc raw"""
+    PoincareBallPoint <: MPoint
+
+A point on the [`Hyperbolic`](@ref) manifold $ℍ^n$ can be represented as a vector of norm
+less than one in $\mathbb R^n$.
+"""
+struct PoincareBallPoint{T} <: MPoint
+    value::Vector{T}
```
mateuszbaran, line 77:
```suggestion
struct PoincareBallPoint{TValue<:AbstractVector} <: MPoint
    value::TValue
```
So that it can wrap around any vector representation.

kellertuer: Cool. I will adapt the others, too.

kellertuer:
I noticed something interesting:

While to define your own allocate_result boils down to defining a function for `allocate_result(M, log, p, q)`, i.e. arguments after the function name),
for `allocate_result_type` one has to define `allocate_result_type(M, log, (p,q))`, i.e. putting the arguments in a tuple. Would it be reasonable to unify that to the first? Makes defining those nicer. But actually it only affects log/inverse_retract since they are specific in the type, since from two `MPoint`s we have to specify the fitting `TVector`.
Maybe that could be globally done by defining a map from `MPoint` to `TVector` for these types?

mateuszbaran:
That's a good idea. There may be some minor things to work out though. Type conversion would have to be more precise (forwarding the type argument of the wrapper) like `vector_type(::Type{HyperboloidPoint{T}}) where {T} = HyperboloidTVector{T}` (not tested).

kellertuer:
So now the main thing missing might be either look for log/exp in the different 2 other models or derive conversions for tangent vectors accordingly (taking the derivative of the conversion functions). And it might be worth doing both.

kellertuer:
despite two conversions being done as double conversions, one conversion is missing, namely how to transform Poincaré ball tangent vectors to Hyperboloid ones. Still have to look for that formula (or derive it from the push forward of the isometry myself).

kellertuer:
All methods should now be implemented, all formula documented (derived the last one by hand just now); still have to check whether the remaining exp/log/retract methods are failing due to a typo in the formulae or whether that's numerical instability.

kellertuer:
It seems, that conversion between both Poincaré models still has a bug.

kellertuer:
I also checked for the nightlies, locally, the latest Nightly on my machine does not have an error with the ambiguities...

mateuszbaran:
> I also checked for the nightlies, locally, the latest Nightly on my machine does not have an error with the ambiguities...

Nightlies stopped having this flood of ambiguity errors since that PR: https://github.com/JuliaLang/julia/pull/37484 was merged. Now it's just the old problem with `Grassmann`.

kellertuer:
> Nightlies stopped having this flood of ambiguity errors since that PR: [JuliaLang/julia#37484](https://github.com/JuliaLang/julia/pull/37484) was merged. Now it's just the old problem with `Grassmann`.

On todays Nightly also Grassmann tests pass :) So I think the GitHub Nightly is a little outdated and we just have to wait.

src/groups/group.jl
```diff
@@ -153,6 +153,7 @@ Base.copyto!(e::TE, ::TE) where {TE<:Identity} = e
 Base.copyto!(p, ::TE) where {TE<:Identity} = copyto!(p, e.p)
 Base.copyto!(p::AbstractArray, e::TE) where {TE<:Identity} = copyto!(p, e.p)
 
+Base.isapprox(e1::Identity, e2::Identity; kwargs...) = isapprox(e.group, e1, e2; kwargs...)
```
mateuszbaran, line 156:
```suggestion
Base.isapprox(e1::Identity{G}, e2::Identity{G}; kwargs...) where {G<:AbstractGroupManifold} = true
```
Could you also add a test for this?

mateuszbaran: Also other new methods in this file seem to lack tests.

src/manifolds/Hyperbolic.jl
```diff
@@ -21,91 +21,193 @@ Note that while the [`MinkowskiMetric`](@ref) renders the [`Lorentz`](@ref) mani
...
+Generate the Hyperbolic manifold of dimension `n`.
 """
 struct Hyperbolic{N} <: AbstractEmbeddedManifold{ℝ,DefaultIsometricEmbeddingType} end
 
 Hyperbolic(n::Int) = Hyperbolic{n}()
 
+@doc raw"""
+    HyperboloidPoint <: MPoint
+
+In the Hyperboloid model of the [`Hyperbolic`](@ref) $ℍ^n$ points are represented
```
mateuszbaran, line 43: Using ℍ^n for the hyperbolic manifold clashes with our symbol for quaternions.

src/manifolds/Hyperbolic.jl
```diff
@@ -233,9 +367,31 @@ where $⟨\cdot,\cdot⟩_p$ denotes the inner product in the tangent space at `p
...
+        p::$P,
+        X::$T,
+        q::$P,
+        m::ParallelTransport,
+    )
+        Y.value .=
+            convert(
+                $T,
+                (
+                    convert(Vector, q),
```
mateuszbaran, line 383:
```suggestion
                    convert(AbstractVector, q),
```
Similarly below.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -0,0 +1,291 @@
...
+    X::HyperboloidTVector;
+    kwargs...,
+)
+    return check_tangent_vector(M, p.value, X.value; kwargs...)
+end
+
+function convert(::Type{HyperboloidTVector}, X::T) where {T<:AbstractVector}
+    return HyperboloidTVector(X)
+end
+function convert(::Type{HyperboloidTVector}, (p, X)::Tuple{T,T}) where {T<:AbstractVector}
```
mateuszbaran, line 52: What do you need this method for?

mateuszbaran (commented):
There is so much converting going on... I'll take a closer look at it later.

src/groups/group.jl
kellertuer: That was a way trying to reduce ambiguities, that did not work, but I had hoped to have added tests, will check again.

mateuszbaran: I almost gave up reducing ambiguities related to `Identity`. Unless it's an ambiguity in a function call I care about, I just let it be.

src/manifolds/HyperbolicHyperboloid.jl
kellertuer: Classically for conversion of tangents `X` you compute the push forward \pi_*(p)[X]. Thats why you need `(p,X)`as input for tangent vector conversions. To have this unifies between all 4 types (Array, Hyperboloid, PoincareBall, PoincareHalfplane), I implemented conversions from and to any of these 4 types. Yes, this is just wrapping an array in the HyperboloidTVector – for in this case we do not need `p`, but in the general interface we do.

src/groups/group.jl
kellertuer: Then I will remove this just again.

src/manifolds/HyperbolicHyperboloid.jl
mateuszbaran: To be more specific, why is the second argument a tuple?

kellertuer: The first (`p`) is a point, the second (`X`) is the tangent vector. Since Arrays (Vectors to be precise) representing our points/vectors do not have types, its just a plain tuple of two arrays.

src/manifolds/Hyperbolic.jl
mateuszbaran: It would be a good idea to document our usage of ℍ in the notation section of documentation.

src/manifolds/HyperbolicHyperboloid.jl
mateuszbaran: Sorry, I didn't see your response when I wrote my second comment. I'll think about this converting.

src/manifolds/Hyperbolic.jl
kellertuer: Oh, I just copied that from the initial docstring of `Hyperbolic`, but you're right; that's a conflict. That's sad, since I like the symmetry of the Sphere and Hyperbolic both having the math Capital letters.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -0,0 +1,291 @@
...
+"""
+function convert(::Type{HyperboloidPoint}, p::PoincareBallPoint)
+    return HyperboloidPoint(convert(AbstractVector, p))
+end
+function convert(::Type{<:AbstractVector}, p::PoincareBallPoint)
+    return 1 / (1 - norm(p.value)^2) .* [(2 .* p.value)..., 1 + norm(p.value)^2]
+end
+
+@doc raw"""
+    convert(::Type{HyperboloidPoint, p::PoincareHalfSpacePoint)
```
mateuszbaran, line 91:
```suggestion
    convert(::Type{HyperboloidPoint}, p::PoincareHalfSpacePoint)
```

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -0,0 +1,291 @@
...
+    return HyperboloidTVector(convert(AbstractVector, (p, X)))
+end
+function convert(
+    ::Type{T},
+    (p, X)::Tuple{PoincareBallPoint,PoincareBallTVector},
+) where {T<:AbstractVector}
+    t = (1 - norm(p.value)^2)
+    den = 4 * dot(p.value, X.value) / (t^2)
+    c1 = (2 / t) .* X.value + den .* p.value
+    return [c1..., den]
```
mateuszbaran, line 138:
```suggestion
    return vcat(c1, den)
```
This is a more standard way of doing this.

src/manifolds/HyperbolicPoincareBall.jl
```diff
@@ -0,0 +1,237 @@
...
+
+````math
+π(p) = \frac{1}{\lVert \tilde p \rVert^2 + (p_n+1)^2}
+\begin{pmatrix}2p_1\\⋮\\2p_{n-1}\\\lVert p\rVert^2 - 1\end{pmatrix}.
+````
+"""
+function convert(::Type{PoincareBallPoint}, p::PoincareHalfSpacePoint)
+    return PoincareBallPoint(
+        1 / (norm(p.value[1:(end - 1)])^2 + (last(p.value) + 1)^2) .*
+        [2 .* p.value[1:(end - 1)]..., norm(p.value)^2 - 1],
```
mateuszbaran, line 49:
```suggestion
        vcat(2 .* p.value[1:(end - 1)], norm(p.value)^2 - 1),
```
Same as above

src/manifolds/HyperbolicPoincareBall.jl
```diff
@@ -0,0 +1,237 @@
...
+    ::Type{PoincareBallTVector},
+    (p, X)::Tuple{PoincareHalfSpacePoint,PoincareHalfSpaceTVector},
+)
+    den = norm(p.value[1:(end - 1)])^2 + (last(p.value) + 1)^2
+    scp = dot(p.value, X.value)
+    c1 =
+        (2 / den .* X.value[1:(end - 1)]) .-
+        (4 * (scp + last(X.value)) / (den^2)) .* p.value[1:(end - 1)]
+    c2 = 2 * scp / den - 2 * (norm(p.value)^2 - 1) * (scp + last(X.value)) / (den^2)
+    return PoincareBallTVector([c1..., c2])
```
mateuszbaran, line 159:
```suggestion
    return PoincareBallTVector(vcat(c1, c2))
```
Same as above

src/manifolds/HyperbolicPoincareHalfspace.jl
```diff
@@ -0,0 +1,242 @@
...
+
+````math
+π(p) = \frac{1}{\lVert \tilde p \rVert^2 - (p_n-1)^2}
+\begin{pmatrix}2p_1\\⋮\\2p_{n-1}\\1-\lVert p\rVert^2\end{pmatrix}.
+````
+"""
+function convert(::Type{PoincareHalfSpacePoint}, p::PoincareBallPoint)
+    return PoincareHalfSpacePoint(
+        1 / (norm(p.value[1:(end - 1)])^2 + (last(p.value) - 1)^2) .*
+        [2 .* p.value[1:(end - 1)]..., 1 - norm(p.value)^2],
```
mateuszbaran, line 31:
```suggestion
        vcat(2 .* p.value[1:(end - 1)], 1 - norm(p.value)^2),
```
Same as above.

src/manifolds/HyperbolicPoincareHalfspace.jl
```diff
@@ -0,0 +1,242 @@
...
+    ::Type{PoincareHalfSpaceTVector},
+    (p, X)::Tuple{PoincareBallPoint,PoincareBallTVector},
+)
+    den = norm(p.value[1:(end - 1)])^2 + (last(p.value) - 1)^2
+    scp = dot(p.value, X.value)
+    c1 =
+        (2 / den .* X.value[1:(end - 1)]) .-
+        (4 * (scp - last(X.value)) / (den^2)) .* p.value[1:(end - 1)]
+    c2 = -2 * scp / den - 2 * (1 - norm(p.value)^2) * (scp - last(X.value)) / (den^2)
+    return PoincareHalfSpaceTVector([c1..., c2])
```
mateuszbaran, line 95:
```suggestion
    return PoincareHalfSpaceTVector(vcat(c1, c2))
```

src/manifolds/Hyperbolic.jl
kellertuer: I changed the hyperbolic to `$\mathcal H$` and added real, complex, quaternions to the notations list.

docs/src/misc/notation.md
```diff
@@ -9,6 +9,7 @@ Within the documented functions, the utf8 symbols are used whenever possible, as
 
 | Symbol | Description | Also used | Comment |
 |:--:|:--------------- |:--:|:-- |
+| $ℝ, ℂ, ℍ$ | the set of real, complex and quaternion numbers, respectively. |
```
mateuszbaran, line 12: Shouldn't this respect the alphabetical ordering of the table?

kellertuer: I was not really sure, where to put those (and did not feel like doing three lines). Where do you propose to put this line?

mateuszbaran: Real and complex numbers are already under "field", so maybe comment in that row could just be expanded a bit?

kellertuer: ah, thanks for noticing, did exactly that.

src/manifolds/HyperbolicHyperboloid.jl
mateuszbaran: Could you change the signature here to something like `convert(::Hyperbolic, ::Type{HyperboloidTVector}, p, X)`? We don't have to necessarily take two arguments in `convert`. This method as written is incompatible with typical usage of two-argument `convert` anyway.

kellertuer: Why is It incompatible? Because the Tuple introduces too many ambiguities?
For sure we can `de-Tuple` this, since just with `X` it does not work anyways, we are breaking a little bit anyways;

So as long as we agree on a unified way, I am also fine with `p, X` instead of `(p,X)`, both are fine to dispatch on for me. And you'Re right the first avoids the ambiguity with several `Tuple` converts

kellertuer: Should I then also remove all these combined ones? I tried to then also do a combined “convert both p and X” methods here, since `p` is converted anyways.

mateuszbaran: > Why is It incompatible? Because the Tuple introduces too many ambiguities?

Primarily because someone who wants to use that methods has to make that tuple anyway. Such tuples don't naturally occur in Manifolds.jl. Technically there shouldn't be any ambiguities (in the multiple dispatch sense) I think.

mateuszbaran: > Should I then also remove all these combined ones? I tried to then also do a combined “convert both p and X” methods here, since `p` is converted anyways.

Combined ones would be fine I think but in a slightly different format:
```
convert(::Type{Tuple{HyperboloidPoint,HyperboloidTVector}}, (p, X)::Tuple{T,T}) where {T<:AbstractVector}
```
If both the argument you convert *and* and type you convert to are tuples then it's also fine to return a tuple I think.

mateuszbaran: So I think it's fine to have tuples in all three places (type, converted argument, returned object) or in none.

kellertuer: Din line 160 I have
````julia
function convert(
     ::Type{Tuple{HyperboloidPoint,HyperboloidTVector}},
     (p, X)::Tuple{PoincareBallPoint,PoincareBallTVector},
 )
````
well internally it called the “just `X`” one – but that one is ok then (similarly for a tuple fo arrays of course)?

That would mean, that a you can never `convert` to just a vector, right?
Because if you convert `(p,X)` to `(q,Y)` converting `X` might require `p` but only rarely requires `q` (on Hyperbolic it never requires `q`).
 So if you just want to convert a tangent, oi.e. get `Y` from `(p,X)` one can spare the computation to convert `p` to `q`. That's my intention of this method.

But sure we can keep the tuple ones we have and do a `convert(::HyperboloidTVector, p, X)`for that, or not?

**Comment by kellertuer** [comment]
> There is so much converting going on... I'll take a closer look at it later.

I convert everything to everything (and its 4 types) as well as points to points, point/vector to vector and point&vector to point&vector. We can omit the last case if that is not needed, especially if I change the point/vector->vector interface now anyways.

File: src/manifolds/HyperbolicHyperboloid.jl (hunk 5 above)
Comment (mateuszbaran) [reply src/manifolds/HyperbolicHyperboloid.jl:52]: > well internally it called the “just `X`” one – but that one is ok then (similarly for a tuple fo arrays of course)?

Yes, this looks fine.

> That would mean, that a you can never `convert` to just a vector, right?

With a 2-argument `convert` that requires the point? No, we could not. I think for the case of converting just the tangent vector when the point is also required we should use a three-argument `convert` like you wrote in the last sentence of your message.

> Because if you convert `(p,X)` to `(q,Y)` converting `X` might require `p` but only rarely requires `q` (on Hyperbolic it never requires `q`).

In which case `q` isn't uniquely defined by `p`?

Comment (kellertuer) [reply src/manifolds/HyperbolicHyperboloid.jl:52]: Oh, its always uniquely determined by `p`, sorry if I wrote that wrong. We can just omit the computation to _compute_ `q` (memory allocation and computational time) if its not needed.

And ok, I will convert the tangent vector converts to 3-argument ones. That seems to be the best.

Comment (kellertuer) [reply src/manifolds/HyperbolicHyperboloid.jl]: Thanks, this refactoring made much of the tuple code much clearer and removed 3 ambiguities the tuple-nontuple conversios introduced.

Comment (mateuszbaran) [reply src/manifolds/HyperbolicHyperboloid.jl]: That's great :slightly_smiling_face: .

File: docs/src/features/utilities.md [hunk 13]
```diff
@@ -1,7 +1,15 @@
+# Ease of notation
+
+The following terms introduce a nicer notation for some operations, line using $p ∈ \mathcal M$ to determine, whether $p$ is a point on the [`Manifold`](@ref) $\mathcal M$.
```
mateuszbaran, line 3:
```suggestion
The following terms introduce a nicer notation for some operations, for example using the ∈ operator, $p ∈ \mathcal M$, to determine whether $p$ is a point on the [`Manifold`](@ref) $\mathcal M$.
```

docs/src/misc/notation.md
```diff
@@ -19,7 +19,7 @@ Within the documented functions, the utf8 symbols are used whenever possible, as
 | $d(\cdot,\cdot)$ | (Riemannian) distance | $d_{\mathcal M}(\cdot,\cdot)$ | |
 | $\exp_p X$ | exponential map at $p \in \mathcal M$ of a vector $X \in T_p \mathcal M$ | $\exp_p(X)$ | |
 | $F$ | a fiber | | see [`VectorBundleFibers`](@ref) |
-| $\mathbb F$ | a field | | field a manifold or a basis is based on, usually $\mathbb F \in \{\mathbb R,\mathbb C\}$ |
+| $\mathbb F$ | a field, usually $\mathbb F \in \{\mathbb R,\mathbb C, \mathbb H\}$, i.e. the real, complex, and quaternion numbers, respectively. | |field a manifold or a basis is based on |
```
mateuszbaran, line 22: We could also add `\mathcal H` to this table.

src/manifolds/Hyperbolic.jl
```diff
@@ -1,116 +1,221 @@
...
 
 Hyperbolic(n::Int) = Hyperbolic{n}()
 
+@doc raw"""
+    HyperboloidPoint <: MPoint
+
+In the Hyperboloid model of the [`Hyperbolic`](@ref) $\mathcal H^n$ points are represented
+as vectors in $ℝ^{n+1}$ with [`MinkowskiMetric`](@ref) equal to $-1$.
+
+This representation is the default, i.e. vectors are assumed to have this repesentation.
```
mateuszbaran, line 46:
```suggestion
This representation is the default, i.e. `AbstractVector`s are assumed to have this repesentation.
```
This should be more clear here.

src/manifolds/Hyperbolic.jl
```diff
@@ -1,116 +1,221 @@
...
+less than one in $\mathbb R^n$.
+"""
+struct PoincareBallPoint{TValue<:AbstractVector} <: MPoint
+    value::TValue
+end
+
+@doc raw"""
+    PoincareBallTVector <: TVector
+
+In the Poincaré ball model of the [`Hyperbolic`](@ref) $\mathcal H^n$ tangent vctors are represented
```
mateuszbaran, line 78:
```suggestion
In the Poincaré ball model of the [`Hyperbolic`](@ref) $\mathcal H^n$ tangent vectors are represented
```

src/manifolds/Hyperbolic.jl
```diff
@@ -1,116 +1,221 @@
...
+    value::TValue
+end
+
+@doc raw"""
+    PoincareHalfPlaneTVector <: TVector
+
+In the Poincaré half plane model of the [`Hyperbolic`](@ref) $\mathcal H^n$ tangent vctors are
+represented as vectors in $ℝ^{n}$.
+"""
+struct PoincareHalfSpaceTVector{TValue<:AbstractVector} <: MPoint
```
mateuszbaran, line 101:
```suggestion
struct PoincareHalfSpaceTVector{TValue<:AbstractVector} <: TVector
```

src/manifolds/Hyperbolic.jl
```diff
@@ -1,116 +1,221 @@
...
-function check_manifold_point(M::Hyperbolic, p; kwargs...)
-    mpv =
-        invoke(check_manifold_point, Tuple{supertype(typeof(M)),typeof(p)}, M, p; kwargs...)
-    mpv === nothing || return mpv
-    if !isapprox(minkowski_metric(p, p), -1.0; kwargs...)
-        return DomainError(
-            minkowski_metric(p, p),
-            "The point $(p) does not lie on $(M) since its Minkowski inner product is not -1.",
-        )
+struct HyperboloidTVector{TValue<:AbstractVector} <: MPoint
```
mateuszbaran, line 61:
```suggestion
struct HyperboloidTVector{TValue<:AbstractVector} <: TVector
```

src/manifolds/Hyperbolic.jl
```diff
@@ -1,116 +1,221 @@
...
+half plane, i.e. $x ∈ ℝ^n$ with $x_d > 0$.
+"""
+struct PoincareHalfSpacePoint{TValue<:AbstractVector} <: MPoint
+    value::TValue
+end
+
+@doc raw"""
+    PoincareHalfPlaneTVector <: TVector
+
+In the Poincaré half plane model of the [`Hyperbolic`](@ref) $\mathcal H^n$ tangent vctors are
```
mateuszbaran, line 98:
```suggestion
In the Poincaré half plane model of the [`Hyperbolic`](@ref) $\mathcal H^n$ tangent vectors are
```

src/manifolds/Hyperbolic.jl
kellertuer: Oops. Thanks for noticing, luckily our implementations are not that dependent, so everything still works.

docs/src/misc/notation.md
kellertuer: Currently we do not list symbols for manifolds, so I would maybe not add a `\mathcal H` here, since it would refer to the hyperbolic space?

mateuszbaran: Ah, right, OK.

src/manifolds/HyperbolicPoincareBall.jl
```diff
@@ -0,0 +1,237 @@
...
+    if !(norm(p.value) < 1)
+        return DomainError(
+            norm(p.value),
+            "The point $(p) does not lie on $(M) since its norm is not less than 1.",
+        )
+    end
+end
+
+@doc raw"""
+    convert(::Type{PoincareBallPoint}, x::HyperboloidPoint)
```
mateuszbaran, line 13: Why do you use `x` here instead of `p`?

kellertuer: No special reason, just missed that after transferring the formula, I changed it.

mateuszbaran (approved):
Looks OK now 👍 .

mateuszbaran:
Oh well... it fails on MacOS but not Ubuntu or Windows...

mateuszbaran:
I think it works now :tada:.

Issue #159, opened 2020-04-09
Projective spaces


mateuszbaran:
Maybe we should add a `Projective space` page to the list of basic manifolds that explains that projective spaces should be represented by `Grassmann` with `k=1`? We could also add faster `exp!` and `log!` for that case.

sethaxen:
That's a good idea, although to fully support projective spaces, e.g. for directional statistics, it might make more sense to do make it its own type. e.g. how `Sphere` is implemented separately from `Stiefel`. This would let us keep the usual notation, that is, `ProjectiveSpace(n, field=𝔽)` would be represented in `𝔽^{n+1}`, as opposed to `Grassmann(n, 1, field=𝔽)`, which is represented in `𝔽^{n}`

Now that I think of it, maybe `Stiefel` should have some `k=1` overloads that call the same functions for the `Sphere` where it's more efficient.

mateuszbaran:
Yes, making projective spaces their own type may be even better.

PR #220, opened 2020-09-18
Add ProjectiveSpace manifold

sethaxen:
This PR resolves https://github.com/JuliaManifolds/Manifolds.jl/issues/159 by adding a `ProjectiveSpace(n)` manifold that is to `Grassmann(n+1, 1)` what `Sphere(n)` is to `Stiefel(n + 1, 1)`. It mostly adapts functionality from `Grassmann`, implemented more efficiently. Modeled after `Sphere`/`ArraySphere`, it uses an `AbstractProjectiveSpace` to implement also an `ArrayProjectiveSpace`.

One minor difference in convention from `Grassmann` is that the final `qr` in `Grassmann`'s `exp`onential standardizes the point (i.e. deterministically selects one of the equivalent points). `exp!` for `ProjectiveSpace` does not perform this step, since it isn't necessary.

to-do:
- [x] add tests
- [x] add docstrings
- [x] implement basis functions
- [x] implement vector transport functions

edit: Along the way, I made several fixes to docstrings in `Grassmann` and `Sphere`, and since many of the implementations are similar to `Sphere`, I also improved efficiency and stability in several `Sphere` methods. It changes the `DefaultOrthonormalBasis` for `Sphere` to be a different more numerically stable implementation, so this is a breaking change.

### DIFF
--- a/src/manifolds/ProjectiveSpace.jl
+++ b/src/manifolds/ProjectiveSpace.jl
@@ -188,42 +188,50 @@ function get_basis(::ProjectiveSpace{n,ℝ}, p, B::DiagonalizingOrthonormalBasis
 end
 
 @doc raw"""
-    get_coordinates(M::AbstractProjectiveSpace{ℝ}, p, X, B::DefaultOrthonormalBasis)
+    get_coordinates(M::AbstractProjectiveSpace, p, X, B::DefaultOrthonormalBasis{ℝ})
 
 Represent the tangent vector `X` at point `p` from the [`AbstractProjectiveSpace`](@ref) `M`
-in an orthonormal basis by rotating the vector `X` using the rotation matrix
-$2\frac{q q^\mathrm{T}}{q^\mathrm{T} q} - I$ where $q = p + (1, 0, …, 0)$, which takes `p`
-to $(1, 0, …, 0)$.
+in an orthonormal basis by rotating the vector `X` to `Y`:
+````math
+\begin{pmatrix} 0 \\ Y \end{pmatrix} = \left(2\frac{q q^\mathrm{H}}{q^\mathrm{H} q} - I\right) X,
+````
+where $q = p λ + (1, 0, …, 0)$, where $λ=\frac{\overline{p_1}}{|p_1|}$, and $\overline{⋅}$
+denotes complex or quaternionic conjugation.
 """
 get_coordinates(::AbstractProjectiveSpace{ℝ}, p, X, ::DefaultOrthonormalBasis)
 
-function get_coordinates!(M::AbstractProjectiveSpace{ℝ}, Y, p, X, ::DefaultOrthonormalBasis)
-    n = manifold_dimension(M)
-    pend, Xend = view(p, 2:(n + 1)), view(X, 2:(n + 1))
-    factor = X[1] / (1 + p[1]) # 2 (q'X)/(q'q)
+function get_coordinates!(M::AbstractProjectiveSpace, Y, p, X, ::DefaultOrthonormalBasis{ℝ})
+    m = length(p)
+    p1 = p[1]
+    λ = sign(p1')
+    pend, Xend = view(p, 2:m), view(X, 2:m)
+    factor = λ * X[1] / (1 + abs(p1)) # 2 λ (q'X)/(q'q)
     Y .= pend .* factor .- Xend
     return Y
 end
 
 @doc raw"""
-    get_vector(M::AbstractProjectiveSpace{ℝ}, p, X, B::DefaultOrthonormalBasis)
+    get_vector(M::AbstractProjectiveSpace, p, X, B::DefaultOrthonormalBasis{ℝ})
 
 Convert a one-dimensional vector of coefficients `X` in the basis `B` of the tangent space
 at `p` on the [`AbstractProjectiveSpace`](@ref) `M` to a tangent vector `Y` at `p`, given by
 ````math
-Y = \left(2\frac{q q^\mathrm{T}}{q^\mathrm{T} q} - I\right) \begin{pmatrix} 0 \\ X \end{pmatrix},
+Y = \left(2\frac{q q^\mathrm{H}}{q^\mathrm{H} q} - I\right) \begin{pmatrix} 0 \\ X \end{pmatrix},
 ````
-where $q = p + (1, 0, …, 0)$.
+where $q = p λ + (1, 0, …, 0)$, $λ=\frac{\overline{p_1}}{|p_1|}$, and $\overline{⋅}$
+denotes complex or quaternionic conjugation.
 """
-get_vector(::AbstractProjectiveSpace{ℝ}, p, X, ::DefaultOrthonormalBasis)
+get_vector(::AbstractProjectiveSpace, p, X, ::DefaultOrthonormalBasis{ℝ})
 
-function get_vector!(M::AbstractProjectiveSpace{ℝ}, Y, p, X, ::DefaultOrthonormalBasis)
-    n = manifold_dimension(M)
-    pend = view(p, 2:(n + 1))
-    Y1 = dot(pend, X)
-    Y[1] = Y1
-    factor = Y1 / (1 + p[1])
-    Y[2:(n + 1)] .= pend .* factor .- X
+function get_vector!(M::AbstractProjectiveSpace, Y, p, X, ::DefaultOrthonormalBasis{ℝ})
+    m = length(p)
+    p1 = p[1]
+    λ = sign(p1)
+    pend = view(p, 2:m)
+    pX = dot(pend, X)
+    Y[1] = λ * pX
+    factor = pX / (1 + abs(p1)) # 2 (q'X)/(q'q)
+    Y[2:m] .= pend .* factor .- X
     return Y
 end
 
--- a/src/manifolds/ProjectiveSpace.jl
+++ b/src/manifolds/ProjectiveSpace.jl
@@ -203,10 +203,10 @@ function get_coordinates!(M::AbstractProjectiveSpace, Y, p, X, ::DefaultOrthonor
     m = length(p)
     z = p[1]' # p'[1,0,…,0]
     cosθ = abs(z)
-    λ = sign_from_abs(z, cosθ)
     pend, Xend = view(p, 2:m), view(X, 2:m)
     factor = λ * X[1] / (1 + cosθ) # 2 λ (q'X)/(q'q)
     Y .= pend .* factor .- Xend
+    λ = nzsign(z, cosθ)
     return Y
 end
 
@@ -228,8 +228,8 @@ function get_vector!(M::AbstractProjectiveSpace, Y, p, X, ::DefaultOrthonormalBa
     m = length(p)
     z = p[1] # [1,0,…,0]'p
     cosθ = abs(z)
-    λ = sign_from_abs(z, cosθ)
     pend = view(p, 2:m)
+    λ = nzsign(z, cosθ)
     pX = dot(pend, X)
     Y[1] = λ * pX
     factor = pX / (1 + cosθ) # 2 (q'X)/(q'q)
@@ -326,7 +326,7 @@ log(::AbstractProjectiveSpace, p, q)
 function log!(M::AbstractProjectiveSpace, X, p, q)
     z = dot(q, p)
     cosθ = abs(z)
-    λ = sign_from_abs(z, cosθ)
+    λ = nzsign(z, cosθ)
     X .= (q .* λ .- cosθ .* p) ./ usinc_from_cos(cosθ)
     return project!(M, X, p, X)
 end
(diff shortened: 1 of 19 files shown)


sethaxen:
Anyone know why real centered matrices tests are failing? They're not touched by this PR. See e.g. https://github.com/JuliaManifolds/Manifolds.jl/runs/1132846172?check_suite_focus=true#step:5:155

mateuszbaran:
Cool! Thanks for working on it :+1: 

> Anyone know why real centered matrices tests are failing?

Sorry, this may be caused by https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/48 . We are currently figuring out how to improve embedded manifolds. See https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/49 for the ongoing work.

sethaxen:
Okay, this PR is ready for review.

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+eval(
+    quote
+        @invoke_maker 1 Manifold injectivity_radius(
+            M::AbstractProjectiveSpace,
+            rm::AbstractRetractionMethod,
+        )
+    end,
+)
+
+function inverse_retract!(
```
kellertuer, line 210: Could you add a docstring to document similarities and differences to the sphere?

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+    signz = sign(dot(p, q))
+    X .= q .* signz' .- p
+    return X
+end
+
+function Base.isapprox(::AbstractProjectiveSpace, p, q; kwargs...)
+    return isapprox(abs2(dot(p, q)), 1; kwargs...)
+end
+
+function log!(M::AbstractProjectiveSpace, X, p, q)
```
kellertuer, line 226: similar to exp!, please document.

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+Generate a distribution in the tangent space at `p` by generating a
+normal distribution in ambient space with standard deviation `σ`
+projected to the tangent space at `p`.
+"""
+function normal_tvector_distribution(M::ProjectiveSpace{n,ℝ}, p, σ) where {n}
+    d = Distributions.MvNormal(zero(p), σ)
+    return ProjectedFVectorDistribution(TangentBundleFibers(M), p, d, project!, p)
+end
+
+project!(::AbstractProjectiveSpace, q, p) = (q .= p ./ norm(p))
```
kellertuer, line 289: Please document, here the same as for distance with the quotient manifold holds as a remark.

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+@doc raw"""
+    representation_size(M::AbstractProjectiveSpace)
+
+Return the size points on the [`AbstractProjectiveSpace`](@ref) `M` are represented as,
+i.e., the representation size of the embedding.
+"""
+@generated representation_size(::ArrayProjectiveSpace{N}) where {N} = size_to_tuple(N)
+@generated representation_size(::ProjectiveSpace{N}) where {N} = (N + 1,)
+
+function retract!(
```
kellertuer, line 302: For me, this just looks like the `ProjectionRetraction`, why should it also be triggered by QU or Polar? I can't see that here.

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+"""
+function uniform_distribution(M::ProjectiveSpace{n,ℝ}, p) where {n}
+    d = Distributions.MvNormal(zero(p), 1.0)
+    return ProjectedPointDistribution(M, d, project!, p)
+end
+
+@doc raw"""
+    vector_transport_to(M::AbstractProjectiveSpace, p, X, q, method::ParallelTransport)
+
+Parallel transport a vector `X` from the tangent space at a point `p` on the
```
kellertuer, line 334: A formula (maybe one that already holds for the Abstract one in general) would be nice,
Otherwise (since this should default back to direction?) There is no additional information in this string compared to the `Manifold` case?

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+[`AbstractProjectiveSpace`](@ref) `M` by interpreting it as an element of the embedding and
+then projecting it onto the tangent space at `q`.
+"""
+vector_transport_to(::AbstractProjectiveSpace, ::Any, ::Any, ::Any, ::ProjectionTransport)
+
+@doc raw"""
+    vector_transport_direction(M::AbstractProjectiveSpace, p, X, d, method::ParallelTransport)
+
+Parallel transport a vector `X` from the tangent space at a point `p` on the
+[`AbstractProjectiveSpace`](@ref) `M` along the geodesic in the direction indicated by the
```
kellertuer, line 353: could we add the formula here, for completeness?

src/manifolds/Sphere.jl
```diff
@@ -365,7 +366,7 @@ norm for the case $m>1$.
 """
 project(::AbstractSphere, ::Any)
 
-project!(S::AbstractSphere, q, p) = copyto!(q, p ./ norm(p))
+project!(::AbstractSphere, q, p) = (q .= p ./ norm(p))
```
kellertuer, line 369: I though we wanted to switch to `copyto!`? Or is that only if we speak about identity operations?

src/manifolds/Sphere.jl
```diff
@@ -412,12 +413,12 @@ function Base.show(io::IO, ::ArraySphere{N,𝔽}) where {N,𝔽}
...
 """
-    uniform_distribution(M::AbstractSphere, p)
+    uniform_distribution(M::Sphere{n,ℝ}, p) where {n}
 
-Uniform distribution on given [`AbstractSphere`](@ref) `M`. Generated points will be of
+Uniform distribution on given [`Sphere`](@ref) `M`. Generated points will be of
 similar type as `p`.
 """
-function uniform_distribution(M::AbstractSphere, p)
+function uniform_distribution(M::Sphere{n,ℝ}, p) where {n}
```
kellertuer, line 421: I had thought with the projection, this should also work for the `AbstractSphere`; why was this reduced to the sphere?

src/tests/tests_general.jl
```diff
@@ -257,9 +257,8 @@ function ManifoldTests.test_manifold(
 
         X3 = log(M, pts[1], pts[3])
 
-        Test.@test real(inner(M, pts[1], X1, X3)) ≈ real(inner(M, pts[1], X3, X1))
-        Test.@test imag(inner(M, pts[1], X1, X3)) ≈ -imag(inner(M, pts[1], X3, X1))
-        Test.@test imag(inner(M, pts[1], X1, X1)) ≈ 0
+        Test.@test inner(M, pts[1], X1, X3) ≈ conj(inner(M, pts[1], X3, X1))
```
kellertuer, line 260: Nice fix!

kellertuer (changes requested):
Thanks, this already looks quite well implemented

* a few more formulae would be cool; one thing I personally like about our docs is, that I can always also look up the math formulae in a concise way. So my main point is, a little more docstrings.
* a test for `get_vector`/`get_coordinates` would be great
*  injectivity_radius seems to not be covered, but that might be a problem we have in most manifolds.

but overall test coverage is already great.

src/manifolds/ProjectiveSpace.jl
```diff
@@ -0,0 +1,392 @@
...
+space, where `field` can also be used to generate the complex- and right-quaternionic
+projective spaces.
+"""
+struct ArrayProjectiveSpace{N,𝔽} <: AbstractProjectiveSpace{𝔽} where {N<:Tuple} end
+function ArrayProjectiveSpace(n::Vararg{Int,I}; field::AbstractNumbers = ℝ) where {I}
+    return ArrayProjectiveSpace{Tuple{n...},field}()
+end
+
+function allocation_promotion_function(::AbstractProjectiveSpace{ℂ}, f, args::Tuple)
+    return complex
```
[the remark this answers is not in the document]
sethaxen: It is tested at https://github.com/JuliaManifolds/Manifolds.jl/blob/d8ef7564ba68cc37ada49afc747af6f7f333c324/test/projective_space.jl#L105, but codecov does not register it for some reason. TBH, I'm not certain what this function does; I copied it from `Grassmann{n,1}`.

src/manifolds/ProjectiveSpace.jl
sethaxen: For `Grassmann(n,1)`, all three retractions converge (with maybe different sign conventions between `QRRetraction` and `PolarRetraction`). The `PolarRetraction` for `Grassmann` is I am quite sure also its `ProjectionRetraction` (`project` isn't implemented for `Grassmann`, but it is for `GeneralizedGrassmann`, and that's what `PolarRetraction` is doing).

src/manifolds/ProjectiveSpace.jl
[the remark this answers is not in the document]
kellertuer: I know what it does: for a complex manifold the `eltype` of the return value from `exp!` is complex; that's why I think it can be implemented on a `Manifold{ℂ}` - level. But you're right, it sometimes does not get found by codecov, sadly.

Maybe I try that idea later to move this to the supertype.

src/manifolds/ProjectiveSpace.jl
kellertuer: With “converge” you mean they are all the same? (Just my english maybe, I think of a limit-thingy of analysis/continuity when reading converge and would call this, they “fall together” would be quite german maybe, so maybe coincide?)

Then a small note in the docs would be nice, that they do coincide.

src/tests/tests_general.jl
sethaxen: Necessary for testing quaternion manifolds, for which `real` and `conj` are well-defined, but `imag` is not.

src/manifolds/Sphere.jl
sethaxen: Distributions is fairly aggressive with checking types in the constructor, so `MvNormal` will error if passed any `p` that is not a real vector. So I guess could I could make the type `M::Union{Sphere{n,ℝ},ArraySphere{Tuple{n},ℝ}}`.

One way to extend to the full `AbstractSphere` would be to have a function that represents `p` as a real vector by flattening and interleaving (e.g. complex numbers become 2 real numbers), then replacing `project!` with a function that first un-flattens then projects. But that seems outside the scope of this PR.

src/manifolds/Sphere.jl
sethaxen: Yeah just for identity. If we use `copyto!` here, it will allocate and then copy, whereas this is non-allocating.

src/manifolds/ProjectiveSpace.jl
sethaxen: Yeah, that's right, coincide is perhaps a better way of saying it.

src/manifolds/Sphere.jl
kellertuer: Ah, so theoretically it would work but `MvNormal` is a little too strictly typed. No let's not make it too complicated, it can stay this way then.

src/manifolds/Sphere.jl
kellertuer: Cool, thanks for the explanation. As I already mentioned, I will migrate code when I see it.

src/manifolds/ProjectiveSpace.jl
```diff
@@ -207,27 +252,84 @@ eval(
     end,
 )
 
+@doc raw"""
+    inverse_retract(M::AbstractProjectiveSpace, p, q, method::ProjectionInverseRetraction)
+    inverse_retract(M::AbstractProjectiveSpace, p, q, method::PolarInverseRetraction)
+    inverse_retract(M::AbstractProjectiveSpace, p, q, method::QRInverseRetraction)
+
+Compute the inverse retraction for the [`ProjectionInverseRetraction`](@ref) $𝔽ℙ^n$,
+[`PolarRetraction`](@ref), and [`QRInverseRetraction`](@ref)
```
kellertuer, line 249:
```suggestion
[`PolarInverseRetraction`](@ref), and [`QRInverseRetraction`](@ref)
```

sethaxen:
This PR should be ready for another review.
Note that it slightly changes the behavior of `get_vector(::Sphere, ...)` to maintain consistency with the formula in the docstring, which could be considered a breaking change.

sethaxen:
I noticed that the `DefaultOrthonormalBasis` for `Sphere` was called a "rotation", but it's actually a Householder transformation `H(p+x)`, where `H(u) = I - 2/(u'u) * u*u'`, i.e. a reflection, and `x` is the `x`-axis. Because it is a reflection, if `p` is actually aligned with `x`, it still negates all the vectors to make the basis. This is easily fixed. The transformation `H(2x)H(p+x)` is unitary and takes `p` to `x`. I'd like to use it for `ProjectiveSpace`, and it makes sense to implement it for `Sphere` too.

kellertuer:
Sure, sounds reasonable.

kellertuer (approved):

PR #221, opened 2020-09-20
Fixes Complex Tangent Spaces for Grassmann and Stiefel

kellertuer:
Thanks to @sethaxen for noticing.


src/manifolds/Stiefel.jl
```diff
@@ -90,9 +91,9 @@ function check_tangent_vector(
         kwargs...,
     )
     mpv === nothing || return mpv
-    if !isapprox(p' * X + X' * p, zeros(k, k); kwargs...)
+    if !isapprox(p' * X + conj(X' * p), zeros(k, k); kwargs...)
```
sethaxen, line 94: Wouldn't this be safer, since equality checks with zeros often require much stricter tolerances (so with default tolerances these will often fail)?
```suggestion
    if !isapprox(p' * X, -conj(X' * p); kwargs...)
```

src/manifolds/Grassmann.jl
```diff
@@ -278,12 +278,12 @@ In this formula the $\operatorname{atan}$ is meant elementwise.
...
 log(::Grassmann, ::Any...)
 
-function log!(M::Grassmann, X, p, q)
+function log!(::Grassmann{n,k}, X, p, q) where {n,k}
     z = q' * p
     At = q' - z * p'
     Bt = z \ At
     d = svd(Bt')
-    return copyto!(X, d.U * Diagonal(atan.(d.S)) * d.Vt)
+    return copyto!(X, d.U[:, 1:k] * Diagonal(atan.(d.S[1:k])) * d.Vt[1:k, :])
```
sethaxen, line 286: Perhaps something like this to reduce number of allocations?
```suggestion
    return X .= view(d.U, :, 1:k) * Diagonal(atan.(view(d.S, 1:k))) * view(d.Vt, 1:k, :))
```

src/manifolds/Grassmann.jl
```diff
@@ -149,7 +149,7 @@ function distance(M::Grassmann, p, q)
     p ≈ q && return zero(real(eltype(p)))
     a = svd(p' * q).S
     a[a .> 1] .= 1
-    return sqrt(sum((acos.(a)) .^ 2))
+    return sqrt(sum((real(acos.(a))) .^ 2))
```
sethaxen, line 152: The above inequality check suggests that `a` contains reals (inequalities fail for complex). `acos` will then only produce reals, so why is the call to `real` needed?
Also, you could use `sum(x -> abs2(real(acos(x))), a)` to avoid allocating `acos.(a)`.

src/manifolds/Grassmann.jl
```diff
@@ -215,7 +215,7 @@ g_p(X,Y) = \operatorname{tr}(X^{\mathrm{H}}Y),
 
 where $\cdot^{\mathrm{H}}$ denotes the complex conjugate transposed or Hermitian.
 """
-inner(::Grassmann, p, X, Y) = dot(X, Y)
+inner(::Grassmann, p, X, Y) = dot(abs.(X), abs.(Y))
```
sethaxen, line 218: I thought we inherited the metric from the embedding, and this change differs from manopt's implementation. Can you explain this?

kellertuer: Oh, sorry the abs - es are from some checks, they were left there by error.

src/manifolds/Grassmann.jl
kellertuer: That's merely for numerical precision to add the real, I think.

src/manifolds/Stiefel.jl
kellertuer: nice, sure.

src/manifolds/Grassmann.jl
sethaxen: The line `a[a .> 1] .= 1` will error if `eltype(a) <: Complex`:

```julia
julia> (1.0 + 0im) < 3
ERROR: MethodError: no method matching isless(::Complex{Float64}, ::Int64)
Closest candidates are:
  isless(::Missing, ::Any) at missing.jl:87
  isless(::AbstractFloat, ::Real) at operators.jl:167
  isless(::Real, ::Real) at operators.jl:355
```
How does `real(x::Real)` add numerical precision? Isn't that a no-op?

kellertuer: For the first: good observation, I will look how to fix that.

For the second: Until now we mostly thought of this lines in the real case, yes. The real stems from here https://www.manopt.org/reference/manopt/manifolds/grassmann/grassmanncomplexfactory.html - line 63, if you have rounding errors and end up with complex numbers I think.

kellertuer: maybe it would then be better to do a `real(a)`?

sethaxen:
Apparently I can no longer comment on lines not changed on GitHub. I had noticed something else you could include in this PR if you wish.
https://github.com/JuliaManifolds/Manifolds.jl/blob/bfd2445c9cc7bfde95c955e66dfd8cd1d009a11c/src/manifolds/Grassmann.jl#L385-L388

could be improved to

```julia
    # avoid promoting to Float64 unless necessary
    # also sign is idempotent
    D = Diagonal(sign.(d .+ 1//2))
    # avoid filling q before overwriting
    # multiply in-place
    mul!(q, Array(qrfac.Q), D)
    return q
```

kellertuer:
I don't know where the double sign comes from, but that's why its good to review code from time to time; I will change it to your lines.

src/manifolds/Grassmann.jl
sethaxen: Hm, perhaps matlab handles `acos(1.1)` by extending to the complex plane with `acos(1.1+0im) => 0.0 - 0.4435682543851153im`? Since the code does not clamp `cos_princ_angle`, then one would need to call `real` on the output, which is effectively the same as our clamping.

That being said, I am pretty sure singular values are always real numbers, so calling `real(a)` should not be necessary.

kellertuer: Ok, I left it out again.

src/manifolds/Grassmann.jl
```diff
@@ -145,11 +145,11 @@ b_{i}=\begin{cases}
 \end{cases}
 ````
 """
-function distance(M::Grassmann, p, q)
+function distance(::Grassmann, p, q)
     p ≈ q && return zero(real(eltype(p)))
     a = svd(p' * q).S
     a[a .> 1] .= 1
-    return sqrt(sum((acos.(a)) .^ 2))
+    return sqrt(sum(x -> abs2(acos(x)), a))
```
sethaxen, line 152: Actually, minor point, but this would be even better than clamping in-place:
```suggestion
    return sqrt(sum(x -> abs2(acos(min(x, 1))), a))
```

kellertuer: And then leave out the line 151? Sure. I could also use `clamp(x,-1,1)`?

src/manifolds/Grassmann.jl
```diff
@@ -278,12 +278,12 @@ In this formula the $\operatorname{atan}$ is meant elementwise.
...
 log(::Grassmann, ::Any...)
 
-function log!(M::Grassmann, X, p, q)
+function log!(::Grassmann{n,k}, X, p, q) where {n,k}
     z = q' * p
     At = q' - z * p'
     Bt = z \ At
     d = svd(Bt')
-    return copyto!(X, d.U * Diagonal(atan.(d.S)) * d.Vt)
+    return X .= view(d.U, :, 1:k) * Diagonal(atan.(view(d.S, 1:k))) * view(d.Vt, 1:k, :)
```
sethaxen, line 285: One other simplification I had noticed in the above lines. They could be changed to
```julia
z = p' * q
A = q - p * z
B = A / z
d = svd(B)
```
And perhaps, depending on what we know about `z` could even do `B = q / z - p`, avoiding allocating `A` entirely.

kellertuer: What would we need to know about z for the last line? It looks really sweet!

src/manifolds/Grassmann.jl
sethaxen: I'm pretty sure singular values are also always non-negative by construction, so the lower bound check should never be necessary, but I don't think it will make much of a performance difference.

sethaxen:
> that's why its good to review code from time to time

or just have someone who's never used a manifold try to wrap his head around it for a few days. 🙂

kellertuer:
Hm, formatter is a little inconsistent recently; despite the but I recently fixed I again can not reproduce, what GitHub does on my machine, where formatter runs nicely and returns true...?

src/manifolds/Grassmann.jl
kellertuer: Then I'll stick with clamp for now.

kellertuer:
> > that's why its good to review code from time to time
> 
> or just have someone who's never used a manifold try to wrap his head around it for a few days. 🙂

Or use the complex care for the first time at all ;) No, the performance tips you are giving here, look really nice and noticing errors sometimes really only happens if a case is used – tests cover lines not all cases.

src/manifolds/Grassmann.jl
kellertuer: So neither of your proposed code (the for lines or the three lines with avoiding `A`) do pass the tests for the complex case one obtains non-tangent results

````
  Expression: isapprox(M, pts[1], exp(M, pts[2], X2); atol = atolp1p2, rtol = rtolp1p2)
   Evaluated: isapprox(Grassmann(3, 2, ℂ), Complex{Float64}[0.5 + 0.5im 0.5 + 0.5im; 0.5 + 0.5im -0.5 - 0.5im; 0.0 + 0.0im 0.0 + 0.0im], Complex{Float64}[-0.5 - 0.49999999999999994im 0.5000000000000001 + 0.5000000000000001im; -0.5000000000000001 - 0.5im -0.5 - 0.5im; 6.162975822039155e-33 + 3.0814879110195774e-33im 5.551115123125784e-17 + 0.0im]; atol = 2.220446049250313e-13, rtol = 0)
````

It seems to be at least numerically less precise?

Comment (kellertuer) [reply src/manifolds/Grassmann.jl:285]: To be precise, above error message is the one with `A`, the one without fails even a lot, with a check for the tangent vector yields nearly pi instead of zero.

Comment (sethaxen) [reply src/manifolds/Grassmann.jl:285]: `z` is square, so all you'd need is for `z/z == one(z)`. According to the docs `w = z/z` is defined as `w * z = z`, which _could_ produce a pseudoinverse `w != one(z)` if `z` is not invertible. So I think you'd need to know that `z` is invertible.

Comment (sethaxen) [reply src/manifolds/Grassmann.jl:285]: > It seems to be at least numerically less precise?

Hmm, strange. I don't see why that would be. Did you catch the change to the definition of `z`?

Comment (kellertuer) [reply src/manifolds/Grassmann.jl:285]: Oh. Sorry. Maybe I should take a break soon.

Comment (kellertuer) [reply src/manifolds/Grassmann.jl:284]: So to be precise again.
For the new variant with `A` I get the error message from above, which just means its numerically not as stable as the previous variant.
The variant without `A` does the job actually! So not only less allocations but as precise as before. Nice, thanks!

Comment (sethaxen) [reply src/manifolds/Grassmann.jl:284]: Weird, and a little bit confusing. I'll take it though!

Comment (kellertuer) [reply src/manifolds/Grassmann.jl:284]: Well _with_ your one-line-B and without `A` is the variant we now have and it seems as precise as the original, so that's nice.

Comment (sethaxen) [reply src/manifolds/Grassmann.jl:284]: Interestingly, these lines are now just computing the `inverse_retraction!` with `PolarInverseRetraction`.

**Comment by kellertuer** [comment]
For the last optimization on the logarithmic map, we have to decide: less allocations or more precision - and yes exp/log are the same as the polar retraction, so we could also do the numerically less exact one in the retractions.

**Comment by sethaxen** [comment]
> For the last optimization on the logarithmic map, we have to decide: less allocations or more precision

It would be good to know if this is actually numerically less exact and, if so, why, since the math is the same (assuming `z` is invertible). Adjoints should make absolutely no difference. e.g. it is possible that the tolerances are just too strict anyways. On your branch and my MacOS with Julia 1.4.2, I get a host of different failures before even getting to the Grassmann tests that we don't see on the Ubuntu CI. And if I only run the Grassmann tests, everything passes, which makes me think it could be an issue of the tolerances just being bad for Ubuntu, which might cause some implementations to randomly fail and some to randomly pass for a pair of points without one being inherently more numerically (un)stable. Could you temporarily disable fast fail (allows MacOS to keep running even if Ubuntu fails)? Just add the bottom line below to `ci.yml`

```
    strategy:
      matrix:
        julia-version: ["1.0","1.4"]
        os: [ubuntu-latest, macOS-latest, windows-latest]
      fail-fast: false
```

> and yes exp/log are the same as the polar retraction, so we could also do the numerically less exact one in the retractions.

It looks like `exp!` and `log!` begin with the polar retraction and inverse retraction but then perform an additional step.

One other thing, I noticed that `GeneralizedGrassmann` and `GeneralizedStiefel` implementations are very similar to those of `Grassmann` and `Stiefel` and probably need some of the same fixes used here.

**Comment by kellertuer** [comment]
I also added Julia 1.5 for completeness; you're right, especially the Polar Retraction looks quite different on Grassmann

For sure, if it works for Grassmann I can transfer these to Stiefel, and both their generalisations.

**Comment by kellertuer** [comment]
So, it's always the same test, Julia 1.4 and 1.5, both on ubuntu and 1.5 also on windows. It's an `isapprox` of points on `Grassmann` to verify exp and log cancel out.

**Comment by sethaxen** [comment]
Strange. Does it pass or fail on your machine? Here's a quick and dirty benchmark I did.

```julia
using Manifolds, Random, LinearAlgebra, StatsPlots

function log_old(::Grassmann{n,k}, p, q) where {n,k}
    z = q' * p
    At = q' - z * p'
    Bt = z \ At
    d = svd(Bt')
    return view(d.U, :, 1:k) * Diagonal(atan.(view(d.S, 1:k))) * view(d.Vt, 1:k, :)
end

function log_new(::Grassmann{n,k}, p, q) where {n,k}
    z = p' * q
    B = q / z - p
    d = svd(B)
    return view(d.U, :, 1:k) * Diagonal(atan.(view(d.S, 1:k))) * view(d.Vt, 1:k, :)
end

function rand_point(M, p, scale)
    X = zero_tangent_vector(M, p)
    randn!(X)
    X .*= scale
    return exp(M, p, project!(M, X, p, X))
end

function distance_old_new(M, p, scale)
    q1 = rand_point(M, p, scale)
    q2 = rand_point(M, p, scale)
    dold = distance(M, q2, exp(M, q1, log_old(M, q1, q2)))
    dnew = distance(M, q2, exp(M, q1, log_new(M, q1, q2)))
    return (dold, dnew)
end

M = Grassmann(3, 2, ℂ)
p = [0.5+0.5im 0.5+0.5im; 0.5+0.5im -0.5-0.5im; 0.0 0.0]
dists = [distance_old_new(M, p, 1) for _ in 1:100000]
dold = first.(dists)
dnew = last.(dists)
ecdfplot(dold; label = "Old log")
ecdfplot!(dnew; label = "New log")
plot!(xlabel = "Distance", ylabel = "Cumulative Probability")
```

The result shows that the distribution of error in `distance` (which `isapprox` checks is 0) from the two implementations is basically identical.
grassmann_log_ecdf
Granted, this is on my machine, for which the tests pass, as on CI.

I do think however that without knowing for certain that `z = p'q` will always be invertible, we should probably allocate both `B` and `A`. All I was able to work out is that `p'q` is a square `k × k` block of an `n × n` unitary matrix, but I can't think of a reason that would guarantee invertibility.

kellertuer:
Well I am in the Julia 1.5.1 MacOS regime, so it works, as the CI on GitHub does,
but yes maybe its better for stability to just return to the version with allocating `A`. Thanks for the check.

kellertuer:
Interesting, so the old variant _with_ allocation would've passed all tests, but since I added 1.5, ubuntu on that actually does still fail the same test.

sethaxen:
What does the tolerance work out to? It's possible that system- and julia-specific details could move a pair of points to a different part of the error distribution without necessarily shifting the error distribution. The above benchmark seems to indicate that if you want the test to pass for all points, you'll need an absolute tolerance of ~5e-8.

kellertuer:
It was already prepared (even for all manifolds) in tests for nightly, so the current commit “just” shifts this to also include Julia 1.5.

kellertuer:
We should maybe switch to more squash merges? Or maybe even in general to squash merges; what do you think?

Despite this, the branch should be ready, I fixed Stiefel, Generalized Stiefel and Generalized Grassmann as well

mateuszbaran:
I'm fine with switching to squash merges by default.

src/manifolds/Grassmann.jl
```diff
@@ -114,16 +114,16 @@ function check_tangent_vector(
...
-            norm(p' * X + X' * p),
+            norm(p' * X + conj(X' * p)),
             "The matrix $(X) does not lie in the tangent space of $(p) on $(M), since p'X + X'p is not the zero matrix.",
         )
     end
     return nothing
 end
 
-decorated_manifold(M::Grassmann{N,K,𝔽}) where {N,K,𝔽} = Euclidean(N, K; field = 𝔽)
+decorated_manifold(::Grassmann{N,K,𝔽}) where {N,K,𝔽} = Euclidean(N, K; field = 𝔽)
```
sethaxen, line 126: A number of lines add the `M` variable to functions where it is not used. I actually prefer to remove variable names where unused unless it makes the code much less clear so it's more obvious when an arg is used only for dispatch, but I don't think we've agreed on a standard. Thoughts?

src/tests/tests_general.jl
```diff
@@ -202,7 +202,19 @@ function ManifoldTests.test_manifold(
             atol = atolp1p2,
             rtol = rtolp1p2,
         )
-        Test.@test isapprox(M, pts[1], exp(M, pts[2], X2); atol = atolp1p2, rtol = rtolp1p2)
+        if VERSION >= v"1.5" && isa(M, Union{Grassmann,GeneralizedStiefel})
```
sethaxen, line 205: Since the test suite passed before these fixes, it would be nice to have some tests that would have failed before the fixes and now pass.

src/manifolds/GeneralizedGrassmann.jl
```diff
@@ -156,7 +156,7 @@ function distance(M::GeneralizedGrassmann, p, q)
     p ≈ q && return zero(real(eltype(p)))
     a = svd(p' * M.B * q).S
     a[a .> 1] .= 1
```
sethaxen, line 158: This step is no longer needed, due to the `clamp`.
```suggestion
```

sethaxen (changes requested):
Looks good! Just a few minor things.

src/tests/tests_general.jl
kellertuer: Actually, the old test suite (1.0 and 1.4) would pass again, it's just nightly (for which this if was already in) and 1.5 (which is now in the test suite) that require this.

sethaxen:
> We should maybe switch to more squash merges? Or maybe even in general to squash merges; what do you think?

Yes, I always squash merge. It makes for much cleaner and easier to follow commit histories on master.

src/manifolds/Grassmann.jl
kellertuer: Yes, I am currently removing occurrences of unused variables, whenever I notice them. I also think that it's nicer without stating unused variables.
I just was not aware of that, say, half a year ago; but now they are even nicely grayed out in VS Code, so they are easy to spot.

src/tests/tests_general.jl
sethaxen: The above comment is not about this line of code but about tests in general. There was a bug, but the test suite passed. So there are two separate problems. So far from what I can tell this PR fixes the bug but not the inadequacy in the test suite that allowed us to miss the bug.

src/manifolds/Grassmann.jl
sethaxen: Oh I see now, I read the diff backwards. Yes we are on the same page.

src/tests/tests_general.jl
kellertuer: Ah, you're right – I was too stuck to this line. I will add something adapted from your example then.

sethaxen: Thanks! I'll go ahead and approve. Once you're happy with that, feel free to merge!

sethaxen (approved):

Issue #148, opened 2020-04-02
Don't always deploy preview docs


sethaxen:
Not every PR adds to the docs enough to warrant deploying a preview build. One reason we wouldn't want to always deploy PR docs is that it can substantially increase the repo size over time (see e.g. ).

It might be good to deactivate deploying preview docs by default but enable it to be turned on for specific PRs. One solution is to only deploy docs on PRs if a special label is set (see https://github.com/tkf/MyPlayground.jl/pull/28 for an implementation). Another would be if something like `[preview docs]` was included in the commit message.

kellertuer:
I was thinking, we could also manually delete old PRs? That might not decrease the size, but avoid too much clutter.

I like the label solution, you can set that if there is a certain amount or some serious changes to the docs

sethaxen:
> I was thinking, we could also manually delete old PRs? That might not decrease the size, but avoid too much clutter.

You mean delete old preview docs from `gh-pages`? Yes, I think we should do that anyways.

> I like the label solution, you can set that if there is a certain amount or some serious changes to the docs

Yeah, I like that if someone other than us submits a PR, we can mark it for building preview docs without needing to commit ourselves.

kellertuer:
Exactly, as far as I read on slack, they currently just stay there.

Exactly, that's what I thought, one can even at one point decide: Oh, I wanted to check the docs.

sethaxen:
With GitHub Actions, this looks straightforward to do with labels. With Travis, right now the only option is to trigger with text in the commit message: https://docs.travis-ci.com/user/conditions-v1#examples

kellertuer:
Since we switched to GitHub Actions in between, do we want to activate doing the docs only when push-preview label is set? I like the idea.

PR #222, opened 2020-09-21
Conditional deploy docs.

kellertuer:
Let's first test that this does currently not deploy (trying to solve #148).

We should maybe at some point also delete old PRs preview docs.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,6 +2,7 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 HybridArrays = "1baab800-613f-4b0a-84e4-9cd3431bfbb9"
+JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 Manifolds = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 ManifoldsBase = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,7 +2,6 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 HybridArrays = "1baab800-613f-4b0a-84e4-9cd3431bfbb9"
-JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 Manifolds = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 ManifoldsBase = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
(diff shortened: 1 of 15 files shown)


kellertuer:
Check 1 was successful, it does not push when the label is not present.

kellertuer:
Reopening with the label and the docs are pushed. Great!

docs/make.jl
```diff
@@ -81,4 +81,21 @@ makedocs(
...
+    yes = "deploydocs" in labels
+    if yes
+        @info "Trying to push preview as label `deploydocs` is specified." labels
+    else
+        @info "Not pushing preview as label `deploydocs` is not specified." labels
+    end
+    return yes
+end
+
+deploydocs(
```
kellertuer, line 98: One could even speed this up and avoid deploying by adding ` should_push_preview() &&` upfront this line; what do you think?

sethaxen:
I'm not a big fan of the Github Actions-specific logic being in `make.jl`. I think it would be cleaner for `should_push_preview` to just read an environment variable, something like `DOCUMENTER_PUSH_PREVIEW`, defaulting to false if the variable is undefined or not `TRUE` or something like that. (I always forget how to write booleans in shell script). Then update `documenter.yml` to set that environment variable if the label is present.

docs/make.jl
sethaxen: If it's not deploying, I doubt calling `deploydocs` with `push_preview=false` takes much time at all.

kellertuer:
It is not Github-Actions intervowen, it actually just reads Github Environment variables (but not a `DOCUMENTER_PUSH_PREVIEW`), namely the labels that are attached to a PR. If its not a PR, I think the `event_path` is not given (since we are on master) and docs are pushed anyways.

docs/make.jl
kellertuer: You're right. Anyways with your other remark and now an approach to a pure GitHub Actions approach this point is not necessary anymore :)

kellertuer:
Now this PR contains a solution purely based on GitHub Actions:
Documenter now contains two actions:
* Documenter, which only runs if the `preview docs` label is present
* a second action (Documentation deployment note), that only runs if the label is _not_ present

this second action has two purposes: (1) it informs a user how to turn docs-preview on if they are off and (2) the action is not empty when the label is not present, because an empty action would fail

Despite that there is only again something with formatter – even updating it locally here and running it still reports errors on the CI; seems like formatter changes its defaults a lot currently..., which is a little sad.

docs/Project.toml
```diff
@@ -2,6 +2,7 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 HybridArrays = "1baab800-613f-4b0a-84e4-9cd3431bfbb9"
+JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
```
sethaxen, line 5: Is JSON still needed?
```suggestion
JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
```

docs/make.jl
```diff
@@ -1,4 +1,4 @@
-using Manifolds, ManifoldsBase, Documenter
+using Manifolds, ManifoldsBase, Documenter, JSON
```
sethaxen, line 1: Likewise, do we still need this import?
```suggestion
using Manifolds, ManifoldsBase, Documenter
```

sethaxen (approved):
Wow this ended up being really clean. LGTM, once the JSON dependency is clarified or removed. I'll let you address and merge when ready.

docs/make.jl
kellertuer: Thanks for noticing, no we don't.

kellertuer:
Thanks for noticing the JSON dependency, that is of course not necessary;
and well – I saw the label conditional for github actions and you were right, decoupling it as it is now is of course much cleaner.

PR #229, opened 2020-10-05
Caley Retraction

kellertuer:
adds a pade approximation based retraction as well as its special case the caley retraction – as well as a corresponding vector transport.

We could check whether a `VectorTransport` that stems from a retraction (its differentiation) is a reasonable own type.

* [x] Tests
* [x] Check documentation.

### DIFF
--- a/src/manifolds/Stiefel.jl
+++ b/src/manifolds/Stiefel.jl
@@ -36,23 +36,24 @@ struct Stiefel{n,k,𝔽} <: AbstractEmbeddedManifold{𝔽,DefaultIsometricEmbedd
 Stiefel(n::Int, k::Int, field::AbstractNumbers = ℝ) = Stiefel{n,k,field}()
 
 @doc raw"""
-    PadéRetraction{m} <: AbstractRetractionMethod
+    PadeRetraction{m} <: AbstractRetractionMethod
 
 A retraction based on the Padé approximation of order $m$
 """
-struct PadéRetraction{m} <: AbstractRetractionMethod end
+struct PadeRetraction{m} <: AbstractRetractionMethod end
 
-function PadéRetraction(m::Int)
-    (m < 1) && error("The Padé based retraction is only available for positive orders, not for order $m.")
-    return PadéRetraction{m}()
+function PadeRetraction(m::Int)
+    (m < 1) &&
+        error("The Padé based retraction is only available for positive orders, not for order $m.")
+    return PadeRetraction{m}()
 end
 @doc raw"""
     CaleyRetraction <: AbstractRetractionMethod
 
-A retraction based on the Caley transform, which is realized by calling the
-[`PadéRetraction`](@ref)`{1}`.
+A retraction based on the Caley transform, which is realized by using the
+[`PadeRetraction`](@ref)`{1}`.
 """
-const CaleyRetraction = PadéRetraction{1}
+const CaleyRetraction = PadeRetraction{1}
 
 """
     CaleyVectorTransport <: AbstractVectorTransportMethod
@@ -61,7 +62,7 @@ A vector transport that one obtains by differentiating a [`CaleyRetraction`](@re
 """
 struct CaleyVectorTransport <: AbstractVectorTransportMethod end
 
-function allocation_promotion_function(M::Stiefel{n,k,ℂ}, f, args::Tuple) where {n,k}
+function allocation_promotion_function(::Stiefel{n,k,ℂ}, ::Any, ::Tuple) where {n,k}
     return complex
 end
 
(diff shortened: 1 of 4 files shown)


sethaxen:
I'll try to give more substantive comments later, but I prefer "Pade" without the accent. Yes the name technically has the accent, but having the only way to access a function, class, or const be with a unicode character can make it hard for certain users to use the package, e.g. those who are calling it from Python, or any user who doesn't know how to make the accent character.

kellertuer:
your arguments are reasonable, i changed the name. Maybe I was too much thinking of all our ´ ℝ` and `ξ`; but sure the first is just a short hand and the second is our notation, used consistently but not enforced to be used by others.

Do you think that Python users might use this package? That would be neat!

kellertuer:
One thing I am looking forward to in the review is reduction of allocations.

src/manifolds/Stiefel.jl
```diff
@@ -287,6 +373,20 @@ where $\operatorname{sgn}(p) = \begin{cases}
 """
 retract(::Stiefel, ::Any, ::Any, ::QRRetraction)
 
+function retract!(::Stiefel, q, p, X, ::PadeRetraction{m}) where {m}
+    Pp = p * p'
+    Pp .= one(Pp) - Pp
```
mateuszbaran, line 378:
```suggestion
    Pp = I - p * p'
```
This should be faster.

kellertuer: Thanks, I also fixed the second occurence.

src/manifolds/Stiefel.jl
```diff
@@ -421,8 +420,7 @@ tangent space at $q=\operatorname{retr}_p(d)$ using the [`CaleyRetraction`](@ref
 vector_transport_direction(M::Stiefel, p, X, d, ::CaleyVectorTransport)
 
 function vector_transport_direction!(::Stiefel, Y, p, X, d, ::CaleyVectorTransport)
-    Pp = p * p'
-    Pp .= one(Pp) - 1 // 2 * Pp
+    Pp = I - p * p'
```
mateuszbaran, line 423: Didn't you want `Pp = I - 1 // 2 * p * p'`? This changes the behavior of code you replaced.

kellertuer: Oops. Yes, the formulae in the papers read with 1/2, though that surprises me a little, I'll add those.

src/manifolds/Stiefel.jl
```diff
@@ -298,6 +397,36 @@ function retract!(::Stiefel, q, p, X, ::QRRetraction)
...
+Since this is the differentiated retraction as a vector transport, the result will be in the
+tangent space at $q=\operatorname{retr}_p(d)$ using the [`CaleyRetraction`](@ref).
+"""
+vector_transport_direction(M::Stiefel, p, X, d, ::CaleyVectorTransport)
+
+function vector_transport_direction!(::Stiefel, Y, p, X, d, ::CaleyVectorTransport)
+    Pp = I - 1 // 2 * p * p'
+    Wpd = Pp * d * p' - p * d' * Pp
+    WpX = Pp * X * p' - p * X' * Pp
+    q1 = one(Wpd) - 1 // 2 * Wpd
```
mateuszbaran, line 426:
```suggestion
    q1 = I - 1 // 2 * Wpd
```

src/manifolds/Stiefel.jl
```diff
@@ -287,6 +373,19 @@ where $\operatorname{sgn}(p) = \begin{cases}
 """
 retract(::Stiefel, ::Any, ::Any, ::QRRetraction)
 
+function retract!(::Stiefel, q, p, X, ::PadeRetraction{m}) where {m}
+    Pp = I - 1 // 2 * p * p'
+    WpX = Pp * X * p' - p * X' * Pp
+    pm = sum([
```
mateuszbaran, line 379: I think a simple loop would be the fastest solution here. Something like
```julia
pm = similar(WpX) # to make sure that `pm` is mutable
copyto!(pm, I)
qm = similar(WpX)
copyto!(qm, I)
WpXk = similar(WpX)
copyto!(WpXk, factorial(m) / factorial(2m) * I)
for k in 0:m
    WpXk .*= (some simplification of these factorials)
    pm .+= WpXk
    if k%2 == 0
        qm .+= WpXk
    else
        qm .-= WpXk
    end
    WpXk *= WpX
end
```

kellertuer: Thanks for the sketch, but I am not really able to figure out the simplifications of the factor. To me it looks like it should be
```
WpXk  .*= ( k==0 ? 1 : (m-k+1)/( (2*m-k+1) * k) )
```
the first “removes” that factor from the 1/(m-k)! (namely the (m-(k-1)) factor) the second the similar term from the nominator and we have to divide by k for building k!. However, this does not yield the same results as before.

kellertuer: Found the errors. For k=0 it has to be 2, and both `pm`, `qm` have to be initialised to be zero (not `I`). Now it works, thanks for the sketch. This part was actually one area where I hoped for an optimization.

mateuszbaran:
> Do you think that Python users might use this package? That would be neat!

Well, @krzysztofrzecki already does that :slightly_smiling_face: . We may prepare an instruction somewhere at some point.

kellertuer:
I think this can be merged, if approved by one of you.

mateuszbaran (approved):

PR #231, opened 2020-10-08
Some fixes for Julia 1.6 nightly

mateuszbaran:
Let's try making tests pass on Julia 1.6 nightly. For now I'm ignoring the issue about ambiguities.

### DIFF
--- a/test/fixed_rank.jl
+++ b/test/fixed_rank.jl
@@ -9,69 +9,36 @@ include("utils.jl")
     v = UMVTVector([0.0 0.0; 0.0 0.0; 1.0 1.0], [1.0 0.0; 0.0 1.0], zeros(2, 2))
     @test repr(M) == "FixedRankMatrices(3, 2, 2, ℝ)"
     @test repr(Mc) == "FixedRankMatrices(3, 2, 2, ℂ)"
-    if VERSION >= v"1.6.0-DEV.772"
-        @test sprint(show, "text/plain", x) == """
-        SVDMPoint{Matrix{Float64}, Vector{Float64}, Matrix{Float64}}
-        U factor:
-         3×2 Matrix{Float64}:
-          1.0  0.0
-          0.0  1.0
-          0.0  0.0
-        singular values:
-         2-element Vector{Float64}:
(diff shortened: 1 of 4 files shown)


test/fixed_rank.jl
```diff
@@ -9,69 +9,36 @@ include("utils.jl")
...
-          0.0  1.0
-        Vt factor:
-         2×2 Array{Float64,2}:
-          0.0  0.0
-          0.0  0.0"""
-    end
+    @test sprint(show, "text/plain", x) == """
+    $(sprint(show, SVDMPoint{Matrix{Float64}, Vector{Float64}, Matrix{Float64}}))
+    U factor:
+     3×2 $(sprint(show, Matrix{Float64})):
```
kellertuer, line 15: You could also use the print of `x.U` for the whole block lines 15--18?

kellertuer:
Do you have an idea why there are now so many ambiguities since 1.6?

test/fixed_rank.jl
mateuszbaran: OK, I'll try that.

mateuszbaran:
> Do you have an idea why there are now so many ambiguities since 1.6?

There were some changes about ambiguity detection, for example here: https://github.com/JuliaLang/julia/pull/37616 . I don't understand what exactly was changed though.

test/fixed_rank.jl
mateuszbaran: I don't how to print `x.U` to get the correct spacing :slightly_frowning_face: .

kellertuer: Ah, I see, because we add an indent (1 or 2 spaces?), see https://github.com/JuliaManifolds/Manifolds.jl/blob/67ee0f103338ec205017fa4d9b448b2c1b22874f/test/vector_bundle.jl#L121
on a place where I tricked something like that, too.

mateuszbaran: I think we can do that in another PR.

kellertuer: Ok.

PR #232, opened 2020-10-08
Faster retraction and inverse retractions for Stiefel

mateuszbaran:
Inverse QR retraction is now much faster for k=1 and k=2 :slightly_smiling_face:.


kellertuer:
Cool. I will look at the details the next days.
Nice, are the missing parts in code coverage covered but just not detected?

Is it maybe time to also do a benchmark ci for these things?

mateuszbaran:
Thanks! Missing coverage was actually missing, I've pushed a new commit with some tests. I'm actually not even sure why CI thinks the k>2 case in inverse QR retraction is covered, I don't know which test covers it.

Yes, benchmark CI would be great. I'll think about adding it.

mateuszbaran:
I've also added faster distance on SPD manifold with linear-affine metric. It's a bit more explicit because Julia had trouble with efficiently figuring out that it's a generalized symmetric positive definite eigenproblem. It would be even faster if StaticArrays had right matrix division by triangular matrices (it only has the left variant) but it's a problem for another day.

src/manifolds/Stiefel.jl
```diff
@@ -178,18 +178,38 @@ function inverse_retract!(::Stiefel, X, p, q, ::PolarInverseRetraction)
...
+    if k == 1
+        @inbounds R = SMatrix{1,1}(inv(A[1, 1]))
+        mul!(X, q, R)
+    elseif k == 2
+        R11 = inv(A[1, 1])
+        @inbounds R =
+            hcat(SA[R11, zero(ElT)], A[SOneTo(2), SOneTo(2)] \ SA[-R11 * A[2, 1], one(ElT)])
+        mul!(X, q, UpperTriangular(R))
+    else
+        if A isa StaticArray
```
kellertuer, line 198: And we can't do this via dispatch on p/q because for A you want to promote the type first, right?

It's just a matter of style, that I would prefer shorter functions and dispatch if possible.

kellertuer (approved):
With my only remark on the length of the function – I'll let you decide whether thats possible but already approve this PR.

src/manifolds/Stiefel.jl
mateuszbaran: I can split this into more functions, yes.

mateuszbaran: Now we have dispatch and lots of small functions instead :smile: .

mateuszbaran:
I will tag a new minor version after this is merged.

mateuszbaran:
These optimizations require StaticArrays master unfortunately. Would you be OK with disabling these tests until next version of StaticArrays is released?

kellertuer:
Can we until then document that as a note in the docstring? We can remove that note once it works (i.e. when we enable the tests)?

mateuszbaran:
Hm... I'm thinking about either making it work with StaticArrays 0.12 in some way or putting the note.

mateuszbaran:
I've disabled the optimization that requires StaticArrays master for now (it's commented out). I'm going to merge & tag when CI passes.

PR #234, opened 2020-10-10
Move formatter config to configuration file

kellertuer:
This also makes running formatter locally easier because you don't have to pass the style anymore.

If this is ok, I would also apply this change to our other packages (just doing one commit to master).

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,6 @@
+short_to_long_function_def = true
+always_for_in = true
+whitespace_ops_in_indices = true
+pipe_to_function_call = true
+import_to_using = true
+always_use_return = true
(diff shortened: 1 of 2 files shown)


kellertuer:
We could actually also shorten this to just have

````
style = blue
````

see https://domluna.github.io/JuliaFormatter.jl/dev/blue_style/
what do you think?

mateuszbaran:
It looks reasonable. That `style = blue` seems like a work in progress right now (new blue style rules are still being added) so maybe we could hold off on that to not disrupt PRs to Manifolds.jl with style tweaks.

kellertuer:
Yes, it seems to be still work in progress, I enabled it that way n Manopt just to try, but the reasonable one is the one currently in this PR.

mateuszbaran (approved):

PR #235, opened 2020-10-10
Fixes Nested PowerManifold of a PowerManifold to not concatenate sizes.

kellertuer:
This case was not documented that way, so to be precise this might be a breaking PR (comments welcome). This fixes the following situation:

Imagine you have a `N = PowerManifold(M,n)` (maybe nested, but not required to be)
and you form a `O = PowerManifold(N, NestedPowerRepresentation(), m)`

I would expect that the nested-ness of `O` specifies, that elements are vectors of length `m` (or arrays if `m` is not a number but a tuple), since we assume nestedness.
Previously this was not the case an `O` would be a `PowerManifold(M,n,m)`, which one would not expect.

Now this is documented to work as described above, using a new constructor and its test is also added.


mateuszbaran (approved):
This looks fine. Whether you consider this a breaking change or a bugfix is up to you :slightly_smiling_face: .

PR #236, opened 2020-10-15
Hyperbolic ONBs

kellertuer:
This PR introduces both a `DefaultOrthonormalBasis`as well as a `DiagonalizingOrthonormalBasis` for `Hyperbolic`, to be precise, the Hyperboloid model.

This also introduces a `_gram_schmidt!` method that we actually could generalise to arbitrary manifolds, make public or even move over to `ManifoldsBAse`, what do you think?


kellertuer:
So, the last commit was a local formatter run on the same branch, so I am again a little lost where the difference comes from...

mateuszbaran:
We already have Gram-Schmidt (and SVD) orthonormalization in base: https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/master/src/bases.jl#L248 and https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/8a9bed5731ba132e2d18270dd77735440c5c0cf1/src/bases.jl#L271 . By the way I guess we could also have rank revealing QR-based orthonormalization.

kellertuer:
Thanks for pointing that out, I did not find it as a get_basis. should we maybe write a `gram_schmidt`?

mateuszbaran:
OK, a separate `gram_schmidt` could be useful.

kellertuer:
Ok, so when this is finsihed I will think about a generic gram schmidt on base.

Your svd based method is more efficient than my for loop?

Is it reasoable to maybe put the `_hyperbolize` publuc (without the `_`)? Note that this is not a project, just a nice helper for the last component.

mateuszbaran:
I'm not sure about speed of SVD vs Gram-Schmidt, I didn't test it. Primarily SVD should be more numerically stable, and rank-revealing QR should be the fastest option here (I think).

> Is it reasoable to maybe put the `_hyperbolize` publuc (without the `_`)? Note that this is not a project, just a nice helper for the last component.

Maybe, I don't know how useful it would be. Generally helper functions are not supposed to be exported although removing the underscore would be some indication that it can be safely used from outside of this package.

kellertuer:
Ok, so then I would say after we finish this one we can check for the rank revealing qr and provide a nice gram schmidt.

I am not sure `hyperbolic`(surely for export without `_`) would be that useful in general, it's nice to have for hyperbolic if you want to add the last component accordingly.

kellertuer:
Shall. we wait with this one until JuliaManifolds/ManifoldsBase.jl#52 is merged and published?

mateuszbaran:
Yes, I think we should wait.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -239,6 +239,120 @@ function exp!(M::Hyperbolic, q, p, X)
...
+
+@doc raw"""
+    _hyperbolize(M,q)
+
+Given the [`Hyperbolic`](@ref)`(n)` manifold using the hyperboloid model, a point from the
+$q\in ℝ^n$ can be set onto the manifold by computing its last component such that for the
+resulting `p` we have that its [`minkowski_metric`](@ref) is $⟨p,p⟩_{\mathrm{M}} = - 1$,
+i.e. $p_{n+1} = \sqrt{\lVert q \rVert^2-^}$
+"""
+_hyperbolize(M::Hyperbolic, q) = [q..., sqrt(norm(q)^2 + 1)]
```
mateuszbaran, line 317:
```suggestion
_hyperbolize(M::Hyperbolic, q) = vcat(q, sqrt(norm(q)^2 + 1))
```

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -239,6 +239,120 @@ function exp!(M::Hyperbolic, q, p, X)
...
+@doc raw"""
+    _hyperbolize(M, p, Y)
+
+Given the [`Hyperbolic`](@ref)`(n)` manifold using the hyperboloid model and a point `p`
+thereon, we can put a vector $Y\in ℝ^n$  into the tangent space by computing its last
+component such that for the
+resulting `p` we have that its [`minkowski_metric`](@ref) is $⟨p,X⟩_{\mathrm{M}} = 0$,
+i.e. $X_{n+1} = \frac{⟨\tilde p, Y⟩}{p_{n+1}}$, where $\tilde p = (p_1,\ldots,p_n)$.
+"""
+_hyperbolize(M::Hyperbolic, p, Y) = [Y..., dot(p[1:(end - 1)], Y) / p[end]]
```
mateuszbaran, line 328:
```suggestion
_hyperbolize(M::Hyperbolic, p, Y) = vcat(Y, dot(p[1:(end - 1)], Y) / p[end])
```

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -239,6 +239,120 @@ function exp!(M::Hyperbolic, q, p, X)
...
+
+function get_basis(M::Hyperbolic, p, B::DiagonalizingOrthonormalBasis)
+    n = manifold_dimension(M)
+    X = B.frame_direction
+    V = [_hyperbolize(M, p, [i == k ? 1 : 0 for k in 1:n]) for i in 1:n]
+    κ = -ones(n)
+    if norm(M, p, X) != 0
+        placed = false
+        for i in 1:n
+            if norm(M, p, V[i] - inner(M, p, X, V[i]) .* V[i]) ≈ 0 # is X a multiple of V[i]?
```
mateuszbaran, line 257: Wouldn't `inner(M, p, X, V[i]) ≈ 1` be equivalent?

kellertuer: I hope not, since I try to check whether V[i] is parallel to X, to avoid that the default placement (V[1|=X) does not yield a linear dependence.

mateuszbaran: Then I think `inner(M, p, X, V[i]) ≈ norm(M, p, X)` looks correct. Or maybe `norm(M, p, X[i] - inner(M, p, X, V[i]) .* V[i]) ≈ 0` but the first variant look cleaner to me.

kellertuer: Is `X[i]` a `V[i]` in your second formula? But you're right, the first looks cleaner and should also work.

mateuszbaran: Hm, it was actually supposed to be just `X`. For example `[2, 0] - dot([2, 0], [1, 0]) .* [1, 0]` is approximately 0.

kellertuer: Right, that should also work, but the norm line is nicer, will add that tomorrow.

kellertuer: Oh, I think your inner line is missing an abs on the left to account for opposing directions (then inner would be minus the norm)

mateuszbaran: Yes, there should be `abs`. It's good that you noticed.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -239,6 +239,93 @@ function exp!(M::Hyperbolic, q, p, X)
...
+function get_basis(M::Hyperbolic, p, B::DefaultOrthonormalBasis)
+    n = manifold_dimension(M)
+    V = [_hyperbolize(M, p, [i == k ? 1 : 0 for k in 1:n]) for i in 1:n]
+    return CachedBasis(B, gram_schmidt(M, p, V))
+end
+
+function get_basis(M::Hyperbolic, p, B::DiagonalizingOrthonormalBasis)
+    n = manifold_dimension(M)
+    X = B.frame_direction
+    V = [_hyperbolize(M, p, [i == k ? 1 : 0 for k in 1:n]) for i in 1:n]
```
mateuszbaran, line 251:
```suggestion
    V = [_hyperbolize(M, p, [i == k ? one(eltype(p)) : zero(eltype(p)) for k in 1:n]) for i in 1:n]
```

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -239,6 +239,93 @@ function exp!(M::Hyperbolic, q, p, X)
     return copyto!(q, cosh(vn) * p + sinh(vn) / vn * X)
 end
 
+function get_basis(M::Hyperbolic, p, B::DefaultOrthonormalBasis)
+    n = manifold_dimension(M)
+    V = [_hyperbolize(M, p, [i == k ? 1 : 0 for k in 1:n]) for i in 1:n]
```
mateuszbaran, line 244:
```suggestion
    V = [_hyperbolize(M, p, [i == k ? one(eltype(p)) : zero(eltype(p)) for k in 1:n]) for i in 1:n]
```

src/manifolds/HyperbolicHyperboloid.jl
kellertuer: Ah, yes, I have to check for those more carefully.

mateuszbaran (approved):

mateuszbaran (approved):

PR #237, opened 2020-10-17
Introduce a basis for tangent spaces of Stiefel

kellertuer:
I just saw an idea how to construct this. I still have to add tests, but feel free to already look at the code.

### DIFF
--- a/src/manifolds/Stiefel.jl
+++ b/src/manifolds/Stiefel.jl
@@ -215,10 +215,10 @@ function get_vectors(::Stiefel{n,k,ℝ}, p, ::DefaultOrthonormalBasis{ℝ}) wher
     p⊥ = nullspace([p zeros(n, n - k)])
     an = div(k * (k - 1), 2)
     bn = (n - k) * k
-    V = [
-        [p * vec2skew(1 / sqrt(2) .* _euclidean_unit_vector(an, i), k) for i in 1:an]...,
-        [p⊥ * reshape(_euclidean_unit_vector(bn, j), (n - k, k)) for j in 1:bn]...,
-    ]
+    V = vcat(
+        [p * vec2skew(1 / sqrt(2) .* _euclidean_unit_vector(an, i), k) for i in 1:an],
+        [p⊥ * reshape(_euclidean_unit_vector(bn, j), (n - k, k)) for j in 1:bn],
+    )
     return V
 end
 
--- a/src/manifolds/Stiefel.jl
+++ b/src/manifolds/Stiefel.jl
@@ -163,6 +163,56 @@ function exp!(::Stiefel{n,k}, q, p, X) where {n,k}
 end
 
 @doc raw"""
+    get_basis(M::Stiefel{n,k,ℝ}, p, B::DefaultBasis) where {n,k}
+
+Create the default basis using the parametrization for any $X ∈ T_p\mathcal M$.
+Set $p_\bot \in ℝ^{n\times(n-k)}$ the matrix such that the $n\times n$ matrix of the common
+columns $[p\ p_\bot]$ is an ONB.
+For any skew symmetric matrix $a ∈ ℝ^{k\times k}$ and any $b ∈ ℝ^{(n-k)\times k}$ the matrix
+
+````math
+X = pa + p_\bot b ∈ T_p\mathcal M
+````
+
+and we can use the $\frac{1}{2}k(k-1) + (n-k)k = nk-\frac{1}{2}k(k+1)$ entries
+of $a$ and $b$ to specify a basis for the tangent space.
+using unit vectors for constructing both
+the upper matrix of $a$ to build a skew symmetric matrix and the matrix b, the default
+basis is constructed.
+"""
+function get_basis(::Stiefel{n,k,ℝ}, p, B::DefaultBasis{ℝ}) where {n,k}
+    p⊥ = nullspace([p zeros(n,n-k)])
+    an = div(k*(k-1),2)
+    bn = (n-k)*k
+    V = [
+        [p*_vec2skew(_euclidean_unit_vector(an,i),k) for i=1:an]...,
+        [p⊥*reshape(_euclidean_unit_vector(bn,j),(n-k,k)) for j=1:bn]...,
+    ]
+    return CachedBasis(B,V)
+end
+
+_euclidean_unit_vector(n,i) = [ k==i ? 1.0 : 0.0 for k ∈ 1:n]
+"""
+    vec2skew(v,k)
+
+create a skew symmetric matrix from a vector, for example for `v=[1,2,3]` and `k=3` this
+yields
+````julia
+[  0  1  2;
+  -1  0  3;
+  -2 -3  0
+]
+````
+"""
+function _vec2skew(v,k)
+    m=0;
+    n = div(k*(k-1),2)
+    length(v) < n && error("The vector $(v) is too short (expected $(n) got $(length(v))).")
+    A = [ i<j ? (m+=1; v[m]) : 0.0 for i=1:k, j=1:k ]
+    return A-A'
+end
+
+@doc raw"""
     inverse_retract(M::Stiefel, p, q, ::PolarInverseRetraction)
 
 Compute the inverse retraction based on a singular value decomposition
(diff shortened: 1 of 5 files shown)


src/manifolds/Stiefel.jl
```diff
@@ -162,6 +162,56 @@ function exp!(::Stiefel{n,k}, q, p, X) where {n,k}
     )
 end
 
+@doc raw"""
+    get_basis(M::Stiefel{n,k,ℝ}, p, B::DefaultBasis) where {n,k}
+
+Create the default basis using the parametrization for any $X ∈ T_p\mathcal M$.
+Set $p_\bot \in ℝ^{n\times(n-k)}$ the matrix such that the $n\times n$ matrix of the common
+columns $[p\ p_\bot]$ is an ONB.
```
sethaxen, line 170: Can this be a `DefaultOrthonormalBasis` then?

src/manifolds/Stiefel.jl
```diff
@@ -162,6 +162,56 @@ function exp!(::Stiefel{n,k}, q, p, X) where {n,k}
...
+create a skew symmetric matrix from a vector, for example for `v=[1,2,3]` and `k=3` this
+yields
+````julia
+[  0  1  2;
+  -1  0  3;
+  -2 -3  0
+]
+````
+"""
+function _vec2skew(v, k)
```
sethaxen, line 207: Can this be unified with the machinery used by `Rotations` to make a skew-symmetric matrix?

src/manifolds/Stiefel.jl
kellertuer: No. the [p\ p_\bot] is an ONB of R^n (that is the characteristic of p anyways and the goal of constructing p\bot) but all that is just a helper to construct the tangent basis and has nothing to do with the resulting basis.

In fact the resulting basis is neither orthogonal nor normalised.

src/manifolds/Stiefel.jl
kellertuer: In `Rotations.jl` I only saw a 4d angles to skew matrix conversion, this takes a vector puts it in the upper triangular (without signal) part and then generates the corresponding skew symmetric matrix. If you can point out what exactly can be unified, I'd happily do that.
When used with more than one manifold, we surely also can put this function into the `utils.jl`.

sethaxen: Ah, it's the implementation of `get_vector!`:
https://github.com/JuliaManifolds/Manifolds.jl/blob/5ad6f46523b3516978bc67723fcd3d64829ea464/src/manifolds/Rotations.jl#L318-L342
The one difference is this generates the upper 3x3 matrix as customary for `so(3)` and then fills the upper triangle for the rest, negating for the lower triangle.

An argument for keeping the same function is that `Rotations` is a submanifold of `Stiefel`. An argument for keeping them separate is that perhaps for `Stiefel` one would not want to use a convention like this that privileges a certain submanifold.

src/manifolds/Stiefel.jl
kellertuer: I just noticed, that `a` and `b` are indeed orthonormal coordinates, so yes it is an ONB, I will adapt that.

src/manifolds/Stiefel.jl
kellertuer: Ah, sure I can use the new function for those. I will provide also a mutating variant then.

kellertuer: Hm, looking closely, I am not sure why you exchange the third and second entry  (lines 323&324) compared to linearly writing them in. This currently makes the tests fail. Can you elaborate on this specific order?

kellertuer: To be precise for the case of dimension 3 and a `v=[a,b,c]` I do
````julia
[ 0 a b;
  -a 0 c;
 -b -c 0 ]
````
and you currently (also in this sub part for larger ones)
````julia
[ 0 -c b;
  c 0 -a;
 -b a 0 ]
````
so if this is a convention usually used for SO(n) and its first 3 entries, we can not use my method here.

Comment (sethaxen) [reply src/manifolds/Stiefel.jl]: Yes, this is the `so(3)` convention, chosen because the resulting 3-vector is the axis of rotation multiplied by the angle of rotation for the rotation one would get following the geodesic from the identity element in the direction of the corresponding tangent vector for unit time. Likewise, if only the upper 2x2 matrix is non-zero, then the result is a rotation in the xy-plane. There's perhaps a way to generalize the choice of sign convention for SO(n), but I haven't seen it.

Comment (kellertuer) [reply src/manifolds/Stiefel.jl]: Thanks for the explanation, but then we can not easily unify both; however, we now have `vec2skew` and ´vec2skew!` available.

Comment (mateuszbaran) [reply src/manifolds/Stiefel.jl]: Isn't SO(n) only diffeomorphic (not isomorphic) to a case of Stiefel with the Euclidean metric? If so I don't see much point in making vectorized representations of skew-symmetric matrices identical for both manifolds.

File: src/utils.jl [hunk 3]
```diff
@@ -91,6 +91,35 @@ Converts a size given by `Tuple{N, M, ...}` into a tuple `(N, M, ...)`.
...
+]
+````
+"""
+function vec2skew!(X, v)
+    k = size(X)[1]
+    size(X)[2] != k && error("X is of wrong size, expected ($k,$k) got $(size(X)).")
+    n = div(k * (k - 1), 2)
+    length(v) < n && error("The vector $(v) is too short, expected $(n) got $(length(v)).")
+    m = 0
+    X .= [i < j ? (m += 1; v[m]) : 0.0 for i in 1:k, j in 1:k]
```
mateuszbaran, line 113:
```suggestion
    X .= [i < j ? (m += 1; v[m]) : zero(eltype(v)) for i in 1:k, j in 1:k]
```

src/manifolds/Stiefel.jl
```diff
@@ -162,6 +162,68 @@ function exp!(::Stiefel{n,k}, q, p, X) where {n,k}
...
+end
+
+function get_vectors(::Stiefel{n,k,ℝ}, p, ::DefaultOrthonormalBasis{ℝ}) where {n,k}
+    p⊥ = nullspace([p zeros(n, n - k)])
+    an = div(k * (k - 1), 2)
+    bn = (n - k) * k
+    V = [
+        [p * vec2skew(1 / sqrt(2) .* _euclidean_unit_vector(an, i), k) for i in 1:an]...,
+        [p⊥ * reshape(_euclidean_unit_vector(bn, j), (n - k, k)) for j in 1:bn]...,
+    ]
```
mateuszbaran, line 221:
```suggestion
    V = vcat(
        [p * vec2skew(1 / sqrt(2) .* _euclidean_unit_vector(an, i), k) for i in 1:an],
        [p⊥ * reshape(_euclidean_unit_vector(bn, j), (n - k, k)) for j in 1:bn],
    )
```

src/manifolds/Stiefel.jl
kellertuer: That's also my conclusion, we should stick to the default SO(n) uses and the one Stiefel uses; sure they are diffeomorphic and there is no point enforcing them to be identical.

src/utils.jl
```diff
@@ -91,6 +91,35 @@ Converts a size given by `Tuple{N, M, ...}` into a tuple `(N, M, ...)`.
...
+    size(X)[2] != k && error("X is of wrong size, expected ($k,$k) got $(size(X)).")
+    n = div(k * (k - 1), 2)
+    length(v) < n && error("The vector $(v) is too short, expected $(n) got $(length(v)).")
+    m = 0
+    X .= [i < j ? (m += 1; v[m]) : zero(eltype(v)) for i in 1:k, j in 1:k]
+    X .= X - X'
+    return X
+end
+function vec2skew(v, k)
+    X = zeros(eltype(v), k, k)
```
mateuszbaran, line 118:
```suggestion
    X = similar(v, k, k)
```
I'd also like a special case for `v` being a `StaticArray` that does `similar(v, Size(k, k))` instead. And since `k` is passed as a type parameter in `Stiefel` it should probably be passed like that here as well.

mateuszbaran (approved):
Looks OK to me.

src/utils.jl
kellertuer: I don't understand your last sentence, for this function we do not have a manifold per se in the arguments so we ave to type to parametrize?

And – how different would the static array variant be? can you maybe help and provide it?

mateuszbaran: I was thinking a bit more about this PR and actually that wouldn't really be a bottleneck. Let's leave this for now.

kellertuer: Then I'll merge now and publish a new version.

PR #238, opened 2020-10-22
Some new test tolerances

mateuszbaran:
I've added a couple of new tolerances that I need for https://github.com/JuliaManifolds/FunManifolds.jl/pull/10 .

### DIFF
--- a/src/tests/tests_general.jl
+++ b/src/tests/tests_general.jl
@@ -69,7 +69,6 @@ function ManifoldTests.test_manifold(
     basis_has_specialized_diagonalizing_get = false,
     basis_types_to_from = (),
     basis_types_vecs = (),
-    copyto_atol = 0,
     default_inverse_retraction_method = LogarithmicInverseRetraction(),
     default_retraction_method = ExponentialRetraction(),
     exp_log_atol_multiplier = 0,
@@ -594,7 +593,7 @@ function ManifoldTests.test_manifold(
         for (p, X) in zip(pts, tv)
             p2 = allocate(p)
             copyto!(p2, p)
-            Test.@test isapprox(M, p2, p; atol = copyto_atol)
+            Test.@test isapprox(M, p2, p)
 
             X2 = allocate(X)
             if default_inverse_retraction_method === nothing
--- a/src/tests/tests_general.jl
+++ b/src/tests/tests_general.jl
@@ -69,6 +69,7 @@ function ManifoldTests.test_manifold(
     basis_has_specialized_diagonalizing_get = false,
     basis_types_to_from = (),
     basis_types_vecs = (),
+    copyto_atol = 0,
     default_inverse_retraction_method = LogarithmicInverseRetraction(),
     default_retraction_method = ExponentialRetraction(),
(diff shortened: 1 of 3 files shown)


src/tests/tests_general.jl
```diff
@@ -593,7 +594,7 @@ function ManifoldTests.test_manifold(
         for (p, X) in zip(pts, tv)
             p2 = allocate(p)
             copyto!(p2, p)
-            Test.@test isapprox(M, p2, p)
+            Test.@test isapprox(M, p2, p; atol = copyto_atol)
```
kellertuer, line 597: This seems a little counter intuitive, why should a copy require a tolerance? 

Maybe you could clarify this in the docs? Similarly documenting the other two new group tolerances would be great.

mateuszbaran: I would prefer if it didn't have to add these tolerances but I'm not quite sure how to avoid that. Distance calculation on let's say a 100 node long approximation to the Hilbert sphere has numerical issues. I'm using fixed nodes and quadrature weights currently. `distance(M, p, p)` can be close to  `sqrt(eps(eltype(p)))` and I don't see a way to make `isapprox` more precise than that.

I will update the docs though.

mateuszbaran: > I would prefer if it didn't have to add these tolerances but I'm not quite sure how to avoid that. Distance calculation on let's say a 100 node long approximation to the Hilbert sphere has numerical issues. I'm using fixed nodes and quadrature weights currently. `distance(M, p, p)` can be close to `sqrt(eps(eltype(p)))` and I don't see a way to make `isapprox` more accurate than that.
> 
> I will update the docs though.

kellertuer: Ah! So it's not copy but the distance calculation within `isapprox`, I see. But for this case, would it be a solution to check if the two points (either in distance or in is approx) could be checked if they are equal to then directly return zero (or true)? I do something similar in SymmetricPositiveDefinite if I remember correctly. Oh wait, I think I don't but I should, I just saw why one of my calculations is off (same, due to rounding errors)...

mateuszbaran: Hm... one problem with that is the equality check gets non-monotonic. Very close to zero the equality (either == or isapprox on vectors) returns true (let's say atol=1e-15), then it returns false. Distance is on the order of 1e-8 though so for atols between these two point comparison returns false (even though it returns true for larger tolerances).

Anyway, I've found a much more accurate formula for distances that doesn't have this problem.

mateuszbaran: OK, it looks like `atol_ident_compose` is actually all I need. Does this PR look good now?

kellertuer: Documentation for that one would be nice, and you have to got for `0.4.4`, since I released `0.4.3` with Stiefel basis today.

mateuszbaran: I've added documentation for `atol_ident_compose`. I don't need to tag right after merging this, I still need #239 to finish my work on discretized curves. Something else may turn out to be needed as well.

kellertuer (approved):

PR #239, opened 2020-10-22
Vector space manifold

mateuszbaran:
For https://github.com/JuliaManifolds/FunManifolds.jl/pull/10 I need a way to represent tangent space at a point as a manifold (for example for curves in the tangent space) so here I re-use `VectorSpaceAtPoint` for that.

Does this look reasonable?

### DIFF
--- a/docs/src/manifolds/vector_bundle.md
+++ b/docs/src/manifolds/vector_bundle.md
@@ -14,6 +14,7 @@ The important difference between functions operating on `VectorBundle` and `Vect
 `VectorBundleFibers` refers to the whole set of fibers of a vector bundle.
 There is also another type, [`VectorSpaceAtPoint`](@ref), that represents a specific fiber at a given point.
 This distinction is made to reduce the need to repeatedly construct objects of type [`VectorSpaceAtPoint`](@ref) in certain usage scenarios.
+This is also considered a manifold.
 
 ## FVector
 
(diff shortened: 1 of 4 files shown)


kellertuer:
Yes, sure, this looks nice – could you also adapt the documentation accordingly?

mateuszbaran:
I will, and actually I should also implement all manifold-related functions...

mateuszbaran:
I've made some updates. Do you have any suggestions to make the documentation less confusing?

kellertuer:
The storage is a little tricky, since you store the manifold to the vector space within the fibre and then fix the point-like part to just be p, right? That is a little tricky to explain. Here is a first idea to a new docstring

````julia
"""
    VectorSpaceAtPoint{
        𝔽,
        TFiber<:VectorBundleFibers{<:VectorSpaceType,<:Manifold{𝔽}},
        TX,
    } <: Manifold{𝔽}   

A vector space at a point `p` on the manifold.
This is modelled using [`VectorBundleFibers`](@ref) with only a vector-like part
and fixing the point-lik part to be just `p`.

This vector space itself is also a `manifold`. Especially, it's flat and hence isometric to to the [`Euclidean`](@ref) manifold.

# Constructor
    VectorSpaceAtPoint(fiber::VectorBundleFibers, p)

A vector space (fiber type `fiber` of a vector bundle) at point `p` from
the manifold `fiber.manifold`.
"""
````

mateuszbaran:
Thanks, I'll change the docstring.

PR #240, opened 2020-10-23
Introduce ParallelTransport for Oblique

kellertuer:
I think we should introduce `ParallelTransport`(and all other transports) a Manifold obtains from it “child” within an `AbstractPowerManifold` i.e. wrapping the `PowerTransportMethod`for ease of use for the user.

This PR introduces this for the Oblique manifold but surely we can check for more manifolds where this would be nice.

### DIFF
--- a/src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
+++ b/src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
@@ -23,6 +23,8 @@ where $\operatorname{Log}$ denotes the matrix logarithm and
 $\lVert\cdot\rVert_{\mathrm{F}}$ denotes the matrix Frobenius norm.
 """
 function distance(M::SymmetricPositiveDefinite{N}, p, q) where {N}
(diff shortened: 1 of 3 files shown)


src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
```diff
@@ -23,6 +23,8 @@ where $\operatorname{Log}$ denotes the matrix logarithm and
 $\lVert\cdot\rVert_{\mathrm{F}}$ denotes the matrix Frobenius norm.
 """
 function distance(M::SymmetricPositiveDefinite{N}, p, q) where {N}
+    # avoid numerical instabilities in cholesky
+    norm(p - q) < eps(eltype(p .+ q)) && return 0
```
mateuszbaran, line 27:
```suggestion
    norm(p - q) < eps(eltype(p .+ q)) && return zero(eltype(p))
```

mateuszbaran (approved):

PR #242, opened 2020-10-27
fix typo

ho-oto:


mateuszbaran (approved):
Thanks!

src/manifolds/Grassmann.jl
```diff
@@ -365,7 +365,7 @@ Compute the QR-based retraction [`QRRetraction`](@ref) on the
 ````
 where D is a $m × n$ matrix with
 ````math
-D = \operatorname{diag}( \operatorname{sgn}(R_{ii}+0,5)_{i=1}^n ).
+D = \operatorname{diag}( \operatorname{sgn}(R_{ii}+0.5)_{i=1}^n ).
```
sethaxen, line 368: I'm partial to using fractions:
```suggestion
D = \operatorname{diag}\left( \operatorname{sgn}\left(R_{ii}+\frac{1}{2}\right)_{i=1}^n \right).
```

sethaxen (approved):

src/manifolds/Grassmann.jl
kellertuer: I would be a fan for using `\bigl` or `\Bigl` (and `\Bigr` of course) instead of `\left` and ´\right`, if that's ok.

kellertuer: And sure- if I am the only one favouring manually scaled brackets we can also leave it at left/right.

PR #245, opened 2020-10-31
Plotting Recipe for data on S2 and H2

kellertuer:
I am not yet sure how `PlottingRecipes` – I think I am going for a user recipe – actually works. Got stuck in this situation, which generates a sphere (wire and/or filled) and plots data as scatter or piecewise geodesics.

* However, this does not yet work and I still have to understand why.
* Also I am not sure, how one would implement a `plot!(scene, M, pts)`, that not necessarily redraws the sphere
* not yet sure, for which generic packages this might work or can be generalized to.


kellertuer:
axis are now correct, `plot(M,pts)` and `plot!(M,pts)` seem to work, I just could not yet convince the sphere (solid or wireframe or both) to show up.

kellertuer:
plotting points and geodesics works now (with `plotly()`, somehow not with `gr()`), but since there is no quiver3, plotting tangent vectors is not so easy.

.zenodo.json
```diff
@@ -0,0 +1,28 @@
...
+            "name": "Baran, Mateusz",
+            "orcid": "0000-0001-9667-5579"
+        },
+        {
+            "affiliation": "Technische Universität Chemnitz",
+            "name": "Bergmann, Ronny",
+            "orcid": "0000-0001-8342-7218"
+        }
+    ],
+    "description": "Manopifolds.jl is a library of Riemannian manifolds in Julia",
```
mateuszbaran, line 19:
```suggestion
    "description": "Manifolds.jl is a library of Riemannian manifolds in Julia",
```

mateuszbaran: Unless we want to change the name :smile:

kellertuer: I had to laugh about that name, but I hope I changed that on master (I also accidentally committed this one here and cherry picked it over); sure will change that; today I just merged master back here to start the next days.

kellertuer:
I unified the scheme with the previous 2D plots and also added a Hyperboliod version (i.e. `Hyperbolic` with vectors in R3). The remaining part ist tests and documentation. And these do not work in every backend, I have to check and document that, too (see https://github.com/JuliaPlots/Plots.jl/issues/3126 for details).

kellertuer:
Up to a little bit of tests, this is done, but I am not yet sure, why the Github Actions documentation building fails, it runs fine locally.

mateuszbaran:
> The reason for the sphere to not work with GR is that their surface&wireframe seem to not work with meshes while it does with plotly for example

It's not great that it doesn't fully work with either GR or plotly. How about maybe matplotlib backend?

kellertuer:
They work (i.e. not throw errors) with plotly, just the wireframe does not have the right number of wires. Sadly GR throws an error for wire and surfaces that I did (for the sphere), I think those only work for functions * z=f(x,y)` so not for the sphere.

I can surely check matplotlib and I have to do the VisualRegressionTests still.

kellertuer:
In principle the documentation on the sphere now uses `peplos` as backend and that looks nice.
I also wrote the tests for these 3D plots but my local version currently errors instead of asking me to save the first reference image. When those exist and are checked in this PR should be done.

kellertuer:
I adapted the CI to properly work with the GR backend. For the pyplot backend (both in tests and documenter) I still have to find a way to also have python and matplotlib in there. Locally here (with matplotlib/python on my machine) the sphere plots are really nice, here's a sketch of geodesics, points and a tangent vector quiver.

Unknown

Not that the top point vanishes, since it is a 2D circle covered within the wireframe sphere. As soon as I see how to do 3D point markers, I'll surely do that.

kellertuer:
Since now the metrics are failing, but only on windows, I am slowly a little lost on getting this to work here. The error seems completely unrelated to what I did, so...any help appreciated. 

The graphics tests for this new feature work fine locally and finished on a few CIs already.

mateuszbaran:
Thanks for finishing it, the plots look nice :+1: .

docs/src/manifolds/hyperbolic.md
```diff
@@ -26,6 +26,46 @@ Pages = ["manifolds/HyperbolicHyperboloid.jl"]
...
+### [Visualization of the Hyperboloid](@id hyperboloid_plot)
+
+For the case of [`Hyperbolic`](@ref)`(2)` there is a plotting available based on a [PlottingRecipe](https://docs.juliaplots.org/latest/recipes/) you can easily plot points, connecting geodesics as well as tangent vectors.
+
+!!! note
+    The recipes are only loaded if [Plots.jl](http://docs.juliaplots.org/latest/) or
+    [RecipesBase.jl](http://juliaplots.org/RecipesBase.jl/stable/) is loaded.
+
+If we consider a set of points, we can first plot these and their connecting
+geodesics using the `geodesic_interpolation` For the points. This variable specifies with how many points a geodesic between two successive points is sampled (per default it's `-1`, which deactivates geodesics) and the line style is set to be a path.
```
mateuszbaran, line 38:
```suggestion
geodesics using the `geodesic_interpolation` for the points. This variable specifies with how many points a geodesic between two successive points is sampled (per default it's `-1`, which deactivates geodesics) and the line style is set to be a path.
```

docs/src/manifolds/hyperbolic.md
```diff
@@ -26,6 +26,46 @@ Pages = ["manifolds/HyperbolicHyperboloid.jl"]
...
+
+In general you can plot the surface of the hyperboloid either as wireframe (`wireframe=true`) additionally specifying `wires` (or `wires_x` and `wires_y`) for more precision and a `wireframe_color`. The same holds for the plot as a `surface` (which is `false` by default) and its `surface_resolution` (or `surface_resolution_x` or `surface_resolution_y`) and a `surface_color`.
```
mateuszbaran, line 40:
```suggestion
In general you can plot the surface of the hyperboloid either as wireframe (`wireframe=true`) additionally specifying `wires` (or `wires_x` and `wires_y`) to change the density of wires and a `wireframe_color`. The same holds for the plot as a `surface` (which is `false` by default) and its `surface_resolution` (or `surface_resolution_x` or `surface_resolution_y`) and a `surface_color`.
```
Is my change correct?

docs/src/manifolds/hyperbolic.md
```diff
@@ -26,6 +26,46 @@ Pages = ["manifolds/HyperbolicHyperboloid.jl"]
...
+
+```@example hyperboloid
+using Manifolds, Plots
+M = Hyperbolic(2)
+pts =  [ [0.85*cos(φ), 0.85*sin(φ), sqrt(0.85^2+1)] for φ ∈ range(0,2π,length=11) ]
+scene = plot(M, pts; geodesic_interpolation=100)
+```
+
+To just plot the points atop, we can just omit the `geodesic_interpolation` parameter to obtain a scatter plot. NOte that we avoid redrawing the wireframe in the following `plot!`s.
```
mateuszbaran, line 49:
```suggestion
To just plot the points atop, we can just omit the `geodesic_interpolation` parameter to obtain a scatter plot. Note that we avoid redrawing the wireframe in the following `plot!` calls.
```

docs/src/manifolds/hyperbolic.md
```diff
@@ -26,6 +26,46 @@ Pages = ["manifolds/HyperbolicHyperboloid.jl"]
...
+
+```@example hyperboloid
+plot!(scene, M, pts; wireframe=false)
+```
+
+We can further generate tangent vectors in these spaces and use a plot for there. Keep in mind, that a tangent vector in plotting always requires its base point
```
mateuszbaran, line 55:
```suggestion
We can further generate tangent vectors in these spaces and use a plot for there. Keep in mind that a tangent vector in plotting always requires its base point.
```

docs/src/manifolds/sphere.md
```diff
@@ -17,9 +17,42 @@ To create the unit sphere of $3×2$ real-valued matrices, write `ArraySphere(3,2
...
 ```
+
+## Visualization on `Sphere{2,ℝ}`
+You can visualize both points and tangent vectors on the sphere.
+
+!!! note
+    There seems to be no unified way to draw spheres in the backends of [Plots.jl](http://docs.juliaplots.org/latest/).
+    This recipe currently uses the `seriestype` `wireframe` and `surface`, which does not yet work with the default backend [`GR`](https://github.com/jheinen/GR.jl).
+
+In general you can plot the surface of the hyperboloid either as wireframe (`wireframe=true`) additionally specifying `wires` (or `wires_lat` and `wires_lon`) for more precision and a `wireframe_color`. The same holds for the plot as a `surface` (which is `false` by default) and its `surface_resolution` (or `surface_resolution_lat` or `surface_resolution_lon`) and a `surface_color`.
```
mateuszbaran, line 34: Similar comment to the one for hyperboloid applies here.

docs/src/manifolds/sphere.md
```diff
@@ -17,9 +17,42 @@ To create the unit sphere of $3×2$ real-valued matrices, write `ArraySphere(3,2
...
+
+```@example sphereplot1
+using Manifolds, Plots
+pyplot()
+M = Sphere(2)
+pts = [ [1.0, 0.0, 0.0], [0.0, -1.0, 0.0], [0.0, 0.0, 1.0], [1.0, 0.0, 0.0] ]
+scene = plot(M,pts; wireframe_color=colorant"#CCCCCC", markersize=10)
```
mateuszbaran, line 41:
```suggestion
scene = plot(M, pts; wireframe_color=colorant"#CCCCCC", markersize=10)
```

docs/src/manifolds/sphere.md
```diff
@@ -17,9 +17,42 @@ To create the unit sphere of $3×2$ real-valued matrices, write `ArraySphere(3,2
...
+```
+
+which scatters our points. We can also draw connecting geodesics, which here is a geodesic triangle. Here we discretize each geodesic with 100 points along the geodesic.
+The default value is `geodesic_interpolation=-1` which switches to scatter plot the data.
```
mateuszbaran, line 45:
```suggestion
The default value is `geodesic_interpolation=-1` which switches to scatter plot of the data.
```

docs/src/manifolds/sphere.md
```diff
@@ -17,9 +17,42 @@ To create the unit sphere of $3×2$ real-valued matrices, write `ArraySphere(3,2
...
+
+```@example sphereplot1
+plot!(scene, M, pts; wireframe=false, geodesic_interpolation=100, linewidth=2)
+```
+
+And we can also add tangent vectors, for example tangents pointing to the middle
```
mateuszbaran, line 51:
```suggestion
And we can also add tangent vectors, for example tangents pointing towards the geometric center of given points.
```

kellertuer:
I am not sure why the test coverage has not triggered, but I think even that should now work, which PyPlot being available in the tests, too.

src/manifolds/ProductManifold.jl
```diff
@@ -129,7 +129,7 @@ end
     check_manifold_point(M::ProductManifold, p; kwargs...)
 
 Check whether `p` is a valid point on the [`ProductManifold`](@ref) `M`.
-If `p` is not a point on `M` a [`CompositeManifoldError`](@ref) consisting of all error messages of the
```
mateuszbaran, line 132: What's the problem with that `@ref`?

mateuszbaran:
> I am not sure why the test coverage has not triggered, but I think even that should now work, which PyPlot being available in the tests, too.

I don't know either, sometimes it just seems to not work.

src/manifolds/ProductManifold.jl
kellertuer: Could work, I think since it moved to ManiofldsBase, for some time it might not have worked.

mateuszbaran (approved):

Issue #225, opened 2020-09-23
Add more matrix groups


sethaxen:
#178 mentioned adding the general linear group. This is basically what the abstract type `AbstractGroupManifold{𝔽,MultiplicationOperation}` represents, and we already have this with something like
```julia
struct GeneralLinear{N,𝔽} <: AbstractGroupManifold{𝔽,MultiplicationOperation} end
```

It would also be good to have some more subgroups. I'm thinking in particular `SpecialLinear`, `Unitary`, and `SpecialUnitary`, with `Orthogonal` just being `Unitary` over the real numbers. Most of the methods would automatically come from `AbstractGroupManifold{𝔽,MultiplicationOperation}`, but for at least `SpecialUnitary(2)`, `Unitary(2)` and `Orthogonal(3)`, we can implement more efficient group exponentials.

The annoying thing is that ideally `SpecialOrthogonal` would be `SpecialUnitary` over the reals, but `SpecialOrthogonal` is already defined as a `GroupManifold`. So there will just be some inconsistency, which as long as we document, shouldn't be a huge issue.

Are there other matrix subgroups we should add at the same time?

kellertuer:
I will read this more carefully tomorrow, but this sounds great (also that #178 does exist).

But SpecialOrthogonal could just get an AbstractGroupManifold, what's so bad to change that? It would be a breaking change, but despite that...I would be fine with that.

sethaxen:
> But SpecialOrthogonal could just get an AbstractGroupManifold, what's so bad to change that? It would be a breaking change, but despite that...I would be fine with that.

My proposal is to implement something like
```julia
struct SpecialUnitary{N,𝔽} <: AbstractGroupManifold{𝔽,MultiplicationOperation} end

const SpecialOrthogonal{N} = SpecialUnitary{N,ℝ}
```
Currently `SpecialOrthogonal` forwards a lot to `Rotations` automatically because `GroupManifold` is a decorator manifold. It really is just `Rotations` with some group-specific functions. But here `SpecialOrthogonal` would not be a decorator manifold, so we might need to duplicate a lot of code to forward the methods. Not certain though, I haven't reviewed the decorator machinery in a while.

kellertuer:
Ah, thanks for remaining me. Of course; we always handles goup Manifolds as being decorators of a manifold.
But can't we just do that further? I mean all other matrix groups also have a manifold, right?

[...]

PR #249, opened 2020-11-11
Add general and special linear groups

sethaxen:
This PR adds the General Linear group and Special Linear groups, along with a default metric and an additional left-invariant metric.

All of the functionality is there for `GeneralLinear`, but I need to write tests and add docstrings. Per discussions on slack, getting this working may require changing adding something like `EmbeddingType` to `AbstractGroupManifold`, since its the first case implementing a group directly as an `AbstractGroupManifold` and not decorated by a `GroupManifold`.

Also, almost all `SpecialLinear` methods will be identical to `GeneralLinear` (including geodesics; only `check_manifold_point` and `check_tangent_vector` are different), so if possible I'd like to somehow forward everything from `SpecialLinear` to `GeneralLinear` by default.

Relates #225


mateuszbaran:
> Also, almost all `SpecialLinear` methods will be identical to `GeneralLinear` (including geodesics; only `check_manifold_point` and `check_tangent_vector` are different), so if possible I'd like to somehow forward everything from `SpecialLinear` to `GeneralLinear` by default.

Maybe it would then make sense to make `AbstractGroupManifold` be a subtype of `AbstractEmbeddedManifold`? You could then embed `GeneralLinear` in `Euclidean` and `SpecialLinear` in `GeneralLinear`, I think. @kellertuer, would that make sense to you?

kellertuer:
You don't have to subtype. `SpecialLinear` can just be a `AbstractEmbeddedManifold` with embedding `GeneralLinear` (and `TransparentIsometricEmbeddingType`?).

mateuszbaran:
But then `SpeicalLinear` won't be a group manifold? And it doesn't solve the problem of embedding of `AbstractGroupManifold`?

mateuszbaran:
> _Edit_: Oh. That would not give is group functions to also pass on, hm, then, indeed, @mateuszbaran s idea would be easiest here, I think, oh we could introduce for this double model a `AbstractEmbeddedGroupManifold`?

Yes, exactly. Do we actually need a separate `AbstractEmbeddedGroupManifold` though? Could we have an `AbstractGroupManifold` that's not an `AbstractEmbeddedGroupManifold`?

kellertuer:
Sure, maybe we don't. I just wanted to be careful, since it's “only us” not seeing such a case – but a `AbstractGroupManifold` subtyping `AbstractEmbeddedManifold` would be fine with me.

src/groups/general_linear.jl
```diff
@@ -0,0 +1,238 @@
...
+"""
+struct GeneralLinear{n,𝔽} <: AbstractGroupManifold{𝔽,MultiplicationOperation} end
+
+GeneralLinear(n, 𝔽::AbstractNumbers = ℝ) = GeneralLinear{n,𝔽}()
+
+@doc raw"""
+    GLInvariantMetric{T<:Real,D<:ActionDirection} <: Metric
+
+If the matrix is normal, then the resulting geodesic at the identity element is equivalent
+to the group exponential map.
```
kellertuer, line 21: I think you have to reference both in the text to show up?

Also documenting the constructor and the three fields would be a great extension.

src/groups/general_linear.jl
```diff
@@ -0,0 +1,238 @@
...
+inverse_translate(::GeneralLinear, p, q, ::RightAction) = q / p
+
+inverse_translate_diff(::GeneralLinear, p, q, X, ::LeftAction) = X
+inverse_translate_diff(::GeneralLinear, p, q, X, ::RightAction) = p * X / p
+
+function inverse_translate_diff!(G::GeneralLinear, Y, p, q, X, conv::ActionDirection)
+    return copyto!(Y, inverse_translate_diff(G, p, q, X, conv))
+end
+
+function log!(G::GeneralLinear, X, p, q)
```
mateuszbaran, line 181: Maybe it would make sense to write a more generic `inverse_retraction` with an `GeneralLinearInverseRetraction` type that allows specifying parameters for that nonlinear solver? Or does the solver more or less works well enough with these default parameters?

src/utils.jl
```diff
@@ -68,6 +68,27 @@ converted to a `Matrix` before computing the log.
...
 end
 
+sqrt_safe(x) = sqrt(x)
+sqrt_safe(x::Real) = sqrt(Complex(x))
+
+exp_2x2(X) = exp(X)
+exp_2x2(X::AbstractMatrix{T}) where {T<:Union{Real,Complex}} = exp_2x2!(similar(X), X)
+
+exp_2x2!(Y, X) = copyto!(Y, exp(X))
+function exp_2x2!(Y, X::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
```
mateuszbaran, line 78: `StaticArrays` also has this, why don't you just use that implementation?

src/groups/general_linear.jl
kellertuer: One could even do this more generally by using a `ApproximateLogarithmicRetraction` type for these parameters?

mateuszbaran: Yes, that's right. And then it could also be used for example in `Stiefel`.

kellertuer: Exactly. We could adapt the variant from GeomStats using the same type. For generality maybe with a dictionary of options for the solver? Something like that.

sethaxen: > Exactly. We could adapt the variant from GeomStats using the same type. For generality maybe with a dictionary of options for the solver? Something like that.

Adapt what variant? GeomStats' `log` for general linear is equivalent to our `GroupLogarithmicInverseRetraction `.

sethaxen: I think maybe an `ApproximateInverseRetraction{<:AbstractRetractionMethod}` (or maybe `SolvedInverseRetraction`?) would be useful, and then, yes, we could have the kwargs be forwarded to the solver.

kellertuer: Oh, I was referring to the approximate (numerical optimization) log of GeomStats `Stiefel`, sorry for not being precise.

Ah a typed ApproximateRetraction would be a better idea, yes.

sethaxen:
> Maybe it would then make sense to make `AbstractGroupManifold` be a subtype of `AbstractEmbeddedManifold`? You could then embed `GeneralLinear` in `Euclidean` and `SpecialLinear` in `GeneralLinear`, I think. @kellertuer, would that make sense to you?

So I changed the signatures to
```julia
abstract type AbstractGroupManifold{𝔽,O<:AbstractGroupOperation,T<:AbstractEmbeddingType} <:
              AbstractEmbeddedManifold{𝔽,T} end

struct GroupManifold{𝔽,M<:Manifold{𝔽},O<:AbstractGroupOperation} <:
       AbstractGroupManifold{𝔽,O,DefaultIsometricEmbeddingType}
    manifold::M
    op::O
end

struct GeneralLinear{n,𝔽} <:
       AbstractGroupManifold{𝔽,MultiplicationOperation,DefaultEmbeddingType} end
```
I think `DefaultIsometricEmbeddingType` makes sense because the default metric for a `GroupManifold` is the same as the underlying manifold. But this causes tons of test failures for just about every group.

I don't understand any of the machinery for embedding and decorating. I hate to ask his, but could one of you try implementing this change for `AbstractGroupManifold`? It should probably be its own PR anyways.

mateuszbaran:
I'll check what could cause failures here.

mateuszbaran:
I've done some work on this issue here: https://github.com/JuliaManifolds/Manifolds.jl/pull/251 .

sethaxen:
Using `DefaultEmbeddingType` for `GeneralLinear` after #251 mostly works now, but I noticed two issues. First, `distance` is forwarding to `Euclidean`, which seems like it shouldn't happen, as `distance` is dependent on the metric. Also, `mean` is forwarded to `Euclidean` (I didn't check if other statistics functions were as well).

mateuszbaran:
Interesting, `DefaultEmbeddingType` should forward these methods to `Euclidean`.

mateuszbaran:
@kellertuer `EmbeddedManifold` I think `distance` should be `:transparent` only for transparent embeddings and `:parent` otherwise? Now it looks like it's always `:transparent`.

src/groups/special_linear.jl
```diff
@@ -0,0 +1,130 @@
...
+    1            & \text{ if } i ≠ n \\
+    \det(p)^{-1} & \text{ if } i = n
+\end{cases}.
+````
+"""
+project(::SpecialLinear, p)
+
+function project!(::SpecialLinear{n}, q, p) where {n}
+    detp = det(p)
+    if !isapprox(detp, 1)
```
mateuszbaran, line 88: You should still `copyto!` if it's 1.

src/nlsolve.jl
```diff
@@ -0,0 +1,86 @@
...
+    m::ApproximateInverseRetraction,
+)
+
+@doc raw"""
+    inverse_retract(
+        M,
+        p,
+        q
+        method::ApproximateInverseRetraction;
+        X0,
```
mateuszbaran, line 39: I think these kwargs should be stored in `ApproximateInverseRetraction`. Otherwise it's impossible to use them in generic code.

mateuszbaran (commented):
I haven't reviewed everything yet but here are a few comments.

src/groups/group.jl
```diff
@@ -1050,3 +1049,8 @@ function group_exp!(G::MultiplicationGroup, q, X)
     X isa Union{Number,AbstractMatrix} && return copyto!(q, exp(X))
     return error("group_exp! not implemented on $(typeof(G)) for vector $(typeof(X)) and element $(typeof(q)).")
 end
+
+function group_log!(G::MultiplicationGroup, X, q)
+    q isa Union{Number,AbstractMatrix} && return copyto!(X, log_safe(q))
+    return error("group_log! not implemented on $(typeof(G)) for element $(typeof(q)) and vector $(typeof(X)).")
```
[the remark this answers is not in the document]
sethaxen: If I do restrict the arguments, then whenever anyone wants to overload `group_log!` downstream generically, they will need to implement two methods instead of one to avoid ambiguity. This approach reduces code duplication.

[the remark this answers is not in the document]
mateuszbaran: I see, OK.

sethaxen:
IIRC, I paused this because I found `log` to be fussy. e.g. even when I know it should produce real numbers, error creeps into the imaginary parts, causing `InexactError`s when storing it in-place, but real-ifying can introduce separate errors. Part of the problem here IIRC is that LinearAlgebra's `log` implementation does not use the best available algorithm for real logarithm. That's something I'll be addressing before continuing work here.

mateuszbaran:
If the real variants work then maybe it would be worth merging this PR without the complex case and add it in a new PR?

sethaxen:
Actually, it's the complex case that works, because `log: ℂ → ℂ`. It's the real one that's tricky. I could definitely get it working right for the complex case and then come back to it later.

mateuszbaran:
OK, I misunderstood. I think it would be nice to merge it even without `log` for real variants.

sethaxen:
If merged, https://github.com/JuliaLang/julia/pull/39973 will handle the issue of the real logarithm becoming complex for new Julia versions. I've added a simple workaround `log_safe!` here in the meantime. Assuming the test suite passes remotely (they do locally), I think all that's left here is to refactor `ApproximateInverseRetraction` as @mateuszbaran recommended.

src/groups/group.jl
```diff
@@ -1061,3 +1060,8 @@ function group_exp!(G::MultiplicationGroup, q, X)
         "group_exp! not implemented on $(typeof(G)) for vector $(typeof(X)) and element $(typeof(q)).",
     )
 end
+
+function group_log!(G::MultiplicationGroup, X, q)
+    q isa AbstractMatrix && return log_safe!(X, q)
```
mateuszbaran, line 1065: Why don't you put `::AbstractMatrix` in method signature? I think we generally want to get rid of "method not implemented" errors.

sethaxen: If we define `group_log!(G::MultiplicationGroup, X, q::AbstractMatrix)` then any time we want to overload `group_log!` for a single group `MyGroup` under the multiplication operation, we then have to overload `group_log!(G::MyGroup, X, q::AbstractMatrix)` and `group_log!(G::MyGroup, X, q)` to not have ambiguity. At the moment it's not a problem, because only `SpecialOrthogonal` has a custom overload, but it could become a nuisance in the future. What do you think?

sethaxen: I went ahead and changed it. We can always switch it back in the future.

kellertuer: I think this makes – for now – the interface/function signature nice.
My remark on your point would be: Would a `MyGroup` not be `<:AbstractGroup` and hence be more like a sister type of `MultiplicationGroup` such that we would not have the ambiguity problem? I am not 100% sure, but we can surely improve the signature later, if need be.

mateuszbaran: I think it's quite unlikely we would ever want to define `group_log!(G::MyGroup, X, q)` with unrestricted `q`, and in any case we can adjust this method later.

sethaxen:
For future reference, I was trying to work out parallel transport for the general linear group but was only able to work out transport of a tangent vector along the geodesic it defines. There's a working example here: https://gist.github.com/sethaxen/1cca1a79ba95e1c553bf98d71dca78e1. I don't think this maps to one of our functions though.

sethaxen:
Alright, assuming I didn't break something with the latest changes, this PR is ready for review.

mateuszbaran:
> For future reference, I was trying to work out parallel transport for the general linear group but was only able to work out transport of a tangent vector along the geodesic it defines. There's a working example here: https://gist.github.com/sethaxen/1cca1a79ba95e1c553bf98d71dca78e1. I don't think this maps to one of our functions though.

No, we currently don't have a function for geodesic flow. Since you need to solve an ODE for that anyway, maybe we could have a new generic function for that? Of course this would be for another PR, and it requires #325 .

sethaxen:
> No, we currently don't have a function for geodesic flow. Since you need to solve an ODE for that anyway, maybe we could have a new generic function for that? Of course this would be for another PR, and it requires #325 .

Actually, `solve_exp_ode` in `ode.jl` already solves that ODE; we just discard the transported tangent vector for `exp`. It's just that it only works with chart representations, so you're right, we do need #325 to make it general; even then it won't be terribly efficient using Christoffel symbols without some more work.

In this case, however, we can compute the ODE using coordinates of the embedding, so it will be much faster (~200 μs on my machine; Schild's ladder is 10x slower).

mateuszbaran:
Yes, you're right. This makes sense. Currently I use neither geodesic flows nor the general linear group so I don't know how useful that would be.

sethaxen:
> Yes, you're right. This makes sense. Currently I use neither geodesic flows nor the general linear group so I don't know how useful that would be.

In my current planned design for https://github.com/sethaxen/ManifoldHMC.jl, I'll eventually need geodesic flows for implementing symplectic integrators, but that's a long ways off. And for that application, it may only be useful for the few cases where we have closed-form expressions for geodesic flow.

I refactored the code to explicitly have an abstract `ApproximateInverseRetraction` type and a specific `NLsolveInverseRetractionMethod`. For now these are unexported, since we may want to iterate on their design, so for now they're implementation details and not part of the API. The approach works surprisingly well for `GL(n)` but can be a bit temperamental, even for the complex `Circle`.

Note that `distance` for `GeneralLinear` still forwards incorrectly to `Euclidean` (I added an overload to stop this, but it's a bug in the decorator dispatch). Likewise, `mean` incorrectly forwards to `Euclidean`.

src/approx_inverse_retraction.jl
```diff
@@ -0,0 +1,103 @@
...
+then the resulting point is projected after the retraction. `nlsolve_kwargs` are keyword
+arguments passed to `NLsolve.nlsolve`.
+"""
+struct NLsolveInverseRetraction{TR<:AbstractRetractionMethod,TV} <:
+       AbstractInverseRetractionMethod
+    retraction::TR
+    X0::TV
+    project_tangent::Bool
+    project_point::Bool
+    nlsolve_kwargs
```
mateuszbaran, line 34: Why is this field abstractly typed?

sethaxen: Because they're used as kwargs, their type can be anything that expands correctly when splatted, such as an `AbstractDict`, iterator of `Pair`s, or `NamedTuple`, but potentially other types. We don't include the type in the signature, because, e.g. if the user provides a `NamedTuple`, then `inverse_retract` would specialize on the kwargs provided, even though `nlsolve` itself would not specialize on these kwargs. I _think_ this is the right thing to do here.

mateuszbaran: As far as I know `nlsolve` itself would actually specialize on that (unless you pass functions or types but that should not be the case). I just tried a simple benchmark:
```julia
julia> struct A{T}
           kwargs::T
       end

julia> using NLsolve

julia> f(t) = sin(t[1])
f (generic function with 2 methods)

julia> hh(a::A) = nlsolve(f, [2.0]; a.kwargs...)
hh (generic function with 1 method)

julia> a2 = A((; xtol=1e-12))
A{NamedTuple{(:xtol,),Tuple{Float64}}}((xtol = 1.0e-12,))

julia> using BenchmarkTools

julia> @benchmark hh($a2)
```
and it's 5.0 us for concretely typed `kwargs` vs 5.2 us for abstractly typed `kwargs` so it's really hard to tell what kind of performance impact it could have, especially when (in the abstractly typed case) the result of optimization is also abstractly typed.

sethaxen: I changed the kwargs to be concretely typed.

src/groups/general_linear.jl
```diff
@@ -0,0 +1,241 @@
...
+This initial tangent vector is then refined using the  [`NLsolveInverseRetraction`](@ref).
+
+For `GeneralLinear(n, ℂ)`, the logarithmic map is instead computed on the realified
+supergroup `GeneralLinear(2n)` and the resulting tangent vector is then complexified.
+
+Note that this implementation is experimental.
+"""
+log(::GeneralLinear, p, q)
+
+# TODO: Look into using geodesic shooting method in §6.2 of http://hdl.handle.net/2078.1/132587
```
kellertuer, line 183: Should we keep a todo in our code? Should this maybe become an issue to keep better track of it?

kellertuer (commented):
This seems to be a really nice PR. Here's a few remarks and it would be great if you could check for the (few) remaining code coverage lines missing.

src/approx_inverse_retraction.jl
```diff
@@ -0,0 +1,104 @@
+"""
+    ApproximateInverseRetraction <: AbstractInverseRetractionMethod
+
+An abstract type for representing approximate inverse retraction methods.
+"""
+abstract type ApproximateInverseRetraction <: AbstractInverseRetractionMethod end
```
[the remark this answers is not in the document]
sethaxen: Perhaps it's a confusion on my part. Aren't most of our inverse retractions exact inverses of some retraction within some radius? Because that is the sense in which this an approximate inverse retraction. The idea is that you may be able to write some retraction in closed-form but not its inverse, and subtypes of this type would be general-purpose approaches to approximately invert that retraction. Which I guess would end up being a first-order approximation to the log, but indirectly.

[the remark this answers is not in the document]
sethaxen: This is also basically a generalization of geodesic shooting to not just the exponential but also any retraction.

src/approx_inverse_retraction.jl
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+Constructs an approximate inverse retraction for the retraction `method` with initial guess
+`X0`, defaulting to the zero vector. If `project_tangent` is `true`, then the tangent
+vector is projected before the retraction using `project`. If `project_point` is `true`,
+then the resulting point is projected after the retraction. `nlsolve_kwargs` are keyword
+arguments passed to `NLsolve.nlsolve`.
+"""
+struct NLsolveInverseRetraction{TR<:AbstractRetractionMethod,TV,TK} <:
+       AbstractInverseRetractionMethod
```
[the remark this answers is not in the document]
sethaxen: Good catch, yes I did.

src/utils.jl
```diff
@@ -68,6 +68,143 @@ converted to a `Matrix` before computing the log.
...
+Given a complex matrix $X = A + iB ∈ ℂ^{n × n}$, compute its realified matrix
+$Y ∈ ℝ^{2n × 2n}$, written
+where
+````math
+Y = \begin{pmatrix}A & -B \\ B & A \end{pmatrix}.
+````
+"""
+function realify!(Y, X, ::typeof(ℂ), n=LinearAlgebra.checksquare(X))
+    axul, axlr = 1:n, (n + 1):(2n)
+    @views begin
```
sethaxen, line 175: Oh cool, this is marked as not covered because when expanded, the `@views` is a no-op:
```julia
julia> @macroexpand @views begin
               Y[axul, axul] .= Y[axlr, axlr] .= real.(X)
               Y[axlr, axul] .= imag.(X)
               Y[axul, axlr] .= .-imag.(X)
           end
quote
    #= REPL[153]:2 =#
    Y[axul, axul] .= (Y[axlr, axlr] .= real.(X))
    #= REPL[153]:3 =#
    Y[axlr, axul] .= imag.(X)
    #= REPL[153]:4 =#
    Y[axul, axlr] .= (.-)(imag.(X))
end
```

src/groups/general_linear.jl
sethaxen: Yeah, perhaps an issue would be more appropriate. The goal here is that a user/developer reading the code understands that there are alternative approaches that might be better. Unless a user browses our issues, they may not find it. A compromise would be to create an issue and have a comment linking to the issue in the code.

sethaxen: I moved it to an issue and opted to remove the comment. Also, the point in the comment about normal matrices was incorrect. It is not the case I believe that the exponential of a normal matrix is normal. And many normal matrices will be invalid for our purposes. e.g. we need the projected matrix to also have a positive determinant and be non-singular, which is not guaranteed.

src/approx_inverse_retraction.jl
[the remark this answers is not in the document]
kellertuer: ah so its a general approximation method, ten its ok, for sure.

sethaxen:
Is this ready for approval?

kellertuer (approved):
...even with a negative number on the change of missed covered lines in tests. nice.

PR #251, opened 2020-11-12
Some work on making groups embedded

mateuszbaran:
Related to https://github.com/JuliaManifolds/Manifolds.jl/pull/249#issuecomment-725789794 . This seems so mostly work but there are some failures with `get_vector` and `get_coordinates` (it looks like there is some problem with dispatch transparency). @kellertuer , could you help with that?

### DIFF
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -81,7 +81,7 @@ function check_manifold_point(M::Rotations{N}, p; kwargs...) where {N}
     if !isapprox(transpose(p) * p, one(p); kwargs...)
         return DomainError(
             norm(transpose(p) * p - one(p)),
-            "$p has to be orthogonal but it's not at kwargs $kwargs",
+            "$p must be orthogonal but it's not at kwargs $kwargs",
         )
     end
     return nothing
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -79,7 +79,10 @@ function check_manifold_point(M::Rotations{N}, p; kwargs...) where {N}
         return DomainError(det(p), "The determinant of $p has to be +1 but it is $(det(p))")
     end
     if !isapprox(transpose(p) * p, one(p); kwargs...)
-        return DomainError(norm(p), "$p has to be orthogonal but it's not")
+        return DomainError(
+            norm(transpose(p) * p - one(p)),
+            "$p has to be orthogonal but it's not at kwargs $kwargs",
+        )
     end
     return nothing
 end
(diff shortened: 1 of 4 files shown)


mateuszbaran:
Alternatively we could also leave `AbstractGroupManifold` as-is and add `AbstractEmbeddedGroupManifold` but that would mean we have two separate type hierarchies for group manifolds (since `AbstractEmbeddedGroupManifold` can't inherit both `AbstractGroupManifold` and `AbstractEmbeddedManifold`).

kellertuer:
no I think the current approach is the good one, you already convinced me. I will take look probably in the evening.

kellertuer:
While for Embedded Manifolds we surely do not want `get_vector!` and `get_coordinates!` to be transparent, for groups we surely want that - so I changed that.

Now only an `is_tangent_vector` test for `special_orthogonal` seems to fail and I don't see directly why.

mateuszbaran:
I've slightly increased tolerance and that special orthogonal test passed locally for me.

src/manifolds/Rotations.jl
```diff
@@ -79,7 +79,10 @@ function check_manifold_point(M::Rotations{N}, p; kwargs...) where {N}
         return DomainError(det(p), "The determinant of $p has to be +1 but it is $(det(p))")
     end
     if !isapprox(transpose(p) * p, one(p); kwargs...)
-        return DomainError(norm(p), "$p has to be orthogonal but it's not")
+        return DomainError(
+            norm(transpose(p) * p - one(p)),
+            "$p has to be orthogonal but it's not at kwargs $kwargs",
```
sethaxen, line 84:
```suggestion
            "$p must be orthogonal but it's not at kwargs $kwargs",
```

test/groups/special_orthogonal.jl
```diff
@@ -90,7 +90,7 @@ include("group_utils.jl")
             inverse_retraction_methods = inverse_retraction_methods,
             exp_log_atol_multiplier = 20,
             retraction_atol_multiplier = 12,
-            is_tangent_atol_multiplier = 1,
+            is_tangent_atol_multiplier = 1.2,
```
sethaxen, line 93: Any ideas why the change in tolerance is necessary? I'm wondering if a different default is being hit now than before.

mateuszbaran: I have no idea. It's a very small change so maybe there is no reason to worry about it?

sethaxen: Probably!

sethaxen (approved):
I think this just needs a patch version bump and then looks good to merge to me. Thanks!

kellertuer (approved):
If you have a moment of time to look for the one line the tests are not covering, that would be great, but in general I approve this.

mateuszbaran:
I've looked at a few coverage failures and they look like false positives so I'll leave them.

PR #255, opened 2020-11-16
Fix transparency of get_vector! and get_coordinates! for AbstractGroupManifold

mateuszbaran:
I think this was wrongly defined in #251 .


kellertuer:
Yes, sure, I was not precise enough there.

kellertuer (approved):

PR #256, opened 2020-11-17
Stabilize SPD distance and a small fix to TangentBundle

kellertuer:
This is just a small bug fix PR.

We noticed that `cholesky` gets very unstable when the matrix is not symmetric (and one disables the checks). With an added `Symmetric` it works fine and is accurate again.

Furthermore for tangent bundle the inner was not working. A tangent vector to the tangent bundle is basically two tangent vectors, so the second (tangent components) inner has to also use the base-base point as inner. @mateuszbaran can you check whether this also holds for the vector bundle? Then we could unify both implementations again.

### DIFF
--- a/src/manifolds/VectorBundle.jl
+++ b/src/manifolds/VectorBundle.jl
@@ -633,7 +633,7 @@ function inner(B::VectorBundle, p, X, Y)
     px, Vx = submanifold_components(B.manifold, p)
     VXM, VXF = submanifold_components(B.manifold, X)
     VYM, VYF = submanifold_components(B.manifold, Y)
-    return inner(B.manifold, px, VXM, VYM) + inner(B.fiber, Vx, VXF, VYF)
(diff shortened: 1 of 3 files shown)


src/manifolds/VectorBundle.jl
```diff
@@ -635,6 +635,12 @@ function inner(B::VectorBundle, p, X, Y)
     VYM, VYF = submanifold_components(B.manifold, Y)
     return inner(B.manifold, px, VXM, VYM) + inner(B.fiber, Vx, VXF, VYF)
```
mateuszbaran, line 636:
```suggestion
    return inner(B.manifold, px, VXM, VYM) + inner(VectorSpaceAtPoint(B.fiber, px), Vx, VXF, VYF)
```

It was wrong here. For arbitrary vector bundles it should be this way I think. `Vx` is a point on the vector space manifold and `VXF` and `VYF` are tangent vectors to that vector space at point.

mateuszbaran (approved):
Thanks for fixing it :+1: .

PR #257, opened 2020-11-17
Tangent bundle parallel transport and maintenance

mateuszbaran:
I've added `ParallelTransport` for tangent bundles, and two minor things discussed with @kellertuer .

### DIFF
--- a/src/manifolds/VectorBundle.jl
+++ b/src/manifolds/VectorBundle.jl
@@ -994,15 +994,15 @@ function vector_transport_to!(M::TangentBundle, Y, p, X, q, m::VectorBundleVecto
     vector_transport_to!(M.manifold, VYF, px, VXF, qx, m.method_vector)
     return Y
 end
-function vector_transport_to!(M::VectorBundle, Y, p, X, q, m::ParallelTransport)
-    return vector_transport_to!(
-        M,
-        Y,
-        p,
-        X,
-        q,
-        VectorBundleVectorTransport(ParallelTransport(), ParallelTransport()),
-    )
+function vector_transport_to!(
+    M::TangentBundle,
+    Y,
+    p,
+    X,
+    q,
+    m::AbstractVectorTransportMethod,
+)
+    return vector_transport_to!(M, Y, p, X, q, VectorBundleVectorTransport(m, m))
 end
 function vector_transport_to!(M::TangentSpaceAtPoint, Y, p, X, q)
     return copyto!(Y, X)
--- a/src/manifolds/VectorBundle.jl
+++ b/src/manifolds/VectorBundle.jl
@@ -1004,6 +1004,14 @@ function vector_transport_to!(
 )
     return vector_transport_to!(M, Y, p, X, q, VectorBundleVectorTransport(m, m))
 end
+@invoke_maker 6 AbstractVectorTransportMethod vector_transport_to!(
+    M::TangentBundle,
+    Y,
+    p,
+    X,
+    q,
+    m::PoleLadderTransport,
+)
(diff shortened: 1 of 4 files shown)


kellertuer:
I think we even can do this for any other transport, too? Though we somehow have to exclude `VectorBundleVectorTransport`for sure.

mateuszbaran:
I'm not sure it's a good idea. If you had a tangent bundle to a tangent bundle, then you wouldn't be sure what to do with `VectorBundleVectorTransport(transport1, transport2)`. Do you pass these inner transports down separately to the inner tangent bundle or the whole transport? I'd prefer to limit this forwarding, at least for now, to an explicit list of vector transport methods.

kellertuer:
Ok.
I would have just passed them down to both,
so on Tangent of Tangent, using a method would mean to use the same on all 4 parts.

But we can skip that surely to a time when we actually need it.

kellertuer:
I just thought about this for a while; we could really just do your approach of doubling for any `AbstractVectorTransportMethod` and the bundle case with different ones is covered already anyways.

That way
* If you do not want to think about it too much all “usual” methods apply
* If you do not want to think about it at all, it works as all others (global default Paralleltransport works and there is no need for a special case here)
* If you have something special in mind, also for nested bundles, you can use (nested) bundle transports.

I think that's fine, what do you think?

mateuszbaran:
One problem here is, I don't know how to do the "double first and then unwrap" approach. By default the reverse would be done (unwrap first and then double).

kellertuer:
But could we just do
```
vector_transport_to(M::TangentBundle, p, X, q, m::A) where {A <: AbstractVectorTransportMethod} = vector_transport_to(M, p, X, q, VectorBundleTransportMethod(m,m))
```
or am I missing something? that would pass it on to the method we already have that then splits this double case?

mateuszbaran:
> But could we just do
> 
> ```
> vector_transport_to(M::TangentBundle, p, X, q, m::A) where {A <: AbstractVectorTransportMethod} = vector_transport_to(M, p, X, q, VectorBundleTransportMethod(m,m))
> ```
> 
> or am I missing something? that would pass it on to the method we already have that then splits this double case?

We can do that and that does exactly the unwrap-first-then-double approach, since the method for `VectorBundleTransportMethod` is more specific. We could maybe in theory take "depth" of vector bundles (how many are wrapped around each other) into account but it would be really ugly.

kellertuer:
Oh, but that's all i meant to do – I don't see the unwrap in there, I just see the double.
For me this fits quite well, especially if you interpret the vector bundle transport method as more like an implementation detail (for people who want more detail).

mateuszbaran:
I am OK with it working this way but in the previous comment: https://github.com/JuliaManifolds/Manifolds.jl/pull/257#issuecomment-729190362 you wrote that you would prefer the opposite approach then the one offered the method from https://github.com/JuliaManifolds/Manifolds.jl/pull/257#issuecomment-729605577 .

mateuszbaran:
Or maybe I misunderstood something.

kellertuer:
Also, maybe I missphrased that, what I meant was what I proposed now here ;)

mateuszbaran:
OK, it should work this way now.

kellertuer:
Would the old `vector transport_to`now be obsolete since PT is now safely usable as default? (old lines 972-981) as well as the default? Similarly the default for the mutating variant?
Or do you want to keep it default to use the internal one from the vector bundle?

mateuszbaran:
Probably not obsolete. PT is now safely usable only for tangent bundles and all these methods are reported as covered by tests, so maybe it's better to leave it?

kellertuer:
Ah, sure.

PR #264, opened 2020-11-20
Make the normal retraction, inverse retraction and vector transport methods available for PowerManifold

### LINKED ISSUE #263: Get rid of PowerX
I am would like to get rid of the
* PowerRetraction
* PowerVectorTransport
* InversePowerRetraction (we might currently miss to export this)

The reason is easy. When I wrote my newest algorithm, I used exp/log/parallel transport. All was easy to call and nicely working.
Yesterday I generalized that to use arbitrary retractions/inverse_retractions/vector_transports,
so now for the problem on a `PowerManifold` to properly work as before I generalized the code I have to explicitly state the keywords
```
        vector_transport_method = PowerVectorTransport(ParallelTransport()),
        retracttion_method = PowerRetraction(ExponentialRetraction()),
        inverse_retraction = InversePowerRetraction(LogarithmicInverseRetraction()),
```
in my algorithm, which I find not nice. Maybe like we do for Basis we can resolve this ambiguity for all 3 types explicitly, though that means to store a list of the types within?

kellertuer:
The idea to resolve #263 is to

Introduce these methods as something that passes `m::AbstractRetractionMethod` further on to the abstract power manifold as a `PowerRetractionMethod(m)`. While I think the function get called, there are a lot of tests failing and I am not yet sure why. So this is WIP, trying to figure this out. Any ideas welcome.


mateuszbaran:
Power manifold tests passed locally for me after these changes so let's see if CI passes.

kellertuer:
Cool. Could we merge this one? Then I am quite close to finishing JuliaManifolds/Manopt.jl#40.

src/manifolds/PowerManifold.jl
```diff
@@ -124,18 +124,34 @@ tangent space of the power manifold.
 struct PowerMetric <: Metric end
 
 """
-    PowerRetraction(retraction::AbstractRetractionMethod)
+    PowerRetraction{TR<:AbstractRetractionMethod} <: AbstractRetractionMethod
+
+The `PowerRetraction` avoids ambiguities between dispatching on the [`AbstractPowerManifold`](@ref)
+and dispatching on the [`AbstractRetractionMethod`](@ref) and ecapsulates this.
```
mateuszbaran, line 130:
```suggestion
and dispatching on the [`AbstractRetractionMethod`](@ref) and encapsulates this.
```

src/manifolds/PowerManifold.jl
```diff
@@ -124,18 +124,34 @@ tangent space of the power manifold.
...
+This container should only be used in rare cases outide of this package. Usually an
```
mateuszbaran, line 131:
```suggestion
This container should only be used in rare cases outside of this package. Usually a
```

src/manifolds/PowerManifold.jl
```diff
@@ -124,18 +124,34 @@ tangent space of the power manifold.
...
     retraction::TR
 end
 
 """
-    InversePowerRetraction(inverse_retractions::AbstractInverseRetractionMethod...)
+    InversePowerRetraction{TR<:AbstractInverseRetractionMethod} <: AbstractInverseRetractionMethod
+
+The `InversePowerRetraction` avoids ambiguities between dispatching on the [`AbstractPowerManifold`](@ref)
+and dispatching on the [`AbstractInverseRetractionMethod`](@ref) and encapsulates this.
+This container should only be used in rare cases outide of this package. Usually an
```
mateuszbaran, line 148:
```suggestion
This container should only be used in rare cases outside of this package. Usually a
```

src/manifolds/PowerManifold.jl
```diff
@@ -173,9 +189,18 @@ struct PowerFVectorDistribution{
 end
 
 """
-    PowerVectorTransport(method::AbstractVectorTransportMethod)
+    PowerVectorTransport{TR<:AbstractVectorTransportMethod} <:
+       AbstractVectorTransportMethod
+
+The `PowerVectorTransport` avoids ambiguities between dispatching on the [`AbstractPowerManifold`](@ref)
+and dispatching on the [`AbstractVectorTransportMethod`](@ref) and encapsulates this.
+This container should only be used in rare cases outide of this package. Usually an
```
mateuszbaran, line 197:
```suggestion
This container should only be used in rare cases outside of this package. Usually a
```

mateuszbaran (approved):
I think we can merge this after fixing these typos.

PR #265, opened 2020-11-20
Semantic access to points on tangent and cotangent bundle

kellertuer:
This PR aims to provide
* `p[M, :base]`
* `p[M, :Tangen]`and `p[M, :cotangent]`, respectively,

but today seems to be the case of ideas, that are too clever for me.

* ProductRepr and ProductArray are only loaded after bundle and I am not sure why, but I can't change the order just now
* I am not sure a `setindex!`would work with `ProductRepr`
* I am not sure a `ProductArray` is useful for (co)tangent bundles, since I cound find no documentation on how to construct/use `ProductArray`s (only that `ProductRepr`which I used until now may be slower than the thing I can't construct).

### DIFF
--- a/docs/src/manifolds/vector_bundle.md
+++ b/docs/src/manifolds/vector_bundle.md
@@ -21,6 +21,21 @@ This is also considered a manifold.
 For cases where confusion between different types of vectors is possible, the type [`FVector`](@ref) can be used to express which type of vector space the vector belongs to.
 It is used for example in musical isomorphisms (the [`flat`](@ref) and [`sharp`](@ref) functions) that are used to go from a tangent space to cotangent space and vice versa.
 
+## Example
+
+The following code defines two points on a tangent bundle of the sphere $S^2$ and calculates distance between them, distance between their base points and norm of one of these tangent vectors.
+
+```@example
+using Manifolds
+M = Sphere(2)
+TB = TangentBundle(M)
+p = ProductRepr([1.0, 0.0, 0.0], [0.0, 1.0, 3.0])
+q = ProductRepr([0.0, 1.0, 0.0], [2.0, 0.0, -1.0])
+println("Distance between p and q: ", distance(TB, p, q))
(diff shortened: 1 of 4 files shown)


mateuszbaran:
Can I change it to `p[M, :vector]` for the vector part and `p[M, :point]` for the point part for all vector bundles?

kellertuer:
Sure, I don't have a strong preference for naming, we should just also document that somewhere.

mateuszbaran:
> * I am not sure a `setindex!`would work with `ProductRepr`

It could in a way, that is copying the contents of arrays instead of arrays themselves.

> * I am not sure a `ProductArray` is useful for (co)tangent bundles, since I co�und find no documentation on how to construct/use `ProductArray`s (only that `ProductRepr`which I used until now may be slower than the thing I can't construct).

`ProductArray` could be useful for tangent bundles but not for cotangent bundles. BTW, `ProductRepr` is slower only in a couple of specific scenarios so I think it's fine to use it essentially everywhere. I'm no longer sure if it's even worth it to have `ProductArray` but I don't consider it particularly important to evaluate now, since new versions of Julia may impact any such comparison.

kellertuer:
Thanks for the explanations. My problem with `ProductArray`, is that I always struggle to initialise those, `ProductRepr` is easy to use indeed.

mateuszbaran:
What do you think about the new version I've just commited?

kellertuer:
That was what I had in mind. Can you explain why I need the `Val()`?
That seems to be the major difference. In my Manopt-code I did not need the `Val`.

mateuszbaran:
`Val` is just to ensure constant propagation of these indices. It would most likely be propagated anyway but this surely works even if inlining decides to do weird things.

mateuszbaran:
I've added some docs for this feature.

kellertuer:
Looks great, thanks for your help!

PR #272, opened 2020-11-25
Update to TagBot.yml

kellertuer:
See https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249 for details.


kellertuer:
If I only could write config files. I am sure there is some space too much, but first I have to teach now. Will check later.

PR #274, opened 2020-11-26
Moving power manifold to base

mateuszbaran:
Some work to resolve https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/54#issuecomment-734222843 .


kellertuer:
I think, this, too can be merged – would just be a queston whether my idea of 0.5.0 is ok or 0.4.11 would also be good, since it's just an nternal change?

mateuszbaran:
Yes, this looks more or less ready. I've just made a small cleanup related to imports. 0.4.11 should be OK too, since that's just an internal change.

kellertuer:
Since this is again teamwork, I am not sure whether I shall merge this or you, but I would be fine with merging.

mateuszbaran:
I think I can merge it :slightly_smiling_face:

PR #275, opened 2020-11-27
CompatHelper: bump compat for "SpecialFunctions" to "1.0"

sethaxen:
This pull request changes the compat entry for the `SpecialFunctions` package from `0.8, 0.9, 0.10` to `0.8, 0.9, 0.10, 1.0`.


kellertuer:
I think we have to still wait for this one https://github.com/JuliaDiff/ReverseDiff.jl/pull/155. Also with ambiguities and nightly it seems to change quite often, maybe we could deactivate ambiguity testing for nightlies?

mateuszbaran:
Yes, we have to wait until ReverseDiff is compatible with SpecialFunctions 1.0. I also think it's a good idea to disable ambiguity testing on nightlies, it changes too often to track.

kellertuer:
new version, see #295 I still wonder why CompatHelper triggers every day a new one if the version is the same? Even asked https://github.com/JuliaRegistries/CompatHelper.jl/issues/261 and tried adapting our settings, but did not get an answer.

mateuszbaran:
It seems to finally pick the new version of SpecialFunctions so I'll merge it.

PR #277, opened 2020-11-27
support specialfunctions 1.0

stevengj:


kellertuer:
Thanks for your PR, we have #275 and are waiting for ForwardDiff basically, for example https://github.com/JuliaDiff/ForwardDiff.jl/pull/474. As soon as that's updated we can surely update here, too.

stevengj:
Great!

PR #295, opened 2020-12-03
CompatHelper: bump compat for "SpecialFunctions" to "1.1"

sethaxen:
This pull request changes the compat entry for the `SpecialFunctions` package from `0.8, 0.9, 0.10` to `0.8, 0.9, 0.10, 1.1`.


mateuszbaran:
I think we should still use 1.0 in compat instead of this, this PR makes it so that 1.0 is not compatible. Definitely not a big issue though.

kellertuer:
Ah sure, you're right.

PR #317, opened 2020-12-21
Better consistency with the blue style

mateuszbaran:
I've added a few more rules that are now available in JuliaFormatter for the Blue Style. `whitespace_in_kwargs = false` has the biggest impact here.


kellertuer (approved):
Great, thanks for updating. I hope they finish their Blue Style issue soon, so that it will be just `style=Blue` in the `.toml`. Optimistic as I was, I started with that in Manopt.jl a while back.

mateuszbaran:
There is still quite a lot of missing things unfortunately: https://github.com/domluna/JuliaFormatter.jl/issues/283 so for now `style=Blue` doesn't look like a good idea.

PR #318, opened 2020-12-24
Introduce a common type for differentiated retractions as vector transport and two such transports for Stiefel

kellertuer:
Here's my little Christmas project. 🎄
I read a little bit about two new differentiated retractions (they are actually available in ROPTLIB (https://github.com/whuang08/ROPTLIB), but not that well documented.

- [x] Testing
- [x] Check Documentation
- [x] Check whether large interims matrices can be avoided (currently for example `d*d'` should be 1000x100 for `Stiefel(1000,2)` which should be avoided by changing order of compuations.


src/manifolds/Stiefel.jl
```diff
@@ -619,3 +596,221 @@ function uniform_distribution(M::Stiefel{n,k,ℝ}, p) where {n,k}
 
     return ProjectedPointDistribution(M, d, project!, p)
 end
+
+@doc raw"""
+    vector_transport_direction(::Stiefel, p, X, d, ::DifferentiatedRetraction{CaleyRetraction})
```
mateuszbaran, line 601: Shouldn't it "Cayley" instead of "Caley"?

src/manifolds/Stiefel.jl
```diff
@@ -55,12 +55,21 @@ A retraction based on the Caley transform, which is realized by using the
 """
 const CaleyRetraction = PadeRetraction{1}
 
+# This could maybe also be moved to Base.
 """
-    CaleyVectorTransport <: AbstractVectorTransportMethod
+    DifferentiatedRetraction{R<:AbstractRetractionMethod} <: AbstractVectorTransportMethod
 
-A vector transport that one obtains by differentiating a [`CaleyRetraction`](@ref).
+A type to specify a vector transport that is given by differentiating a retraction
```
mateuszbaran, line 62: Could you add a generic formula here in docs? It looks like it would work for any manifold. Additionally, isn't `ParallelTransport` a `DifferentiatedRetraction` for the exponential retraction?

Also, maybe a different name would be better, `DifferentiatedRetraction` doesn't clearly indicate that it's a vector transport and is a bit vague.

src/manifolds/Stiefel.jl
kellertuer: Thanks for noticing the typo.

src/manifolds/Stiefel.jl
kellertuer: Yes, this would work for any manifold and yes, as far as I see, the DifferentiatedRetraction of exp is PT; just for fun we could also add this.
Currently it's placed within `Stiefel` since I did not find a good place to put it within `Manifolds.jl` since it is generic enough to even go over to `ManifoldsBase.jl` I think.

I changed the name to `DifferentiatedRetractionVectorTransport` – I initially felt that to be too long, but you're right that otherwise one does not see what it is. I also added the formula from Absil Mahony Sepulchre.

mateuszbaran: Nice, it looks much more clear now. I think we can definitely put `DifferentiatedRetractionVectorTransport` in `ManifoldsBase.jl` and define `ParallelTransport` as an alias for differentiated retraction of the exponential map.

src/manifolds/Stiefel.jl
```diff
@@ -48,19 +48,52 @@ function PadeRetraction(m::Int)
...
+``p\in\mathcal M`` a point, and ``X,Y\in T_p\mathcal M`` denote two tangent vectors at ``p``.
+
+Given a retraction (cf. [`AbstractRetractionMethod`](@ref)) ``\operatorname{retr}``,
+the vector transport of `X` in direction `Y` (cf. [`vector_transport_direction`](@ref))
+by differentiation this retraction, is given by
+
+```math
+\mathcal T^{\operatorname{retr}}_{p,Y}X
+= D_Y\operatorname{retr}_x(Y)[X]
+= \frac{\mathrm{d}}{\mathrm{d}t}\operatorname{retr}_x(Y+tX)\Bigr|_{t=1}.
```
mateuszbaran, line 74:
```suggestion
= D_Y\operatorname{retr}_p(Y)[X]
= \frac{\mathrm{d}}{\mathrm{d}t}\operatorname{retr}_p(Y+tX)\Bigr|_{t=0}.
```
I think that's the correct formula?

src/manifolds/Stiefel.jl
kellertuer: Ok, I will do that.

src/manifolds/Stiefel.jl
```diff
@@ -619,3 +620,227 @@ function uniform_distribution(M::Stiefel{n,k,ℝ}, p) where {n,k}
...
+    M::Stiefel,
+    Y,
+    p,
+    X,
+    d,
+    ::DifferentiatedRetractionVectorTransport{QRRetraction},
+)
+    q = retract(M, p, d, QRRetraction())
+    rf = qr(p + d).R
+    return copyto!(Y, q * matUpper2skew((q' * X) / rf) + (X - q * (q' * X)) / rf)
```
mateuszbaran, line 747:
```suggestion
    rf = qr(p + d).R
    Xrf = X / rf
    qtXrf = q' * Xrf
    return copyto!(Y, q * matUpper2skew(qtXrf) + Xrf - q * qtXrf)
```
I've extracted some common computations. Could you check if that's correct? Additionally, `UpperTriangular(qtXrf) - UpperTriangular(qtXrf)'` is probably be a more standard way of writing `matUpper2skew`.

kellertuer: Sure we can also switch to your suggestion for the skew. I wasn't aware of that. For your rewriting – are you sure that holds for non-square matrices, that one can factor our the `q'`?

mateuszbaran: This rewriting relies only on associativity and distributivity of matrix multiplication and this holds for non-square matrices.

By the way, for Stiefel the R factor of QR decomposition is I think an upper triangular matrix so it would be worth applying the `UpperTriangular` wrapper (to dispatch to the more efficient solver for `/`).

mateuszbaran:
```
(X - q * (q' * X)) / rf = X / rf - q * (q' * X) / rf
q * (q' * X) / rf = q * (q' * (X / rf))
```

kellertuer: I will check for Uppertriangular after writing a small test on Base for the new branch there.

mateuszbaran:
The failure on Julia nightly is now caused by this: https://github.com/JuliaPy/PyCall.jl/issues/873 .

kellertuer:
Thanks for investigating, I did not yet have the time to investigate (and was happy to get pyplot-stuff running in the first place, which took some time and tries).

mateuszbaran:
No problem, I think it's good to see why nightly builds are failing once in a while.

BTW, I just noticed that the description of scaled transport has some issues: https://juliamanifolds.github.io/Manifolds.jl/previews/PR318/interface.html#ManifoldsBase.ScaledVectorTransport .

kellertuer:
Thanks for noticing there is a space missing; I will push that fix to master and it will then get into the next version,

mateuszbaran (approved):
OK, I don't have any other comments here :slightly_smiling_face: .

kellertuer:
Cool! I just bumped the version and will then merge in a few minutes.

PR #319, opened 2020-12-26
Improve Test style and coverage

### LINKED ISSUE #259: Improve test suite organization
Currently if one of the test sets fails, the test suite does not run to completion. The way to fix this is to place all `include` statements within a giant `@testset`. The downside is that if you are then running in the REPL, you don't get instant feedback about when a test set completes. The way around this is to print a message upon include.

https://github.com/JuliaDiff/ChainRules.jl/pull/308 did a nice job of this, also printing out time taken to execute a testset.

kellertuer:
This PR resolves #259 and
* I checked all files and for the lines not mentioned below we do have tests that do not trigger somwhow/why.
* I removed all >= 1.3 checks since we have 1.3 as a minimium requirement already.
* should we exclude the manifold test functions form coverage? Currently a few `@testset`s are missed and a few nonmutating else-cases, 

The following is a list of lines where I am not sure how to cover those, but it would be great to have these covered, too.
Sure with our current coverage this is maybe not that necessary, I just would like to have some feedback in them :)

- [x] What do we do about the statistics default warning?
- [x] -> #312 Can we do a test for https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/groups/metric.jl#L216
- [x] -> #312 How can we best test the `det_local_metric`? For example
  - https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/PowerManifold.jl#L73
  - https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#211
- [x] -> #312 How can we best test the `Powerdistribution`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/PowerManifold.jl#L191 
- [x] Can we test allocation promotion on Product? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#109
- [x] -> #312 Can we test `ProductFVectorDistribution`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#799
- [x] -> #312] Can we test `ProductManifold`s `_rand`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#861
- [x] -> #312] Which errors do appear in `Rotations`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L438
- [x] -> #312 Can we make up a sopecial case for this situation in `Rotations` `log!`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L504
- [x] I hope there is an easy way for the special 1D rotations matrix random? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L662
- [x] For Stiefel I am not sure when this case happens? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Stiefel.jl#L338
- [x] -> #312 Do we have a tangent space fibre bundle test do add this line? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/VectorBundle.jl#L617
- [x] -> #312 Similarly a tangent space type fiber should cover the norm, too? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/VectorBundle.jl#L736
- [x] -> #312 similarly the generic project https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/VectorBundle.jl#L820
- [x] -> #312 Distributions are missing a non-implemented check https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/distributions.jl#L62
- [x] is there an easy test for projected point distributions support ? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/projected_distribution.jl#L45
- [x] -> #312 How far is the Riemannian gradient function stuff, we could add differential! and gradient! tests? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/riemannian_diff.jl#L27
- [x] do we ever reach the last line in `mean_and_var` and how can we get that better? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/statistics.jl#L860
- [x] -> #312 from the utils: Are `ziptuples`, `eiogvals` (for static arrays) and `eigen_safe` for a `StaticMatrix` still needed or are they now part of `StaticArrays in some form? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/utils.jl#L54


mateuszbaran:
Wow, thanks for working on it. That's a nice cleanup. I will try to find some time to give you some feedback here.

mateuszbaran:
> * should we exclude the manifold test functions form coverage? Currently a few `@testset`s are missed and a few nonmutating else-cases,

I wouldn't exclude them but covering them seems like a low priority task.

> Sure with our current coverage this is maybe not that necessary, I just would like to have some feedback in them :)

Sure, we have a good coverage but of course it can be improved :slightly_smiling_face: .

> [ ] What do we do about the statistics default warning?

Which default warning?

> [ ] Can we do a test for https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/groups/metric.jl#L216

Sure, we just need a metric group manifold with a non-bi-invariant metric.

> How can we best test the `det_local_metric`? For example
> * https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/PowerManifold.jl#L73
> * https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#211

I think the whole metric stuff needs a larger rework aimed at addressing #6. IIRC this quantity depends on the selection of a chart so this is somewhat broken now. I think this is perhaps supposed to return this for normal coordinates which is I think well defined but I don't like it anyway.

> * [ ]  How can we best test the `Powerdistribution`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/PowerManifold.jl#L191
> * [ ]  Can we test allocation promotion on Product? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#109
> * [ ]  Can we test `ProductFVectorDistribution`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#799
> * [ ]  Can we test `ProductManifold`s `_rand`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/ProductManifold.jl#861

That looks doable, I'll take a look at it later.

> [ ] Which errors do appear in `Rotations`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L438

I don't know, `sylvester` for custom array types could throw something else so I wouldn't touch it.

> * Can we make up a sopecial case for this situation in `Rotations` `log!`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L504

Probably yes but it looks like a very special case.

> * I hope there is an easy way for the special 1D rotations matrix random? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L662

Sure, I can add that.

kellertuer:
Thanks for the long response. I think we do not have to solve all of my points, for example if the det local metric is not yet finished, there is surely no need to do a test.

> > [ ] What do we do about the statistics default warning?
> 
> Which default warning?

Sorry, that I forgot to link something, this one
https://github.com/JuliaManifolds/Manifolds.jl/pull/319/checks?check_run_id=1611331687#step:5:395

> > [ ] Which errors do appear in `Rotations`? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/manifolds/Rotations.jl#L438
> 
> I don't know, `sylvester` for custom array types could throw something else so I wouldn't touch it.

I did mean, to we have a case where we expect the rethrow so we can check that line? Or can we do that in a one-line instead of with an if?

mateuszbaran:
> Thanks for the long response. I think we do not have to solve all of my points, for example if the det local metric is not yet finished, there is surely no need to do a test.

No problem, I'll go through other points too :slightly_smiling_face: .

> Sorry, that I forgot to link something, this one
> https://github.com/JuliaManifolds/Manifolds.jl/pull/319/checks?check_run_id=1611331687#step:5:395

I'd leave it for now.

> I did mean, to we have a case where we expect the rethrow so we can check that line? Or can we do that in a one-line instead of with an if?

It can be either tested by providing a custom method for `sylvester` for some custom types. It can also be rewritten in one line using the ternary operator ` ? : ` but the rethrow still wouldn't be covered. Ideally https://github.com/JuliaLang/julia/issues/7026 would solve this.

mateuszbaran:
> * from the utils: Are `ziptuples`, `eiogvals` (for static arrays) and `eigen_safe` for a `StaticMatrix` still needed or are they now part of `StaticArrays in some form? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/utils.jl#L54

`eigvals` from StaticArrays.jl should be good enough now, I've removed it. `ziptuples` is still needed since it's not in StaticArrays.jl I think? `eigen_safe` can probably be removed as well.

> [ ] is there an easy test for projected point distributions support ? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/projected_distribution.jl#L45

There seems to be an actual major problem with projected point distributions (they seem to not work at all).

mateuszbaran:
> [ ] How far is the Riemannian gradient function stuff, we could add differential! and gradient! tests? https://codecov.io/gh/JuliaManifolds/Manifolds.jl/src/master/src/riemannian_diff.jl#L27

These are just some fallbacks that aren't used. Riemannian gradients should work, let me know if there are any problems. You should be able to use them in Manopt.jl.

kellertuer:
Cool, I hopefully find time to check it or write an example about that then.

mateuszbaran:
I've updated formatting and hopefully fixed the projected distributions. I don't know why they had the decorator dispatch but it seems to only break things. Maybe it would be nice to fix it but `:parent` dispatch on metric manifold, which I thought should do the right thing, apparently unwrapped the metric decorator in the Euclidean manifold tests.

kellertuer:
Thanks for working on that. I hopefully find some time this week to also add a test or two. You can also mark things that we discussed.

Are there tests available for the fallbacks of the Riemannian Gradient or do we leave those untested?

mateuszbaran:
> Thanks for working on that. I hopefully find some time this week to also add a test or two. You can also mark things that we discussed.

Thanks, I've marked a few things but some are only partially resolved.

> Are there tests available for the fallbacks of the Riemannian Gradient or do we leave those untested?

I'd vote for merging this PR before any other new work since it touches quite a few files, and any remaining points could be put in an issue. Actually, this PR may be worth splitting into a few PRs with a smaller scope instead of solving here every issue you've listed in the first comment.

kellertuer:
Ok, sure, we can just mark them atop that we will do them later and finish this PR. I just came across making that list since I had rewritten our tests a little bit and an hour left to go through uncovered lines.

kellertuer:
I think I just did 4 more lines. Can we mark the ones we move in the initial comment (maybe with an X after the box or an -> new PR at the end?) and check the box upfront?

src/statistics.jl
```diff
@@ -857,7 +879,6 @@ function StatsBase.mean_and_var(
             return y, v
         end
     end
-    return y, v
```
mateuszbaran, line 860: Is this function still type stable after removing this line? Compiler may think it could return `nothing` after removing this line.

kellertuer: But that line is really never reached, so how can we resolve this? Look at the injectivty Radius for y upfront, if that does not trigger the return after that, one of the ifs _in_ the loop will trigger a return.

mateuszbaran:
> I think I just did 4 more lines. Can we mark the ones we move in the initial comment (maybe with an X after the box or an -> new PR at the end?) and check the box upfront?

Maybe let's merge this PR and copy all points that are not checked as done?

kellertuer:
Ok. Would you mind copying those and whereto? An Issue or multiple issues?

src/statistics.jl
mateuszbaran: I think the last line would be reached if `method.radius` in `NaN`. There may be other cases for custom floating point types. After the loop we need to either return `y, v` or throw an error.

mateuszbaran:
Sure, I'll open a new issue.

src/statistics.jl
kellertuer: Ok, should we catch that upfront? Because we don't have a chance to test this, have we? So this line that you currently propose at the end will never be covered by a test then?

mateuszbaran: We can't really catch that upfront, and we could test the last line by calling this method with geodesic interpolation within radius `NaN`, which is easy enough. I'm just not sure if this should result in an error or just return `y, v`.

kellertuer: I am not sure which variant is best, I just wanted to get a line of test covered or remove it if it is never reached, I did not expect that it gets so difficult.

mateuszbaran: So maybe let's just leave that line there for now and add this issue to #321 ?

kellertuer: ok.

mateuszbaran:
OK, so can we merge this now?

kellertuer:
In my opinion, yes.

PR #320, opened 2021-01-06
Essential manifold

kellertuer:
Adds the Essential manifold. This was mainly written by @r-dornig, thanks!, I just finished a little bit of testing.


kellertuer:
We are missing one line, since the new version of Formatter prefers to add a `return nothing` line after a `return` line somehow.

src/manifolds/EssentialManifold.jl
```diff
@@ -0,0 +1,516 @@
...
+    return Y
+end
+
+"""
+    representation_size(M::EssentialManifold)
+
+Return the array dimensions required to represent an element on the
+[`EssentialManifold`](@ref) `M`, i.e. the vector of all array dimensions.
+"""
+@generated representation_size(::EssentialManifold) = (3, 3, 2)
```
mateuszbaran, line 457: That would be correct for the `ArrayPowerRepresentation` but `NestedPowerRepresentation` doesn't have `representation_size`.

src/manifolds/EssentialManifold.jl
```diff
@@ -0,0 +1,516 @@
...
+\text{dist}([p],[q]) = \| \text{log}_{[p]} [q] \| = \| \log_p (S_z(t_{\text{opt}})q) \|,
+````
+where $S_z ∈ H_z = \{(R_z(θ),R_z(θ))\colon θ \in [-π,π) \}$ in which $R_z(θ)$ is the rotation around the z axis with angle
+$θ$ and $t_{\text{opt}}$ is the minimizer of the cost function
+````math
+f(t) = f_1 + f_2, \quad f_i = \frac{1}{2} θ^2_i(t), \quad θ_i(t)=d(R_{p_i},R_z(t)R_{b_i}) \text{ for } i=1,2,
+````
+where $d(⋅,⋅)$ is the distance function in $SO(3)$[^TronDaniilidis2017].
+"""
+distance(M::EssentialManifold, p, q) = norm(M, p, log(M, p, q))
```
mateuszbaran, line 129: That's just the default implementation. By the way, does `EssentialManifold` assume that the horizontal component of all tangent vectors is zero?

src/manifolds/EssentialManifold.jl
kellertuer: Ah, thanks for noticing, I thought this was necessary for some allocations. Then we should remove this.

src/manifolds/EssentialManifold.jl
kellertuer: Then this could be removed, too, I'll check that the tests still work then; 

No the horizontal component is not necessary zero, and that is also never assumed.

kellertuer: Ah, this is actually more complicated, since with the power manifold, the default ist _first_ to go down to the elements and call their logs, so to avoid that and call the correct log here, we had to “reset” it from the Power default to the Manifold default.

mateuszbaran: > No the horizontal component is not necessary zero, and that is also never assumed.

Right, it always confuses me when operations on a quotient manifold are implemented in a way where `inner` from the embedding just works.

kellertuer: Sure, that would require to always store the points in a way that the horizontal component is zero; though having both projections (horizontal and vertical) would be nice, one should not assume aways having such a form.

src/manifolds/EssentialManifold.jl
```diff
@@ -0,0 +1,508 @@
...
+    #compute rotations
+    q211 = q[1] * p[1]'
+    q212 = q[2] * p[2]'
+
+    #compute values from Proposition 9
+    t_break1, c1, m1, Φ1 = dist_min_angle_pair_discontinuity_distance(q211)
+    t_break2, c2, m2, Φ2 = dist_min_angle_pair_discontinuity_distance(q212)
+
+    #check if cost is constant
+    tolMZero = 1e-15
```
mateuszbaran, line 246: Could we calculate tolerances in this function from `eps(eltype(q))`? 1e-15 is on the order of `eps(eltype(q))` and 1e-8 is about `sqrt(eps(eltype(q)))`.

mateuszbaran: Similarly in `dist_min_angle_pair_df_newton`.

kellertuer: Sure, that's a good idea.

src/manifolds/EssentialManifold.jl
```diff
@@ -0,0 +1,508 @@
...
+        p;
+        kwargs...,
+    )
+end
+
+"""
+    check_tangent_vector(M::EssentialManifold, p, X; check_base_point = true, kwargs... )
+
+Check whether `X` is a tangent vector to manifold point `p` on the [`EssentialManifold`](@ref) `M`,
+i.e. `X` has to be a Tuple of two `3`-by-`3` skew-symmetric matrices. The optional parameter `check_base_point`
```
mateuszbaran, line 85: I think it has to be a two-element array, not a tuple. Similarly in docstring for `check_manifold_point`.

kellertuer: Yes, sure, since it's nested it's an array.

mateuszbaran:
Performance here needs some attention but I don't want to spend time on it now, and I don't want to stall this PR either. We can merge it once these two comments are resolved and I'll open an issue about performance.

kellertuer:
To be honest, I was glad we could reproduce the results from the Matlab implementation and I am not yet 100% happy with how the log is designed, since, sure, it could see some improvements for sure, but it's not so easy to match this newtons iteration to even the formulae from the paper.

src/manifolds/EssentialManifold.jl
```diff
@@ -0,0 +1,509 @@
...
+````
+where $R_z(θ)$ is the rotation around the z axis with angle $θ$. Points in $H_z$ are denoted by
+$S_z$. Then, the logarithm is defined
+as
+````math
+\log_p (S_z(t_{\text{opt}})q) = [\text{Log}(R_{p_i}^T R_z(t_{\text{opt}})R_{b_i})]_{i=1,2},
+````
+where $\text{Log}$ is the [`logarithm`](@ref log(::Rotations, ::Any...)) on $SO(3)$. For more
+details see [^TronDaniilidis2017].
+"""
```
mateuszbaran, line 184: This docstring should be attached to `log`, not `log!`.

kellertuer: Ah, yes, thanks for noticing.

mateuszbaran:
OK, so feel free to merge this :slightly_smiling_face: .

PR #322, opened 2021-01-08
Compute the Median with Weiszfeld algorithm

kellertuer:
This small PR introduces another stable method for the median; the Weiszfeld algorithm (I saw it was easy to do). Still missing are
- [x] check rendering of the docs.
- [x] Test case (I was looking for the default test to CyclicProximalPoint to just extend that maybe).

Yes, this is looking a little larger due to the JuliaFormatter change, sorry for that.


kellertuer:
I haven't added tests yet so I am not yet sure, why we have 4 failing tests here, there seems to be something that changed type from Float to Int in either Std or var?!

src/statistics.jl
```diff
@@ -657,6 +721,49 @@ function Statistics.median!(
...
+    copyto!(q, p0)
+    yold = allocate_result(M, median, q)
+    ytmp = copy(yold)
+    d = zeros(n)
+    v = zero_tangent_vector(M, q)
+    wv = convert(AbstractVector, w) ./ sum(w)
+    for i in 1:stop_iter
+        d .= [distance(M, q, xi) for xi in x] # compute distances
+        # compute new weights / exclude points xi=q
+        d .= [di > 0 ? wi / di : 0 for (di, wi) in zip(d, w)]
```
mateuszbaran, line 754:
```suggestion
        d .= [di > 0 ? wi / di : zero(typeof(wi / di)) for (di, wi) in zip(d, w)]
```
Just to make it type stable.

mateuszbaran (approved):

PR #323, opened 2021-01-11
CompatHelper: bump compat for "FiniteDifferences" to "0.12"

sethaxen:
This pull request changes the compat entry for the `FiniteDifferences` package from `0.9, 0.10, 0.11` to `0.9, 0.10, 0.11, 0.12`.


kellertuer:
I am not that familiar with `FiniteDifferences`but I hope this https://github.com/JuliaManifolds/Manifolds.jl/pull/323/checks?check_run_id=1678527660#step:5:724 is not that complicated to fix @mateuszbaran ?
this is again one of the Pos that would pop up every day since it is only in our extra reps and the CompatHeper still seems to not check whether the extra dips updates where already opened as PRs.

mateuszbaran:
I think it's just a problem with that one test. Just in case I've restricted compat for FiniteDifferences to not have to deal with two variants of that test (as far as I can tell there is no way to check order of a method that works both pre-0.12 and in 0.12).

kellertuer (approved):

Issue #261, opened 2020-11-19
Cotangent vectors


mateuszbaran:
After recently reading a bit about use of cotangent vectors in AD (for example here: https://gist.github.com/Keno/4a6507b75288b1fe671e9d1cc683014f ) I think we handle them wrong. Now I think there are only two valid ways to represent them:

1. a function/functor/something callable that accepts tangent vectors and returns a number,
2. coordinates in a basis.

At the moment our `sharp` and `flat` functions do something different (especially for `MetricManifold`) and it makes me quite uneasy.

kellertuer:
Well, some call the functions that we call `get_vector`(flat) and `get_coordinates` (sharp) that way, I think, so it agrees with your second idea in that sense.

I think until now we just don't represent them, do we?

mateuszbaran:
> Well, some call the functions that we call `get_vector`(flat) and `get_coordinates` (sharp) that way,

I don't quite understand what you mean by that. `get_vector` and `flat` don't seem particularly related?

> I think until now we just don't represent them, do we?

We mostly don't but still we have things like this: https://github.com/JuliaManifolds/Manifolds.jl/blob/master/src/manifolds/MetricManifold.jl#L614-L618 that don't integrate well with what the rest of `Manifolds.jl` does with tangent vectors.

kellertuer:
but that sharp is you 1., just sorting the linear functional using a vector representation.

mateuszbaran:
No, it's very much not. It's actually sort of my 2nd point. Sort of because I don't see a point in talking about local metric without specifying a basis.

My first point is something we don't have at the moment. In my first point `flat` would be just taking given tangent vector as a Riesz representer of the cotangent (and wrapping in a struct that indicates this).

kellertuer:
Ah, right! Then I misread your first variant. I actually like both, in our recent paper we used variant 1, but in the implementation (I am currently finishing), I was lazy and removed all sharp and flats (also since that was possible) and just used tangent vectors.

But it would be very neat to have something like this.

kellertuer:
We had a discussion today about variant 1 (after I did not look at this for quite some time) – so good that we have this still open.
So I would still think this is nice to have if realised in a nice and consistent way.

PR #325, opened 2021-01-13
Atlases and charts

mateuszbaran:
This is supposed to provide minimal support for atlases and charts, mainly aimed at solving out current problems with metric manifold being inconsistent about it. This should solve #6 .

One point where I'm not sure what to do is `flat` and `sharp`. We don't have any specific representation of cotangent vectors. They can be represented by either functions, functors or coordinates in a basis of the cotangent space, and https://github.com/JuliaManifolds/Manifolds.jl/blob/77aecb49a57a7c7da9f9d6558412059a42aa0d69/src/manifolds/MetricManifold.jl#L336-L345 works only in one very specific case. The question here is: how do we know which basis? My first idea is to store the basis in `FVector` but I'd like to hear your opinion.


src/atlases.jl
```diff
@@ -0,0 +1,123 @@
...
+    TInvRetr<:AbstractInverseRetractionMethod,
+    TRetr<:AbstractRetractionMethod,
+    TBasis<:AbstractBasis,
+} <: AbstractAtlas
+    invretr::TInvRetr
+    retr::TRetr
+    basis::TBasis
+end
+
+function RetractionAtlas(
```
kellertuer, line 29: We could provide and document a default, that uses `exp/log`?

kellertuer:
Concerning `flat!`, I think we could do something similar to what we do with `MPoint` and `TVector`, i.e. if you just provide an array we treat it as an default cotangent representation, which is an array (as we do with vector) and I would introduce a type for the linear function representation (maybe similar to the `Dual` type?).

src/atlases.jl
mateuszbaran: Yes, that's a good idea.

mateuszbaran:
Well, I'm currently leaning towards making functions/functors the default cotangent vector representation (see https://github.com/JuliaManifolds/Manifolds.jl/issues/261 ), and arrays of coefficients as an alternative. Then flat and sharp have many possible variants, depending on received and target representation of tangent and cotangent vectors. Some problems would be solved by storing the basis in `FVector`, i.e. we would have the same number of arguments in all variants.

mateuszbaran:
`flat` and `sharp` work more or less as intended now on `Circle`, what do you think about this approach?

src/atlases.jl
```diff
@@ -0,0 +1,157 @@
...
+
+"""
+    select_chart(M::Manifold, A::AbstractAtlas, p)
+
+Select a chart from atlas `A` for manifold `M` that is suitable for representing
+neighborhood of point `p`.
+"""
+select_chart(::Manifold, ::AbstractAtlas, ::Any)
+
+select_chart(::Manifold, ::RetractionAtlas, p) = p
```
kellertuer, line 83: You here return the index of the chart in the `::RetractionAtlas`, right? Would it maybe be reasonable to actually return the char, i.e. a function?

kellertuer:
> `flat` and `sharp` work more or less as intended now on `Circle`, what do you think about this approach?

Looks nice to me; just a question that came to my mind: Is there or could we introduce a way to automatically convert array<->Vector? In all other parts of the package we do not constrain tangent vectors, but here we do. Maybe that's necessary, I am not yet 100% sure, but it would be nice if those could work easily, i.e. with auto conversion.

src/atlases.jl
mateuszbaran: Returning the index is more general, for example one may want to use it to do `get_point_coordinates`, `get_point` or in `transition_map`. We can always add a convenience function for that later.

mateuszbaran:
There are a few separate issues.

> Is there or could we introduce a way to automatically convert array<->Vector?

What do you mean by array <-> vector conversion? Conversion between the default representation of a tangent vector and a vector of its coefficients in a basis? Any such conversion requires manifold and a point which are always stored separately.

> In all other parts of the package we do not constrain tangent vectors, but here we do.

I'd say that all other parts of the package don't really support `FVector`s, and I think `FVector`s are primarily for dealing with vector representation in a basis. Currently you can give `nothing` as a basis and then it is supposed to mean the default representation but maybe that's more confusing than helpful. I don't see here any constraining other than special-casing `FVector`s and changing default cotangent vector representation.

src/cotangent_space.jl
```diff
@@ -0,0 +1,65 @@
...
+function flat!(M::Manifold, ξ::CoTFVector{<:Any,Nothing}, p, X::TFVector{<:Any,Nothing})
+    return flat!(M, ξ.data, p, X.data)
+end
+function flat!(
+    M::Manifold,
+    ξ::CoTFVector{<:Any,<:AbstractBasis},
+    p,
+    X::TFVector{<:Any,<:AbstractBasis},
+)
+    # TODO: check handle basis change here? Right now it assumes dual basis.
```
kellertuer, line 42: You can access both bases in ξ and X and perform the change of basis if they are different. I think that is necessary here.

kellertuer:
Currently we often have `X` just as an array (say a vector for the sphere), but for `flatland `sharp` we need them specifically as `TFVector`s. Ah I see for flat you do also accept arrays. My main goal here is, that if you stumble upon using sharp and flat, it would be nice if for start you can just plug in what you know already, i.e. your normal `X` (for flat that is). I am not sure whether something that easy is possible for sharp, when or cotangent vectors are functions by default and not vectors.

But due to the duality pairing I sometimes think the same array representation (i.e. not yet a coordinate vector in a basis) as for tangents should be possible for cotangents. And it would be nice to have an array typed function for them as well? It maps to some default of course, as does the flat with an array `X` (maybe a default that even returns an array?).

Again this is just to help users enter our regime of the sharp/flat world, which I really like to have, since I think there is no package yet doing that rigorously.

src/atlases.jl
kellertuer: Then I would maybe change the name to `get_chart_index`? Just semantically I would `get_chart` (sure a convenience function, that could then be added after) to actually return a chart.

mateuszbaran: Sure, I'll change the name.

src/cotangent_space.jl
mateuszbaran: Right. I think I'll just by default use `get_vector` and `get_coordinates` here.

mateuszbaran:
> Currently we often have `X` just as an array (say a vector for the sphere), but for `flatland `sharp`we need them specifically as`TFVector`s. Ah I see for flat you do also accept arrays. My main goal here is, that if you stumble upon using sharp and flat, it would be nice if for start you can just plug in what you know already, i.e. your normal `X` (for flat that is). I am not sure whether something that easy is possible for sharp, when or cotangent vectors are functions by default and not vectors.

Yes, you can just put an array in `flat` and by default you will get a `RieszRepresenterCotangentVector`. The other way, i.e. being able to give any function to `sharp` should be possible but it's not as easy, so for now only `RieszRepresenterCotangentVector` (or `CoTFVector`) are accepted. I think `sharp`ing an arbitrary function is beyond the scope of this PR.

> But due to the duality pairing I sometimes think the same array representation (i.e. not yet a coordinate vector in a basis) as for tangents should be possible for cotangents. And it would be nice to have an array typed function for them as well? It maps to some default of course, as does the flat with an array `X` (maybe a default that even returns an array?).

Could you give an example of what for example `ξ[i]` would be for the array representation of cotagnents, for example on S^2?

kellertuer:
Thanks, that's already great. For sharp I was thinking of the following: If you have X in some TpM on S2, then you can take (in the embedding, R3), `x -> dot(X,x)` as a Dual, but this can also be uniquely represented by `X` (using the default ONB). so setting `ξ=X` would be reasonable and an array representation of the cotangent vector. Sure, such a default has to be documented.

So if that such a default is not too dangerous, I think using that (similar to the flat case) would be nice for users stating here.

mateuszbaran:
For now I've removed `size` and indexing into `FVector`. I've also removed the `nothing` basis thing for `FVector` so that `FVector` is only for coefficients in a basis. I think the interface is cleaner now.

mateuszbaran:
My current thinking is that `inner` on a metric manifold should use `local_metric` only if the tangent vectors are given as `FVector`s in a basis defined by a chart. Does that sound reasonable?

BTW, I'm not quite sure how `BaseManifold` from tests was supposed to work regarding metrics...

kellertuer:
I think base manifold should just check the possibilities to implement a second metric and have fallbacks working correctly to a default metric?

kellertuer:
to use the local metric as you propose sounds reasonable, yes.

mateuszbaran:
Locally, everything except the tests marked as broken works now. I'm not sure how to fix them in a minimal way. The old way it's done is kinda broken anyway.

kellertuer:
Interesting, I haven't had the time to check but I thought they were just meant to check for proper dispatch.

mateuszbaran:
The broken ones also check for the presence of
```julia
function inner__intransparent(M::MMT, p, X, Y) where {MMT<:MetricManifold}
    return dot(X, local_metric(M, p) * Y)
end
```
which is very wrong. This can only work for `X`, `Y` being `FVector`s in a basis defined by a chart (until `local_metric` is added to general bases), but that's not what the tests expect. And currently there is no API to check that both `FVector`s are defined in the same basis.

mateuszbaran:
I've figured out a way to fix these broken tests. What do you think? I don't quite like that there are two ways to call `local_metric` now, we probably only want these metric functions for bases (since a chart can uniquely determine a basis but not the other way around). I haven't completely thought it through though yet.

kellertuer:
At first glance it looks like a very good approach.

mateuszbaran:
I think this is more or less feature-complete for this PR. `StereographicAtlas` illustrates how an atlas could be added and used. What do you think? There is of course a lot that could be added but it already fixes some issues with metric manifolds. By the way, I've added a second chart to `StereographicAtlas`.

If you think this is enough code, I'll just add some documentation and fix tests.

kellertuer:
I think the features are fine and you can surely start documenting.

Please check my comment on the stereographic example – and having that ,maybe also as a tutorial would be great, too.

mateuszbaran:
I have one minor design problem. What do we do about `zero_vector` and `zero_tangent_vector`? The first one is kind of necessary but we almost exclusively use `zero_tangent_vector` which won't work for other vector spaces. I think it would make the most sense to just merge `zero_tangent_vector` into `zero_vector`, though it would be a breaking change. What do you think?

kellertuer:
Since from the signature (`zero_vector(M,p)`) it would be (to the known user) clear that this is a tangent vector, I'd be fine with it, though at first glance the `_tangent` might help new users more.

mateuszbaran:
We can also keep both functions, though it would be then less clear which one should be overloaded for new manifolds.

kellertuer:
No, keeping both is irritating for all users then, I think with a careful documentation, we can switch to `zero_vector`, though that is really a large change affecting a lot of things. I use that often for initialisation in Manopt for example.

mateuszbaran:
OK, so I'll switch to `zero_vector` as a part of this PR. I'm planning to make `zero_vector(::Manifold, p)` return tangent vector and keep `zero_vector(::VectorBundleFibers, p)` for all other vector spaces.

mateuszbaran:
I just replaced `zero_tangent_vector` with `zero_vector` and all tests still pass. What would you change in documentation?

kellertuer:
> I just replaced `zero_tangent_vector` with `zero_vector` and all tests still pass. What would you change in documentation?

Ah that would actually be over in ManifoldBase, the `zero_vector!` could be extended similarly long as the `zero_vector` general one and I will check whether one can emphasise the tangent part more in the text.

mateuszbaran:
I've changed that in ManifoldsBase.jl as well: https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/59 :slightly_smiling_face: .

src/atlases.jl
```diff
@@ -0,0 +1,242 @@
...
+
+Select a chart from an [`AbstractAtlas`](@ref) `A` for manifold `M` that is suitable for
+representing neighborhood of point `p`.
+"""
+get_chart_index(::Manifold, ::AbstractAtlas, ::Any)
+
+get_chart_index(::Manifold, ::RetractionAtlas, p) = p
+
+"""
+    transition_map(M::Manifold, A_from::AbstractAtlas, i_from, A_to::AbstractAtlas, i_to, x)
```
kellertuer, line 109: One could use `i` and `j` for the indices and maybe `A1` and `A2` or `A` and `B` for the atlases (though `B` might be confused with a basis maybe)?

src/manifolds/Euclidean.jl
```diff
@@ -280,13 +274,21 @@ inner(::Euclidean, ::Any...)
...
 @inline inner(::MetricManifold{𝔽,<:Manifold,EuclideanMetric}, p, X, Y) where {𝔽} = dot(X, Y)
 
-function inverse_local_metric(M::MetricManifold{𝔽,<:Manifold,EuclideanMetric}, p) where {𝔽}
-    return local_metric(M, p)
+function inverse_local_metric(
+    M::MetricManifold{𝔽,<:Manifold,EuclideanMetric},
+    B::InducedBasis{𝔽,TangentSpaceType,<:RetractionAtlas},
+    p,
+) where {𝔽}
+    return local_metric(M, B, p)
```
kellertuer, line 282: I know this is an old code line, but I am confused that the inverse of the local metric actually calls the local metric?

src/manifolds/Sphere.jl
```diff
@@ -461,3 +471,103 @@ function vector_transport_to!(::AbstractSphere, Y, p, X, q, ::ParallelTransport)
...
+    return Y
+end
+
+function local_metric(
+    M::Sphere{n,ℝ},
+    B::InducedBasis{ℝ,TangentSpaceType,StereographicAtlas,Symbol},
+    p,
+) where {n}
+    x = get_point_coordinates(M, B.A, B.i, p)
+    return (4 / (1 + dot(x, x))^2) * I
```
kellertuer, line 572: Does `I` here “just work”? Can Julia infer its size automatically?

test/groups/metric.jl
```diff
@@ -102,12 +102,13 @@ invariant_metric_dispatch(::TestDefaultInvariantMetricManifold, ::RightAction) =
         Y = hat(SO3, Identity(SO3, e), [3.0, 4.0, 1.0])
 
         G = MetricManifold(SO3, lmetric)
-        @test inner(G, p, X, Y) ≈ dot(X, Diagonal([1.0, 2.0, 3.0]) * Y)
-        @test norm(G, p, X) ≈ sqrt(inner(G, p, X, X))
+        @test_broken inner(G, p, X, Y) ≈ dot(X, Diagonal([1.0, 2.0, 3.0]) * Y)
```
kellertuer, line 105: We should check why these are broken and hopefully fix them

kellertuer (commented):
Overall this looks very nice.

A few remarks
* I would love to also have a tutorial how to uses atlases. Maybe I try to come up with one on the weekend.
* Similarly also for the basis, since this PR changes the type, it would be a good opportunity to illustrate their use
* Should we introduce `♭` for `flat` and `♯` for `sharp` and similarly `♭!` and `♯!` ? the first two could be postfix operators?
* I like that Christoffel symbols got further, also for those I would love to see a tutorial on how to use them.

mateuszbaran:
> Overall this looks very nice.

Thanks :slightly_smiling_face: .

> * I would love to also have a tutorial how to uses atlases. Maybe I try to come up with one on the weekend.

That would be great. Are you thinking about something standalone or as a part of metric manifold stuff?

> * Similarly also for the basis, since this PR changes the type, it would be a good opportunity to illustrate their use

That's right. I didn't want to push too far in that direction though since for spaces other than the tangent one it's sort of an unknown territory. Still, there is an example of a basis of the contangent space. It would be nice to have an idea how to integrate this into the tensor ecosystem: https://discourse.julialang.org/t/ann-tensoralgebra-jl-taking-covariance-seriously/43587/49?u=mateuszbaran .

> * Should we introduce `♭` for `flat` and `♯` for `sharp` and similarly `♭!` and `♯!` ? the first two could be postfix operators?

I prefer to discuss it separately since it's not entirely obvious what object would they be called on.

> * I like that Christoffel symbols got further, also for those I would love to see a tutorial on how to use them.

At this point there is only one use case for them, `solve_exp_ode`. What other uses would you like to demonstrate?

src/atlases.jl
```diff
@@ -0,0 +1,242 @@
...
+end
+
+function get_point_coordinates(M::Manifold, A::RetractionAtlas, i, p)
+    return get_coordinates(M, i, inverse_retract(M, i, p, A.invretr), A.basis)
+end
+
+"""
+    get_point(M::Manifold, A::AbstractAtlas, i, x)
+
+Calculate point at coordinates `x` on manifold `M` in chart from an [`AbstractAtlas`](@ref)
```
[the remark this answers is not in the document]
mateuszbaran: Yes, I'm currently using `x` for coordinates in a chart. I can definitely use some other notation but I don't know which symbol would be most clear. `x` can indeed by confused with `X`.

src/atlases.jl
```diff
@@ -0,0 +1,242 @@
...
+end
+
+function get_point!(M::Manifold, p, A::RetractionAtlas, i, x)
+    return retract!(M, p, i, get_vector(M, i, x, A.basis), A.retr)
+end
+
+"""
+    get_chart_index(M::Manifold, A::AbstractAtlas, p)
+
+Select a chart from an [`AbstractAtlas`](@ref) `A` for manifold `M` that is suitable for
```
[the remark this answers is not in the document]
mateuszbaran: Yes, that's a good point.

src/atlases.jl
mateuszbaran: I think in this case more descriptive argument names are useful to be able to easily tell which chart is which (is it from 1 to 2 or from 2 to 1?)

src/manifolds/Euclidean.jl
mateuszbaran: Well, both should just be unit matrices in this basis I think?

src/manifolds/Sphere.jl
mateuszbaran: "Just working" may depend on the particular operation. If Julia can infer the size from other arguments of an operation, it usually does so:
```julia
julia> a = 2*I
UniformScaling{Int64}
2*I

julia> a * [1 2; 3 4]
2×2 Array{Int64,2}:
 2  4
 6  8
```

kellertuer: Ah ok, and since we think that this would only be used in these cases it should be fine – ok.

src/atlases.jl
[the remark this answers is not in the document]
kellertuer: We could use something like `a`. and `b` for them?

src/atlases.jl
kellertuer: ok.

src/manifolds/Euclidean.jl
kellertuer: I am not 100% sure.

src/atlases.jl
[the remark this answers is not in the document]
mateuszbaran: OK, I'll do that.

test/groups/metric.jl
mateuszbaran: These are... interesting failures. I've fixed two so far. The "exp τ-invariant" failures are caused by ODE method for `exp` defaulting to the induced basis. I don't know how to fix this. Perhaps there should be an automatic definition for that based on the `local_metric` that's already defined? Or maybe `local_metric` should just be defined for `InducedBasis` in the test file?

These "inner/norm" tests are broken because there is currently no `translate_diff` for the right action on `SpecialOrthogonal` with `TFVector` argument. I'm not sure whether this ever worked correctly?

mateuszbaran: @sethaxen could you take a look at this issue?

src/atlases.jl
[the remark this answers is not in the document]
mateuszbaran: Done in https://github.com/JuliaManifolds/Manifolds.jl/pull/325/commits/d134eef919ecff5cfa3ae145a75016a90ac210d5

test/groups/metric.jl
sethaxen: IIRC, the ODE method only ever worked for manifolds whose default representation was in a global chart. So it maybe could work for a manifold whose default representation is in embedded coordinates, but it should not be expected to. Most metric-related methods probably only work within charts as well, and `local_metric` likewise requires being in a chart or should take a basis. I haven't had the time to review this PR, but is that consistent with the failures you're seeing?

kellertuer:
> > * I would love to also have a tutorial how to uses atlases. Maybe I try to come up with one on the weekend.
> 
> That would be great. Are you thinking about something standalone or as a part of metric manifold stuff?

Either would be fine. Also a tutorial just on atlasses and charts would be nice.

> > * Should we introduce `♭` for `flat` and `♯` for `sharp` and similarly `♭!` and `♯!` ? the first two could be postfix operators?
> 
> I prefer to discuss it separately since it's not entirely obvious what object would they be called on.

Sure, no problem

> 
> > * I like that Christoffel symbols got further, also for those I would love to see a tutorial on how to use them.
> 
> At this point there is only one use case for them, `solve_exp_ode`. What other uses would you like to demonstrate?

Then let's wait with this one, sure.

test/groups/metric.jl
mateuszbaran: I've changed all metric-related functions to work with respect to bases (not charts because they are not strictly needed to express coordinates of these metric functions). In this PR I try to be consistent about usage of coordinates and these are just two cases that are tricky to fix.

kellertuer (approved):

mateuszbaran:
I'm generally satisfied with this code. The three remaining issues are:
1) Making CI pass (but that's annoying without tagging a release of ManifoldsBase.jl with https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/59 merged).
2) A tutorial for atlases would be nice but I'm not sure what to put there. Basic usage is described in docs.
3) There are four broken tests that I don't know how to fix.

I'd be fine with solving 2 and 3 in new PRs though so I think we could merge https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/59 ?

mateuszbaran:
One more TODO (for another PR): make `induced_basis` work with arbitrary atlases.

mateuszbaran:
TODO (maybe?) : complex atlases. Do you think they could be useful? I mean atlases that map to ℂ^n instead of ℝ^2n.

kellertuer:
I think that would be useful. But we would restrict that to the domain being real-valued and only the range would be complex? Otherwise it might get too messy (cf. our bases for tangent spaces, where it is already a little bit of technical details).

sethaxen:
I'll try to review this today.

mateuszbaran:
I was thinking about something like the atlas described here: http://www.mat.unimi.it/users/geemen/Ferrari_complexmanifolds.pdf (definitions 2.1 and 2.2) but maybe it's... too complex? :smile: That may make sense though, with complex bases of tangent spaces corresponding to complex atlases (i.e. induced basis of the tangent space from a complex atlas would be a complex basis of the tangent space). Similarly the retraction atlas from a complex basis would be a complex atlas. Would there be a problem with that?

kellertuer:
I think that sounds great, my hope is that using only the reals as domain, we would avoid this Complex tangents with real coefficitents versus real basis with complex coefficients that we have in basis (which is a little technical concerning the dimensions). But for sure complex domain would mathematically also be fine, just complicated here maybe?

The rest – as said – sounds reasonable and would be nice; I don't see a problem with that.

src/atlases.jl
```diff
@@ -0,0 +1,251 @@
...
+Calculate coordinates of point `p` on manifold `M` in chart from an [`AbstractAtlas`](@ref)
+`A` at index `i`.
+
+# See also
+
+[`get_point`](@ref), [`get_chart_index`](@ref)
+"""
+get_point_coordinates(::Manifold, ::AbstractAtlas, ::Any, ::Any)
+
+function get_point_coordinates(M::Manifold, A::AbstractAtlas, i, p)
```
sethaxen, line 56: Should the decorator machinery be used here at all?

src/atlases.jl
```diff
@@ -0,0 +1,251 @@
...
+
+function get_point!(M::Manifold, p, A::RetractionAtlas, i, a)
+    return retract!(M, p, i, get_vector(M, i, a, A.basis), A.retr)
+end
+
+"""
+    get_chart_index(M::Manifold, A::AbstractAtlas, p)
+
+Select a chart from an [`AbstractAtlas`](@ref) `A` for manifold `M` that is suitable for
+representing neighborhood of point `p`. This selection should be deterministic, although
```
sethaxen, line 110:
```suggestion
representing the neighborhood of point `p`. This selection should be deterministic, although
```

src/atlases.jl
```diff
@@ -0,0 +1,251 @@
...
+
+"""
+    local_metric(M::Manifold, B::InducedBasis, p)
+
+Compute the local metric tensor for vectors expressed in terms of coordinates
+in basis `B` on manifold `M`. The point `p` is not checked.
+"""
+local_metric(::Manifold, ::InducedBasis, ::Any)
+
+function allocate_result(M::PowerManifoldNested, f::typeof(get_point), x)
```
sethaxen, line 240: Should this be defined with the rest of the `PowerManifold` code?

src/manifolds/Euclidean.jl
sethaxen: It should be the identity matrix when the basis is orthonormal. If the basis is only orthogonal, then the local metric is a diagonal matrix, not the identity matrix, and it is not equal to its inverse.

src/manifolds/MetricManifold.jl
```diff
@@ -58,13 +59,14 @@ where $g_{ij,k}=\frac{∂}{∂ p^k} g_{ij}$ is the coordinate
 derivative of the local representation of the metric tensor. The dimensions of
 the resulting multi-dimensional array are ordered $(i,j,k)$.
 """
-christoffel_symbols_first(::MetricManifold, ::Any)
+christoffel_symbols_first(::MetricManifold, B::AbstractBasis, ::Any)
```
sethaxen, line 62: Our basis-related functions all put the basis argument as last. Should we do that here as well for consistency? A bonus would be that it can then have a default value.

test/sphere.jl
```diff
@@ -166,4 +169,35 @@ include("utils.jl")
...
+            p2 = randn(3)
+            p3 = randn(3)
+            X2 = log(M, p, p2)
+            X3 = log(M, p, p3)
+            B = induced_basis(M, A, i, TangentSpace)
+
+            X2B = get_coordinates(M, p, X2, B)
+            X3B = get_coordinates(M, p, X3, B)
+
+            @test inner(M, p, X2, X3) ≈ X2B' * local_metric(M, B, p) * X3B
```
sethaxen, line 195:
```suggestion
            @test inner(M, p, X2, X3) ≈ dot(X2B, local_metric(M, B, p) * X3B)
```

sethaxen (commented):
I had some questions and suggestions. It's a big PR, and I don't think I understand everything yet. I'll go through it again tomorrow or whenever the preview docs have built.

src/atlases.jl
kellertuer: Nice observation – indeed, this function could be transparent for metrics as well as groups, since the atlas could be used for the manifold without having any of this in mind.

docs/src/features/atlases.md
```diff
@@ -0,0 +1,37 @@
+# [Atlases and charts](@id atlases_and_charts)
+
+Atlases on an $n$-dimensional manifold $\mathcal M$ are collections of charts $\{(U_i, \varphi_i) \colon i \in I\}$ such that $U_i \subseteq \mathcal M$ and each chart $\varphi_i$ is a function from $U_i$ to $\mathbb{R}^n$.
+They provide a basic connection between manifolds and Euclidean spaces.
+
+Most operations on manifolds in `Manifolds.jl` avoid operating in a chart through appropriate embeddings and formulas derived for particular manifolds, though atlases provide the most general way of working with manifolds.
+They are extensively used in metric-related functions on [`MetricManifold`](@ref Main.Manifolds.MetricManifold)s.
```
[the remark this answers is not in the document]
mateuszbaran: I don't quite understand how Documenter.jl resolves names when it generates documentation based on multiple packages. IIRC that didn't link correctly without `Main.Manifolds`. Anyway, it's also used in `manifold.md` and some docstrings from `ManifoldsBase.jl`.

src/atlases.jl
```diff
@@ -0,0 +1,251 @@
...
+
+[`AbstractAtlas`](@ref), [`AbstractInverseRetractionMethod`](@ref),
+[`AbstractRetractionMethod`](@ref), [`AbstractBasis`](@ref)
+"""
+struct RetractionAtlas{
+    TInvRetr<:AbstractInverseRetractionMethod,
+    TRetr<:AbstractRetractionMethod,
+    TBasis<:AbstractBasis{<:Any,TangentSpaceType},
+} <: AbstractAtlas
+    invretr::TInvRetr
```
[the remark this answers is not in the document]
mateuszbaran: Sure, I'll swap them.

src/atlases.jl
```diff
@@ -0,0 +1,251 @@
+
+"""
+    abstract type AbstractAtlas end
+
+An abstract class for atlases.
+"""
+abstract type AbstractAtlas end
+
+"""
+    struct RetractionAtlas{
```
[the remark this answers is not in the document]
mateuszbaran: OK, I'll add some documentation.

src/atlases.jl
mateuszbaran: That's an interesting point. `RetractionAtlas` does actually depend on the metric though so maybe it shouldn't always be transparent for metric decorators?

src/atlases.jl
```diff
@@ -0,0 +1,251 @@
...
+
+function get_point_coordinates(M::Manifold, A::AbstractAtlas, i, p)
+    a = allocate_result(M, get_point_coordinates, p)
+    get_point_coordinates!(M, a, A, i, p)
+    return a
+end
+
+function allocate_result(M::Manifold, f::typeof(get_point_coordinates), p)
+    T = allocate_result_type(M, f, (p,))
+    return allocate(p, T, manifold_dimension(M))
```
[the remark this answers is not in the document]
mateuszbaran: That's one point that was discussed recently in discussion for this PR: https://github.com/JuliaManifolds/Manifolds.jl/pull/325#issuecomment-813519462 and I think I'll stick to atlases with real coordinates for this PR. I'll make this point more clear in documentation.

src/manifolds/Euclidean.jl
mateuszbaran: Yes, right, so I think I'll just restrict these two methods for `inverse_local_metric` and `local_metric` to orthonormal bases (in the retraction atlas).

src/manifolds/MetricManifold.jl
mateuszbaran: That's a good idea, I'll change order of arguments. What would be the right default value? Just `DefaultOrthonormalBasis()`? Another route would be through `get_default_atlas`, `get_chart_index` and `induced_basis`.

mateuszbaran:
> I'll go through it again tomorrow or whenever the preview docs have built.

Preview docs are somewhat problematic, as is CI in general for this PR. This PR requires changes from https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/59 but I don't really want to merge that too soon. On the other hand running CI with unreleased branches of dependencies is quite poorly supported (IIRC at some point I managed to do it with custom Manifest.toml files but it'd be much easier to work on a released version of ManifoldsBase.jl)

src/atlases.jl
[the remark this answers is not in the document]
mateuszbaran: Actually, I'm thinking about adding field type parameter to `AbstractAtlas` in this PR just to not have to revisit all docstrings when support for complex atlases is added.
cc @kellertuer

src/atlases.jl
mateuszbaran: So I think `get_default_atlas`, `get_point_coordinates`, `get_point`, `get_chart_index`, `transition_map` and `induced_basis` should be forwarded through the group decorator?

src/manifolds/Euclidean.jl
mateuszbaran: I think for that we need an orthonormality trait for bases so that we don't have to special-case bases. `InducedBasis` really increases the problem of determining whether a given basis is orthonormal or not.

src/manifolds/MetricManifold.jl
sethaxen: > What would be the right default value? Just `DefaultOrthonormalBasis()`?

I think `DefaultBasis()`? Because there's no need for the basis to be orthonormal or even orthogonal, it should just ideally be fast, and `DefaultBasis` is meant to be the fastest. Besides, it forwards to `DefaultOrthogonalBasis` and `DefaultOrthonormalBasis` anyways.

> Another route would be through `get_default_atlas`, `get_chart_index` and `induced_basis`.

Can you elaborate on this route?

mateuszbaran: I think `DefaultBasis` is primarily supposed to be the fastest at `get_vector` and `get_coordinates`, which doesn't necessarily correspond to being the fastest at computing coefficients of higher order tensors.

Anyway, this question touches an important point, specifically if one wants to specify a metric tensor, which basis are they supposed to do it in? The examples are currently for induced bases but that can be changed. You can take a look at the "scaled Sphere metric" test case where I take the default atlas for the sphere (`get_default_atlas`), then for a given point `p` I call `get_chart_index` to get the index of the chart from the atlas that is most suitable for representing neighborhood of `p`, and then I work in the basis induced by that chart (`induced_basis`).

IIRC a basis of the tangent space is insufficient to express coefficients of things like Christoffel symbols, unless it's an induced basis in which case you also have an atlas. Right now there is one underspecified piece here, that is I do not strictly enforce the basis of what vector space has to be passed to metric functions. An earlier design here just had atlas and chart parameters instead of a basis parameter but later I found it somewhat inflexible so I generalized it. Still, the core idea is that you would usually pass something that indicates a chart to metric-related functions.

src/atlases.jl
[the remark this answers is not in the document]
mateuszbaran: I've added number system parameter to `AbstractAtlas` and some documentation.

src/atlases.jl
mateuszbaran: Hm, I tried adding some decorator annotations and it just reminded me that I still don't know how it's supposed to work :confused: .

mateuszbaran:
I realized the relation between metric-related functions and bases of vector spaces may be a little unclear so I've added a section to documentation about it.

mateuszbaran:
I've renamed `Metric` to `AbstractMetric` here.

test/euclidean.jl
```diff
@@ -229,24 +229,27 @@ using Manifolds: induced_basis
     @testset "Euclidean metric tests" begin
         M = Euclidean(2)
         p = zeros(2)
-        C1 = christoffel_symbols_first(M, p)
+        A = Manifolds.get_default_atlas(M)
```
kellertuer, line 232: This is a trivial atlas for sure, but just these lines 232-234 look _so_ nice in code now. I just wrote a similar exercise recently and I might (just might, not yet sure) try to write an atlas tutorial on a subset of the Euclidean manifold based on that.

mateuszbaran: Thanks, this finally lets us take metrics seriously :slightly_smiling_face: . For a bit more complicated atlas you may look at `StereographicAtlas` on sphere (this may make a bit nicer tutorial).

mateuszbaran:
The documenter job is weirdly unaware that ManifoldsBase.jl 0.11.0 has been released and for this reason is currently failing.

mateuszbaran:
Anyway, I'm working on resolving remaining issues.

mateuszbaran:
Hm, I just noticed that we don't have a function that converts to/from `FVector`. Should we just recommend using `get_vector` and `get_coordinates` for that?

kellertuer:
> Hm, I just noticed that we don't have a function that converts to/from `FVector`. Should we just recommend using `get_vector` and `get_coordinates` for that?

If we follow the lazy-model, i.e. only to use `FVector`s when it is necessary and use arrays otherwise, recommending `get_vector`/`get_coordinate` makes sense, yes.

mateuszbaran:
`FVector` is only going to grow in significance as it's the only way to work with more advanced metric stuff. Anyway, `get_vector`/`get_coordinates` seems like a good solution.

test/euclidean.jl
kellertuer: Sure that might be nicer since it is not just one chart and one can explain a bit more.

kellertuer:
We loose a few lines of code but we also loose a few more lines of coverage – any idea why?

mateuszbaran:
In some cases I don't really know what could cause the coverage loss -- for example I have no idea where was `local_metric` for `MinkowskiMetric` tested. Similarly I have no idea why some lines in `check_vector` aren't covered any more. There are a few places I could improve but I'd like to finally get it merged.

kellertuer:
I would like to not loose (percentage of) coverage. I will take a closer look tomorrow, but it might only be in the afternoon.

kellertuer:
I think I found the main reason when checking torus, I will adapt the tests today.

The problem is that for example

```Julia-repl
julia> M = Torus(2)
Torus(2)

julia> is_vector(M, 9.0, 0.0, true)
ERROR: DomainError with 1:
The number of elements in `p` (1) does not match the dimension of the torus (2).
```
previously entered `check_tangent_vector`, since the base point check was only done inside. Now this is all done in a central place (good!) so in order to actually enter `check_vector` we have to use a correct point (dimension and values), for example `is_vector(M, [0.0, 0.0], 0.0, true)` triggers the right error that `X` is of wrong size.

I think that might be the reason for the other missing tests, too.

kellertuer:
Furthermore for

```
local_metric(M::MetricManifold{𝔽<::AbstractManifold,EuclideanMetric}, p, B::InducedBasis{𝔽,TangentSpaceType,&lt;:RetractionAtlas})
```
and the corresponding inverse local metric you should add a test.
The `InducedBasis` is also currently not exported.



&nbsp; | 278 | &nbsp; | + | )
-- | -- | -- | -- | --

kellertuer:
Finally `induced_basis` should also be covered (both variants) with a test?

src/manifolds/PowerManifold.jl
```diff
@@ -55,50 +55,56 @@ struct PowerFVectorDistribution{
...
 function det_local_metric(
     M::MetricManifold{PowerMetric,𝔽,<:AbstractPowerManifold{𝔽}},
     p::AbstractArray,
+    B::DefaultOrthonormalBasis,
 ) where {𝔽}
     result = one(number_eltype(p))
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
-        result *= det_local_metric(M.manifold, _read(M, rep_size, p, i))
+        result *= det_local_metric(M.manifold, _read(M, rep_size, p, i), B)
```
kellertuer, line 85: When I tried to write a test for this I encountered two problems
(1) `PowerMetric` is the default metric for an `AbstractPowerManifold` so by transparency rules this seems to pass down to Metricmanifold.jl line 346
(2) I think this line has to have (if we reach it after solving (1)) `M.manifold.manifold`, since `M` is a metric manifold, then `M.manifold is the power manifold and we need its `.manifold` here, actually.

src/manifolds/ProductManifold.jl
```diff
@@ -234,8 +227,11 @@ end
 function det_local_metric(
```
kellertuer, line 227: Here the same seems to hold as for power
(1) `M.manifold is the product manifold, so it should be `M.manifold.manifolds`
(2) due to `ProductMetric` being the default metric this seems to be transparent – in the most recent commit there is a commented out line that fails for comparison (bottom of product manifold tests).

kellertuer:
I hopefully finished this nearly – see the two remarks at the code.
I have one further (hopefully really final) question.

Can we call `get_point_coordinates` also just `get_coordinates`? That would be neatly symmetric:
- `get_coordinates`/`get_vector` have the same signature `(M, p, X, B)`/`(M,p,a,B)` when acting on TpM
- `get_coordinates`/`get_point` have the same signature `(M, A, i, p)`/`(M,A,i,a)`

So similar to `project` I think `get_coordinates` should be clear from the signature, which coordinates one wants to obtain (first case: M,p and a basis so its a tangent vector, second case M,A,i indicates the chart, so its a point).

mateuszbaran:
> previously entered `check_tangent_vector`, since the base point check was only done inside. Now this is all done in a central place (good!) so in order to actually enter `check_vector` we have to use a correct point (dimension and values), for example `is_vector(M, [0.0, 0.0], 0.0, true)` triggers the right error that `X` is of wrong size.
> 
> I think that might be the reason for the other missing tests, too.

Yes, this makes sense :+1: .

> Furthermore for
> 
> ```
> local_metric(M::MetricManifold{𝔽<::AbstractManifold,EuclideanMetric}, p, B::InducedBasis{𝔽,TangentSpaceType,&lt;:RetractionAtlas})
> ```
> 
> and the corresponding inverse local metric you should add a test.
> The `InducedBasis` is also currently not exported.

I haven't exported `InducedBasis` yet because I'm not sure it's the right user-facing abstraction (maybe we should rely more on `DefaultOrthogonalBasis` than on `InducedBasis`? I'm not sure). I'll take a look at that method.

> Finally `induced_basis` should also be covered (both variants) with a test?
> 
> **edit:** `det_local_metric` is now untested in power and product manifold, and I have not yet seen why. I think that's the main tree points I miss (`local_metric` above, `induced_basis` and said `det_local_metric`s).

For power and product we should have many more metric-related tests and functionality anyway. I could delete them for now.

> I hopefully finished this nearly – see the two remarks at the code.
> I have one further (hopefully really final) question.

Thanks for your help! I'll take a look at your remarks.

> Can we call `get_point_coordinates` also just `get_coordinates`? That would be neatly symmetric:
> 
> * `get_coordinates`/`get_vector` have the same signature `(M, p, X, B)`/`(M,p,a,B)` when acting on TpM
> * `get_coordinates`/`get_point` have the same signature `(M, A, i, p)`/`(M,A,i,a)`
> 
> So similar to `project` I think `get_coordinates` should be clear from the signature, which coordinates one wants to obtain (first case: M,p and a basis so its a tangent vector, second case M,A,i indicates the chart, so its a point).

I'm afraid of method ambiguities this may introduce so I'd prefer to keep `get_point_coordinates` and `get_coordinates` separate. `project` is more simple because methods for points and vectors differ by the number of arguments.

src/manifolds/PowerManifold.jl
mateuszbaran: (1) I guess we should have this method for `AbstractPowerManifold` directly instead of `MetricManifold`? Or maybe for both?
(2) Yes, that seems right.

src/manifolds/ProductManifold.jl
mateuszbaran: Yes, I think both need the exact same fix.

kellertuer:
Oh I think I broke something else.

I think concerning test coverage just the power/product issues are left, I hope I got the rest.

Concerning the function names – uff – that‘s a difficult decision, because currently it is _really_ asymmetric with get_coordinates and get_point_coordinates, I really dislike that, but I see your point, too, for sure.

src/manifolds/PowerManifold.jl
kellertuer: With abstract, the transparency still does the same, I think? Can you try?

mateuszbaran:
> Oh I think I broke something else.
> 
> I think concerning test coverage just the power/product issues are left, I hope I got the rest

OK, I can check that.

> Concerning the function names – uff – that‘s a difficult decision, because currently it is _really_ asymmetric with get_coordinates and get_point_coordinates, I really dislike that, but I see your point, too, for sure.

We could also rename `get_coordinates` to `get_vector_coordinates` to be symmetric :smile: .

test/product_manifold.jl
```diff
@@ -595,4 +596,17 @@ end
...
+        p = [zeros(2), ones(2)]
+        X = [ones(2), 2 .* ones(2)]
+        A = RetractionAtlas()
+        a = get_point_coordinates(M, A, p, p)
+        p2 = get_point(M, A, p, a)
+        @test all(p2 .== p)
+        MM = MetricManifold(M, PowerMetric())
+        # det_local_metric(MM, p, DefaultOrthonormalBasis()) to check transparency!
+    end
+    e
```
mateuszbaran, line 611: I don't think this `e` is supposed to be here.

kellertuer: Ah, maybe I was getting tired after an hour of reverse-engineering how the new methods are meant to be, sure, that should not be there.

kellertuer:
> > The `InducedBasis` is also currently not exported.

Well now, since it has a test, it is.

mateuszbaran:
> > > The `InducedBasis` is also currently not exported.
> 
> Well now, since it has a test, it is.

Ah, that's OK :slightly_smiling_face: .

src/manifolds/PowerManifold.jl
mateuszbaran: OK, I'll try that.

mateuszbaran:
Oh no, it looks like metric functions have wrong propagation:
```julia
julia> local_metric(M1, [1.0, 0.0, 0.0], B)
ERROR: MethodError: no method matching local_metric(::Euclidean{Tuple{3}, ℝ}, ::Vector{Float64}, ::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType})
Closest candidates are:
  local_metric(::Euclidean, ::Any, ::InducedBasis{𝔽, ManifoldsBase.TangentSpaceType, var"#s116", TI} where {var"#s116"<:RetractionAtlas, TI}) where 𝔽 at /home/mateusz/.julia/dev/Manifolds/src/manifolds/Euclidean.jl:298
  local_metric(::AbstractDecoratorManifold, ::Any, ::AbstractBasis; kwargs...) at /home/mateusz/.julia/dev/ManifoldsBase/src/DecoratorManifold.jl:358
  local_metric(::MetricManifold{𝔽, var"#s116", EuclideanMetric} where var"#s116"<:AbstractManifold, ::Any, ::InducedBasis{𝔽, ManifoldsBase.TangentSpaceType, var"#s115", TI} where {var"#s115"<:RetractionAtlas, TI}) where 𝔽 at /home/mateusz/.julia/dev/Manifolds/src/manifolds/Euclidean.jl:291
  ...
Stacktrace:
 [1] local_metric__transparent(M::Sphere{2, ℝ}, p::Vector{Float64}, B::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType}; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Manifolds ~/.julia/dev/ManifoldsBase/src/DecoratorManifold.jl:377
 [2] local_metric__transparent(M::Sphere{2, ℝ}, p::Vector{Float64}, B::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType})
   @ Manifolds ~/.julia/dev/ManifoldsBase/src/DecoratorManifold.jl:377
 [3] local_metric(M::Sphere{2, ℝ}, p::Vector{Float64}, B::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType}; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Manifolds ~/.julia/dev/ManifoldsBase/src/DecoratorManifold.jl:363
 [4] local_metric(M::Sphere{2, ℝ}, p::Vector{Float64}, B::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType})
   @ Manifolds ~/.julia/dev/ManifoldsBase/src/DecoratorManifold.jl:359
 [5] top-level scope
   @ REPL[33]:1
```

mateuszbaran:
Anyway, maybe we should just return `1` as `det_local_metric` for `AbstractOrthonormalBasis` (see https://mathoverflow.net/questions/353860/taylor-expansion-of-determinant-of-riemannian-metric-in-normal-coordinates-up-to )

mateuszbaran:
I've removed these two methods for now, they are not that easy to fix and we can always add them back later.

kellertuer:
Ok, that‘s also fine with me.

kellertuer:
Ok, fine. Since I am still unsure about the coordinate naming, let‘s for now just keep it as it is.

So then, as far as I see we could merge this?

mateuszbaran:
Yes, I think we can merge this :tada: .

PR #330, opened 2021-02-11
Euclidean numbers

### LINKED ISSUE #327: Support Euclidean() to be similar to PositiveNumbers()
I just started playing around with mutating functions in Manopt and noticed, I will have to make a few special variants for `Circle()` and `ProsiviteNumbers()` since they work on `Float`s and hence can not have mutating gradient functions – or do not have a `retract!` for example.

I noticed that `Euclidean` can be initialised as `Euclidean()` though that currently does not make much sense.
Would it be reasonable to say that `Euclidean()` is the real numbers (i.e. providing a `Float` based implementation similar to the `PositiveNumers()`) and the 1-element-arrays are `Euclidean(1)` ?

kellertuer:
Provides a thorough distinction between 1-element arrays (`Euclidean(1)`) and immutable points/numbers (`Euclidean()`). This resolves #327.

### DIFF
--- a/src/manifolds/Euclidean.jl
+++ b/src/manifolds/Euclidean.jl
@@ -195,7 +195,7 @@ Transform a tangent vector `X` into a cotangent. Since they can directly be iden
 tangent space of `p` on `M`.
 """
 flat(::Euclidean, ::Any...)
-function flat(::Euclidean{Tuple{}}, ::T, X::TFVector) where {T<:Number}
+function flat(::Euclidean{Tuple{}}, ::Number, X::TFVector)
     return FVector(CotangentSpace, X.data)
 end
 
@@ -295,10 +295,10 @@ function inverse_retract(M::Euclidean{Tuple{}}, x::T, y::T) where {T<:Number}
 end
 function inverse_retract(
     M::Euclidean{Tuple{}},
-    x::T,
-    y::T,
+    x::Number,
+    y::Number,
     ::LogarithmicInverseRetraction,
-) where {T<:Number}
+)
     return log(M, x, y)
 end
 
@@ -312,7 +312,7 @@ which in this case is just
 ````
 """
 Base.log(::Euclidean, ::Any...)
-Base.log(::Euclidean{Tuple{}}, p::T, q::T) where {T<:Number} = q - p
+Base.log(::Euclidean{Tuple{}}, p::Number, q::Number) = q - p
 
 log!(::Euclidean, X, p, q) = (X .= q .- p)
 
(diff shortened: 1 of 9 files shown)


src/manifolds/Euclidean.jl
```diff
@@ -122,13 +127,15 @@ manifold `M`, i.e. for vectors it's just the norm of the difference, for matrice
 and higher order arrays, the matrix and ternsor Frobenius norm, respectively.
 """
 distance(::Euclidean, p, q) = norm(p .- q)
+distance(::Euclidean{Tuple{}}, p::T, q::T) where {T<:Number} = abs(p - q)
```
mateuszbaran, line 130:
```suggestion
distance(::Euclidean{Tuple{}}, p::Number, q::Number) = abs(p - q)
```
We don't need to require both `p` and `q` to have the same type.

src/manifolds/Euclidean.jl
```diff
@@ -175,7 +182,8 @@ Compute the exponential map on the [`Euclidean`](@ref) manifold `M` from `p` in
 \exp_p X = p + X.
 ````
 """
-exp(::Euclidean, ::Any...)
+Base.exp(::Euclidean, ::Any...)
+Base.exp(::Euclidean, p::T, q::T) where {T<:Number} = p + q
```
mateuszbaran, line 186:
```suggestion
Base.exp(::Euclidean, p::Number, q::Number) = p + q
```
Same as above.

src/manifolds/Euclidean.jl
```diff
@@ -275,6 +290,18 @@ function local_metric(::MetricManifold{𝔽,<:Manifold,EuclideanMetric}, p) wher
     return Diagonal(ones(SVector{size(p, 1),eltype(p)}))
 end
 
+function inverse_retract(M::Euclidean{Tuple{}}, x::T, y::T) where {T<:Number}
+    return inverse_retract(M, x, y, LogarithmicInverseRetraction())
+end
+function inverse_retract(
```
mateuszbaran, line 296: Why do we need these methods?

kellertuer: It basically reverses the order:

The first (line 293) would allocate first and call `inverse_retract!` where the default retraction method is then resolved.
The second (here 296) would otherwise allocate and call `retract!` before then calling `log!`.
For both we don't want to go down the mutating route.

src/manifolds/Euclidean.jl
kellertuer: Should I remove the `T` everywhere and write `::Number` instead? For me writing the `T<:Number` is maybe an old habit, where that was beneficial for some reason, I don't remember completely; you're right, these days it would mainly enforce them to have the same type.

mateuszbaran: Yes, I think you should only prefer `::T`... `where T<:Number` over `::Number` when you actually need to use `T` in body of that method.

src/manifolds/Euclidean.jl
mateuszbaran: Yes, I see.

src/manifolds/Euclidean.jl
kellertuer: Yes, thanks. I hopefully just unified that.

src/manifolds/Euclidean.jl
kellertuer: Thanks for checking/asking, it's always good to still check that the ideas are correct.

mateuszbaran (approved):

PR #331, opened 2021-02-12
Adds an integration test with Manopt

### LINKED ISSUE #326: Add integration test with Manopt
We should have an integration test with Manopt, that is, an action that runs Manopt's test suite using the latest changes to Manifolds. This will ensure that potentially breaking changes are identified before being merged. This workflow is really nice: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/105

kellertuer:
adopted (or in other words mainly copied) from  https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/105.

This resolves #326

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -1,0 +1,47 @@
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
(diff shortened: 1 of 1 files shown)


kellertuer:
The test seems to work, since the most recent change to power manifolds `zero_tantent_vector`(now using `zero_tangent_vector` per element instead of power-manifold-log) introduces minor wiggles here and there, that I did not have in mind).

mateuszbaran:
Thanks, that's a good idea. We should merge #330 first though since it solves the tolerance problems.

mateuszbaran (approved):

kellertuer:
Cool thanks. Since this is not a real feature, I would just merge this and it gets into the next release whenever that is?

mateuszbaran:
Yes, I don't think we need to tag a new release with this.

PR #334, opened 2021-03-05
Reading from a power manifold zero-representation point with integer …

mateuszbaran:
…indices yields a number

Counterpart of https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/62


kellertuer:
\o/ This fixes my Power manifold Circle example, thanks!

src/manifolds/PowerManifold.jl
```diff
@@ -104,6 +104,21 @@ function flat!(M::AbstractPowerManifold, ξ::CoTFVector, p, X::TFVector)
     return ξ
 end
 
+Base.@propagate_inbounds function Base.getindex(
```
kellertuer, line 107: Why is the `AbstractPowerManifold` case here in manifolds?
I feel, it could be in base, or does that introduce ambiguities?

mateuszbaran: It's because it can't really be tested in `ManifoldsBase.jl` without the multidimensional array representation or some reduced version of it. Maybe I should just restrict it to the multidimensional case then.

kellertuer: Yes, you're right, I also already noticed, that a test might be tricky in Base. That's a good reason.

I think it can still be the abstract case, such that it will be default to a potentially third representation automatically.

mateuszbaran: I think restricting it is a safer choice for now, if tests pass locally I'll make a commit with this change.

kellertuer: Ok, that's also fine with me, sure.

kellertuer (approved):

PR #335, opened 2021-03-06
Implement the canonical metric on the Stiefel manifold

kellertuer:
This PR adds a second metric to the Stiefel Manifold that provides both a exp and a log map.

* [x] check documentation (I seem to have locally a problem with Plots/PyPlot)
* [x] implement tests


src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+q = \exp_p X = pC + QB.
+```
+For more details, see [^EdelmanAriasSmith1998][^Zimmermann2017].
+"""
+exp(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, ::Any...) where {n,k}
+
+function exp!(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, Q, p, X) where {n,k}
+    A = p' * X
+    QR = qr(X - p * A)
+    BC_ext = exp([A -QR.R'; QR.R, zeros(k, k)])
```
sethaxen, line 57:
```suggestion
    BC_ext = exp([A -QR.R'; QR.R 0*I])
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+    q .= [p Matrix(QR.Q)] * BC_ext[:, 1:k]
```
sethaxen, line 58: This removes an allocation. Could you also use `mul!` to make this line entirely non-allocating
```suggestion
    q .= [p Matrix(QR.Q)] * @view(BC_ext[:, 1:k])
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+
+compute the inner procuct on the [`Stiefel`](@ref) manifold with respect to the
+[`CanonicalMetric`](@ref). The formula reads
+
+```
+g_p(X,Y) = \operatorname{tr}\bigl( X^{\mathrm{T}}(I_n - \frac{1}{2}pp^{\mathrm{T}})Y \bigr).
+```
+"""
+function inner(::MetricManifold{ℝ,Stiefel{n,k,ℝ},X,CanonicalMetric}, p, X, Y) where {n,k}
+    return tr(X' * (Matrix(I, n, n) - 0.5 * p * p') * Y)
```
sethaxen, line 73: What about something like this?
```suggestion
    return dot(X, (Y .- (p * (p'Y)) ./ 2))
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+function log!(
+    ::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric},
+    p,
+    q;
+    tolerance=1e-9,
+    maxiter=1e5,
+) where {n,k}
+    M = p' * q
+    QR = qr(q - p * M)
+    V = qr([M; Matrix(QR.R)]).Q * Matrix(I, 2 * k, 2 * k)
```
sethaxen, line 103: I think this should work.
```suggestion
    V = Matrix(qr([M; QR.R]).Q)
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+    S = svd(V[(k + 1):(2 * k), (k + 1):(2k)]) #bottom right corner
+    V[:, (k + 1):(2 * k)] = V[:, (k + 1):(2 * k)] * (S.V * S.U')
```
sethaxen, line 105: What about this to cut back on allocations?
```suggestion
    Vcorner = @view V[(k + 1):(2 * k), (k + 1):(2k)] #bottom right corner
    S = svd!(Vcorner)
    mul!(Vcorner, Vcorner * S.V, S.U')
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+    LV = log(V)
+    C = view(LV, (k + 1):(2 * k), (k + 1):(2 * k))
+    i = 0
+    while (i < maxiter) && (norm(C) > tolerance)
+        LV = log(V)
+        V[:, (k + 1):(2 * k)] *= exp(-C)
+    end
+    X .= p * LV[1:k, 1:k] + QR.Q * LV[(k + 1):(2 * k), 1:k]
```
sethaxen, line 113: Again we can eliminate allocations here
```suggestion
    mul!(X, p, @view(LV[1:k, 1:k]))
    mul!(X, QR.Q, @view(LV[(k + 1):(2 * k), 1:k]), true, true)
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+    tolerance=1e-9,
+    maxiter=1e5,
+) where {n,k}
+    M = p' * q
+    QR = qr(q - p * M)
+    V = qr([M; Matrix(QR.R)]).Q * Matrix(I, 2 * k, 2 * k)
+    S = svd(V[(k + 1):(2 * k), (k + 1):(2k)]) #bottom right corner
+    V[:, (k + 1):(2 * k)] = V[:, (k + 1):(2 * k)] * (S.V * S.U')
+    LV = log(V)
+    C = view(LV, (k + 1):(2 * k), (k + 1):(2 * k))
```
sethaxen, line 107:
```suggestion
    C = view(LV, (k + 1):(2 * k), (k + 1):(2 * k))
    expnC = exp(-C)
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+    i = 0
+    while (i < maxiter) && (norm(C) > tolerance)
+        LV = log(V)
+        V[:, (k + 1):(2 * k)] *= exp(-C)
```
sethaxen, line 111:
```suggestion
        copyto!(Vcorner, Vcorner*expnC)
```

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -0,0 +1,144 @@
...
+"""
+exp(::Stiefel, ::Any...)
+
+function exp!(::Stiefel{n,k}, q, p, X) where {n,k}
+    return copyto!(
+        q,
+        [p X] *
+        exp([p'X -X'*X; one(zeros(eltype(p), k, k)) p'*X]) *
+        [exp(-p'X); zeros(eltype(p), k, k)],
+    )
```
sethaxen, line 37: This is simpler and avoids a redundant product
```suggestion
    A = p'*X
    return copyto!(q, [p X] * exp([A -X'*X; I p'*X]) * [exp(-A); 0*I])
```

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -0,0 +1,144 @@
...
+
+function get_coordinates!(
+    M::Stiefel{n,k,ℝ},
+    c,
+    p,
+    X,
+    B::DefaultOrthonormalBasis{ℝ},
+) where {n,k}
+    V = get_vectors(M, p, B)
+    c .= [inner(M, p, v, X) for v in V]
```
sethaxen, line 77: Avoids an allocation
```suggestion
    c .= inner.(Ref(M), Ref(p), V, Ref(X))
```

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -0,0 +1,144 @@
...
+````
+
+where $\operatorname{Sym}(q)$ is the symmetrization of $q$, e.g. by
+$\operatorname{Sym}(q) = \frac{q^{\mathrm{H}}+q}{2}$.
+"""
+project(::Stiefel, ::Any...)
+
+function project!(::Stiefel, Y, p, X)
+    A = p' * X
+    copyto!(Y, X - p * Hermitian((A + A') / 2))
```
sethaxen, line 138: Usually `Hermitian` makes things less efficient unless it enables dispatch to an algorithm that leverages the hermitian-ness. matmul I believe does not. An alternative that only allocates once
```suggestion
    T = eltype(Y)
    copyto!(Y, X)
    mul!(Y, p, A + A', T(-0.5), true)
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,115 @@
...
+[^Zimmermann2017]:
+    > Zimmermann, R.: _A matrix-algebraic algorithm for the Riemannian logarithm on the Stiefel manifold under the canoncial metric.
+    > SIAM Journal on Matrix Analysis and Applications 28(2), pp. 322-342, 2017.
+    > doi: [10.1137/16M1074485](https://doi.org/10.1137/16M1074485),
+    > arXiv: [1604.05054](https://arxiv.org/abs/1604.05054).
+"""
+log(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, ::Any...) where {n,k}
+
+function log!(
+    ::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric},
```
sethaxen, line 95:
```suggestion
    ::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric},
    X,
```

sethaxen (changes requested):
I haven't looked at this super closely, but I had some suggestions to reduce allocations and improve efficiency

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: I will try `mul!`, when I write the tests.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Thanks, I definitely have to get more used to thinking in `views`. I was already proud to have seen the view for `C`.

src/manifolds/StiefelEuclideanMetric.jl
kellertuer: Ah, cool that you also looked at the older code (I just reorganised per metric).

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -0,0 +1,142 @@
...
+
+where $\operatorname{Exp}$ denotes matrix exponential,
+$\cdot^{\mathrm{H}}$ denotes the complex conjugate transpose or Hermitian, and $I_k$ and
+$0_k$ are the identity matrix and the zero matrix of dimension $k × k$, respectively.
+"""
+exp(::Stiefel, ::Any...)
+
+function exp!(::Stiefel{n,k}, q, p, X) where {n,k}
+    A = p'*X
+    return copyto!(q, [p X] * exp([A -X'*X; I p'*X]) * [exp(-A); 0*I])
```
sethaxen, line 33:
```suggestion
    return copyto!(q, [p X] * exp([A -X'*X; I A]) * [exp(-A); 0*I])
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,119 @@
...
+
+compute the inner procuct on the [`Stiefel`](@ref) manifold with respect to the
+[`CanonicalMetric`](@ref). The formula reads
+
+```
+g_p(X,Y) = \operatorname{tr}\bigl( X^{\mathrm{T}}(I_n - \frac{1}{2}pp^{\mathrm{T}})Y \bigr).
+```
+"""
+function inner(::MetricManifold{ℝ,Stiefel{n,k,ℝ},X,CanonicalMetric}, p, X, Y) where {n,k}
+    return dot(X, (Y .- (p * (p'Y)) ./ 2))
```
sethaxen, line 73: This cuts out an allocation and is more efficient when `n > k`
```suggestion
    return dot(X, Y) - dot(p'X, p'Y) / 2
```

sethaxen: Actually, since `p*p' == I`, this is the same as
```suggestion
    return dot(X, Y) / 2
```

sethaxen: What about something like this?
```suggestion
    T = Base.promote_eltype(p, X, Y)
    s = T(dot(X, Y))
    if n != k
        s -=  dot(p'X, p'Y) / 2
    end
    return s
```

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,119 @@
...
+
+```math
+q = \exp_p X = pC + QB.
+```
+For more details, see [^EdelmanAriasSmith1998][^Zimmermann2017].
+"""
+exp(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, ::Any...) where {n,k}
+
+function exp!(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, Q, p, X) where {n,k}
+    A = p' * X
```
sethaxen, line 55:
```suggestion
    A = p' * X
    n == k && return mul!(q, p, exp(A))
```

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Thanks for checking all these. When I first got the email with the second suggestion, I directly though: No. Columns are orthogonal not rows ... ;)

But you noticed yourself and I Luke the last one, though for n=k one would most probably use the orthogonal group anyways.

kellertuer:
Hm, setting plots to `"1.6, =1.10.5" loads for the extra of `Plots` in the tests - ... 1.10.6. Here locally tests run fine with 1.10.5, on Julia 1.4 this loads the last working version (1.6), but for Julia 1.5 – or maybe in general? compat constraints for extra test packages seem to be ignored.

kellertuer:
I just checked your code – at one point the small/economic Q or QR was expanded to the full one, which I would avoid using a view – and I forgot one line of code.

Now testing should work and cover all new functions.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Here, on thing did not work: `svd!` changed `Vcorner` which yielded very strange side effects. Also I could not find a reference what is done in place, so switching back to `svd`, the algorithm works.

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,132 @@
...
+    n == k && return mul!(q, p, exp(A))
+    QR = qr(X - p * A)
+    BC_ext = exp([A -QR.R'; QR.R 0*I])
+    q .= [p Matrix(QR.Q)] * @view(BC_ext[:, 1:k])
+    return q
+end
+
+@doc raw"""
+    inner(M::MetricManifold{ℝ, Stiefel{n,k,ℝ}, X, CanonicalMetric}, p,, X, Y)
```
mateuszbaran, line 64:
```suggestion
    inner(M::MetricManifold{ℝ, Stiefel{n,k,ℝ}, X, CanonicalMetric}, p, X, Y)
```

test/stiefel.jl
```diff
@@ -296,4 +301,21 @@ include("utils.jl")
...
+    @testset "Canonical Metric" begin
+        M3 = MetricManifold(Stiefel(3, 2), CanonicalMetric())
+        p = [1.0 0.0; 0.0 1.0; 0.0 0.0]
+        X = [0.0 0.0; 0.0 0.0; 1.0 1.0]
+        q = exp(M3, p, X)
+        Y = [0.0 0.0; 0.0 0.0; -1.0 1.0]
+        r = exp(M3, p, Y)
+        @test isapprox(M3, p, log(M3, p, q), X)
+        @test isapprox(M3, p, log(M3, p, r), Y)
+        @test inner(M3, p, X, Y) == 0
```
mateuszbaran, line 314: Could you also add a test for `inner` for non-orthogonal vectors?

test/stiefel.jl
```diff
@@ -296,4 +301,21 @@ include("utils.jl")
...
+        a = 1.4
+        Z = [0.0 2a; -2a 0.0; -a/2 a/2]
+        s = exp(M3, p, Z)
+        Z2 = log(M3, p, s)
+        @test isapprox(M3, p, Z, Z2)
```
mateuszbaran, line 319: It would be nice to test `distance` (we had (have?) a bug, where `distance` worked for Euclidean Stiefel but returned a wrong answer).

kellertuer: I hope we had that bug since I changed the transparency in this PR for that.

I can add a test for sure.

test/stiefel.jl
kellertuer: It's sometimes hard to come up with numbers,

I also need a test that actually needs the iterative process, currently all `log`s are already precise enough after the procrustes prestep; I just can not find a good example where that Is needed. If you have a good idea, feel free to propose a test (see currently the 4 uncovered lines).

mateuszbaran: I think you could just try something like `inner(M3, p, X, 2*X+3*Y)`? I have no idea how to test that case of `log` though other than randomly trying different points until the right code path is hit.

kellertuer: Ok, I will add that, but what should I compare that to?

As I wrote above, I even already talked to Ralf about that, I will try to find one.

mateuszbaran: Well, it should be equal to `2*norm(M3, p, X)^2` I think.

kellertuer: Ah, I alredy forgot about the orthogonality of X and Y, sorry; yes, you're right. I will add that.

kellertuer: For the Euclidean distance, I changed transparency and added a test, that it really does not work.
We can change that of course once we get a log for Euclidean metric.

kellertuer:
One thing we should maybe do in a next PR, currently this `log` needs to pass down two warst to `log!`. We should maybe allow that on the generic level?

mateuszbaran:
What is "warst"?

kellertuer:
> What is "warst"?

some strange autocorrect (German past participle of “sein” -  to be) who I meant to type `kwargs`.

mateuszbaran:
My only issue with introducing that on a generic level is that last time I checked Julia did a relatively poor job at optimizing out kwargs. Maybe it's better now, I'll have to check it carefully before this change is made.

kellertuer:
Sure. I am also not 100% sure a global level would be that good, also for further usage (in algorithms say). We can keep it the current way for not, surely.

mateuszbaran:
Regarding the failure on Julia 1.4, I think it's fine to check ambiguities only on Julia >=1.5. By the way, Julia nightly also has a similar problem: https://github.com/JuliaLang/julia/issues/37711 .

kellertuer:
Ok, I till change that as soon as I find time to look for the remaining code lines to be covered.

sethaxen:
Perhaps unrelated, but the ambiguity test stackoverflows for me on Windows only on Julia 1.5: https://github.com/JuliaManifolds/Manifolds.jl/pull/249/checks?check_run_id=2093697330#step:5:329.

kellertuer:
Yes, here in the checks the same happens for Julia 1.4, seems to be unrelated, though I have not yet seen how to fix that.

kellertuer:
So with the commit from today this should be feature complete and all new lines tested. If you could take another look and check if all changes are good and the changes are good, that would be great.

Concerning the avoidance of allocations, the only thing I had to revert is the use of `svd!` (back to `svd`) since the input matrix is still needed afterwards.

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
+@doc raw"""
+    CanonicalMetric <: Metric
+
+The Canonical Metric is name used on several manifolds, for example for the [`Stiefel`](@ref)
+manifold, see for example[^EdelmanAriasSmith1998].
+
```
sethaxen, line 6: If this is intended to be general, then perhaps this definition should be in `metric.jl` or another file? Then the reference should be moved to the docs page.

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
...
+[^EdelmanAriasSmith1998]:
+    > Edelman, A., Ariar, T. A., Smith, S. T.:
+    > _The Geometry of Algorihthms with Orthogonality Constraints_,
+    > SIAM Journal on Matrix Analysis and Applications (20(2), pp. 303–353, 1998.
+    > doi: [10.1137/S0895479895290954](https://doi.org/10.1137/S0895479895290954)
+    > arxiv: [9806030](https://arxiv.org/abs/physics/9806030)
+"""
+struct CanonicalMetric <: RiemannianMetric end
+
+function distance(M::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, q, p) where {n,k}
```
sethaxen, line 16: Since this is the standard definition of `distance`, is there a reason this is limited to the reals?

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
...
+For more details, see [^EdelmanAriasSmith1998][^Zimmermann2017].
+"""
+exp(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, ::Any...) where {n,k}
+
+function exp!(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, q, p, X) where {n,k}
+    A = p' * X
+    n == k && return mul!(q, p, exp(A))
+    QR = qr(X - p * A)
+    BC_ext = exp([A -QR.R'; QR.R 0*I])
+    q .= [p Matrix(QR.Q)] * @view(BC_ext[:, 1:k])
```
sethaxen, line 63: If you blockwise multiply here, then instead of the 3 allocations here, I think you can do with none.

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
...
+    tolerance=1e-9,
+) where {n,k}
+    M = p' * q
+    QR = qr(q - p * M)
+    V = Matrix(qr([M; QR.R]).Q * I)
+    Vcorner = @view V[(k + 1):(2 * k), (k + 1):(2k)] #bottom right corner
+    Vpcols = @view V[:, (k + 1):(2 * k)] #second half of the columns
+    S = svd(Vcorner) # preprocessing: Procrustes
+    mul!(Vpcols, Vpcols * S.U, S.V')
+    V[:, 1:k] .= [M; QR.R]
```
sethaxen, line 136: Why not set the blocks of `V` directly to avoid the rhs allocation?

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
...
+    ::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric},
+    X,
+    p,
+    q;
+    maxiter=1e5,
+    tolerance=1e-9,
+) where {n,k}
+    M = p' * q
+    QR = qr(q - p * M)
+    V = Matrix(qr([M; QR.R]).Q * I)
```
sethaxen, line 131: Same here, why `I`?

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
...
+
+    LV = real.(log(V)) # this can be replaced by log_safe.
+    C = view(LV, (k + 1):(2 * k), (k + 1):(2 * k))
+    expnC = exp(-C)
+    i = 0
+    while (i < maxiter) && (norm(C) > tolerance)
+        i = i + 1
+        LV .= real.(log(V))
+        expnC .= exp(-C)
+        copyto!(Vpcols, Vpcols * expnC)
```
sethaxen, line 146: Can you allocate a temporary outside the loop and then multiply into it? e.g. `mul!(tmp, Vpcols, expnC)`

src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -0,0 +1,153 @@
...
+    end
+    LV .= real.(LV)
```
sethaxen, line 148: What does this line do? Since `LV` in all previous calls is already constrained to have real entries.

src/manifolds/StiefelCanonicalMetric.jl
sethaxen: Is it because the metric-related functions below are only defined for the real case? Is there any reason they can't also be generalized to the complex manifold?

src/manifolds/SymmetricPositiveDefinite.jl
```diff
@@ -170,4 +170,4 @@ definite matrix `x` on the [`SymmetricPositiveDefinite`](@ref) manifold `M`.
 """
 zero_tangent_vector(::SymmetricPositiveDefinite, ::Any)
 
-zero_tangent_vector!(M::SymmetricPositiveDefinite{N}, v, x) where {N} = fill!(v, 0)
+zero_tangent_vector!(::SymmetricPositiveDefinite{N}, v, x) where {N} = fill!(v, 0)
```
sethaxen, line 173:
```suggestion
zero_tangent_vector!(::SymmetricPositiveDefinite, v, x) = fill!(v, 0)
```

sethaxen (changes requested):
Just a few comments and a few more allocation-reductions. There are more allocations that can be removed using e.g. `qr!` but I don't think they'll have a huge impact on performance, and it's less likely that `qr!` will be overloaded for some special matrix than `qr` will be.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Sure we can move it to the metric.jl file and docs. I am not 100% sure what you mean by moving the reference? It is shown in REPL this way and shows up on the docs page where it is used. Sure I would prefer a cite-style referencing, but that package/addon for Documenter is not yet ready.

docs/src/manifolds/stiefel.md
```diff
@@ -1,9 +1,48 @@
...
+
+```@autodocs
+Modules = [Manifolds]
+Pages = ["manifolds/StiefelEuclideanMetric.jl"]
+Order = [:function]
+```
+
+## The canonical metric
+
+```@autodocs
```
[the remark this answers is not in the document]
kellertuer: Good idea.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Exactly, I am just not 100% sure what the complex case would look like. If you have a reference that this all also works for the complex case, we can surely do that.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: How would I do that? I twiddled _so much_ with these few lines of code, before and after your allocation reduction (about 2 days to find that we can not use the `svd!`) that I don't see directly how to do this block wise.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: This is due to the reduced form the QR would return, see https://github.com/JuliaLang/julia/issues/27397, at least without that, the Q factor is rectangular this way, it should be quadratic.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: In earlier times I tried to fiddle around whether this works for complex and where I might have to switch back to reals (initially it seemed that within the loop all had to be complex), but now, you'Re right, this is obsolete.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: At point of writing (see last comment above) it was easier to write this than do think about the sized of the blocks. But sure, I will do that.

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Hm, looking at it closely, I am not sure what it would be in the generic sense. Because canonical metric is usually always the euclidean one. So can we maybe only move it to a global case and move the reference for example to the exponential map, once we have a second canonical-non-Euclidean-metric case?

src/manifolds/StiefelCanonicalMetric.jl
kellertuer: Yes, confirmed it again, with `I` the matrix `V` is 2k x 2k (in our tests 4 x 4), otherwise only 2k x k (4x2).

kellertuer:
> Just a few comments and a few more allocation-reductions. There are more allocations that can be removed using e.g. `qr!` but I don't think they'll have a huge impact on performance, and it's less likely that `qr!` will be overloaded for some special matrix than `qr` will be.

I don't feel comfortable currently trying that after it took me two full days changing to `svd!` to see that it had too many side effects and made the calculation wrong, since the input matrix was still used afterwards. So my fear is not the overwriting but the side effects from working in place. Debugging this `log!` here is not really much fun. So if you want to give it a try, feel free to do so. I would like to not do that, to not again take a few full days getting `log!` back up working (possibly in the form it is now).

kellertuer:
Should we maybe deactivate the windows test or at least ambiguities thereon until it has come to conclusion not to do stack overflows when looking for those?

src/manifolds/StiefelCanonicalMetric.jl
sethaxen: No, I don't have one, just curious.

src/manifolds/StiefelCanonicalMetric.jl
sethaxen: I'll take a stab at it.

src/manifolds/StiefelCanonicalMetric.jl
sethaxen: `Matrix(A)` creates a copy even if `A` is already a `Matrix` (which it should be after multiplying by `I`). Changing to `convert(Matrix, A)` avoids the copy unless `A` happens to not be a `Matrix`. I'll push the change as well.

sethaxen:
> > Just a few comments and a few more allocation-reductions. There are more allocations that can be removed using e.g. `qr!` but I don't think they'll have a huge impact on performance, and it's less likely that `qr!` will be overloaded for some special matrix than `qr` will be.
> 
> I don't feel comfortable currently trying that after it took me two full days changing to `svd!` to see that it had too many side effects and made the calculation wrong, since the input matrix was still used afterwards. So my fear is not the overwriting but the side effects from working in place. Debugging this `log!` here is not really much fun. So if you want to give it a try, feel free to do so. I would like to not do that, to not again take a few full days getting `log!` back up working (possibly in the form it is now).

Oh, I meant to say that these changes are probably not worth the code complexity or effort. I'm sorry for not being clear.

sethaxen:
> Should we maybe deactivate the windows test or at least ambiguities thereon until it has come to conclusion not to do stack overflows when looking for those?

This is probably a good idea.

sethaxen (approved):
Assuming tests pass and you approve of the changes, I think this is ready to merge. Nice work!

PR #336, opened 2021-03-19
Copy and broadcasting for ProductRepr

mateuszbaran:
This should fix https://github.com/JuliaManifolds/Manopt.jl/issues/64 .


kellertuer (approved):

kellertuer:
I fear the bug with Plots is not yet fixed, as far as I understand the current test result.

mateuszbaran:
Oh, indeed. This looks like a separate bug though.

kellertuer:
I will check for that bug soon, then.

mateuszbaran:
Thanks!

There seems to be a problem with ambiguities but only on Windows, weird.

PR #340, opened 2021-03-22
Fix projection and retraction in fixed rank matrices

qrebjock:
Hello,

I propose these few changes to address the comments in https://github.com/JuliaManifolds/Manopt.jl/issues/61.

The retraction is equivalent to the one in Algorithm 6 by Vandereycken, but is numerically more stable when the iterates approach matrices of lower rank. (It is the one proposed in Chapter 7 of http://sma.epfl.ch/~nboumal/book/IntroOptimManifolds_Boumal_2020.pdf.)


src/manifolds/FixedRankMatrices.jl
```diff
@@ -81,9 +81,9 @@ Base.:(==)(x::SVDMPoint, y::SVDMPoint) = (x.U == y.U) && (x.S == y.S) && (x.Vt =
 @doc raw"""
     UMVTVector <: TVector
 
-A tangent vector that can be described as a product $UMV^\mathrm{H}$, at least
-together with its base point, see for example [`FixedRankMatrices`](@ref). This
-vector structure stores the additionally (to the point) required fields.
+A tangent vector that can be described as a product $UMV^\mathrm{H} + U_p V^\mathrm{H} + U V_p^\mathrm{H}$,
```
kellertuer, line 84: Just two notation remarks
* it is easier/better to use tow backticks (``) instead of the usual ($), since that is the default Markdown syntax and a little more Julia friendly
* Usually our points are ``p,q\in\mathcal M`` while tangents are ``X,Y\in T_p\mathcal M``, see https://juliamanifolds.github.io/Manifolds.jl/latest/misc/notation.html . I know this is sometimes challenging when they are all matrices, but I would prefer writing ``X= UMV...`` since that's the tangent vector and then ``p = U\SigmaV...``, where I think I used ``S`` instead of a ``\Sigma`` until now here. We can change that last point if you like, but then please consistently.

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +407,21 @@ function zero_tangent_vector!(
     X.Vt .= zeros(eltype(X.Vt), k, n)
     return X
 end
+
+@doc raw"""
+    to_ambient(X::SVDMPoint)
+
+Reconstruct a low-rank matrix from the its `SVDMPoint` representation.
+"""
+function to_ambient(X::SVDMPoint)
```
kellertuer, line 416: This should be called `embed!`, see https://juliamanifolds.github.io/Manifolds.jl/latest/interface.html#ManifoldsBase.embed!-Tuple{Manifold,Any,Any}, where we have two cases:

For a default embedding that one “usually has in mind” one would use as the first parameter just the manifold (`::FixedRankMatrices` obviously has a default embedding in that sense), and for further/other embeddings use an `EmbeddedManifold`, which is “just” a type to bind two manifolds, a Manifold `M` and its embedding (one of the several it might have).

Note that `embed!` works in place of the second parameter, the `embed` then automatically works, too :)

src/manifolds/FixedRankMatrices.jl
qrebjock: I see. I used the notation `U_p` because the columns are perpendicular to the ones of `U` (same for `V_p` and `V`).
Since we have `p = U S V^T` it's a bit confusing to have the tangent vector represented by `(U, M, V^T)`.
Maybe `\tilde{U}`?

src/manifolds/FixedRankMatrices.jl
```diff
@@ -287,16 +284,18 @@ singular values and $U$ and $V$ are shortened accordingly.
...
     ::PolarRetraction,
-) where {M,N,k}
-    s = svd(p.U * Diagonal(p.S) * p.Vt + (p.U * X.M * p.Vt + X.U * p.Vt + X.U * X.Vt))
-    q.U .= s.U[:, 1:k]
-    q.S .= s.S[1:k]
-    q.Vt .= s.Vt[1:k, :]
+) where {m, n, k}
+    QU, RU = qr([p.U X.U])
+    QV, RV = qr([p.Vt' X.Vt'])
+    T = svd(RU * [diagm(p.S) + X.M I; I zeros(k, k)] * RV')
```
sethaxen, line 295: The expression within the `svd(...)` allocates 6 matrices. It would be preferable to perform the multiplication blockwise to avoid allocating so many intermediates. e.g. something like
```suggestion
    @views begin
        RU11 = RU[:, 1:k]
        RU12 = RU[:, (k + 1):(2 * k)]
        RV11 = RV[:, 1:k]
        RV12 = RV[:, (k + 1):(2 * k)]
    end
    tmp = RU11 .* p.S' .+ RU12
    mul!(tmp, RU11, X.M, true, true)
    tmp2 = tmp * RV11'
    mul!(tmp2, RU11, RV12', true, true)
    T = svd(tmp2)
```
would allocate just two temporaries. (perhaps you can think of more descriptive variable names)

src/manifolds/FixedRankMatrices.jl
```diff
@@ -287,16 +284,18 @@ singular values and $U$ and $V$ are shortened accordingly.
...
+    q.U .= QU * T.U[:, 1:k]
+    q.S .= T.S[1:k]
+    q.Vt .= T.Vt[1:k, :] * QV'
```
sethaxen, line 298: Can you use views here to avoid allocating when you slice?
Also you could use `mul!` to multiply in-place instead of multiplying and then copying. This should reduce the 5 allocations here to 0.
```suggestion
    mul!(q.U, QU, @view(T.U[:, 1:k]))
    q.S .= @view(T.S[1:k])
    mul!(q.Vt, @view(T.Vt[1:k, :]), QV')
```

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +407,21 @@ function zero_tangent_vector!(
...
     return X
 end
+
+@doc raw"""
+    to_ambient(X::SVDMPoint)
+
+Reconstruct a low-rank matrix from the its `SVDMPoint` representation.
+"""
+function to_ambient(X::SVDMPoint)
+    return X.U * diagm(X.S) * X.Vt
```
sethaxen, line 417: This avoids an allocation
```suggestion
    return X.U * Diagonal(X.S) * X.Vt
```

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +407,21 @@ function zero_tangent_vector!(
...
+end
+
+@doc raw"""
+    to_ambient(X::SVDMPoint, T::UMVTVector)
+
+Reconstruct a tangent vector from its `UMVTVector` representation.
+"""
+function to_ambient(X::SVDMPoint, T::UMVTVector)
+    return X.U * T.M * X.Vt + T.U * X.Vt + X.U * T.Vt
```
sethaxen, line 426: You can do this with just two allocations:
```suggestion
    Y = X.U * T.M * X.Vt
    mul!(Y, T.U, X.Vt, true, true)
    mul!(Y, X.U, T.Vt, true, true)
    return Y
```

sethaxen (changes requested):
Thanks for the PR! I had a few suggestions, mostly to reduce the number of allocations.

src/manifolds/FixedRankMatrices.jl
sethaxen: (definitely check my math on this one)

qrebjock:
Thank you very much for the detailed comments sethaxen.
I am not aware of all these tricks as I just started working with Julia but I am very happy to learn them and see that code can be optimized so much.

I am going to make the changes tomorrow.

Actually I was also concerned by the performances of the QR decomposition.
There is no way to get the thin QR decomposition in Julia as far as I know.
When we do `QU, RU = qr([p.U X.U])` the `QU` factor is in fact `m x 2r`. So it should be possible to make the function scale linearly with `m` (and `n` similarly). But because of the internal representation of the Q factor it's not really the case.
I opened a topic about this issue here https://discourse.julialang.org/t/economy-sized-qr-decomposition/57608/7.
If you have other ideas it could be helpful!

src/manifolds/FixedRankMatrices.jl
qrebjock: I was not aware of this function. I will make the changes tomorrow!

kellertuer: That's completely fine that you were not aware (luckily I knew of `to_ambient`), ours is called `embed` since it might be more general then “just” going into ambient space.

src/manifolds/FixedRankMatrices.jl
kellertuer: Sure, we are a little different from a few low rank papers here, but we try to keep a unified notation over all our manifolds.

What do you think about specifzying both with indices, i.e. `p = U_pS_pV_p^T` is the point on the manifold and the three matrices for the condensed form of the tangent `X` are `U_X`, `M_X` and `V_X` ? Sure this should then also be adapted in the `FixedRankMatrices` doc string, I must have misread that doc string (thinking U_p and V_p belong to p and not that they are orthogonal) yesterday. But I think this only further demonstrates, that it has to be very clear which matrices belong to what element.

kellertuer:
> Thank you very much for the detailed comments sethaxen.

He's an allocation magician, for sure. ;)

> There is no way to get the thin QR decomposition in Julia as far as I know.

The default is the thin one, see for example https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.qr the second to last paragraph (to get the full one, you have to multiply with `I`). Not that the `Q` is not returned as a matrix but as a factor object.

src/manifolds/FixedRankMatrices.jl
qrebjock: I made the changes. It's indeed allocating less memory and is slightly faster.
However the line
`mul!(q.Vt, @view(T.Vt[1:k, :]), QV')`
throws an error because of supposedly dimension mismatchs.

But the dimensions are right, and actually computing the product `@view(T.Vt[1:k, :]) * QV'` raises no error.
I think it's because `q.Vt` is an Adjoint. If I do
```julia
C = zeros(size(q.Vt))
mul!(C, @view(T.Vt[1:k, :]), QV')
```
then it works.

That's weird.

I guess we could just keep the operation `q.Vt .= T.Vt[1:k, :] * QV'` unless there is an easy fix.

kellertuer: When do you get the error? The test cases for FixedRank run fine for me (at least for the retraction).

src/manifolds/FixedRankMatrices.jl
```diff
@@ -261,9 +261,6 @@ function project!(::FixedRankMatrices, Y::UMVTVector, p::SVDMPoint, A::AbstractM
     Y.Vt .= (aTu - p.Vt' * uTav')'
     return Y
 end
-function project!(M::FixedRankMatrices, Y::UMVTVector, p::SVDMPoint, X::UMVTVector)
-    return project!(M, Y, p, X.U * X.M * X.Vt)
```
kellertuer, line 265: We have to adapt the tests, since this was deleted (and sure this method was wrong, one would need the point `q` to whose tangent space the vector `X` belongs to, which we do not have). We could introduce a vector transport by projection using something like this.

src/manifolds/FixedRankMatrices.jl
sethaxen: If `q.Vt` is an `Adjoint`, then you can just perform the adjoint operation:
```julia
mul!(q.Vt', QV, @view(T.Vt'[:, 1:k]))
```

@kellertuer is there a reason `Vt` is stored as an `Adjoint`? In an `SVD` object, usually `Vt` is a `Matrix`, and `V` is the `Adjoint`.

kellertuer: I am not aware of any reason, I wasn't even aware it's stored as an adjoint, but sure, I usually stored `V` I think. if storing `Vt` is easier, we can surely do that.

sethaxen:
> Thank you very much for the detailed comments sethaxen.
> I am not aware of all these tricks as I just started working with Julia but I am very happy to learn them and see that code can be optimized so much.

No problem, and welcome! In Julia, you'll find that there's usually a trade-off between generality and efficiency, and the key is finding the right balance. e.g. in Manifolds we have chosen to make the default implementations of all functions mutate the inputs. The downside to this is that some automatic differentiation packages (notably Zygote), can't differentiate through mutation. So we've chosen efficiency over generality. Likewise, this PR could use the `qr!` and `svd!` functions defined in LinearAlgebra, which mutate the inputs to reduce allocations. However, if a user defines some custom array type, it's far more likely they'll specialize `qr` and `svd` than `qr!` and `svd!`, so if we call `qr!`, we may miss some efficient method defined by the user for their type and instead hit a less efficient default. So in this case we've chosen generality over efficiency.

Add to that that often the most efficient implementation is less human-readable and therefore harder to maintain. We usually handle that by including the more human-readable equation for the function in a docstring.

In this case, views are really general, and `mul!` is very commonly implemented for custom array types. Likewise, broadcasting is generally efficiently implemented, so these particular optimizations are pretty safe and should substantially increase the efficiency.

src/manifolds/FixedRankMatrices.jl
qrebjock: Unfortunately the adjoint operation
```julia
mul!(q.Vt', QV, @view(T.Vt'[:, 1:k]))
```
doesn't work either. It really looks like a bug to me.

It's an adjoint just because of the way `random_tangent` samples the vector. But we can make it a `Matrix` if you think it's better.

qrebjock: I didn't push anything yet kellertuer. That's why you don't see the error. I wanted to make sure that it's ok to keep the last operation as `q.Vt .= T.Vt[1:k, :] * QV'`.

kellertuer: If we could resolve this by not storing the adjoint but the actual matrix (as @sethaxen mentioned) then I would maybe opt for that, since that's easier to handle. I don't know why I originally saved the adjoint and not the matrix in the struct, I feel the matrix would be more natural. So if qe could carefully switch to storing a matrix, then the original comment of this change would work again :)

qrebjock: I tried without the adjoint. The operation doesn't throw an error anymore. But: it doesn't compute the right result.
An example is better than long sentences.
Try to run this code:
```julia
A = randn(5, 4)
Q, R = qr(A)
B = randn(4, 4)
C = randn(2, 5);

mul!(C, @view(B[1:2, :]), Q')
display(C)

display(@view(B[1:2, :]) * Q')
```
The result is pretty surprising.

It seems like the operation `mul!(C, @view(B[1:2, :]), Q')` doesn't override `C` but takes into account what was written inside.

sethaxen: Oh yeah, `qr(A).Q` has really unique behavior because it's an `AbstractMatrix` with a certain size but can be multiplied by matrices of different first dimensions. Probably `Adjoint` here is breaking its behavior for `mul!` only. This works:
`mul!(C, @view(B[1:2, :]), Matrix(Q)')`.

src/manifolds/FixedRankMatrices.jl
qrebjock: Actually in the doc https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/fixedrankmatrices.html you are also using `U_p` and `V_p` for the tangent components.
I could change that to the notation you suggested above if you like.
What do you think?

kellertuer: Yes, because at that time I was confused myself. I would be in favour of changing that then as well, yes.

src/manifolds/FixedRankMatrices.jl
qrebjock: Yes this is working!
So the last operation could become
```julia
mul!(q.Vt, @view(T.Vt[1:k, :]), Matrix(QV)')
```
but it's not faster than
```julia
q.Vt .= T.Vt[1:k, :] * QV'
```
(it's actually slower.)

I'll put `mul!(q.Vt, @view(T.Vt[1:k, :]), Matrix(QV)')` for now but this should probably be updated in the future.

src/manifolds/FixedRankMatrices.jl
qrebjock: Oh ok, actually I misread what you said I think. All good! I'll make the changes.

qrebjock:
Ok here are some changes that should hopefully address your comments.

I did not manage to make `embed` work by defining only `embed!`. So I implemented the function `embed` on top of `embed!`. Not sure why it is not working without this.

src/manifolds/FixedRankMatrices.jl
sethaxen: On my machine, they're equivalent in speed, but the `mul!` one allocates less:
```julia
julia> using LinearAlgebra, BenchmarkTools

julia> A = randn(5, 4);

julia> Q, R = qr(A);

julia> B = randn(4, 4);

julia> C = randn(2, 5);

julia> foo!(C, B, Q) = mul!(C, @view(B[1:2, :]), Matrix(Q)');

julia> bar!(C, B, Q) = (C .= @view(B[1:2, :]) * Q'; C);

julia> @btime $(foo!)($C, $B, $Q);
  1.442 μs (3 allocations: 688 bytes)

julia> @btime $(bar!)($C, $B, $Q);
  1.444 μs (9 allocations: 784 bytes)
```

It probably doesn't matter, but I think this is fine.

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +421,34 @@ function zero_tangent_vector!(
...
 end
+
+@doc raw"""
+    embed(p::SVDMPoint)
+
+Embed the point `p` from its `SVDMPoint` representation.
+"""
+
+function embed!(::FixedRankMatrices, q, p)
+    return q .= p.U * Diagonal(p.S) * p.Vt
```
sethaxen, line 432:
```suggestion
    return mul!(q, p.U * Diagonal(p.S), p.Vt)
```

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +421,34 @@ function zero_tangent_vector!(
...
+end
+function embed(M::FixedRankMatrices{m,n,k}, p) where {m,n,k}
```
sethaxen, line 434: Do we need this method defined, or is the default `embed` in ManifoldsBase sufficient?
https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/v0.10.4/src/ManifoldsBase.jl#L320-L324

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +421,34 @@ function zero_tangent_vector!(
...
+    q = zeros(m, n)
+    return embed!(M, q, p)
+end
+
+@doc raw"""
+    embed(::FixedRankMatrices, Y, p, X)
+
+Embed the tangent vector `X` at point `p` in `M` from
+its `UMVTVector` representation.
+"""
```
sethaxen, line 444: Our convention is to document the non-mutating function (even if not defined).

```suggestion
@doc raw"""
    embed(M::FixedRankMatrices, p, X)

Embed the tangent vector `X` at point `p` in `M` from
its [`UMVTVector`](@ref) representation.
"""
embed(::FixedRankMatrices, p, X)

```

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +421,34 @@ function zero_tangent_vector!(
...
+function embed!(::FixedRankMatrices, Y, p, X)
+    Y = p.U * X.M * p.Vt
+    mul!(Y, X.U, p.Vt, true, true)
+    return mul!(Y, p.U, X.Vt, true, true)
+end
+
+function embed(M::FixedRankMatrices{m,n,k}, p, X) where {m,n,k}
```
sethaxen, line 451: Same, is the default sufficient? https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/v0.10.4/src/ManifoldsBase.jl#L363-L369

src/manifolds/FixedRankMatrices.jl
qrebjock: Hmm interesting. For small dimensions I have the same result. But when dimensions are larger `bar!` becomes 2 times faster even though it allocates more:
```julia
using LinearAlgebra, BenchmarkTools

n = 1000
m = 800
p = m ÷ 2

A = randn(n, m);
Q, R = qr(A);
B = randn(m, m);
C = randn(p, n);

foo!(C, B, Q) = mul!(C, @view(B[1:p, :]), Matrix(Q)');
bar!(C, B, Q) = (C .= @view(B[1:p, :]) * Q'; C);

@btime $(foo!)($C, $B, $Q);
@btime $(bar!)($C, $B, $Q);

  18.510 ms (14 allocations: 17.09 MiB)
  8.450 ms (20 allocations: 6.10 MiB)
```

src/manifolds/FixedRankMatrices.jl
```diff
@@ -408,3 +421,34 @@ function zero_tangent_vector!(
...
+@doc raw"""
+    embed(::FixedRankMatrices, Y, p, X)
+
+Embed the tangent vector `X` at point `p` in `M` from
+its `UMVTVector` representation.
+"""
+function embed!(::FixedRankMatrices, Y, p, X)
+    Y = p.U * X.M * p.Vt
+    mul!(Y, X.U, p.Vt, true, true)
+    return mul!(Y, p.U, X.Vt, true, true)
```
sethaxen, line 448: I see now we can eliminate one matrix product. Also, we need to mutate the passed `Y`
```suggestion
    tmp = p.U * X.M
    tmp .+= X.U
    mul!(Y, tmp, p.Vt)
    return mul!(Y, p.U, X.Vt, true, true)
```

sethaxen (changes requested):

src/manifolds/FixedRankMatrices.jl
sethaxen: Oh, good spot, then I'd say `bar!` should be preferred. Also perhaps should use `copyto!` instead of broadcasting when filling the whole array, since it's simpler and more general. so e.g.
```julia
copyto!(q.Vt, @view(T.Vt[1:k, :]) * QV')
```

src/manifolds/FixedRankMatrices.jl
qrebjock: For some reason that I don't understand `embed` throws an error if I don't implement it. But kellertuer said that `embed!` should be enough, so I don't know.
```julia
MethodError: no method matching mul!(::SVDMPoint{Array{Float64,2},Array{Float64,1},Array{Float64,2}}, ::Array{Float64,2}, ::Array{Float64,2}, ::Bool, ::Bool)
Closest candidates are:
  mul!(!Matched::StridedArray{T, 2}, ::StridedVecOrMat{T}, ::StridedVecOrMat{T}, ::Number, ::Number) where T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64} at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/matmul.jl:167
  mul!(!Matched::AbstractArray{T,2} where T, ::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T, ::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T, ::Number, ::Number) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/matmul.jl:235
  mul!(!Matched::AbstractArray, !Matched::Number, ::AbstractArray, ::Number, ::Number) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/generic.jl:124
  ...

Stacktrace:
 [1] mul! at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/matmul.jl:208 [inlined]
 [2] embed!(::FixedRankMatrices{50,50,10,ℝ}, ::SVDMPoint{Array{Float64,2},Array{Float64,1},Array{Float64,2}}, ::SVDMPoint{Array{Float64,2},Array{Float64,1},Array{Float64,2}}) at /Users/quentinrebjock/.julia/dev/Manifolds/src/manifolds/FixedRankMatrices.jl:437
 [3] embed(::FixedRankMatrices{50,50,10,ℝ}, ::SVDMPoint{Array{Float64,2},Array{Float64,1},Array{Float64,2}}) at /Users/quentinrebjock/.julia/dev/ManifoldsBase/src/ManifoldsBase.jl:322
 [4] top-level scope at In[3]:1
 [5] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

I experimented a little and I think it's because `embed` doesn't allocate the right type for `q`. And I think something should be added to the manifold `FixedRankMatrices` so that `embed` recognizes the right type to allocate. But I haven't figured out what to do.

sethaxen: That's interesting. Yeah, looks like a bug that should be handled elsewhere first. @kellertuer any ideas what's going wrong here?

kellertuer: Yes, we are missing the correct allocation function https://juliamanifolds.github.io/Manifolds.jl/latest/interface.html#Allocation, which is an advanced variant of `similar`basically, since for example the log has two points as input but has to allocate the right tangent vector as result memory. I know how that works and how to do it, I just have a meeting today first.

kellertuer:
I noticed a few points to get embed to work.
1) The `FixedRankMatrices` should be subtyped to `struct FixedRankMatrices{M,N,K,𝔽} <: AbstractEmbeddedManifold{𝔽,DefaultEmbeddingType} end`. I think this is not necessary, but useful to know

2) We have to allocate the matrix for `embed!` ( see https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/bebd48ec476fc86b813d41e83cd7c3b4e3148c3e/src/ManifoldsBase.jl#L320-L324 for how our generic non mutating functions are usually implemented). I chose to do a variant that directly works for both embedding points and embedding tangents like this
```julia
function allocate_result(::FixedRankMatrices{m,n,k}, ::typeof(embed), v...) where {m,n,k}
    #note that vals is (p,) or (X,p) but both first entries have a U of correct type
    return similar(typeof(v[1].U),m,n)
end
```
what it does is, it determines the type of arrays that are currently used and generates an array of correct size within the generic `embed` which then calls the specific `embed!` you implemented. Maybe something similar should also be done with the `zero_tangent_vector`, where I seem to have been lazy for this manifold. This works fine for example with
```julia
    M = FixedRankMatrices(3, 2, 2)
    x = SVDMPoint([1.0 0.0; 0.0 1.0; 0.0 0.0])
    v = UMVTVector([0.0 0.0; 0.0 0.0; 1.0 1.0], [1.0 0.0; 0.0 1.0], zeros(2, 2))
    embed(M,x)
    embed(M,x,v)
```

when removing your `embed` functions. Note that for the tangent vector `embed!` you forgot to put the signature below the doctoring in line 430, there is a `embed(::FixedRankMatrices, p)`missing.

3) It would be nice to order the functions within a manifold file alphabetically.

kellertuer:
Dear @qrebjock – would you maybe have time to have a look here again? If you are stuck somewhere, just let us know, where we might offer help; I could to a similar second-branch-trick as for the other PR.

kellertuer:
bump @qrebjock – let us know if we can help with anything. The later we want to merge this, the more complicated it might get, when the rest develops further.

kellertuer:
Hm, this seems to stall, should we, @sethaxen, @mateuszbaran maybe restart this as one of our branches o this repo to get it finished? I think there should not be much left to do, I just don‘t see directly what has not yet been addressed I think

1. the renaming in the docs to have `U_X` and `U_p` to address them more clearly
2. finishing the optimisation `retract!`
3. the `project!` deletion was not addressed yet
4. renaming `to_ambient` to `embed!`

I can start a new PR and try to finish 1,3 & 4, I am just not sure about 2, is the final `copyto!` all that‘s left to do?

mateuszbaran:
Sure, I think at this point we may just fork this branch and make a new PR. I'm not quite sure what is left to fix (or improve) here, I haven't taken a closer look at this yet.

kellertuer:
Ok, I will open an new PR maybe on Thursday and try to dress the remarks from above.

kellertuer:
See #396 for a continuation as a branch in this repository.

Issue #10, opened 2019-06-07
Linear algebra for tangent vectors


mateuszbaran:
There are two options to do linear algebra on tangent vectors:

1. Without broadcasting:
```julia
tv1 = 2*tv2 + tv3 - tv4 
```
2. With broadcasting:
```julia
tv1 .= 2.*tv2 .+ tv3 .- tv4 
```

The variant with broadcasting will generally be faster but more troublesome to implement (there will have to be custom broadcast rules as far as I understand this). Do we require both variants to work on all manifolds? I suspect the `MatrixManifold` proposal #8 could be extended with some default broadcasting rules to reduce the effort required to make wrapped tangent vectors that support broadcast arithmetic.

kellertuer:
Very good idea, though I am not sure, which version I favour.

sethaxen:
Is this issue still needed?

mateuszbaran:
I think we could keep it open until the broadcasting/no broadcasting thing is documented. Something like "For the highest compatibility you should use non-broadcasted operations but some manifolds and representations support broadcasting. Refer to documentation of the manifold you are using to check for this".

This is, though, a part of a more general issue: how do we document which manifolds support which operations?

kellertuer:
I was thinking anyways to extend the specific manifold documentaion `.md`s with more details, i.e. a general description and references as well as illustrations for manifolds before delving into the functions. That markdown could be used to also state this property.

We should maybe agree on a generic structure of such a markdown file to have that always in a certain place on these pages?

mateuszbaran:
I think it would be nice if such information about supporting (or not) different features was automatically generated from our test set but it looks like a more long-term project.

kellertuer:
While for examples in the docs there exists `Literate.jl`, I have not really a clue how tests could do that ;) The general description I wanted to do anyways, but right a table of features automatically would be really sweet.

mateuszbaran:
There would have to be a config file that describes what manifolds support what, that is used both by the testset (to specify keyword arguments of `test_manifold`) and in `@eval` blocks in documentation of each manifold to generate a table with support information.

mateuszbaran:
#336 and #339 implemented broadcasting for two tangent vector types so I think it's quite safe to assume it can be used. Probably only `_HyperbolicTangentTypes` need to be fixed?

kellertuer:
Yes they still need broadcasting, but since they just encapsulate an array that should not be too difficult. I will check.

[...]

PR #342, opened 2021-03-26
introduce broadcasting for all _HyperbolicTypes

kellertuer:
by passing on the neccessary functions to the value field. This addresses #10 for the hyperbolic types.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.4.20"
+version = "0.4.21"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -60,8 +60,7 @@ import ManifoldsBase:
     vector_transport_to!,
     zero_tangent_vector,
     zero_tangent_vector!
-import Base:
-    copyto!, convert, in, isapprox, isempty, length, showerror, getindex, setindex!, iterate
+import Base: copyto!, convert, in, isapprox, isempty, length, showerror
 
 using Base.Iterators: repeated
 using Distributions
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -60,7 +60,8 @@ import ManifoldsBase:
     vector_transport_to!,
     zero_tangent_vector,
     zero_tangent_vector!
-import Base: copyto!, convert, in, isapprox, isempty, length, showerror
+import Base:
+    copyto!, convert, in, isapprox, isempty, length, showerror, getindex, setindex!, iterate
 
 using Base.Iterators: repeated
 using Distributions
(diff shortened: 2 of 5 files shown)


kellertuer:
Locally formatter ran just fine. I have to check somewhen, what causes it to even fail (not only report a wrongly formatted code) here.

kellertuer:
@mateuszbaran may I remove tests for Julia 1.6 (or at least ambiguity-tests) until we have reduced/removed the 119 ambiguities?

kellertuer:
...and I ran into a slight problem when writing the tests – currently if we have one of the hyperbolic types, writing `2 .* X` yields and array (with the correctly broadcasted values), but not the type again.

mateuszbaran:
> @mateuszbaran may I remove tests for Julia 1.6 (or at least ambiguity-tests) until we have reduced/removed the 119 ambiguities?

Sure, I suggest not testing for ambiguities on 1.6 at all for now.

> ...and I ran into a slight problem when writing the tests – currently if we have one of the hyperbolic types, writing `2 .* X` yields and array (with the correctly broadcasted values), but not the type again.

I think that can't be fixed without defining custom broadcasting style. Do you want me to fix it?

kellertuer:
Well I tried to see how a broadcast style would have to be done – and I defined it and a `similar` function (cf https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting) just that I did not use an `ArrayStyle` since the hyperbolic types are not abstract arrays – but all I did did not have any effect whatsoever on the broadcast. So sure feel free to illustrate how that is supposed to work

This

````
        Base.BroadcastStyle(::Type{<:$T}) = Broadcast.Style{$T}()
        function Base.similar(bc::Broadcast.Broadcasted{Broadcast.Style{$T}}, ::Type{S}) where {S}
            return $T(simlar(Array{S}, axis(bc)))
        end
````

is how it does _not_ work ;)

mateuszbaran:
OK, a few problems here:
1) `broadcastable` didn't recognize hyperbolic types so it tried collecting it (and that's why you needed `iterate` in the first place). This removed the hyperbolic wrapper so you didn't see it after broadcasting.
2) You didn't have broadcasting mixing rules (so Julia didn't know how to broadcast a hyperbolic vector with a scalar).
3) This style of broadcasting has real trouble preserving the container type (`Array`, `SArray` etc.) so I changed it to something similar to `ProductRepr`.
4) Do you actually want broadcasting for hyperbolic point types?

kellertuer:
1. I don't understand, what's trying tried collecting it here ?
2. Ah, yes, those I missed
3. uff. And you had to introduce so much magic to fix that? While I now understand the point in 2) – lines 149-182 are just indecipherable magic for me. 
4. Sure I want that, if all points/vectors have broadcast that would make the Manopt code nicer (again).

mateuszbaran:
1. You can take a look here: https://github.com/JuliaLang/julia/blob/master/base/broadcast.jl#L685-L690 . `broadcastable` is called on basically every thing you broadcast. Some things are wrapped in `Ref` so they broadcast as scalars, some are untouched (like arrays) and everything else by default is `collect`ed. That is what was happening in your code.
2. :slightly_smiling_face: 
3. I think there is another way to do this but it may turn out even more convoluted than that code. I think someone needs to finally make a cheat sheet for Julia's indexing and broadcasting...
4. OK, I'll leave it then.

kellertuer:
Thanks for the explanation. Do you have a case in mind to cover the last 3 lines, that are new but not yet covered?

mateuszbaran:
Sure, that case should be simple broadcast-assignment like `X2 .= X1` where both are hyperbolic TVectors. I'll add it later, I'm working on something different now.

kellertuer:
Thanks for the explanation and for directly providing it, before I could reach a computer, I had the same idea to easily add that test :)

mateuszbaran (approved):

PR #344, opened 2021-03-31
Make decorator dispatch for mean! be :parent

### LINKED ISSUE #338: distance and mean forward to decorated manifold for DefaultEmbeddingType
#249 added `GeneralLinear` with the following definition:

```julia
struct GeneralLinear{n,𝔽} <:
       AbstractGroupManifold{𝔽,MultiplicationOperation,DefaultEmbeddingType} end

decorated_manifold(::GeneralLinear{n,𝔽}) where {n,𝔽} = Euclidean(n, n; field=𝔽)
```

This signature should result in metric-related methods not forwarding to the decorated manifold. Which is mostly the case. However, `distance` and `mean` both silently forward to `Euclidean`. The former was fixed by a custom definition of `distance`. However, the issue with `mean` also needs to be fixed, and it seems this should be fixed in general.

Is this a ManifoldsBase issue?

sethaxen:
Fixes #338

Still need to add tests.


kellertuer (approved):
Cool, thanks for noticing!

PR #345, opened 2021-04-03
Add unitary and orthogonal groups

sethaxen:
This PR adds `Unitary` as an `AbstractGroupManifold`. `Orthogonal` is just a convenient alias of `Unitary`. Necessarily, most methods of `Rotations` and `SpecialOrthogonal` also apply directly to `Orthogonal`, so it might be worth it to move those methods to the more general `Orthogonal` and have `Rotations` call those methods then. What do you think?

Relates #225


mateuszbaran:
I think moving methods common to `Orthogonal` and `SpecialOrthogonal` to the file that has the implementation of `Orthogonal` is fine.

sethaxen:
It's a bit more of a refactor, but I wonder if it makes more sense to implement `SpecialUnitary` as a manifold transparently isometrically embedded in `Unitary`. Then many methods can be implemented for just `Unitary` and are automatically used by `Unitary`, `Orthogonal`, `SpecialUnitary`, and `SpecialOrthogonal`. And any special-cases implemented for `Orthogonal` are automatically inherited by `SpecialOrthogonal`. Instead of `SpecialOrthogonal` being a `GroupManifold`, it would be an `AbstractGroupManifold`, and we could define `const Rotations{N} = SpecialOrthogonal{N}` as a convenient exported alias.

Another alternative is to implement `Rotations{N,𝔽}` as transparently isometrically embedded in `Unitary{N,𝔽}`, and then `SpecialOrthogonal` and `SpecialUnitary` just decorate `Rotations`. The benefit of this is similarity to what we currently do. The downside is a less clean relationship between manifolds.

src/groups/unitary.jl
```diff
@@ -0,0 +1,220 @@
...
+``p^\mathrm{H}p = pp^\mathrm{H} = I``. All such points satisfy the property ``|\det(p)|=1``.
+
+The tangent vectors ``X_p ∈ T_p \mathrm{U}(n, 𝔽)`` are represented instead as the
+corresponding element ``X_e = p^\mathrm{H} X_p`` of the Lie algebra ``𝔲(n, 𝔽)``, which
+consists of the skew-hermitian matrices, that is, all ``X_e ∈ 𝔽^{n × n}`` where
+``X_e = -X_e^\mathrm{H}``.
+"""
+struct Unitary{n,𝔽} <: AbstractGroupManifold{𝔽,MultiplicationOperation,DefaultEmbeddingType} end
+
+Unitary(n, 𝔽::AbstractNumbers=ℝ) = Unitary{n,𝔽}()
```
sethaxen, line 16: In most cases, I think people will use the term `Orthogonal` to refer to real unitary and `Unitary` with no qualifier to refer to complex unitary, so perhaps it makes sense to use this default:
```suggestion
Unitary(n, 𝔽::AbstractNumbers=ℂ) = Unitary{n,𝔽}()
```

mateuszbaran:
> It's a bit more of a refactor, but I wonder if it makes more sense to implement `SpecialUnitary` as a manifold transparently isometrically embedded in `Unitary`. Then many methods can be implemented for just `Unitary` and are automatically used by `Unitary`, `Orthogonal`, `SpecialUnitary`, and `SpecialOrthogonal`. And any special-cases implemented for `Orthogonal` are automatically inherited by `SpecialOrthogonal`. Instead of `SpecialOrthogonal` being a `GroupManifold`, it would be an `AbstractGroupManifold`, and we could define `const Rotations{N} = SpecialOrthogonal{N}` as a convenient exported alias.
> 
> Another alternative is to implement `Rotations{N,𝔽}` as transparently isometrically embedded in `Unitary{N,𝔽}`, and then `SpecialOrthogonal` and `SpecialUnitary` just decorate `Rotations`. The benefit of this is similarity to what we currently do. The downside is a less clean relationship between manifolds.

Embedding group manifolds into other group manifolds seems quite natural but I'm not sure if we currently have the right propagation rules for group operations. That should be fixable though. So `SpecialOrthogonal` would be embedded in `Orthogonal` which would be embedded in `Unitary`?

kellertuer:
Yes, `SpecialOrthogonal` is - informally speaking – “one half” of the orthogonal matrices (positive determinant, i.e. +1); and yes Orthogonal is the real-valued restriction of Unitary.

I also think the main problem is introducing the transparency rules for group operations, but the rest should be available.

sethaxen:
> So `SpecialOrthogonal` would be embedded in `Orthogonal` which would be embedded in `Unitary`?

We would have `Orthogonal{n} = Unitary{n,ℝ}` as in this PR. The only embedding would be `SpecialUnitary` in `Unitary`.
For that matter, `Unitary` could be isometrically embedded in `GeneralLinear`, but there's no benefit since we override all of its functions.

> Embedding group manifolds into other group manifolds seems quite natural but I'm not sure if we currently have the right propagation rules for group operations. That should be fixable though.

> I also think the main problem is introducing the transparency rules for group operations, but the rest should be available.

Yeah it should not be difficult to add the transparency rules.

mateuszbaran:
> We would have `Orthogonal{n} = Unitary{n,ℝ}` as in this PR. The only embedding would be `SpecialUnitary` in `Unitary`.
> For that matter, `Unitary` could be isometrically embedded in `GeneralLinear`, but there's no benefit since we override all of its functions.

Right, OK, that looks fine :+1: .

sethaxen:
The problem with embedding one group in another using `AbstractGroupManifold` is that the group operation causes all group functions to dispatch to the defaults for `MultiplicationGroup`, so those of the embedding are not seen. But since we already have the notion that we should be able to decorate a group and still get the group behavior, I tried instead to define `SpecialUnitary` as an `AbstractEmbeddedManifold`. There are a few places where we were assuming things be `AbstractGroupManifold` where they shouldn't need to be, and there are some tests I still need to fix, but before I put more time into it, I'm wondering what you think.

mateuszbaran:
I think that seems fine. Let's maybe keep `AbstractGroupManifold` for introducing completely new group structure, and and let's just transparently pass group operations through `AbstractEmbeddedManifold`. That seems to make sense (but it should be documented). @kellertuer , could you also comment on that?

kellertuer:
I must have missed to comment on this one. I would prefer to have transparency on the decorator pattern already, since I think we are misusing the `EmbeddedManifold` here. I am working on that in JuliaManifolds/ManifoldsBase.jl#75, which would make it also easier to introduce more coarse transparency (like transparent for all functions concerning the metric or something like that).

kellertuer:
What is the status here? Can I maybe help to get this finished?

sethaxen:
> What is the status here? Can I maybe help to get this finished?

Hm, I don't remember. I do recall being dissatisfied with this PR. I'd need to go through it again to figure out why though. For now, I can resolve merge conflicts at least.

sethaxen:
Oof, there have been so many merges since then, it might be easier to just start over with a new PR.

kellertuer:
Oh, we changed identity, yes. But here were also very many commits; sure we have to restart this?

kellertuer:
This might need a restart (I could do this), but what was missing when you stopped working on this?

sethaxen:
> This might need a restart (I could do this), but what was missing when you stopped working on this?

Unfortunately, I remember less about this now than I did last year when I wrote:

> > What is the status here? Can I maybe help to get this finished?
> 
> Hm, I don't remember. I do recall being dissatisfied with this PR. I'd need to go through it again to figure out why though. For now, I can resolve merge conflicts at least.

kellertuer:
Ok, I think now it is easier to do this anew (merging might be more time intensive). I can try to start this next (want to check my Grassmannians more first).

PR #347, opened 2021-04-08
Updates to SkewSymmetricMatrices

sethaxen:
This PR makes several non-breaking changes to `SkewSymmetricMatrices` and also reuses its functionality when possible for the tangent vectors of `Rotations`.

### DIFF
--- a/src/manifolds/SkewSymmetric.jl
+++ b/src/manifolds/SkewSymmetric.jl
@@ -208,7 +208,10 @@ where $\cdot^{\mathrm{H}}$ denotes the Hermitian, i.e. complex conjugate transpo
 """
 project(::SkewSymmetricMatrices, ::Any)
 
-project!(M::SkewSymmetricMatrices, q, p) = copyto!(q, (p - p') ./ 2)
+function project!(M::SkewSymmetricMatrices, q, p)
+    q .= (p .- p') ./ 2
+    return q
+end
 
 @doc raw"""
     project(M::SkewSymmetricMatrices, p, X)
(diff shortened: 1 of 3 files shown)


src/manifolds/SkewSymmetric.jl
```diff
@@ -197,20 +180,19 @@ end
...
 ````
 
-where the last $n$ is due to an imaginary diagonal that is allowed $\dim_ℝ 𝔽$ is the [`real_dimension`](@ref) of `𝔽`.
+where ``\dim_ℝ 𝔽`` is the [`real_dimension`](@ref) of ``𝔽``. The first term corresponds to
+only the upper triangular elements of the matrix being unique, and the second term
+corresponds to the constraint that the real part of the diagonal be zero.
 """
 function manifold_dimension(::SkewSymmetricMatrices{N,𝔽}) where {N,𝔽}
-    return div(N * (N - 1), 2) * real_dimension(𝔽) + (𝔽 === ℂ ? N : 0)
+    return div(N * (N + 1), 2) * real_dimension(𝔽) - N
```
kellertuer, line 195: nice reformulation!
I was first thinking, how this captures the complex diagonal adding `N` to the dimension, but sure, it's added via the `+` which adds `2N` in the complex case (since real dimension is 2).

kellertuer (approved):
Thanks for these nice little improvements. For me it looks good so far. Please check for code coverage.

src/manifolds/SkewSymmetric.jl
sethaxen: That's right, the full formula is `(dim(𝔽^{n × n}) - dim(𝔽^n)) / 2 + dim(𝔽^n) - dim(ℝ^n)`, which is generic in that it even covers skew-Hermitian matrices with quaternionic elements.

sethaxen:
All lines previously covered still are.

PR #350, opened 2021-04-18
Introduce copyto for specific points and tangents, and ProductRepr

kellertuer:
This has to wait for JuliaManifolds/ManifoldsBase.jl#65 to be merged
* [x] Eventually manually do transparency for groups and metric manifolds.
* [x] Tests


test/fixed_rank.jl
```diff
@@ -167,6 +174,12 @@ include("utils.jl")
                 oneV = UMVTVector(one(zeros(3, 3)), one(zeros(2, 2)), one(zeros(2, 2)), 2)
                 @test oneV == one(v)
 
+                # copyto
+                w2 = allocate(w)
+                copyto!(M, w2, w)
```
mateuszbaran, line 179: Why does this work on tangent vectors without providing the point they are tangent at?

kellertuer: good observation. This test is wrong and should be `copyto!(M, w2, p, w)` to be precise. The reason it works is the following:

There might be manifolds, when you need information about both the manifold `M` as well as the base point `p` to copy (though I don't have an example in mind, this is the motivation/philosophy for this kind of signature throughout all functions in Manifolds).

But the fallbacks are 
`copyto!(M::Manifold, q, p) = copyto!(q,p)` for points (that is all array based manifolds just use the classical `copyto!`)
and
`copyto!(M::Manifold, Y, p, X) = copyto!(Y,X)`

and even more in the fixed rank manifold case I “just” defined a copyto! for both the `SVDMPoint` and the `UMVTVector`.

These two things make that this here also just works.

mateuszbaran (approved):

kellertuer:
Thanks (I even forgot to un-draft it, since GitHub has now a draftPR feature...).

PR #351, opened 2021-04-25
A getindex for product manifold

kellertuer:
While workingg on JuliaManifolds/Manopt.jl#79 I noticed that an `getindex` is nice when working with manifolds, so similar to accessing `ProductRepr` as `x[M,1]` not accessing the first component of a product manifold `M` is just `M[1]`.

### DIFF
--- a/src/manifolds/ProductManifold.jl
+++ b/src/manifolds/ProductManifold.jl
@@ -26,7 +26,7 @@ end
 
 access the `i`th manifold component from the [`ProductManifold`](@ref) `M`.
 """
-@inline Base.getindex(M::ProductManifold, i) = M.manifolds[i]
+@inline Base.getindex(M::ProductManifold, i::Integer) = M.manifolds[i]
 
 ProductManifold() = throw(MethodError("No method matching ProductManifold()."))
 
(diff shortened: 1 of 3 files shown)


src/manifolds/ProductManifold.jl
```diff
@@ -20,6 +20,14 @@ function ProductManifold(manifolds::Manifold...)
     return ProductManifold{𝔽,typeof(manifolds)}(manifolds)
 end
 
+"""
+    getindex(M::ProductManifold, i)
+    M[i]
+
+access the `i`th manifold component from the [`ProductManifold`](@ref) `M`.
+"""
+@inline Base.getindex(M::ProductManifold, i) = M.manifolds[i]
```
mateuszbaran, line 29:
```suggestion
@inline Base.getindex(M::ProductManifold, i::Integer) = M.manifolds[i]
```
I'd restrict it to integers for now, unless you need indexing with something else? (ranges, colon, etc.)

kellertuer: Sure, sounds reasonable, since product manifolds always have a linear, integer indexed order of their manifolds.

mateuszbaran (approved):

PR #356, opened 2021-05-07
Fix behaviour of christioffel symbol functions and related Jacobians 

kellertuer:
@jlumpe pointed out that

```
using Manifolds
M = MetricManifold(Euclidean(3), EuclideanMetric())
christoffel_symbols_first(M, zeros(3))
```
Throws an error, that `christoffel_symbols_first not implemented on Euclidean(3; field = ℝ)...`.

I was surprised by that and dived a little bit into our dispatch system for Christoffel symbols and their Jacobians.
I think I found the culprit.

In the original code the functions `christoffel_symbols_first`, `christoffel_symbols_second`,  `christoffel_symbols_second_jacobian` and `local_metric_jacobian` where introduced as transparent functions in our decorator system. This meant the following

1. the functions acted completely transparent when called with another decorator (group manifold for example)
2. the original implementation works for non-default metrics
3. the functions acted transparent for default metrics, i.e. falling back to calling `christoffel-symbols_first(M.manifold,...)` for those.

As far as I could see, the default implementation relied on functions that themselves might be transparent or not (`local_metric`, `local_metric_jacobian`,...) but was just a proper implementation for arbitrary metric manifolds anyways. This default implementation did not work for default metrics (as the one above).

So I changed the behaviour to the following: The function is now defined for an arbitrary `MetricManifold`, that is transparent for all others (1. is still valid), but the generic implementation is now called for 2. and 3.
The inner calls are now also working properly (falling back to `local_metric_jacobian` for `Euclidean` when using `MetricManifold(Euclidean(3), EuclideanMetric(3))`.

Now the above example works.

The one point left to discuss is:
The Christoffel functions now still have to be called with `MatricManifold`s, while they would be also valid for `Manifolds` (if one has a default metric in mind). One could try to do something like passing a `::Manifold` based call to the default metric `MetricManifold`. What do you think?

### DIFF
--- a/src/manifolds/MetricManifold.jl
+++ b/src/manifolds/MetricManifold.jl
@@ -4,12 +4,13 @@
 Abstract type for the pseudo-Riemannian metric tensor ``g``, a family of smoothly
 varying inner products on the tangent space. See [`inner`](@ref).
 
-# Constructors
+# Functor
 
-    Metric(M)
+    (metric::Metric)(M::Manifold)
 
-Generate the manifold `M` with the correstponding metric (subtype of this type).
-See also [`MetricManifold`](@ref).
+Generate the `MetricManifold` that wraps the manifold `M` with given `metric`.
+This works for both a variable containing the metric as well as a subtype `T<:Metric`,
+where a zero parameter constructor `T()` is availabe.
 """
 abstract type Metric end
 
@@ -562,7 +563,7 @@ The matrix has the property that ``g(X, Y)=X^\mathrm{T} [g_{ij}] Y = g_{ij} X^i 
 where the latter expression uses Einstein summation convention.
 """
 local_metric(::Manifold, ::Any...)
-@decorator_transparent_signature local_metric(M::Manifold, p; kwargs...)
+@decorator_transparent_signature local_metric(M::AbstractDecoratorManifold, p; kwargs...)
 function decorator_transparent_dispatch(::typeof(local_metric), ::MetricManifold, args...)
     return Val(:parent)
 end
--- a/src/manifolds/MetricManifold.jl
+++ b/src/manifolds/MetricManifold.jl
@@ -70,7 +70,11 @@ function christoffel_symbols_first(
     @einsum Γ[i, j, k] = 1 / 2 * (∂g[k, j, i] + ∂g[i, k, j] - ∂g[i, j, k])
     return Γ
 end
-@decorator_transparent_signature christoffel_symbols_first(M::MD, p; kwargs...) where {MD <:AbstractDecoratorManifold}
+@decorator_transparent_signature christoffel_symbols_first(
+    M::MD,
+    p;
+    kwargs...,
+) where {MD<:AbstractDecoratorManifold}
 
 @doc raw"""
     christoffel_symbols_second(
@@ -100,7 +104,11 @@ function christoffel_symbols_second(
     @einsum Γ₂[l, i, j] = Ginv[k, l] * Γ₁[i, j, k]
     return Γ₂
 end
-@decorator_transparent_signature christoffel_symbols_second(M::MD, p; kwargs...) where {MD <:AbstractDecoratorManifold}
+@decorator_transparent_signature christoffel_symbols_second(
+    M::MD,
+    p;
+    kwargs...,
+) where {MD<:AbstractDecoratorManifold}
 
 @doc raw"""
     christoffel_symbols_second_jacobian(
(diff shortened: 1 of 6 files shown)


kellertuer:
I think there might even be more functions having the same problem, I will check them soon.

mateuszbaran:
Thanks for fixing it :+1: . 

> The one point left to discuss is:
> The Christoffel functions now still have to be called with `MatricManifold`s, while they would be also valid for `Manifolds` (if one has a default metric in mind). One could try to do something like passing a `::Manifold` based call to the default metric `MetricManifold`. What do you think?

I don't remember why we restricted metric-related functions to `MetricManifold`s but they definitely make sense for any manifold. What do you mean by default metric `MetricManifold` here? I don't think I understand your idea.

kellertuer:
Oh I meant the default metric `Manifold` that implicitly acts as a `MetricManifold` with default metric; for example for exp/log.

kellertuer:
The current state does currently not yet work, but my brain also stopped – this decorator stuff is quite challenging.

Missing:
* [ ] make metric functions available for manifolds with their implicit metric
* [ ] check documentation to be complete to this new extension

kellertuer:
While working on this I noticed that most probably all these metric functions lack a basis (that we might have implicitly assumed?), since when you work with the g_ij (elements of the metric matrix G) or the g^ij (its inverse) they are always with respect to a basis and hence the \Gamma_ij^k are, too?

kellertuer:
So. All docs are correct again, I am a little too lazy to add formal to Einstein and curvature tensor, but that would be a nice feature in the future I think.

src/manifolds/MetricManifold.jl
```diff
@@ -1,8 +1,15 @@
...
 
-Abstract type for the pseudo-Riemannian metric tensor $g$, a family of smoothly
+Abstract type for the pseudo-Riemannian metric tensor ``g``, a family of smoothly
 varying inner products on the tangent space. See [`inner`](@ref).
+
+# Constructors
+
+    Metric(M)
+
+Generate the manifold `M` with the correstponding metric (subtype of this type).
```
mateuszbaran, line 11:
```suggestion
Generate the manifold `M` with the corresponding metric (subtype of this type).
```
Do you mean `MetricManifold` by "this type" or something else?

mateuszbaran:
Could you add some test that check that these methods work for, for example, `MetricManifold(Euclidean(3), EuclideanMetric())`?

src/manifolds/MetricManifold.jl
kellertuer: Good that you ask, no, I meant the `Metric`. I wanted to clarify that there is `Metric(M)` for `MetricManifold(M,Metric())` as a constructor.

kellertuer:
> Could you add some test that check that these methods work for, for example, `MetricManifold(Euclidean(3), EuclideanMetric())`?

Which “these methods” do you mean? All of them? (at first glance: christoffel_first, &_jacobian, christoffel_second, &_jacobian, Einstein, Ricci, local metric, inverse local metric, der local metric, inverse det local metric)?

mateuszbaran:
> Which “these methods” do you mean? All of them? (at first glance: christoffel_first, &_jacobian, christoffel_second, &_jacobian, Einstein, Ricci, local metric, inverse local metric, der local metric, inverse det local metric)?

Yes, exactly these, plus `riemann_tensor`.

src/manifolds/MetricManifold.jl
mateuszbaran: Then the documentation is quite unclear, and this isn't exactly a constructor of `Metric` type since it returns a value of type `MetricManifold`. It should be more clearly stated I think. First, I'd add that it required a concrete subtype of `Metric` in the call, a second the description could be a bit more like:
```
    (metric::Metric)(M::Manifold)

Generate the `MetricManifold` that wraps the manifold `M` with given `metric`.
```

mateuszbaran: Or similarly for the other method, `(::Type{T})(M::AbstractManifold) where {T<:AbstractMetric}`.

kellertuer:
Ok, I will check for those in the evening (hopefully).

src/manifolds/MetricManifold.jl
kellertuer: Ah yes, that is a better signature. By the way, just for completeness we could also introduce `(M::Manifold)(m::Metric)` to do the same (which even feels more natural to me).

mateuszbaran: > By the way, just for completeness we could also introduce `(M::Manifold)(m::Metric)` to do the same (which even feels more natural to me).

I don't have any opinions on that but if you like it then sure :slightly_smiling_face: .

kellertuer: Maybe in a later PR to not make this PR too chaotic.

kellertuer:
It was a great idea to write the tests

1) I added `local_metric` and `inverse_local_metric` for `Euclidean` since now we have the proper transparency that passes the default metric (`EuclideanMetric`) down to the manifold itself. That's what we introduced the default-mechanism for.

2) noticed one bug in transparency

What's cool is that with just the local metric and its inverse all other functions worked :)

kellertuer:
I am surprised the codecov patch fails, since overall the code coverage increases. All new functionality (i.e. having the possibility of using all functions with an implicit default metric without the decorator) has tests so this PR is ready to merge.

mateuszbaran:
That coverage failure looks like a false positive.

mateuszbaran (approved):

PR #359, opened 2021-05-16
Interface Documentation rewrite

kellertuer:
This is the companion PR to JuliaManifolds/ManifoldsBase.jl#69 to rewrite the interface documentation.

This first sketch has the one problem that a lot of functions are doubled, but we would have to do separate files for

* Manifold/MPoint/TVector/CoTVector (maybe together with the next)
* all exp/logs
* all retraction related stuff
* all projection related stuff
* maybe also a file just for all remaining interface functions

to not state all functions explicitly (like it is now done partly). But I would like to have some feedback on the structure of the Manifold functions section first, before I continue with that (and the graphics and such).


Affie:
This is great, thanks! It would have saved us hours...
cc @dehann

kellertuer:
thanks for the feedback – and sorry that it took you longer.
While we try to document as much as possible (see also all the formulae at the manifolds themselves) it is sometimes hard to have the right level of detail in mind – so your feedback is valuable here and we then of course try to adapt the docs.

kellertuer:
Since I am rearranging anyways; I am not yet sure how the `ManifoldTest`subpackage is intended to work. I expected the test functions to be included there, but they are not. What is the reason for this? That way the test functions are included in the main scope on Manifolds (since tests is loaded and those are included in this case). With being in the main scope, they are warning for being undocumented. 

Shoul we maybe include the test functions into the Subpackage and only do a `using ManifoldTest` when test is loaded?

mateuszbaran:
I think the intention was to avoid adding a new module through Requires.jl but I don't really know if that's necessary.

kellertuer:
We could keep it as a submodule, but the way it is currently included we can just keep the includes and omit the submodule :)

mateuszbaran:
Hm, that's probably due to those `export`s? Anyway, feel free to change that if you don't like it :slightly_smiling_face: .

kellertuer:
I think I already have an idea.

kellertuer:
Finished the idea and locally (together with the ManifoldsBase PR) the docs run fine and there is no documentation missing; all tests also still run fine.

kellertuer:
I added two figures, one for the projection one for the retraction

https://github.com/JuliaManifolds/Manifolds.jl/blob/7dd0ce6342f9c17af78ce1e4e3ae33515fcfc449/docs/src/assets/images/projection_illustration_600.png

https://github.com/JuliaManifolds/Manifolds.jl/blob/7dd0ce6342f9c17af78ce1e4e3ae33515fcfc449/docs/src/assets/images/retraction_illustration_600.png

They stem from using Manifolds with PGFPLotsX; the only thing I am not yet 100% sure about: The colours are currently chosen to fit both light and dark mode (Paul Tol's colours, but also his gray for the text), which I think is not the best choice for text but I have not much more of an idea.

Affie:
About the project illustration. 

Is `s2` correct? 

Should it not be this one: `project(M::AbstractManifold, p, X)` and be from the ambient space representation of a vector X?

kellertuer:
Thanks for the question.
The `s2` including its whole label is correct. Maybe the name `q` is irritating here.
For `s1` the `q` acts as a point on the manifold R^2 (the embedding) and we project onto the circle.
For `s2` the `q` acts as a vector in the embedding that gets projected onto the tangent plane. So in this case `q` _is_ said `X` in the function `project(M,p,X)`.

kellertuer:
This is now updated and just has to wait for the ManifoldsBase branch.

kellertuer:
I am not sure why documenter is failing here, I updated the `Project.toml` but it seems to be fixed to not be above 0.11.1 somehow (and should be 0.11.3) and hence fails. But there is no occurrence of 0.11.1 in the project, so I don't know why it fails.

mateuszbaran:
The currently failing documenter job picks Manifolds.jl 0.5.0 but the ManifoldsBase.jl version is indeed 0.11.1 and 0.11.3 isn't visible. I've noticed before that it takes some time (even more than a day) before documenter jobs see a newly registered version of a package.

docs/src/misc/notation.md
```diff
@@ -38,6 +39,7 @@ Within the documented functions, the utf8 symbols are used whenever possible, as
 | ``\operatorname{Log}`` | the matrix logarithm | |
 | ``\mathcal P_{q\gets p}X`` | parallel transport | | of the vector ``X`` from ``T_p\mathcal M`` to ``T_q\mathcal M``
 | ``\mathcal P_{p,Y}X`` | parallel transport in direction ``Y`` | | of the vector ``X`` from ``T_p\mathcal M`` to ``T_q\mathcal M``, ``q = exp_pY``
```
mateuszbaran, line 41:
```suggestion
| ``\mathcal P_{p,Y}X`` | parallel transport in direction ``Y`` | | of the vector ``X`` from ``T_p\mathcal M`` to ``T_q\mathcal M``, ``q = \exp_pY``
```
A backslash is missing there.

kellertuer:
Thanks for the help. I can reproduce it here locally activating the docs environment, so we might have to just wait (and it is either ManifoldsBase or even Manifolds, I am not 100% sure.

docs/src/interface.md
```diff
@@ -11,11 +11,109 @@ Depth = 2
...
+
+1. Given a point ``p∈\mathcal N`` we can look for the closest point on the manifold ``\mathcal M`` formally as
+
+```math
+  \operatorname*{arg\,min}_{q\in \mathcal M} d_{\mathcal N}(i(q),p)
+```
+
+And this resulting ``q`` we call the projection of ``p`` onto the manifold ``\mathcal M``.
+
+2. Given a point ``p∈\mathcal M`` and a vector in ``X\inT_{i(p)}\mathcal N`` in the embedding we can similarly look for the closest point to ``Y∈ T_p\mathcal M`` using the push forward ``i_*`` of the embedding.
```
mateuszbaran, line 91:
```suggestion
2. Given a point ``p∈\mathcal M`` and a vector in ``X\inT_{i(p)}\mathcal N`` in the embedding we can similarly look for the closest point to ``Y∈ T_p\mathcal M`` using the pushforward ``i_*`` of the embedding.
```
We currently have a different notation for pushforward: https://juliamanifolds.github.io/Manifolds.jl/stable/misc/notation.html .

docs/src/interface.md
```diff
@@ -11,11 +11,109 @@ Depth = 2
...
+To distinguish different types of retractions, the last argument of the (inverse) retraction
+specifies a type. The following ones are available.
+
+```@autodocs
+Modules = [Manifolds, ManifoldsBase]
+Pages = ["retractions.jl"]
+Order = [:type]
+```
+
+### Projections
```
mateuszbaran, line 71: It may also make sense to illustrate projections on Rotations (even 2D ones) since it's noticeably different.

kellertuer: Can you elaborate a little on this? At first glance I don‘t see how that‘s different – and also I would not want to clutter this too much. It is illustrating one difference (project tangent and project point), so maybe the rotation projection could also be done over on the rotations tutorial or on the rotations manifold page?

docs/src/interface.md
kellertuer: Oh, indeed, I will change those, I somehow missed the notation difference (and recently used this one).

docs/src/interface.md
mateuszbaran: Yes, maybe that's not the best place to explain fine details. The reason I wrote about Rotations is that tangent vectors to rotations are represented in the Lie algebra at identity. Currently rotation to that tangent space (i.e. representation change) is not performed, so I don't think tangent vector projection for rotations is the pushforward of point projections.

kellertuer: oh, if so, a change to the Lie group presentation should be performed, yes (if that has a different representation) but sure there it is a little different, also the parallel transport is (due to the representation, and just the representation) the identity.

mateuszbaran: The issue is already tracked in https://github.com/JuliaManifolds/Manifolds.jl/issues/360 so we can leave it for another PR.

kellertuer: Perfect, cool.

mateuszbaran (approved):
I think this looks very nice now :+1: .

PR #367, opened 2021-05-26
Docs for `NestedReplacingPowerRepresentation`

mateuszbaran:
Docs for the representation introduced in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/73 .


docs/src/manifolds/power.md
```diff
@@ -73,6 +85,27 @@ set_component!(M, p, [0.0, 0.0, 1.0], 4)
 get_component(M, p, 4)
 ```
 
+### `NestedReplacingPowerRepresentation`
+
+The final representation is the [`NestedReplacingPowerRepresentation`](@ref). It is similar to the `NestedPowerRepresentation` but it does not perform mutating operations on the points on the underlying manifold. The example below uses this representation to store points on a power manifold of the [`SpecialEuclidean`](@ref) group in-line in an `Vector` for improved efficiency. Points on this manifold are not represented as arrays, so `ArrayPowerRepresentation` can't be used. Also, `NestedPowerRepresentation` would prevent in-line storage.
```
kellertuer, line 90: I first of all added a few `@ref`s, but also for the last two sentences: Would it maybe more sense to phrase it positively? Now they state, why others __can not__ (or at least should not) be used. Would it maybe be better to mention why this type then _should__ be used?
```suggestion
The final representation is the [`NestedReplacingPowerRepresentation`](@ref). It is similar to the [`NestedPowerRepresentation`](@ref) but it does not perform mutating operations on the points on the underlying manifold. The example below uses this representation to store points on a power manifold of the [`SpecialEuclidean`](@ref) group in-line in an `Vector` for improved efficiency. Points on this manifold are not represented as arrays, so [`ArrayPowerRepresentation`](@ref) can't be used. Also, [`NestedPowerRepresentation`](@ref) would prevent in-line storage.
```

kellertuer (commented):
Just one comment on how to phrase the example.

docs/src/manifolds/power.md
mateuszbaran: These last two sentences actually just re-iterate points made elsewhere. If you don't like this negative phrasing, I could just delete them.

mateuszbaran: I've changed it to more or less your text, except I skipped the reference to `ArrayPowerRepresentation` which may be a bit confusing without further explanation, and removed the "accessing" word since accessing is usually non-mutating :slightly_smiling_face: .

kellertuer:
I did not bump the version for my update, but would it maybe reasonable to bump now such that both updated docs are a patch update to the repo?

kellertuer (approved):

mateuszbaran:
OK, I'll bump the version.

kellertuer (approved):

PR #368, opened 2021-05-27
Deprecate ProductArray

mateuszbaran:
As discussed in https://github.com/JuliaManifolds/Manifolds.jl/pull/366#discussion_r640132557 .


kellertuer (approved):
Looks good to me, though I was surprised that there is no `@deprecate` for types.

kellertuer:
Something in the packages seems to have changes, see also the latest merge/commit on master. Suddenly 5 tests are failing and I haven't understood why.

kellertuer:
Maybe it did, yes.

PR #369, opened 2021-05-27
Update projection illustration to correct image

Affie:


kellertuer:
Oops, copy paste error – thanks for noticing!

PR #371, opened 2021-05-27
rename `get_point_coordinates` to `get_parameters`

### LINKED ISSUE #370: Rename `get_point_coordinates` to `get_parameters`
I was not so happy with the name `get_point_coordinates` back in #325, and while looking for hat/vee the last few minutes, I remembered one terminology that might fix/help here.

If we have a chart (U,φ) then the inverse of the chart is also called a parametrisation. For example do Carmo, Riemannian Manifolds, Chapter 0, Section 2, even defines Manifolds using an atlas of parametrisation.
So if we have a point `p` on the manifold and its image `x=φ(p)` in R^n it can be seen as parameters to this point (and the inverse, the parametrisation brings back the point `p=φ^-1(x)`, where `x` are the coordinates in the chart `φ` or in other words the parameters `x` in the parametrisation `φ^-1`.

So, I feel `get_parameters` would be a very nice name. I would extend the documentation accordingly if this change is accepted (and maybe so early on this can - with a slight misuse) still be seen as non breaking?

kellertuer:
I renamed the function according to our discussion (closes #370).
I also added a few more details to the docs concerning atlasess.
Especially, since I had to think about that for a moment myself, I added how charts and parametrizations look like for the `RetractionAtlas`.

I noticed a small caveat, that I currently did not add, since it might really be just a technical remark.
In the `transition_map`, if one chooses `A_from` and `A_to` such that their union is not an atlas, i.e. they are not compatible, then they represent a different structure and in the new parametrisation – informally said – smoothness on `M` might be different.

### DIFF
--- a/docs/src/features/atlases.md
+++ b/docs/src/features/atlases.md
@@ -1,17 +1,19 @@
 # [Atlases and charts](@id atlases_and_charts)
 
-Atlases on an ``n``-dimensional manifold $\mathcal M$ are collections of charts ``\mathcal A = \{(U_i, \varphi_i) \colon i \in I\}``, where ``I`` is a (finite or infinte) index family, such that ``U_i \subseteq \mathcal M`` is an open set and each chart ``\varphi_i: U_i \to \mathbb{R}^n`` is a homeomorphism. This means, that ``\varphi_i`` is bijecive – sometimes also called one-to-one and onto - and continuous, and its inverse ``\varphi_i^{-1}`` is continuous as well.
-The inverse ``\varphi_i^{-1}`` is called (local) parametrization.
+Atlases on an ``n``-dimensional manifold $\mathcal M$ are collections of charts ``\mathcal A = \{(U_i, φ_i) \colon i \in I\}``, where ``I`` is a (finite or infinte) index family, such that ``U_i \subseteq \mathcal M`` is an open set and each chart ``φ_i: U_i \to \mathbb{R}^n`` is a homeomorphism. This means, that ``φ_i`` is bijecive – sometimes also called one-to-one and onto - and continuous, and its inverse ``φ_i^{-1}`` is continuous as well.
+The inverse ``φ_i^{-1}`` is called (local) parametrization.
+The resulting _parameters_ ``a=φ(p)`` of ``p`` (with respect to the chart ``φ`` are in the literature also called “(local) coordinates”. To distinguish the parameter ``a`` from  [`get_coordinates`](@ref) in a basis, we use the terminology parameter in this package.
+
 For an atlas ``\mathcal A`` we further require that
 
 ```math
 \displaystyle\bigcup_{i\in I} U_i = \mathcal M.
 ```
 
-We say that ``\varphi_i`` is a chart about ``p``, if ``p\in U_i``.
+We say that ``φ_i`` is a chart about ``p``, if ``p\in U_i``.
 An atlas provides a connection between a manifold and the Euclidean space ``\mathbb{R}^n``, since
 locally, a chart about ``p`` can be used to identify its neighborhood (as long as you stay in ``U_i``) with a subset of a Euclidean space.
-Most manifolds we consider are smooth, i.e. any change of charts ``\varphi_i\circ\varphi_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, where ``i,j\in I``, is a smooth function. These changes of charts are also called transition maps.
+Most manifolds we consider are smooth, i.e. any change of charts ``φ_i \circ φ_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, where ``i,j\in I``, is a smooth function. These changes of charts are also called transition maps.
 
 Most operations on manifolds in `Manifolds.jl` avoid operating in a chart through appropriate embeddings and formulas derived for particular manifolds, though atlases provide the most general way of working with manifolds.
 Compared to these approaches, using an atlas is often more technical and time-consuming.
@@ -23,12 +25,12 @@ There are no type restrictions for indices of charts in atlases.
 Operations using atlases and charts are available through the following functions:
 
 * [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point ``p``. This function should work deterministically, i.e. for a fixed ``p`` always return the same chart.
-* [`get_parameters`](@ref Main.Manifolds.get_parameters) converts a point to its local coordinates, also called parameters with respect to the chart in a chart.
+* [`get_parameters`](@ref Main.Manifolds.get_parameters) converts a point to its parameters with respect to the chart in a chart.
 * [`get_point`](@ref Main.Manifolds.get_point) converts parameters (local coordinates) in a chart to the point that corresponds to them.
-* [`induced_basis`](@ref Main.Manifolds.induced_basis) returns a basis of a given vector space at a point induced by a chart ``\varphi``.
-* [`transition_map`](@ref Main.Manifolds.transition_map) converts coordinates of a point between two charts, e.g. computes ``\varphi_i\circ\varphi_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, ``i,j\in I``.
+* [`induced_basis`](@ref Main.Manifolds.induced_basis) returns a basis of a given vector space at a point induced by a chart ``φ``.
+* [`transition_map`](@ref Main.Manifolds.transition_map) converts coordinates of a point between two charts, e.g. computes ``φ_i\circ φ_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, ``i,j\in I``.
 
-While an atlas could store charts as explicit functions, it is favourable, that the [`get_parameters`] actually implements a chart ``\varphi``, [`get_point`](@ref) its inverse, the prametrization ``\varphi^{-1}``.
+While an atlas could store charts as explicit functions, it is favourable, that the [`get_parameters`] actually implements a chart ``φ``, [`get_point`](@ref) its inverse, the prametrization ``φ^{-1}``.
 
 ```@autodocs
 Modules = [Manifolds,ManifoldsBase]
--- a/docs/src/features/atlases.md
+++ b/docs/src/features/atlases.md
@@ -8,13 +8,13 @@ For an atlas ``\mathcal A`` we further require that
 \displaystyle\bigcup_{i\in I} U_i = \mathcal M.
 ```
 
-We say ``\varphi_i`` is a chart about ``p``, if ``p\in U_i``
+We say that ``\varphi_i`` is a chart about ``p``, if ``p\in U_i``.
 An atlas provides a connection between a manifold and the Euclidean space ``\mathbb{R}^n``, since
-locally, a chart about ``p`` can be used to identify its neighborhood (as long as you sta in ``U_i``) “looks like” the Euclidean space.
-Most manifolds we consider are smooth, i.e. any change of charts ``\varphi_i\circ\varphi_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, ``i,j\in I``, is a smooth function. These change of charts are also called transition maps
+locally, a chart about ``p`` can be used to identify its neighborhood (as long as you stay in ``U_i``) with a subset of a Euclidean space.
+Most manifolds we consider are smooth, i.e. any change of charts ``\varphi_i\circ\varphi_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, where ``i,j\in I``, is a smooth function. These changes of charts are also called transition maps.
 
 Most operations on manifolds in `Manifolds.jl` avoid operating in a chart through appropriate embeddings and formulas derived for particular manifolds, though atlases provide the most general way of working with manifolds.
-Compared to these approaches, using an atlas ist often more technical and time-consuming.
+Compared to these approaches, using an atlas is often more technical and time-consuming.
 They are extensively used in metric-related functions on [`MetricManifold`](@ref Main.Manifolds.MetricManifold)s.
 
 Atlases are represented by objects of subtypes of [`AbstractAtlas`](@ref Main.Manifolds.AbstractAtlas).
@@ -22,7 +22,7 @@ There are no type restrictions for indices of charts in atlases.
 
 Operations using atlases and charts are available through the following functions:
 
-* [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point ``p``. This function shoudl work deterministic, i.e. for a fixed ``p`` always return the same chart.
+* [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point ``p``. This function should work deterministically, i.e. for a fixed ``p`` always return the same chart.
 * [`get_parameters`](@ref Main.Manifolds.get_parameters) converts a point to its local coordinates, also called parameters with respect to the chart in a chart.
 * [`get_point`](@ref Main.Manifolds.get_point) converts parameters (local coordinates) in a chart to the point that corresponds to them.
 * [`induced_basis`](@ref Main.Manifolds.induced_basis) returns a basis of a given vector space at a point induced by a chart ``\varphi``, by taking the derivative of the coordinate functions ``\varphi^k``, ``k=1,\ldots,n``.
--- a/docs/src/features/atlases.md
+++ b/docs/src/features/atlases.md
@@ -25,7 +25,7 @@ Operations using atlases and charts are available through the following function
 * [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point ``p``. This function should work deterministically, i.e. for a fixed ``p`` always return the same chart.
 * [`get_parameters`](@ref Main.Manifolds.get_parameters) converts a point to its local coordinates, also called parameters with respect to the chart in a chart.
 * [`get_point`](@ref Main.Manifolds.get_point) converts parameters (local coordinates) in a chart to the point that corresponds to them.
-* [`induced_basis`](@ref Main.Manifolds.induced_basis) returns a basis of a given vector space at a point induced by a chart ``\varphi``, by taking the derivative of the coordinate functions ``\varphi^k``, ``k=1,\ldots,n``.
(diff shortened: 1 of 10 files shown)


mateuszbaran:
> I noticed a small caveat, that I currently did not add, since it might really be just a technical remark.
> In the `transition_map`, if one chooses `A_from` and `A_to` such that their union is not an atlas, i.e. they are not compatible, then they represent a different structure and in the new parametrisation – informally said – smoothness on `M` might be different.

I'd assume that each manifold has a default smooth structure, as given by `default_atlas`. All other atlases should be compatible with it but if someone has a good reason to violate this rule, then I think it's OK.

kellertuer:
Yes, I am sure, you had that in mind and that is also completely fine. It just seems the only thing that is currently not mentioned in the docs, but I also did not feel like adding that details, since it is very technical.

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
-They provide a basic connection between manifolds and Euclidean spaces.
+Atlases on an ``n``-dimensional manifold $\mathcal M$ are collections of charts ``\mathcal A = \{(U_i, \varphi_i) \colon i \in I\}``, where ``I`` is a (finite or infinte) index family, such that ``U_i \subseteq \mathcal M`` is an open set and each chart ``\varphi_i: U_i \to \mathbb{R}^n`` is a homeomorphism. This means, that ``\varphi_i`` is bijecive – sometimes also called one-to-one and onto - and continuous, and its inverse ``\varphi_i^{-1}`` is continuous as well.
+The inverse ``\varphi_i^{-1}`` is called (local) parametrization.
+For an atlas ``\mathcal A`` we further require that
+
+```math
+\displaystyle\bigcup_{i\in I} U_i = \mathcal M.
+```
+
+We say ``\varphi_i`` is a chart about ``p``, if ``p\in U_i``
```
mateuszbaran, line 11:
```suggestion
We say that ``\varphi_i`` is a chart about ``p``, if ``p\in U_i``.
```

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
+An atlas provides a connection between a manifold and the Euclidean space ``\mathbb{R}^n``, since
+locally, a chart about ``p`` can be used to identify its neighborhood (as long as you sta in ``U_i``) “looks like” the Euclidean space.
```
mateuszbaran, line 13:
```suggestion
locally, a chart about ``p`` can be used to identify its neighborhood (as long as you stay in ``U_i``) with a subset of a Euclidean space.
```
Feel free to rephrase it, I just didn't like the original version.

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
+Most manifolds we consider are smooth, i.e. any change of charts ``\varphi_i\circ\varphi_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, ``i,j\in I``, is a smooth function. These change of charts are also called transition maps
```
mateuszbaran, line 14:
```suggestion
Most manifolds we consider are smooth, i.e. any change of charts ``\varphi_i\circ\varphi_j^{-1}: \mathbb{R}^n\to\mathbb{R}^n``, where ``i,j\in I``, is a smooth function. These changes of charts are also called transition maps.
```

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
 
 Most operations on manifolds in `Manifolds.jl` avoid operating in a chart through appropriate embeddings and formulas derived for particular manifolds, though atlases provide the most general way of working with manifolds.
+Compared to these approaches, using an atlas ist often more technical and time-consuming.
```
mateuszbaran, line 17:
```suggestion
Compared to these approaches, using an atlas is often more technical and time-consuming.
```

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
 There are no type restrictions for indices of charts in atlases.
 
 Operations using atlases and charts are available through the following functions:
 
-* [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point.
-* [`get_point_coordinates`](@ref Main.Manifolds.get_point_coordinates) converts a point to its coordinates in a chart.
-* [`get_point`](@ref Main.Manifolds.get_point) converts coordinates in a chart to the point that corresponds to them.
-* [`induced_basis`](@ref Main.Manifolds.induced_basis) returns a basis of a given vector space at a point induced by a chart.
-* [`transition_map`](@ref Main.Manifolds.transition_map) converts coordinates of a point between two charts.
+* [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point ``p``. This function shoudl work deterministic, i.e. for a fixed ``p`` always return the same chart.
```
mateuszbaran, line 25:
```suggestion
* [`get_chart_index`](@ref Main.Manifolds.get_chart_index) can be used to select an appropriate chart for the neighborhood of a given point ``p``. This function should work deterministically, i.e. for a fixed ``p`` always return the same chart.
```

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
+* [`get_parameters`](@ref Main.Manifolds.get_parameters) converts a point to its local coordinates, also called parameters with respect to the chart in a chart.
```
mateuszbaran, line 26: Should we keep using both words (coordinates and parameters) to refer to parameters in a chart? I'd prefer sticking with one term, i.e.  "parameters". If you want to mention coordinates here, could you phrase it more like 'some authors refer to parameters as "local coordinates"' ?

docs/src/features/atlases.md
```diff
@@ -1,21 +1,34 @@
...
+* [`get_point`](@ref Main.Manifolds.get_point) converts parameters (local coordinates) in a chart to the point that corresponds to them.
+* [`induced_basis`](@ref Main.Manifolds.induced_basis) returns a basis of a given vector space at a point induced by a chart ``\varphi``, by taking the derivative of the coordinate functions ``\varphi^k``, ``k=1,\ldots,n``.
```
mateuszbaran, line 28: What do you think about phrasing a bit more formally? Wikipedia defines it nicely for tangent spaces: https://en.wikipedia.org/wiki/Tangent_space#Definition_via_tangent_curves , and "taking the derivative of the coordinate functions" can be a bit confusing.

src/atlases.jl
```diff
@@ -8,16 +8,23 @@ object.
...
         TRetr<:AbstractRetractionMethod,
         TInvRetr<:AbstractInverseRetractionMethod,
         TBasis<:AbstractBasis,
     } <: AbstractAtlas{𝔽}
 
-An atlas indexed by points on a manifold, such that coordinate transformations are performed
-using retractions, inverse retractions, and coordinate calculation for a given basis.
+An atlas indexed by points on a manifold, ``\mathcal M = I`` and parameters (local coordinates)
+are given in ``T_p\mathcal M``.
+This means, that a chart ``\varphi_p = \mathrm{cord}\circ\mathrm{retr}_p^{-1}`` is only locally
```
mateuszbaran, line 21:
```suggestion
This means that a chart ``\varphi_p = \mathrm{cord}\circ\mathrm{retr}_p^{-1}`` is only locally
```

src/atlases.jl
```diff
@@ -57,46 +64,51 @@ function get_default_atlas(::AbstractManifold)
...
-"""
-    get_point_coordinates(M::AbstractManifold, A::AbstractAtlas, i, p)
+@doc raw"""
+    get_parameters(M::AbstractManifold, A::AbstractAtlas, i, p)
 
-Calculate coordinates of point `p` on manifold `M` in chart from an [`AbstractAtlas`](@ref)
-`A` at index `i`. Coordinates are in the number system determined by `A`.
+Calculate parameters (local coordinates) of point `p` on manifold `M` in chart from an [`AbstractAtlas`](@ref)
+`A` at index `i`.
+This function is hence animplementation chart ``\varphi_i(p), i\in I``.
```
mateuszbaran, line 72:
```suggestion
This function is hence an implementation of the chart ``\varphi_i(p), i\in I``.
```

src/atlases.jl
```diff
@@ -57,46 +64,51 @@ function get_default_atlas(::AbstractManifold)
...
+The parameters are in the number system determined by `A`.
+If the point ``p\notin U_i`` is not in the domain of the chart, this method shoudl throw an error.
```
mateuszbaran, line 74:
```suggestion
If the point ``p\notin U_i`` is not in the domain of the chart, this method should throw an error.
```

src/atlases.jl
```diff
@@ -134,17 +146,29 @@ get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any)
...
+@doc raw"""
     transition_map(M::AbstractManifold, A_from::AbstractAtlas, i_from, A_to::AbstractAtlas, i_to, a)
     transition_map(M::AbstractManifold, A::AbstractAtlas, i_from, i_to, a)
 
 Given coordinates `a` in chart `(A_from, i_from)` of a point on manifold `M`, returns
 coordinates of that point in chart `(A_to, i_to)`. If `A_from` and `A_to` are equal, `A_to`
 can be omitted.
 
+Mathematically this function is the transition map or change of charts, but it
+might even be between to atlasses ``A_{\text{from}} = \{(U_i,\varphi_i)\}_{i\in I} `` and ``A_{\text{to}} = \{(V_j,\psi_j)\}_{j\in J}``,
```
mateuszbaran, line 158:
```suggestion
might even be between two atlases ``A_{\text{from}} = \{(U_i,\varphi_i)\}_{i\in I} `` and ``A_{\text{to}} = \{(V_j,\psi_j)\}_{j\in J}``,
```

src/atlases.jl
```diff
@@ -134,17 +146,29 @@ get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any)
...
+and hence ``I, J`` are their index sets.
+We have ``i_{\text{from}}\in I``, ``i_{\text{to}}\in J`` and this method computes
+
+This method then computes
```
mateuszbaran, line 162: You have "this method computes" twice here.

src/atlases.jl
```diff
@@ -134,17 +146,29 @@ get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any)
...
+```math
+\bigl(\psi{i_{\text{to}}}\circ\varphi_{i_{\text{from}}}^{-1}\bigr)(a)
+```
+
+Note that, similar to [`get_parameters`](@ref) this method should fail the same way if ``V_{i_{\text{to}}}\cap U_{i_{\text{from}}}=\emptyset``.
```
mateuszbaran, line 167:
```suggestion
Note that, similarly to [`get_parameters`](@ref), this method should fail the same way if ``V_{i_{\text{to}}}\cap U_{i_{\text{from}}}=\emptyset``.
```

src/atlases.jl
```diff
@@ -216,11 +240,13 @@ function induced_basis(
...
 end
 
-"""
+@doc raw"""
     InducedBasis(vs::VectorSpaceType, A::AbstractAtlas, i)
 
 The basis induced by chart with index `i` from an [`AbstractAtlas`](@ref) `A` of vector
-space of type `vs`.
+space of type `vs`, i.e. if you write the chart ``\varphi_i = (\varphi_i^1,...,\varphi_n^1)^{\mathrm{T}}``.
+as its coordinate functions, then the basis consists of the partial derivatives ``\Bigl(\frac{\partial}{\partial\varphi_i^j}\Bigr)_p``.
```
mateuszbaran, line 248: I think this might be a bit too informal (see also my comment to the `.md` file).

docs/src/features/atlases.md
kellertuer: Your suggestion is fine, my text was too informal, I agree.

docs/src/features/atlases.md
kellertuer: Currently I used parameters (and local coordinates in brackets), but maybe we should only do that here and just once in the function docs. So sure, we can do as you propose.

In general, when writing this, I felt quite good with the clear distinction parameters vs coordinates, especially in the RetractionAtlas docs.

docs/src/features/atlases.md
kellertuer: Sure, I wasn't sure how much to go into detail, especially since one has to also mention the equivalence of curves and such.

kellertuer: I would maybe remove this one, since it only is an overview.

src/atlases.jl
kellertuer: I tried for a while to come up with an idea – I could write a few pages, but that would be too long for a doctoring. Can you provide an idea what you would write?

mateuszbaran: I'll try to think of something. Essentially, ``\Bigl(\frac{\partial}{\partial\varphi_i^j}\Bigr)_p`` IMO doesn't explain too much without  some additional definitions and an explanation why we can define a basis this way -- I'm not even totally sure myself about that notation.

kellertuer: I wrote a more extended explanation.

src/atlases.jl
```diff
@@ -245,10 +245,23 @@ end
...
+For the `vs` a `TangentSpace` this works as  follows:
+
+Let ``n`` denote the dimension of the manifold ``\mathcal M``.
+
+Let the parameter ``a=φ_i(p) ∈ \mathbb R^n`` and ``j∈\{1,…,n\}``.
+We can look at the ``j``th parameter curve ``b_j(t) = a + te_j``, where ``e_j`` denotes the ``j``th unit vector.
+Using the parametrisation we obtain a curve ``c_i(t) = φ^{-1}(b_j(t))`` which fulfills ``c(0) = p``.
+
+Now taking the derivative(s) with respect to ``t`` (and evaluate at ``t=0``),
+we obtain a tangent vector for each `corresponding to an equivalence class of curves (having the same derivative) as
```
mateuszbaran, line 258:
```suggestion
we obtain a tangent vector for each ``j`` corresponding to an equivalence class of curves (having the same derivative) as
```

src/atlases.jl
mateuszbaran: Thanks, this looks very nice now :+1:

src/atlases.jl
```diff
@@ -216,12 +240,29 @@ function induced_basis(
...
 The basis induced by chart with index `i` from an [`AbstractAtlas`](@ref) `A` of vector
 space of type `vs`.
 
+For the `vs` a `TangentSpace` this works as  follows:
+
+Let ``n`` denote the dimension of the manifold ``\mathcal M``.
+
+Let the parameter ``a=φ_i(p) ∈ \mathbb R^n`` and ``j∈\{1,…,n\}``.
+We can look at the ``j``th parameter curve ``b_j(t) = a + te_j``, where ``e_j`` denotes the ``j``th unit vector.
+Using the parametrisation we obtain a curve ``c_i(t) = φ^{-1}(b_j(t))`` which fulfills ``c(0) = p``.
```
mateuszbaran, line 255:
```suggestion
Using the parametrisation we obtain a curve ``c_i(t) = φ_i^{-1}(b_j(t))`` which fulfills ``c(0) = p``.
```

src/atlases.jl
```diff
@@ -216,12 +240,29 @@ function induced_basis(
...
 end
 
-"""
+@doc raw"""
     InducedBasis(vs::VectorSpaceType, A::AbstractAtlas, i)
 
 The basis induced by chart with index `i` from an [`AbstractAtlas`](@ref) `A` of vector
 space of type `vs`.
 
+For the `vs` a `TangentSpace` this works as  follows:
```
mateuszbaran, line 249:
```suggestion
For the `vs` a [`TangentSpace`](@ref) this works as  follows:
```

src/atlases.jl
```diff
@@ -134,17 +146,29 @@ get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any)
...
 can be omitted.
 
+Mathematically this function is the transition map or change of charts, but it
+might even be between two atlases ``A_{\text{from}} = \{(U_i,φ_i)\}_{i\in I} `` and ``A_{\text{to}} = \{(V_j,\psi_j)\}_{j\in J}``,
+and hence ``I, J`` are their index sets.
+We have ``i_{\text{from}}\in I``, ``i_{\text{to}}\in J``.
+
+This method then computes
+```math
+\bigl(\psi{i_{\text{to}}}\circ φ_{i_{\text{from}}}^{-1}\bigr)(a)
```
mateuszbaran, line 164:
```suggestion
\bigl(\psi_{i_{\text{to}}}\circ φ_{i_{\text{from}}}^{-1}\bigr)(a)
```

src/atlases.jl
kellertuer: Thanks for carefully checking, at some point I noticed I used `I` twice and I must have missed this occurence.

docs/src/features/atlases.md
```diff
@@ -1,21 +1,36 @@
 # [Atlases and charts](@id atlases_and_charts)
 
-Atlases on an $n$-dimensional manifold $\mathcal M$ are collections of charts $\{(U_i, \varphi_i) \colon i \in I\}$ such that $U_i \subseteq \mathcal M$ and each chart $\varphi_i$ is a function from $U_i$ to $\mathbb{R}^n$.
-They provide a basic connection between manifolds and Euclidean spaces.
+Atlases on an ``n``-dimensional manifold $\mathcal M$ are collections of charts ``\mathcal A = \{(U_i, φ_i) \colon i \in I\}``, where ``I`` is a (finite or infinte) index family, such that ``U_i \subseteq \mathcal M`` is an open set and each chart ``φ_i: U_i \to \mathbb{R}^n`` is a homeomorphism. This means, that ``φ_i`` is bijecive – sometimes also called one-to-one and onto - and continuous, and its inverse ``φ_i^{-1}`` is continuous as well.
+The inverse ``φ_i^{-1}`` is called (local) parametrization.
+The resulting _parameters_ ``a=φ(p)`` of ``p`` (with respect to the chart ``φ`` are in the literature also called “(local) coordinates”. To distinguish the parameter ``a`` from  [`get_coordinates`](@ref) in a basis, we use the terminology parameter in this package.
```
mateuszbaran, line 5:
```suggestion
The resulting _parameters_ ``a=φ(p)`` of ``p`` (with respect to the chart ``φ``) are in the literature also called “(local) coordinates”. To distinguish the parameter ``a`` from  [`get_coordinates`](@ref) in a basis, we use the terminology parameter in this package.
```

mateuszbaran (approved):

PR #372, opened 2021-06-01
Ensure eigenvalues are positive in SPD matrices

TimQuelch:
By definition symmetric positive definite matrices have positive eigenvalues. Numerical errors may result in infinitesimal negative eigenvalues being computed, resulting in DomainErrors when the sqrt or log of the eigenvalue is calculated.

This change replaces any negative eigenvalues with the smallest positive value to ensure that the sqrt is successful.

This makes it the user's responsibility to ensure that input arguments are valid SPDs before
passing.

This is similar logic to https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/68

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.5.2"
+version = "0.5.3"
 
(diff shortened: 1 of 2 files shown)


TimQuelch:
Does it make more sense to use `floatmin` or `eps` here?

mateuszbaran:
Good idea. I think `floatmin` makes more sense since eigendecomposition can easily return positive eigenvalues smaller than `eps`.

kellertuer:
I think this is a good idea, though it might lead to unexpected results, when the matrix is really not SPD; however this disadvantage might outweigh the positive effect for the user using these methods correctly and just meeting numerical precision problems. Thanks for proposing this change.

kellertuer (approved):
Could you just bump the version as a final step?

TimQuelch:
Done. I bumped the version and fixed some formatting + wording in the commit message

PR #375, opened 2021-06-04
Introduce an ApproximateLogarithmicMap.

kellertuer:
Refactor the one approximate algorithm for a log we have to be an approximate log struct (subtype of approximate inverse retraction).

We should discuss
* backwards compatibility to the old variant of the log with keywords (currently this is the case), but we could also deprecate this already
* the docs are currently in the Stiefel manifold since (similar to PadeRetraction) that is the only manifold where this occurs. We should move it to base when we have a second manifold that needs this (and at that time also add the analogue for exp).

### DIFF
--- a/src/manifolds/StiefelCanonicalMetric.jl
+++ b/src/manifolds/StiefelCanonicalMetric.jl
@@ -156,7 +156,7 @@ function inverse_retract!(
     X,
     p,
     q,
-    a::ApproximateInverseRetraction,
+    a::ApproximateLogarithmicMap,
 ) where {n,k}
     M = p' * q
     QR = qr(q - p * M)
--- a/src/manifolds/StiefelCanonicalMetric.jl
+++ b/src/manifolds/StiefelCanonicalMetric.jl
@@ -20,7 +20,7 @@ An approximate implementation of the logarithmic map, which is an [`inverse_retr
 See [`inverse_retract(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, ::Any, ::Any, ::ApproximateLogarithmicMap) where {n,k}`](@ref) for a use case.
 """
 struct ApproximateLogarithmicMap{T} <: ApproximateInverseRetraction
-    max_iterations::T
+    max_iterations::Int
     tolerance::T
 end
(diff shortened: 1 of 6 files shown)


src/manifolds/StiefelCanonicalMetric.jl
```diff
@@ -13,6 +13,17 @@ manifold, see[^EdelmanAriasSmith1998].
...
 struct CanonicalMetric <: RiemannianMetric end
 
+"""
+    ApproximateLogarithmicMap <: ApproximateInverseRetraction
+
+An approximate implementation of the logarithmic map, which is an [`inverse_retract`](@ref)ion.
+See [`inverse_retract(::MetricManifold{ℝ,Stiefel{n,k,ℝ},CanonicalMetric}, ::Any, ::Any, ::ApproximateLogarithmicMap) where {n,k}`](@ref) for a use case.
+"""
+struct ApproximateLogarithmicMap{T} <: ApproximateInverseRetraction
+    max_iterations::T
```
mateuszbaran, line 23: This could be just an integer.
```suggestion
    max_iterations::Int
```

mateuszbaran (commented):
Good idea! I think we can keep the keywords for now. I think we don't need to worry about using this type for other manifolds until we need it.

mateuszbaran (approved):

test/runtests.jl
```diff
@@ -9,9 +9,8 @@ include("utils.jl")
     include_test("differentiation.jl")
 
     @testset "Ambiguities" begin
-        # TODO: reduce the number of ambiguities for Julia 1.6
         if VERSION.prerelease == () && !Sys.iswindows() && VERSION < v"1.6.0"
-            @test length(Test.detect_ambiguities(ManifoldsBase)) <= 65
+            @test length(Test.detect_ambiguities(ManifoldsBase)) <= 66
```
mateuszbaran, line 13: BTW, does this still need to be bumped?

kellertuer: I think it would be good since it introduces a small new feature, yes.

mateuszbaran: OK.

Issue #376, opened 2021-06-05
Reduce ambiguities with cotangent vectors


kellertuer:
I just took a look at the ambiguities we currently have and it seems most come from
`get_vector`, `get_coordinates` in relation with the cotangent Riesz representer. Is there an easy fix for this?
This yields all the ambiguities wrt ManifoldsBase (66 is the bound we need for 1.5 in 1.6 I think its 47)

Also we should work on Ambiguities of Manifolds (which is zero in 1.5 and this is tested but 189 in 1.6).

mateuszbaran:
Yes, we should definitely reduce the number of ambiguities. I just keep postponing it because they usually correspond to unlikely corner cases (and even if someone hits it there is little harm in that, no wrong answer would be given so it's relatively harmless) or sometimes even calls that wouldn't make much sense.

kellertuer:
Sure, since that is some interplay between ValidationManifold and Basis, I will take a look whether some of these can easily be reduced.

kellertuer:
So a few things I noticed

## Operations with other Packages / Identity
I a few occurrences we have ambiguities with other packages like for example

````
*(s, comp::ChainRulesCore.Tangent) 
in ChainRulesCore at 
ChainRulesCore/fsJxJ/src/differential_arithmetic.jl:190,
vs
*(::Identity{G, PT} where PT, p) where G<:(AbstractGroupManifold{𝔽, MultiplicationOperation, T} where {𝔽, T<:AbstractEmbeddingType})
in Manifolds at
Manifolds.jl/src/groups/group.jl:983)
````

so I think we can not get rid of these?
The arithmetic operators all appear with identity since then one argument is typed the other not.
These are
* 44 with ChainRulesCore.jl (all `+`,`-`,`*`,`/`)
* 4 with DoubleFloats.jl (all `+`,`-`,`*`,`/`)
* 7 with Intervals.jl (all `+`,`-`,`*`,`/`)
* 10 with Polynomials.jl (´+`,`-`+`*`,`/`,ìsapprox`,`mul!`)
* 4 with SpecialFunctions (all `*`)
* 8 LightGraphs (`*`,`mul!*`)

and further
* 4 set/getindex with StaticArrays.jl
* 2 getindex with SparseArrays.jl
* 1 set index with ReverseDiff.jl

'' Own Identity problems

We even manage to do this with ourselves 
````
[...]

PR #377, opened 2021-06-05
Reduce Ambiguities in 1.6

kellertuer:
For now this only reduces by two (186->184) and is not covered in the test (only running for 1.5 with ambiguities). See #376 for an overview of the issues I would like to get solved here.


kellertuer:
Filtering out Identity, we are down to 20 in ManifoldsBase and 28 in Manifolds.

* Most seem to be get/setindex
* two are due to a receipt, which I plan to fix (special case calling Hyperbolic with neither points nor vectors)
* two get_vector/get_coordinate ambituities since we have a specific function in Stiefel.
* some allocations related to the Power Manifold.

for the first and the last I am not so sure how to fix them, for the Stiefel case, I can surely come up with something.

mateuszbaran:
I can take a look at the first and last point.

kellertuer:
Great! I just fixed the last transparency adaptions that got lost in rearranging the code. So now tests should run, up to two failing tests since I set both ambiguity checks for now to 0.

mateuszbaran:
After looking at some `allocate_result` ambiguities, I think the right way to solve them is by implementing the second idea from https://github.com/JuliaManifolds/Manifolds.jl/pull/380#issuecomment-857610393 . Otherwise we are may have to write lots of new code to cover the new disambiguating methods.

kellertuer:
I think, too, that that is reasonable.

mateuszbaran:
I'll change that allocation thing in a separate PR.

kellertuer:
Ok, I already have an idea for the receipt and will look at the remaining vector/coordinate things in the evening. What about the first? For me that seems to be a tricky one, since we type the second input argument, what usually other methods don't do.

mateuszbaran:
There are two setindex/getindex ambiguities that are caused by `ProductArray`, but `ProductArray` is deprecated anyway. The other ones are indeed caused by us dispatching on the second or third argument, and they would only be relevant if someone wanted to use our indexing on arrays from `NamedDims` or `SparseArrays`. I think this is fine.

kellertuer:
Ok, perfect. So then only the receipt and the remaining get_vector/get_coordinates have to be done in this PR.

kellertuer:
I just checked the plotting receipt. That is strange. Both variants for Hyperbolic have optional points and vectors, so with not providing either of them (that is calling `plot(Hyperbolic(2))` is ambiguous, but I can not define a receipt for that it seems, though such a receipt would also just throw an error, since it is not clear which manifold (representation) to draw.

kellertuer:
Ah I found the receipts trick. I saw that we also miss two constructors, but if I try to define what I think might be the ambiguity I always en put defining functions that where already defined, so I am not sure what is missing here, for me it nearly seems an error of the ambiguity, since the argument is once a metric once a manifold and there should not be an ambiguity?

Case 1
```
InvariantMetric{G, D}(metric) where {G<:AbstractMetric, D<:ActionDirection} in Manifolds at /src/groups/metric.jl:37
(::Type{T})(M::AbstractManifold) where T<:AbstractMetric in Manifolds at /src/manifolds/MetricManifold.jl:19
```

Case 2
```
(::Type{T})(M::AbstractManifold) where T<:AbstractGroupOperation in Manifolds at /src/groups/group.jl:128
Manifolds.SemidirectProductOperation{A}(action) where A<:AbstractGroupAction in Manifolds at /src/groups/semidirect_product_group.jl:9
```

mateuszbaran:
> Ah I found the receipts trick. I saw that we also miss two constructors, but if I try to define what I think might be the ambiguity I always en put defining functions that where already defined, so I am not sure what is missing here, for me it nearly seems an error of the ambiguity, since the argument is once a metric once a manifold and there should not be an ambiguity?

That's really weird, the first function in case 1 and the second function in case 2 should have type bounds on arguments. I'm currently working on allocations so I can't check it right now.

mateuszbaran:
It looks like that rework of allocations is a much bigger issue than I anticipated. It revealed how fragile our allocation support for embedded manifolds is and that there are serious design issues. I'd prefer to leave it for later.

kellertuer:
I resolved the metric constructor issues.

I resolved the allocation issues (similar to get_coordinates / get_vector in base now for get_point and get_parameters).

The first should be tested directly for the second I will check later. We are down to (Excluding Identity-ambiguities) 
* 16 ambiguities with ManifoldsBase (haven't looked at these but they might be similar to the ones in the next point)
* 19 in Manifolds
  * 9 `getindex` ones (we might not fix these here, especially since 5 are related to the ProductArray which is basically deprecated)
  * 2 `setindex` (both with `ProductArray`)
  * 2 `get_coordinates!`  (Stiefel versus AbstractDecorator/AbstractManifold) 
  * 2 `get_vector` (AbstractManifold versus ProductManifold)
  * 2 `get_vector!`(same/similar to `get_coordinates!`) above)

So I think we should fix the last 6 ones at least.

mateuszbaran:
Thanks! I've solved the 4 ambiguities related to Stiefel. The two related to `ProductManifold`:
```
get_vector(M::AbstractManifold, p, v, ::DefaultOrthonormalBasis{𝔽, ManifoldsBase.CotangentSpaceType}) where 𝔽 in Manifolds at /home/mateusz/.julia/dev/Manifolds/src/cotangent_space.jl:97,
get_vector(M::ProductManifold, p::ProductRepr, X, B::AbstractBasis) in Manifolds at /home/mateusz/.julia/dev/Manifolds/src/manifolds/ProductManifold.jl:483)
 
get_vector(M::ProductManifold, p::ProductRepr, X, B::AbstractBasis) in Manifolds at /home/mateusz/.julia/dev/Manifolds/src/manifolds/ProductManifold.jl:483,
get_vector(M::ProductManifold, p, v, b::DefaultOrthonormalBasis{𝔽, ManifoldsBase.CotangentSpaceType}) where 𝔽 in Manifolds at /home/mateusz/.julia/dev/ManifoldsBase/src/bases.jl:922)
```
Basically mean that we don't support mixing `ProductRepr` and cotangent spaces, which may be fine? I don't think cotangent vectors are ever supposed to be represented as `ProductRepr`, and that's the only way to cause these ambiguities.

mateuszbaran:
After this, there are only 12 remaining ambiguities in ManifoldsBase, and they are either related to `ProductArray`, or indicate incomplete interoperability between our indexing with manifolds and other libraries.

kellertuer:
Great, you saw the fixes for Stiefel much faster than I would have seen them, I am sure :)

I am not 100% sure we would not mix that (what would be the reason not to? Wouldn't that appear if someone works with cotangent vectors on a product manifold? I can at least imagine someone coming by to try that. But I also have not yet checked how difficult a fix would be.

mateuszbaran:
> Great, you saw the fixes for Stiefel much faster than I would have seen them, I am sure :)

Well, it's actually me who changed our types for bases so that this change is needed :slightly_smiling_face: . I'm sure I'd have noticed that earlier if we didn't have so many ambiguities, so this cleanup will be helpful.

> I am not 100% sure we would not mix that (what would be the reason not to? Wouldn't that appear if someone works with cotangent vectors on a product manifold? I can at least imagine someone coming by to try that. But I also have not yet checked how difficult a fix would be.

My current thinking is that cotangent vectors should be wrapped in some special structure, like `RieszRepresenterCotangentVector` or `FVector` (when storing coefficients). So `ProductRepr` would never be the outer layer. What would it mean to have a cotangent vector as a `ProductRepr`? What would be allowed inside? That requires some designing work.

kellertuer:
The only thing left is to fix the missing 8 lines of code coverage, but otherwise this is good to go.

mateuszbaran:
I think my last change should cover things related to `PowerManifold` allocation.

kellertuer:
`get_parameter`seems to not be covered on Power, but with that we should have enough coverage I think.

mateuszbaran:
OK, this last change should cover the `get_parameter` method.

kellertuer:
Perfect, since it does not break anything nor introduce a new feature I would maybe merge this without doing a new release?

mateuszbaran:
It does make a few small changes so we might as well tag a new release.

mateuszbaran (approved):

Issue #379, opened 2021-06-08
Orthonormality test is too strict


Nikdwal:
I'm implementing the manifold of  tensors of fixed multilinear rank and the test code seems to be much too strict.
I have a `DefaultOrthonormalBasis`, and the normalisation is checked in `src/tests_general.jl` using the following line:
https://github.com/JuliaManifolds/Manifolds.jl/blob/7e90c455804311f1220bcb52878fa5abf86e0abb/src/tests/tests_general.jl#L540

This test gives the following output:
```
Tucker: **Test Failed** at .../Manifolds.jl/src/tests/tests_general.jl:540
  Expression: ≈(norm(M, p, Xs_invs[i]), 1, atol = find_eps(p))
   Evaluated: 0.9999999999999993 ≈ 1 (atol=2.220446049250313e-16)
```

It seems appropriate to change the value of `atol` to something like `√find_eps(p)`, as is done three lines down.

mateuszbaran:
Yes, sure, `sqrt(find_eps(p))` in that place makes sense.

kellertuer:
I second that, sure, reduce the tolerance, that seems reasonable. Just do it on your branch.

And cool that you work on a new manifold for `Manifolds.jl`!
If you would like some feedback on your current state of implementation (haven't looked at your branch yet) feel free to open a PR already early on.
I hope the way how to implement a manifold is understandble; if not, also open the PR and we can help along the way.

mateuszbaran:
Fixed in #380  .

PR #380, opened 2021-06-09
The Tucker manifold

Nikdwal:
This implements the manifold of tensors of fixed multilinear rank. It generalises `FixedRankMatrices`. The `DefaultOrthonormalBasis` is based on research that is coming out very soon.
It also fixes issue #379 . 

I had a few issues and implemented my own tweaks to the basic interface. If you think these are good ideas, I'm more than happy to help implementing them for the rest of the code base. Of course, if there's already a standard solution, I'll do that instead.

**Allocate result**
The non-mutating variant of several functions allocates the wrong type by default if you use specialised data types. For instance, as far as I know, `inverse_retract` has no way of knowing what the data type of the tangent vector is, but it always allocates something of the same type as the manifold point. Because tangent vectors are a different type in this case, i implemented a function `allocate_vector` and added a variant of `allocate_result` that is specific for this manifold. I don't think the documentation is sufficiently clear on this. I wonder if the current `ManifoldsBase` implementation is able to deal with these specialised data types. If not, I suggest doing something like my `allocate_vector` implementation in general and providing documentation for this. 

**foreach on a basis**
I had to loop over the basis vectors several times, but i didn't want to use `get_vectors`, which would be inefficient in terms of memory. I added a method for `Base.foreach` to get around this. This seems to be rather useful in general.

**show for specialised basis datatypes**
By default, the REPL shows an error when creating a CachedBasis of some exotic type, because there is no implementation for `Base.show`. I converted mine to a matrix whose columns are the basis vectors in ambient space coordinates. This is very inefficient is the basis is much larger than what can be displayed on the screen, though.

**inner products with vectors in the ambient space**
I added a method to take inner products between a tangent and non-tangent vector, which is useful for projection. If this were applied in general, this should probably have a different name than `inner`. This is because the general interface is not powerful enough to detect whether a vector is tangent to the manifold or not based on its datatype.


kellertuer:
Thanks for working on this.

Concerning the allocation, we have https://juliamanifolds.github.io/Manifolds.jl/latest/interface.html#ManifoldsBase.allocate_result-Tuple{AbstractManifold,Any,Vararg{Any,N}%20where%20N} in ManifoldsBase.

Concerning the foreach and the show method: Great ideas!

Concerning the inner products:
Whether a vector is a tangent vector or not can be determined using `is_vector` (you yourself implement `check_vector`) and note that for embeddings we have the `EmbeddedManifold`(to couple a manifold and it embedding manifold) or an `AbstractEmbeddedManifold`, if a manifold itself (in the default representation) is embedded. Then `embed` changes a tangent vector from the manifold representation to the one of the embedding. After `embed` you can call the `inner` in the embedding.

Nikdwal:
Yes, you can always do `embed` before doing `inner` and then you have an inner product that works for tangent and non-tangent vectors. The thing is that this may be inefficient (and I think it is in this case). 

And I do think the inner workings of `allocate_result` deserve a more elaborate explanation in the `How to implement a manifold` section. It's not obvious from the documentation when it might be useful to override this

kellertuer:
> Yes, you can always do `embed` before doing `inner` and then you have an inner product that works for tangent and non-tangent vectors. The thing is that this may be inefficient (and I think it is in this case).

that is, you say, that you function doing embed-inner (embed for one argument) is more efficient than coupling the two? Uff. Complicated scenario; one could do `inner(::(Abstract)EmbeddedManifold, embedded_point, embedded_tangent, non_embedded_tangent)` (and by symmetry similarly for the tangent vectors the other way around) – but then the tangent vectors have to be of different type (not just both matrices for example). Sure, that would be something we could provide.

> 
> And I do think the inner workings of `allocate_result` deserve a more elaborate explanation in the `How to implement a manifold` section. It's not obvious from the documentation when it might be useful to override this

Sure, it might be too short currently. Sorry, we should extend that. A lame excuse here is maybe that up to now we did not require that that often yet. Maybe report your difficulties here and we incorporate them :)
Until then, maybe an example like this https://github.com/JuliaManifolds/Manifolds.jl/blob/eb296b51a0c00b6459b277ef28f561a984e06b13/src/manifolds/Hyperbolic.jl#L189 might help? If you have questions, sure, just ask.

Nikdwal:
I see two ways the allocation problem could be resolved:

1. The user is instructed to override `allocate_result` or `allocate_result_type` for every method that could cause difficulties. The problem is that they may not know in advance which methods will cause difficulties. 
2. The user should implement something like `allocate_vector`.  If no implementation is provided, there is a default that does the same thing as what `allocate_result` does. Then functions like `inverse_retract` and `zero_vector` use `allocate_vector` and then apply the mutating variant.

The former requires more thinking on the user's end. The latter requires changing the interface in a rather significant way. If you prefer the first solution, I could do a proposal for the documentation that I would have liked to see.

kellertuer:
> I see two ways the allocation problem could be resolved:
> 
> 1. The user is instructed to override `allocate_result` or `allocate_result_type` for every method that could cause difficulties. The problem is that they may not know in advance which methods will cause difficulties.
> 2. The user should implement something like `allocate_vector`.  If no implementation is provided, there is a default that does the same thing as what `allocate_result` does. Then functions like `inverse_retract` and `zero_vector` use `allocate_vector` and then apply the mutating variant.
> 
> The former requires more thinking on the user's end. The latter requires changing the interface in a rather significant way. If you prefer the first solution, I could do a proposal for the documentation that I would have liked to see.

I like the second idea, but would phrase it a little different, we could do a `allocate_vector_type`(because some places just need the type), a `allocate_vector`, an `allocate_point_type`, and an `allocate_point` just to be complete;
then for sure, especially `log` (which started this allocation type business here ;)) `inverse_retract` and `zero_vector` would do as you propose. Maybe this additional level would make it easier for users to implement manifolds (if not operating on matrices/arrays).

Do you think you can manage here without them and we can do that in a separate/next issue/PR or should we do that here? (also cc @mateuszbaran, because he is the allocation expert among us).

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,741 @@
...
+isapprox(::Tucker, p::TuckerPoint, q::TuckerPoint; kwargs...) = isapprox(p, q; kwargs...)
+function isapprox(M::Tucker, p::TuckerPoint, x::TuckerTVector, y::TuckerTVector; kwargs...)
+    return isapprox(embed(M, p, x), embed(M, p, y); kwargs...)
+end
+
+#=
+Determines whether there are tensors of dimensions n⃗ with multilinear rank r⃗
+=#
+function is_valid_mlrank(n⃗, r⃗)
+    return all(r⃗ .≤ n⃗) && all(ntuple(i -> r⃗[i] ≤ prod(r⃗) ÷ r⃗[i], length(r⃗)))
```
mateuszbaran, line 534: I guess this could also explicitly check that `all(r⃗ .>=1 )`?

Nikdwal:
Sure, I can do solution 1 for the Tucker manifold.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,741 @@
...
+
+Base.:*(s::Number, x::TuckerTVector) = TuckerTVector(s * x.Ċ, s .* x.U̇)
+Base.:*(x::TuckerTVector, s::Number) = TuckerTVector(x.Ċ * s, x.U̇ .* s)
+Base.:/(x::TuckerTVector, s::Number) = TuckerTVector(x.Ċ / s, x.U̇ ./ s)
+Base.:\(s::Number, x::TuckerTVector) = TuckerTVector(s \ x.Ċ, s .\ x.U̇)
+Base.:+(x::TuckerTVector, y::TuckerTVector) = TuckerTVector(x.Ċ + y.Ċ, x.U̇ .+ y.U̇)
+Base.:-(x::TuckerTVector, y::TuckerTVector) = TuckerTVector(x.Ċ - y.Ċ, x.U̇ .- y.U̇)
+Base.:-(x::TuckerTVector) = TuckerTVector(-x.Ċ, map(-, x.U̇))
+Base.:+(x::TuckerTVector) = TuckerTVector(x.Ċ, x.U̇)
+Base.:(==)(x::TuckerTVector, y::TuckerTVector) = (x.Ċ == y.Ċ) && all(x.U̇ .== y.U̇)
```
mateuszbaran, line 154: We should also have the broadcasting thing for `TuckerTVector`. That may be a bit more tricky than for `UMVTVector` so I can write that.

kellertuer:
Cool! Let's open an issue/discuss the second one afterwards, it sounds like a good improvement, I think.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,741 @@
...
+    Uℭ = embed(ℳ, 𝔄)
+    n⃗ = size(Uℭ)
+    for d in 1:D
+        Y .= Y + fold(X.U̇[d] * (𝔄.hosvd.U[d]' * unfold(Uℭ, d)), d, n⃗)
+    end
+    return Y
+end
+
+# Inverse of the k'th unfolding of a size n₁ × ... × n_D tensor
+function fold(𝔄♭::AbstractMatrix{T}, k, n⃗::NTuple{D,Int})::Array{T,D} where {T,D,Int}
```
mateuszbaran, line 336: Could you change this name to, for example, `tucker_fold`? This way it won't be confused with any other folding.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,741 @@
...
+        U = U .* hosvd_core.U
+        𝔄 = hosvd_core.core
+        σ = hosvd_core.σ
+    end
+
+    return HOSVD{T,D}(U, 𝔄, σ)
+end
+
+#Mode-k unfolding of the array 𝔄 of order D ≥ k
+function unfold(𝔄, k)
```
mateuszbaran, line 709: Could you rename it to `tucker_unfold`, similarly to `fold`?

mateuszbaran (commented):
This looks really good. I still have a lot to digest, but here are a few quick comments :slightly_smiling_face: . Thanks for this great contribution!

mateuszbaran:
> Do you think you can manage here without them and we can do that in a separate/next issue/PR or should we do that here? (also cc @mateuszbaran, because he is the allocation expert among us).

I like the idea of making `allocate_result_type` call `allocate_point_type` and `allocate_vector_type` depending on the function. What do you think `allocate_point` would be needed for?

Nikdwal:
I think every function that needs to allocate a point gets a point at an argument, so the default `allocate` should do.

src/manifolds/Tucker.jl
mateuszbaran: here is the code for broadcasting:
```julia
# Tuple-like broadcasting of TuckerTVector

function Broadcast.BroadcastStyle(::Type{TuckerTVector{T,D}}) where {T,D}
    return Broadcast.Style{TuckerTVector{Any,D}}()
end
function Broadcast.BroadcastStyle(
    ::Broadcast.AbstractArrayStyle{0},
    b::Broadcast.Style{<:TuckerTVector},
)
    return b
end

Broadcast.instantiate(bc::Broadcast.Broadcasted{Broadcast.Style{<:TuckerTVector},Nothing}) = bc
function Broadcast.instantiate(bc::Broadcast.Broadcasted{Broadcast.Style{<:TuckerTVector}})
    Broadcast.check_broadcast_axes(bc.axes, bc.args...)
    return bc
end

Broadcast.broadcastable(v::TuckerTVector) = v

@inline function Base.copy(bc::Broadcast.Broadcasted{Broadcast.Style{TuckerTVector{Any,D}}}) where {D}
    return TuckerTVector(
        @inbounds(Broadcast._broadcast_getindex(bc, Val(:Ċ))),
        ntuple(i -> @inbounds(Broadcast._broadcast_getindex(bc, Val((:U̇, i)))), Val(D)),
    )
end

Base.@propagate_inbounds function Broadcast._broadcast_getindex(
    v::TuckerTVector,
    ::Val{I},
) where {I}
    if I isa Symbol
        return getfield(v, I)
    else
        return getfield(v, I[1])[I[2]]
    end
end

Base.axes(::TuckerTVector) = ()

@inline function Base.copyto!(
    dest::TuckerTVector,
    bc::Broadcast.Broadcasted{Broadcast.Style{TuckerTVector{Any,D}}},
) where {D}
    # Performance optimization: broadcast!(identity, dest, A) is equivalent to copyto!(dest, A) if indices match
    if bc.f === identity && bc.args isa Tuple{TuckerTVector} # only a single input argument to broadcast!
        A = bc.args[1]
        return copyto!(dest, A)
    end
    bc′ = Broadcast.preprocess(dest, bc)
    copyto!(dest.U, Broadcast._broadcast_getindex(bc′, Val(:Ċ)))
    for i in 1:D
        copyto!(dest.M, Broadcast._broadcast_getindex(bc, Val((:U̇, i))))
    end
    return dest
end

####
```

kellertuer:
> > Do you think you can manage here without them and we can do that in a separate/next issue/PR or should we do that here? (also cc @mateuszbaran, because he is the allocation expert among us).
> 
> I like the idea of making `allocate_result_type` call `allocate_point_type` and `allocate_vector_type` depending on the function. What do you think `allocate_point` would be needed for?

I just had something like `rand(M)` in mind.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,741 @@
...
+$N_1 \times \dots \times N_D$-array.
+
+    embed(::Tucker, A::TuckerPoint, X::TuckerTVector)
+
+Convert a tangent vector `X` with base point `A` on the Tucker manifold to a full tensor,
+epresented as an $N_1 \times \dots \times N_D$-array.
+"""
+embed(::Tucker, ::Any, ::TuckerPoint)
+
+function embed!(::Tucker, q, p::TuckerPoint)
```
kellertuer, line 322: This might not be necessary if the allocation is correct? Also `Tucker`should maybe be an `AbstractEmbeddedManifold` where `get_embedding` is then defined to return the `Euclidean` of corresponding size (compare e.g. to Sphere or Hyperbolic); oh I noticed, for FixedRank it might also be nice, maybe.

Nikdwal: A `TuckerPoint` is stored in factorised form. This is the only function where the factorisation is evaluated explicitly. So the default `embed!`  implementation won't work.

kellertuer: Sure, you still have to implement `embed!`. But it would be more precisely typed. I am not 100% sure it would bring any actual benefits (see Fixedrank, where it seems missing but nobody noticed yet).

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,741 @@
...
+number_eltype(::TuckerTVector{T,D}) where {T,D} = T
+
+"""
+    project(ℳ::Tucker, 𝔄::TuckerPoint, X)
+
+The least-squares projection of a tensor `X` to the tangent space to `ℳ` at `A`.
+"""
+project(::Tucker, ::Any, ::TuckerPoint, ::Any)
+
+function project!(ℳ::Tucker, Y, 𝔄::TuckerPoint, X)
```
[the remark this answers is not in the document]
Nikdwal: Now that I'm writing the code, I dont think that's a good idea. The default implementation for a lot of methods goes to the `Euclidean` implementation, which assumes that the point or vector is represented as a dense array. I'm having problems with `zero_vector` for instance. If you use other data structures, there is little to gain from these default implementations.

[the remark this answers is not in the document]
kellertuer: You can set the transparency to be quite a few functions only, but I am also fine if you do not do that, if you see to many problems coming up.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,812 @@
...
+inner(M::Tucker, 𝔄::TuckerPoint, x, y::TuckerTVector) = dot(x, embed(M, 𝔄, y))
+
+"""
+    inverse_retract(ℳ::Tucker, A::TuckerPoint, B::TuckerPoint, r::ProjectionInverseRetraction)
+
+The projection inverse retraction on the Tucker manifold interprets `B` as a point in the
+ambient Euclidean space and projects it onto the tangent space at to `ℳ` at `A`.
+"""
+inverse_retract(
+    ::Tucker, ::Any, ::TuckerPoint, ::TuckerPoint, ::ProjectionInverseRetraction
```
kellertuer, line 568:
```suggestion
    ::Tucker, ::TuckerPoint, ::TuckerPoint, ::ProjectionInverseRetraction
```
Note that `inverse_retraction` as correctly documented has a manifold, two points and a retraction type, the any is a parameter too much here.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,812 @@
...
+
+function project!(ℳ::Tucker, Y, 𝔄::TuckerPoint, X)
+    ℬ = get_basis(ℳ, 𝔄, DefaultOrthonormalBasis())
+    coords = Vector{number_eltype(𝔄)}(undef, manifold_dimension(ℳ))
+    f!(i, ℬᵢ) = setindex!(coords, inner(ℳ, 𝔄, ℬᵢ, X), i)
+    foreach(f!, ℳ, 𝔄, ℬ)
+    return get_vector!(ℳ, Y, 𝔄, coords, ℬ)
+end
+
+representation_size(::Tucker{N}) where {N} = N
```
mateuszbaran, line 632: This is incorrect, when points and vectors on a manifold are represented using special types, `representation_size` should return `nothing` (which is the default). Did you need it to make some test pass?

kellertuer:
Cool, the tests are running! Could you check either the annotations in the “Flies Changed“ tab here or on the page here
https://app.codecov.io/gh/JuliaManifolds/Manifolds.jl/compare/380/diff#diff-c3JjL2dyb3Vwcy9ncm91cF9hY3Rpb24uamw=

That you do a few more tests of your new functions?
Currently
* Not all vector operations seem to be covered (multiplication with a scalar from the left, `/`, `\` with a scalar
* The current `allocate` implementations are not used
* a test for one of the the broadcasts seems missing
* Try a `copyto!` (or if you get the current master you can also just test a `copy` which internally uses `copyto!`
* the first copyto! could also be moved down to be sorted in the alphabetic order we usually prefer
* I think the generic `copyto!` as the manifold as first argument, and the `copyto!` for vectors also includes the base point (i.e. a `copyto!(M, Y, p, X)` to copy X to Y in TpM of M).
* Whats the generic `check_point` for arbitrary `x`? Also for the other the size test is never triggered nor the `gram` test.
* `get_vectors` is not testet
* one of the inners (L. 585/586), one is not tested, but since the other is, just use the same setting to check for symmetry)
* Please write a few tests for the show methods. We do this for example here [here](https://github.com/JuliaManifolds/Manifolds.jl/blob/fcc8645e28713c33f4c280f927863c4ce59cb2f5/test/fixed_rank.jl#L12), but I like it better if the inner prints are not explicitly tested, that is, if Julia changes print of an array, this is taken into account in the test automatically, like we do for example [here](https://github.com/JuliaManifolds/Manifolds.jl/blob/fcc8645e28713c33f4c280f927863c4ce59cb2f5/test/power_manifold.jl#L352) – i.e. to use an `sprint` for inner print/show calls within your show.

I know testing is something that takes time and is not much fun. Hang in there :)

Also I added the `preview docs` tag, so that I hope the next push here will pre-render the docs, too, so we can take a look here. Otherwise I will check that some time locally.

Thanks again for your contribution, in general this looks quite nice!

Nikdwal:
I agree with all of these besides the `show`, especially when applied to a `TuckerPoint`. I could either just basically rewrite the `show` code in the test, or I could come up with an example where you can predict what the output is going to be. For instance, you can find tensors whose Tucker factorisation is very sparse. Then you would expect sparse factors to be printed. _However_, the constructor always computes the factorisation up to finite precision. Instead of zeros, you get numbers close to the machine epsilon. Copying this output into the test is not good unit testing in my opinion. I think the best you can hope for is just checking part of the output, but that's a bit silly.

kellertuer:
You are completely right, if we reach numerical precision and the value is not 100% known even for a deterministic constructor (please do not do tests with random), then checking against a pure string is not useful.
Let me sketch the test I have in mind, similar to [here](https://github.com/JuliaManifolds/Manifolds.jl/blob/fcc8645e28713c33f4c280f927863c4ce59cb2f5/test/power_manifold.jl#L352).

You reconstruct the show by constructing the string's `"U factor"`s and the `"Core"` and check that for a point `p` the call `@test sprint(show, "text/plain", Bc)` has these blocks in the right order with the right indent and such.
The important this (similar to the power manifold link) is, that nowhere in your test you have a string containing a number of the factors/core.
But I feel it is important also for show methods that they behave as intended. So they should be covered by a test.

kellertuer:
* For now you can ignore the Nightly failing
* Please check the results from the formatter or run `using JuliaFormatter; format(".")` (if you have that installed that is)
* Could you check for just a few more lines of code coverage? Best would be of course if you could resolve all uncovered ones in  https://app.codecov.io/gh/JuliaManifolds/Manifolds.jl/compare/380/diff#diff-c3JjL21hbmlmb2xkcy9UdWNrZXIuamw= - which to me look like they are not too hard to test, e.g.
  * broadcast is not tested
  * that `is_point`(calling `check_point`) returns true for a point that is on the manifold is not tested
  * the second case of `copyto!` is not tested
* If you want to, add yourself to the `.zenodo.json` which keeps track of contributors, where minor changes/fixes are `Other` and contributing a feature/manifold we consider `ProjectMember`, so that would be the case for you, too, if you like. Be careful, if that is wrongly formatted on merge we might loose a Zenodo version. This file is read by Zenodo and you are listed there then :)
* Finally bump the last of the version numbers, i.e. to `0.5.9` if I see that correctly on master.

I have to check the docs, still, the next few days, but it looks quite good already!

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
+
+@doc raw"""
+Tucker{N, R, D, 𝔽} <: AbstractManifold{𝔽}
```
kellertuer, line 3:
```suggestion
    Tucker{N, R, D, 𝔽} <: AbstractManifold{𝔽}
```
The first line should be indented by 4 spaces to be rendered as code.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+# Representation in HOSVD format
+
+Any tensor $\mathcal{A}$ on the Tucker manifold can be represented in HOSVD
+[^DeLathauwer2000] form
+```math
+\mathcal{A} = (U_1,\dots,U_D) \cdot \mathcal{C}
+```
+where $\mathcal C \in \mathbb{F}^{R_1 \times \dots \times R_D}$ and, for $d=1,\dots,D$,
+the matrix $U_d \in \mathbb{F}^{N_d \times R_d}$ contains the singular vectors of the
+$d$th unfolding of $\mathcal{A}$
```
kellertuer, line 18: At the end of this line a . is missing.

I think it would be good to mention, that F is either the real-valued or the complex-valued numbers (or even quaternions?)

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+T_{\mathcal{A}} \mathcal{M} =
+\bigl\{
+(U_1,\dots,U_D) \cdot \dot{\mathcal{C}}
++ \sum_{d=1}^D \bigl(
+    (U_1, \dots, U_{d-1}, \dot{U}_d, U_{d+1}, \dots, U_D)
+    \cdot \mathcal{C}
+\bigr)
+\bigr\}
+```
+where $\dot{\mathcal{C}}$ is arbitrary, $\dot{U}_d^{\mathrm{H}}$ is the Hermitian adjoint of
```
kellertuer, line 34: Why is \dot U^H the adjoint of U? That reads a little strange.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+$U_d$, and $\dot{U}_d^{\mathrm{H}} U_d = 0$ for all $d$.
+
+# Constructor
+    Tucker(n⃗::NTuple{D, Int}, r⃗::NTuple{D, Int}[, field = ℝ])
```
kellertuer, line 38: It would be great if all input variables could shortly be explained in the text after this line.

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+* σ : singular values of the unfoldings
+=#
+struct HOSVD{T,D}
+    U::NTuple{D,Matrix{T}}
+    core::Array{T,D}
+    σ::NTuple{D,Vector{T}}
+end
+
+"""
+TuckerPoint{T, D}
```
kellertuer, line 75:
```suggestion
    TuckerPoint{T, D}
```

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+
+An order D tensor of fixed multilinear rank and entries of type T. The tensor is
+represented in HOSVD form. See also [`Tucker`](@ref).
+
+# Constructors:
+    TuckerPoint(core::AbstractArray{T, D}, factors::Vararg{MtxT, D}) where {T, D, MtxT <: AbstractMatrix{T}}
+
+A tensor of the form (factors[1], …, factors[D]) ⋅ core
```
kellertuer, line 83: Should "(factors[1], …, factors[D])" maybe be in backtracks?
Would it be reasonable to explain also what T and D are?

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+where it is assumed that the dimensions of the core are the multilinear rank of the tensor.
+
+    TuckerPoint(A::AbstractArray{T, D}, mlrank::NTuple{D, Int}) where {T, D}
+
+The low-multilinear rank tensor arising from the sequentially truncated the higher-order
+singular value decomposition of A [^Vannieuwenhoven2012].
```
kellertuer, line 89: Could you also mention what `mlrank` is?

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+end
+function TuckerPoint(A::AbstractArray{T,D}, mlrank::NTuple{D,Int}) where {T,D}
+    @assert is_valid_mlrank(size(A), mlrank)
+    return TuckerPoint(st_hosvd(A, mlrank))
+end
+
+@doc raw"""
+    TuckerTVector{T, D} <: TVector
+
+Tangent vector to the D-th order Tucker manifold at $x = (U_1,\dots,U_D) ⋅ \mathcal{C}$.
```
kellertuer, line 123: In most cases we use `p` as a point on the manifold. You seem to use usually A here, which is fine, too, but please be consistent then.

Style-wise we often prefer two backticks `` instead of $ for math in Markdown

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,876 @@
...
+
+"""
+    inner(M::Tucker, A::TuckerPoint, x::TuckerTVector, y::TuckerTVector)
+
+The Euclidean inner product between tangent vectors `x` and `y` at the point `A` on
+the Tucker manifold. This is equal to `embed(M, A, x) ⋅ embed(M, A, y)`.
+
+    inner(::Tucker, A::TuckerPoint, x::TuckerTVector, y)
+    inner(::Tucker, A::TuckerPoint, x, y::TuckerTVector)
+
```
kellertuer, line 572: We usually use `p` for points on a manifold `M` and `X` and `Y` for the tangent vectors. I think you used a mathcal M instead often for the manifold.

If you use a different notation, please use it consistently; otherwise it would also be great if you could adapt to the notation we use (but you doing not have to); see https://juliamanifolds.github.io/Manifolds.jl/latest/misc/notation.html

kellertuer (commented):
Thanks for all your work, this starts to look really great. I have a few remarks on the documentation after I checked it out locally and compiled it (sorry, for external PRs our automatic way of generating them does not seem to work.

I hope this will be the last remarks, thanks for taking all the ones before into account already!

src/manifolds/Tucker.jl
```diff
@@ -0,0 +1,884 @@
...
+Construct an order `D` tensor of element type `T` that can be represented as the
+multilinear product `(factors[1], …, factors[D]) ⋅ core`.
+It is assumed that the dimensions of the core are the multilinear
+rank of the tensor and that the matrices `factors` each have full rank. No further
+assumptions are made.
+
+    TuckerPoint(p::AbstractArray{T, D}, mlrank::NTuple{D, Int}) where {T, D}
+
+The low-multilinear rank tensor arising from the sequentially truncated the higher-order
```
kellertuer, line 92: I might have missed this one: Could you explain T and D here, too? Should be similar to the line above, but just for completeness.

src/manifolds/Tucker.jl
kellertuer: If you just could check for math being in `` (instead of $)? That is more markdown-y

Nikdwal: I only use the $ notation whenever I refer to symbols that also occur in a more complicated formula that requires TeX typesetting. This is one example, because the expression for the point is used in the equation below. In all other cases, I used backticks.

kellertuer: Sure, that is reasonable and even preferable to use a math environment when the formulae are better readable that way (actually U even use it whenever I refer to math variables an not to input/output variables).

While in TeX you use `$...$` in Markdown you can also use ` ``...`` `, which is what we use mostly (we haven‘t gone through all the old `$...$`s yet. In rendered HTML (Documentation) both yield the same. So I actually really mean the double backticks here :)

kellertuer (approved):
Thanks for all your contribution and for patiently going through all the remarks we had. Let‘s wait for the tests to run just one more time and maybe a day or two whether someone else wants to comment here – and then we can merge :)

kellertuer:
I am sorry we have to check why Julia 1.4 decided (not only here but also on the other PR) to not finish anymore. I will investigate the next days. So thanks already for your work we will merge as soon as that issue (unrelated to your work) is fixed :)

kellertuer:
We fixed it in another feature that we just merged, so you can merge master to solve the test problem, you just have to set the version again, since you are now 0.5.10 (and might get that as a merge conflict maybe when you merge).

PR #381, opened 2021-06-10
`lie_bracket` and `adjoint_action`

mateuszbaran:
This PR adds two things that seem to be missing from our Lie group/Lie algebra support, as identified in #366 .


kellertuer:
We might need some more documentation, especially, it would be nice to have the general case ( https://en.wikipedia.org/wiki/Lie_bracket_of_vector_fields ) on Manifolds, too (at least have it in mind here).

mateuszbaran:
Yes, I'll add more documentation here. For this PR I'm going to restrict the Lie bracket to Lie algebras, and the more general case can be added later (I don't even know what could it be useful for).

src/groups/circle_group.jl
```diff
@@ -14,6 +14,13 @@ invariant_metric_dispatch(::CircleGroup, ::ActionDirection) = Val(true)
 
 default_metric_dispatch(::MetricManifold{ℂ,CircleGroup,EuclideanMetric}) = Val(true)
 
+adjoint_action(::CircleGroup, p, X) = X
+
+function adjoint_action!(::CircleGroup, Y, p, X)
+    copyto!(Y, X)
+    return Y
+end
```
sethaxen, line 22:
```suggestion
adjoint_action!(::CircleGroup, Y, p, X) = copyto!(Y, X)
```

src/groups/circle_group.jl
```diff
@@ -46,6 +53,13 @@ function inverse_translate(
     return map(/, q, p)
 end
 
+lie_bracket(::CircleGroup, X, Y) = zero(X)
+
+function lie_bracket!(::CircleGroup, Z, X, Y)
+    fill(Z, 0)
+    return Z
+end
```
sethaxen, line 61:
```suggestion
lie_bracket!(::CircleGroup, Z, X, Y) = fill!(Z, 0)
```

src/groups/group.jl
```diff
@@ -484,6 +509,18 @@ function decorator_transparent_dispatch(
     return Val(:intransparent)
 end
 
+"""
+    lie_bracket(G::AbstractGroupManifold, X, Y)
+
+Lie bracket between elements `X`, `Y` of the Lie algebra corresponding to the Lie group `G`.
```
sethaxen, line 515:
```suggestion
Lie bracket between elements `X` and `Y` of the Lie algebra corresponding to the Lie group `G`.
```

src/groups/group.jl
```diff
@@ -951,6 +988,13 @@ Base.:*(e::Identity{G}, p) where {G<:AdditionGroup} = e
 Base.:*(p, e::Identity{G}) where {G<:AdditionGroup} = e
 Base.:*(e::E, ::E) where {G<:AdditionGroup,E<:Identity{G}} = e
 
+adjoint_action(::AdditionGroup, p, X) = X
+
+function adjoint_action!(::AdditionGroup, Y, p, X)
+    copyto!(Y, X)
+    return Y
+end
```
sethaxen, line 996:
```suggestion
adjoint_action!(::AdditionGroup, Y, p, X) = copyto!(Y, X)
```

src/groups/group.jl
```diff
@@ -988,6 +1032,13 @@ group_log(::AdditionGroup, q) = q
 
 group_log!(::AdditionGroup, X, q) = copyto!(X, q)
 
+lie_bracket(::AdditionGroup, X, Y) = zero(X)
+
+function lie_bracket!(::AdditionGroup, Z, X, Y)
+    fill(Z, 0)
+    return Z
+end
```
sethaxen, line 1040:
```suggestion
lie_bracket!(::AdditionGroup, Z, X, Y) = fill!(Z, 0)
```

src/groups/group.jl
```diff
@@ -1015,6 +1066,13 @@ Base.:\(p, ::Identity{G}) where {G<:MultiplicationGroup} = inv(p)
 Base.:\(::Identity{G}, p) where {G<:MultiplicationGroup} = p
 Base.:\(e::E, ::E) where {G<:MultiplicationGroup,E<:Identity{G}} = e
 
+adjoint_action(G::MultiplicationGroup, p, X) = p * X / p
```
sethaxen, line 1069: I think it's better to use the `translate_diff` and `inverse_translate_diff` machinery here and below, especially because e.g. for `SpecialOrthogonal` that is implemented more efficiently.

src/groups/group.jl
```diff
@@ -1062,3 +1120,10 @@ function group_exp!(G::MultiplicationGroup, q, X)
 end
 
 group_log!(::MultiplicationGroup, X::AbstractMatrix, q::AbstractMatrix) = log_safe!(X, q)
+
+lie_bracket(::MultiplicationGroup, X, Y) = X * Y - Y * X
+
+function lie_bracket!(::MultiplicationGroup, Z, X, Y)
+    copyto!(Z, X * Y - Y * X)
```
sethaxen, line 1127:
```suggestion
    mul!(Z, X, Y)
    mul!(Z, Y, X, -1, true)
```

src/groups/group.jl
```diff
@@ -1062,3 +1120,10 @@ function group_exp!(G::MultiplicationGroup, q, X)
 end
 
 group_log!(::MultiplicationGroup, X::AbstractMatrix, q::AbstractMatrix) = log_safe!(X, q)
+
+lie_bracket(::MultiplicationGroup, X, Y) = X * Y - Y * X
```
sethaxen, line 1124:
```suggestion
lie_bracket(::MultiplicationGroup, X, Y) = mul!(X * Y, Y, X, -1, true)
```

src/groups/special_euclidean.jl
```diff
@@ -98,7 +98,14 @@ R & t \\
 \end{pmatrix}.
 ````
 
+This function embeds $\mathrm{SE}(n)$ in the general linear group $\mathrm{GL}(n+1)$.
+It is an isometric embedding and group homomorphism [^RicoMatrinex1988].
```
sethaxen, line 102:
```suggestion
It is an isometric embedding and group homomorphism [^RicoMartinez1988].
```

src/groups/special_euclidean.jl
```diff
@@ -98,7 +98,14 @@ R & t \\
...
+
 See also [`screw_matrix`](@ref) for matrix representations of the Lie algebra.
+
+[^RicoMatrinex1988]:
```
sethaxen, line 106:
```suggestion
[^RicoMartinez1988]:
```

sethaxen (commented):
Looks good! It may be the case that the approach for `SpecialEuclidean` works more generally for semidirect product groups. i.e. the normal subgroup's part of the lie bracket looks like the lie bracket of the normal subgroup. We'd have to think more carefully about the other part of the tangent vector though.

mateuszbaran:
> Looks good! It may be the case that the approach for `SpecialEuclidean` works more generally for semidirect product groups. i.e. the normal subgroup's part of the lie bracket looks like the lie bracket of the normal subgroup. We'd have to think more carefully about the other part of the tangent vector though.

There is such a thing as a semidirect sum of Lie algebras but it looks like in practice the structure can be most easily obtained by embedding is some higher dimensional general linear group. There are some more general details here: https://en.wikipedia.org/wiki/Lie_algebra_extension#By_semidirect_sum .

sethaxen:
> There is such a thing as a semidirect sum of Lie algebras but it looks like in practice the structure can be most easily obtained by embedding is some higher dimensional general linear group. There are some more general details here: https://en.wikipedia.org/wiki/Lie_algebra_extension#By_semidirect_sum .

Yeah that looks identical to what we see in the special case of semidirect products embedded in `GL(n+1)` with the affine representation, it just has the extra Lie bracket for the non-normal subgroup, which produces zero vectors in the case where we can use the affine representation. Seems like a good default to have.

src/groups/group.jl
mateuszbaran: Which machinery do you mean? I just checked that and I don't see anything that would make implementing this using `translate_diff` and `inverse_translate_diff` more efficient.

sethaxen: What I mean is to define
```suggestion
function adjoint_action(G::MultiplicationGroup, p, Xₑ)
    e = make_identity(G, p)
    Xₚ = translate_diff(G, p, e, Xₑ, LeftAction())
    Y = inverse_translate_diff(G, p, p, Xₚ, RightAction())
    return Y
end
```
This is general enough actually to work for non-multiplication groups. But let's take the case of `SO(n)`, where all tangent vectors are represented as their translation to the Lie algebra. What Julia will execute is something like
```julia
function adjoint_action(G::SpecialOrthogonal, p, Xₑ)
    e = one(p)
    Xₚ = Xₑ
    Y = p * Xₚ * p'
    return Y
end
```
The `Xₚ * p'` is much more efficient than `Xₚ / p`, which is what I'm referring to.

If you defines this generally for `AbstractGroup`, then for `TranslationGroup`, this would simplify to
```julia
function adjoint_action(G::TranslationGroup, p, Xₑ)
    e = zero(p)
    Xₚ = Xₑ
    Y = Xₚ
    return Y
end
```
so you almost don't even need to special-case for `AdditionGroup`.

mateuszbaran: Yes, that's a good point.

mateuszbaran:
I've read a bit about the representation theory of the adjoint representation and it is very nontrivial. I am no longer sure this is the right design :confused: .

sethaxen:
> I've read a bit about the representation theory of the adjoint representation and it is very nontrivial. I am no longer sure this is the right design 😕 .

What in particular are your concerns?

mateuszbaran:
> What in particular are your concerns?

Mostly the fact that adjoint representations of many groups and algebras are not faithful, and that we still won't have any generic faithful representations that would work on `FVector`s (if that is even possible?).

mateuszbaran:
OK, after thinking a bit more about it I think these methods can be useful anyway. I'd still like to finish differentials of translations and adjoint action on SE(n) (primarily SE(3)).

mateuszbaran:
@sethaxen Where did you get the formula for `function translate_diff!(G::SemidirectProductGroup, Y, p, q, X, conX::LeftAction)` from? I'm trying to understand it and I'm actually surprised that we can't for some reason get `translate_diff` for SE(n) through embedding it in GL(n+1)? Your code gives a different translation part of the returned tangent vector.

sethaxen:
> @sethaxen Where did you get the formula for `function translate_diff!(G::SemidirectProductGroup, Y, p, q, X, conX::LeftAction)` from? I'm trying to understand it and I'm actually surprised that we can't for some reason get `translate_diff` for SE(n) through embedding it in GL(n+1)? Your code gives a different translation part of the returned tangent vector.

I don't remember or if I worked it out myself. Can you share a snippet that fails?

mateuszbaran:
I guess nearly any invocation of `translate_diff` would differ between your version and the embedding in GL(n+1) but here is what I'm currently using for testing:
```julia
julia> using Manifolds, LinearAlgebra

julia> G = SpecialEuclidean(3);

julia> t = Vector{Float64}.([1:3, 2:4, 4:6]);

julia> ω = [[1.0, 2.0, 3.0], [3.0, 2.0, 1.0], [1.0, 3.0, 2.0]];

julia> p = Matrix(I, 3, 3);

julia> Rn = Rotations(3);

julia> pts = [ProductRepr(ti, exp(Rn, p, hat(Rn, p, ωi))) for (ti, ωi) in zip(t, ω)];

julia> X = ProductRepr([-1.0, 2.0, 1.0], hat(Rn, p, [1.0, 0.5, -0.5]));

julia> q = ProductRepr([0.0, 0.0, 0.0], p);

julia> translate_diff(G, pts[1], q, X)
ProductRepr with 2 submanifold components:
 Component 1 =
  3-element Vector{Float64}:
   2.2112553975587987
   0.5176292379367051
   0.9178287088559302
 Component 2 =
  3×3 Matrix{Float64}:
    0.0  0.5   0.5
   -0.5  0.0  -1.0
   -0.5  1.0   0.0

julia> X
ProductRepr with 2 submanifold components:
 Component 1 =
  3-element Vector{Float64}:
   -1.0
    2.0
    1.0
 Component 2 =
  3×3 Matrix{Float64}:
    0.0  0.5   0.5
   -0.5  0.0  -1.0
   -0.5  1.0   0.0
```
When calculating things using the embedding in GL(n+1) the left `translate_diff` would leave `X` unchanged.

sethaxen:
I wonder if maybe the difference is the choice in representation. For `GL(n+1)`, all tangent vectors `X` at point `p` are represented instead as `p \ X`. And while `SO(n)` adopts the same convention `SE(n)` does not, representing instead the rotational part of the tangent vector as `R^T Ω` but doing nothing to the translational part.

mateuszbaran:
That might be it. Why did you choose that convention? I think having semidirect products easily identifiable as subgrups of GL(M) would make defining new ones much easier and consistent, and this convention makes this identification hard.

sethaxen:
> That might be it. Why did you choose that convention?

Because it involved less divisions by points for `LeftAction`, which would probably be the most commonly used one. Otherwise almost every operation with tangent vectors uses division by a point. For `SL(n)` it also makes checking the validity of a tangent vector easier/cheaper.

mateuszbaran:
I'm slowly working out the more general form of differential of the right translation but it's quite challenging. It seems like we need for that differential of the group action with respect to the group element, and even coming up with notation isn't easy. I'm trying to document things as I go:
image
but this formula at the bottom seems like a quite horrible way to write that differential :confused: .

mateuszbaran:
I've split out the "connection manifold" part of the metric manifold. Ref: https://github.com/JuliaManifolds/Manifolds.jl/issues/387 .

mateuszbaran:
I've finally estabished a homomorphism between SE(n) and GL(n+1) so the differential of the right translation is now verified (and actually derived) this way. I didn't manage to fully derive it generally for semidirect product groups.

mateuszbaran:
@dehann I've added an example of adjoint action on SE(3): https://github.com/JuliaManifolds/Manifolds.jl/pull/381/commits/a3b811b36da7f58769c6bd86095941ca15bc92c6 . It shows that you can go about it two different ways: classic formulas that work on `FVector`s of coefficients (like here: https://github.com/JuliaManifolds/Manifolds.jl/pull/381/files#diff-c856e9189bb4b28cd5d9c2645bc7e8f4ddd98605f91831781dbc1f8217e1b200R87-R95 ) or the general way which uses differentials of left and right translations (and is general for any Lie groups): https://github.com/JuliaManifolds/Manifolds.jl/pull/381/files#diff-10808c095fb9edee2db8898db210f688390ae05268469026dd5001b3e3016008R366-R394 and they both agree as tested here: https://github.com/JuliaManifolds/Manifolds.jl/pull/381/files#diff-691a7da5df480d8515c7004a1809fb2975f5a40ef62f860d679c312be0ecb306R205-R219

dehann:
Hi @mateuszbaran , oh wow -- will work through this thanks!  

The part i'm paying most attention to at the moment is how ProductGroup or SEmidirectProductGroup should be used in general, plus studying up on the Lie bracket.

mateuszbaran:
The testing failure seems to be due to Project.toml incompatibility between ColorTypes and VisualRegressionTests, I've open a PR that will hopefully fix this: https://github.com/JuliaPlots/VisualRegressionTests.jl/pull/32

mateuszbaran:
I've added custom transparency rules for `GroupManifold` but now I'm a bit unsure whether `GroupManifold` itself should conditionally override the connection or, rather, overriding the connection should be done through `ConnectionManifold`. Perhaps the latter would be more consistent?

mateuszbaran:
With the latest commit we now have the right parallel transport on `SO(n)` :tada: . Sadly, for more generic Cartan-Schouten connections, I've only found parallel transport along one-parameter subgroups.

mateuszbaran:
@kellertuer what do you suggest as a temporary workaround to the problem of VisualRegressionTests not being correctly resolved on Julia 1.4? I guess it may take a while before VisualRegressionTests work with the newest ColorTypes (an older version seems to be fine but Julia 1.4 doesn't seem to be able to downgrade it).

I can disable recipe testing, disable testing on Julia 1.4 or add ColorTypes 0.10 as an explicit dependency of Manifolds.jl.

 https://github.com/JuliaPlots/VisualRegressionTests.jl/pull/32

kellertuer:
I feel like temporarily using ColorTypes 0.10 (can we add a comment in the Project.toml to state that its temporary?) would be my favourite option.

mateuszbaran:
For some bizarre reason ColorTypes.jl sometimes causes trouble and sometimes it does not. Weird.

mateuszbaran:
I think this PR adds/changes everything I intended to do for now. It should be safe to release it as non-breaking.

src/manifolds/ConnectionManifold.jl
```diff
@@ -0,0 +1,404 @@
...
+    B = induced_basis(M, A, i, TangentSpace)
+    sol = solve_exp_ode(M, p, X, tspan, B; dense=false, saveat=[1.0])
+    n = length(p)
+    return copyto!(q, sol.u[1][(n + 1):end])
+end
+
+"""
+    gaussian_curvature(M::AbstractManifold, p, B::AbstractBasis; backend::AbstractDiffBackend = diff_backend())
+
+Compute the Gaussian curvature of the manifold `M` at the point `p` using basis `B`.
```
kellertuer, line 160: Can we maybe add a formula here?

src/manifolds/ConnectionManifold.jl
```diff
@@ -0,0 +1,404 @@
...
+    return ricci_curvature(M, p, B; kwargs...) / 2
+end
+@decorator_transparent_signature gaussian_curvature(
+    M::AbstractDecoratorManifold,
+    p,
+    B::AbstractBasis;
+    kwargs...,
+)
+
+function injectivity_radius(M::AbstractConnectionManifold, p)
```
kellertuer, line 173: Could this instead be done with transparency?

src/manifolds/ConnectionManifold.jl
```diff
@@ -0,0 +1,404 @@
...
+end
+function injectivity_radius(M::AbstractConnectionManifold, p, m::ExponentialRetraction)
+    return injectivity_radius(base_manifold(M), p, m)
+end
+
+"""
+    ricci_curvature(M::AbstractManifold, p, B::AbstractBasis; backend::AbstractDiffBackend = diff_backend())
+
+Compute the Ricci scalar curvature of the manifold `M` at the point `p` using basis `B`.
+"""
```
kellertuer, line 193: Can we add a formula here or a reference to the tensor?

src/manifolds/ConnectionManifold.jl
```diff
@@ -0,0 +1,404 @@
...
+    B::AbstractBasis;
+    kwargs...,
+)
+
+@doc raw"""
+    riemann_tensor(M::AbstractManifold, p, B::AbstractBasis; backend::AbstractDiffBackend = diff_backend())
+
+Compute the Riemann tensor ``R^l_{ijk}``, also known as the Riemann curvature
+tensor, at the point `p`. The dimensions of the resulting multi-dimensional
+array are ordered ``(l,i,j,k)``.
```
kellertuer, line 240: Could we add a formula here or a reference to wikipedia?

kellertuer (commented):
This turned out to be an impressive amount of modelling and work; great! And also great to see that we will have connections then.

I have a few remarks but most are documentation and only one on modelling.

docs/src/manifolds/metric.md
```diff
@@ -17,6 +17,8 @@ Pages = ["metric.md"]
 Depth = 2
 ```
 
+Note that a metric manifold is an [`AbstractConnectionManifold`](@ref) with the [`LeviCivitaConnection`](@ref) of the metric $g$, and thus a large part of metric manifold's functionality relies on this.
```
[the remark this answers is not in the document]
mateuszbaran: Yes, I was positively surprised how easy it actually was to introduce connections.

src/groups/circle_group.jl
```diff
@@ -14,6 +14,10 @@ invariant_metric_dispatch(::CircleGroup, ::ActionDirection) = Val(true)
 
 default_metric_dispatch(::MetricManifold{ℂ,CircleGroup,EuclideanMetric}) = Val(true)
 
+adjoint_action(::CircleGroup, p, X) = X
```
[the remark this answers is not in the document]
mateuszbaran: It should be because we may want to work on tangent vectors that are just numbers (and thus immutable).

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
+
+"""
+    AbstractCartanSchoutenConnection
+
+Abstract type for Cartan-Schouten connections, that is connections whose geodesics
+going through group identity are one-parameter subgroups. See [^Pennec2020] for details.
+
+[^Pennec2020]:
+    > X. Pennec and M. Lorenzi, “5 - Beyond Riemannian geometry: The affine connection
```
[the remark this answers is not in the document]
mateuszbaran: Yes, that's quite weird. I've taken this title from https://www.sciencedirect.com/science/article/pii/B9780128147252000121 IIRC and it has the chapter number in front of chapter title.

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
...
+    > setting for transformation groups,” in Riemannian Geometric Statistics in Medical Image
+    > Analysis, X. Pennec, S. Sommer, and T. Fletcher, Eds. Academic Press, 2020, pp. 169–229.
+    > doi: 10.1016/B978-0-12-814725-2.00012-1.
+"""
+abstract type AbstractCartanSchoutenConnection <: AbstractAffineConnection end
+
+"""
```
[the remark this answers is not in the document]
mateuszbaran: It's a bit complicated. Each Lie group has at least a one-parameter family of Cartan-Schouten connections, but if the Lie algebra is commutative then the are all the same. These plus, minus and zero connections belong to this family so we could technically have a struct with a number that represents all of these. In practice these three seem to be the most important so for performance I think they should have different types.

Moreover, some groups have Cartan-Schouten connections that don't belong to this one-parameter family but I have no idea what their significance is.

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
...
+    p,
+    q,
+) where {𝔽}
+    pinvq = compose(M.manifold, inv(M.manifold, p), q)
+    group_log!(M.manifold, Y, pinvq)
+    return translate_diff!(M.manifold, Y, p, Identity(M.manifold, p), Y)
+end
+
+"""
+    vector_transport_to(M::CartanSchoutenMinusGroup, ::Identity, X, q, ::ParallelTransport)
```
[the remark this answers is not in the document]
mateuszbaran: Good point. The formulas in [^Pennec2020] for vector transport in Cartan-Schouten connections are given along one-parameter subgroups but "minus" and "plus" connections give flat geometry so it should be safe to have them defined for `X` tangent at any point.

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
...
+    > doi: 10.1016/B978-0-12-814725-2.00012-1.
+"""
+vector_transport_to(M::CartanSchoutenPlusGroup, Y, ::Identity, X, q, ::ParallelTransport)
+
+function vector_transport_to!(M::CartanSchoutenPlusGroup, Y, p, X, q, ::ParallelTransport)
+    return inverse_translate_diff!(M.manifold, Y, q, p, X, RightAction())
+end
+
+"""
+    vector_transport_direction(M::CartanSchoutenZeroGroup, ::Identity, X, d, ::ParallelTransport)
```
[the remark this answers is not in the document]
mateuszbaran: It could be slightly generalized but the issue is that [^Pennec] only describes parallel transport along one-parameter subgroups and the "zero" connection is in general not flat, so I don't know if there are explicit formulas for vector transport along arbitrary geodesics.

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
...
+Transport tangent vector `X` at identity on the group manifold with the
+[`CartanSchoutenZero`](@ref) connection in the direction `d`. See [^Pennec2020] for details.
+
+[^Pennec2020]:
+    > X. Pennec and M. Lorenzi, “5 - Beyond Riemannian geometry: The affine connection
+    > setting for transformation groups,” in Riemannian Geometric Statistics in Medical Image
+    > Analysis, X. Pennec, S. Sommer, and T. Fletcher, Eds. Academic Press, 2020, pp. 169–229.
+    > doi: 10.1016/B978-0-12-814725-2.00012-1.
+"""
+vector_transport_direction(
```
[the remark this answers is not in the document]
mateuszbaran: Very good question. In [^Pennec] there are formulas for `vector_tranport_to`, and for the plus and minus connections the only way I know to transport in a direction is taking the group logarithm of a point. On the other hand, the formula for the zero connection is quite a bit different -- you don't have to compute any logarithms if you have a direction.

Maybe there is a way to rewrite the formulas for the plus and minus connections to make it more similar, I didn't check.

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
...
+    X,
+    d,
+    ::ParallelTransport,
+)
+    dexp_half = group_exp(M.manifold, d / 2)
+    translate_diff!(M.manifold, Y, dexp_half, p, X, RightAction())
+    return translate_diff!(M.manifold, Y, dexp_half, p, Y, LeftAction())
+end
+
+function vector_transport_to!(
```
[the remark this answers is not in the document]
mateuszbaran: In general it's only correct along one parameter subgroups, so I restricted it to elements of the Lie algebra (even though it's a bit more general).

I didn't document it because, for consistency here, I've only put documentation for `vector_transport_direction` methods and this method of `vector_transport_to` is fairly straightforward. Should I just document this one method or explain the design differently?

src/groups/connections.jl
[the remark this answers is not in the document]
mateuszbaran: There isn't much they have in common except that they all belong to the same one-parameter family, so the improving readability of the name would be the only point of this refactoring. The problem is, `CartanSchouten{<:Symbol}` might suggest that you can get any Cartan-Schouten connection using this type but actually it would only cover these three.

src/groups/semidirect_product_group.jl
```diff
@@ -112,6 +112,23 @@ function compose!(G::GT, x, e::E, ::E) where {GT<:SemidirectProductGroup,E<:Iden
     return identity!(G, x, e)
 end
 
+@doc raw"""
```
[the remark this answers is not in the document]
mateuszbaran: I'll change this, I just didn't pay enough attention here :slightly_smiling_face: .

src/groups/special_euclidean.jl
```diff
@@ -84,6 +84,15 @@ Base.@propagate_inbounds function _padvector!(
     return X
 end
 
+function adjoint_action(::SpecialEuclidean{3}, p, fX::TFVector{<:Any,VeeOrthogonalBasis{ℝ}})
```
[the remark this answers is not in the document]
mateuszbaran: I'll add documentation. This method is mostly about demonstrating possibilities of the interface, and only works on small input, so `SArray` would be the most likely argument type and then the non-mutating version can't be efficiently implemented through the mutating one.

src/groups/special_euclidean.jl
```diff
@@ -386,3 +405,108 @@ function group_log!(G::SpecialEuclidean{3}, X, q)
     @inbounds _padvector!(G, X)
     return X
 end
+
+function lie_bracket(G::SpecialEuclidean, X::ProductRepr, Y::ProductRepr)
```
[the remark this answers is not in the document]
mateuszbaran: I think we're going to need a lot more non-mutating variants explicitly implemented but maybe this one isn't actually particularly important.

src/groups/connections.jl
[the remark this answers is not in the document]
kellertuer: Ah right, with the last argument, you convinced me. While I still feel the names are too long, the parameter approach would yield a wrong impression maybe.

src/groups/special_euclidean.jl
```diff
@@ -386,3 +405,108 @@ function group_log!(G::SpecialEuclidean{3}, X, q)
...
+    @inbounds _padvector!(G, Y)
+end
+function translate_diff!(G::SpecialEuclidean, Y, ::Identity, ::Identity, X, ::RightAction)
+    copyto!(G, Y, X)
+    @inbounds _padvector!(G, Y)
+    return Y
+end
+
+@doc raw"""
+    SE_in_GL
```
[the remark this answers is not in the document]
mateuszbaran: Yes, I don't like this name but `SpecialEuclideanEmbeddedInGeneralLinear` sounds even worse.

src/manifolds/ConnectionManifold.jl
mateuszbaran: Actually I don't think this is even correct, I just copied it from the metric manifold. I'd prefer to resolve this in a separate PR.

kellertuer: Ok, that's also fine with me.

src/groups/connections.jl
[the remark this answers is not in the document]
kellertuer: I am also not sure, I was just surprised that is was documented for a different signature than it was implemented – in the other two cases, here it is ok.

src/groups/semidirect_product_group.jl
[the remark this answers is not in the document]
mateuszbaran: This is fixed now.

src/manifolds/ConnectionManifold.jl
```diff
@@ -190,6 +196,12 @@ end
     ricci_curvature(M::AbstractManifold, p, B::AbstractBasis; backend::AbstractDiffBackend = diff_backend())
 
 Compute the Ricci scalar curvature of the manifold `M` at the point `p` using basis `B`.
+The curvature is computed as the trace of the Ricci curvature tensor with respect to
+the metric, that is ``R=g^{ij}_R_{ij}`` where ``R`` is the scalar Ricci curvature at `p`,
```
kellertuer, line 200: What is `g^{ij}_R_{ij}` supposed to be? Currently this would yield a double subscripts error.

mateuszbaran: Just a typo, I'm fixing it right now.

mateuszbaran: It should be fine now.

src/manifolds/ConnectionManifold.jl
mateuszbaran: Done :slightly_smiling_face:

src/groups/connections.jl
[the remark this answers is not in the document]
kellertuer: Ok, then let's leave it at that for now.

mateuszbaran:
I think I've addressed all issues now, except the `SE_in_GL` which I don't know what to do about -- it looks like a great use case for `EmbeddedManifold` but there seems to be no nice name for it.

kellertuer:
What do we do about the vector transports that are documented for identity but implemented more general? I hope I have marked all resolved questions as resolved, so could you once more check the ones I did not yet mark?

For the name: What about `SpecialEuclideanInGeneralLinear`? Maybe a little too long but I thin nicer than the short form?

mateuszbaran:
I've improved documentation of vector transports today, is it still inconsistent somewhere? The currently unresolved comments are:

1. This: https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664315395 where the conclusion seems to be that the current design is OK.
2. Here: https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664316401, https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664316760, and this: https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664318000 which should be (I hope) resolved by changes I've pushed today.
3. Here: https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664575128, and I'll rename that constant to `SpecialEuclideanInGeneralLinear`.
4. https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664322782 and there: https://github.com/JuliaManifolds/Manifolds.jl/pull/381#discussion_r664326751 I've added references today, and I think it should be fine to have some explicit non-mutating variants for the special Euclidean, special orthogonal and translations groups for better interoperability with `SArray` and AD.

src/groups/connections.jl
```diff
@@ -0,0 +1,182 @@
...
+Transport tangent vector `X` at point `p` on the group manifold `M` with the
+[`CartanSchoutenPlus`](@ref) connection to point `q`. See [^Pennec2020] for details.
+
+[^Pennec2020]:
+    > X. Pennec and M. Lorenzi, “5 - Beyond Riemannian geometry: The affine connection
+    > setting for transformation groups,” in Riemannian Geometric Statistics in Medical Image
+    > Analysis, X. Pennec, S. Sommer, and T. Fletcher, Eds. Academic Press, 2020, pp. 169–229.
+    > doi: 10.1016/B978-0-12-814725-2.00012-1.
+"""
+vector_transport_to(M::CartanSchoutenPlusGroup, Y, ::Identity, X, q, ::ParallelTransport)
```
kellertuer, line 124:
```suggestion
vector_transport_to(M::CartanSchoutenPlusGroup, ::Identity, X, q, ::ParallelTransport)
```

src/groups/connections.jl
```diff
@@ -0,0 +1,174 @@
...
+    > doi: 10.1016/B978-0-12-814725-2.00012-1.
+"""
+vector_transport_to(M::CartanSchoutenMinusGroup, Y, ::Identity, X, q, ::ParallelTransport)
+
+function vector_transport_to!(M::CartanSchoutenMinusGroup, Y, p, X, q, ::ParallelTransport)
+    return inverse_translate_diff!(M.manifold, Y, q, p, X, LeftAction())
+end
+
+"""
+    vector_transport_to(M::CartanSchoutenPlusGroup, ::Identity, X, q, ::ParallelTransport)
```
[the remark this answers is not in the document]
kellertuer: Well this is – both in the documentation and the signature for the documentation, still done for `::Identity`, while the generic implementation is still done for a.. `p`, so my question remains whether this might be inconsistent?

kellertuer:
Thanks for the summary, the ones where the line got changed too much did not show up in the “Files Changed” view so I must have missed them that way; I only have the one open one left I just commented on. Thanks for carefully working on all my remarks.

src/groups/connections.jl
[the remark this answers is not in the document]
mateuszbaran: Right, it should be fixed now.

mateuszbaran:
Thanks for a review. This is a quite significant set of changes so it's good to have them checked :slightly_smiling_face:.

kellertuer (approved):

kellertuer:
Once the tests have passed, I think this can be merged, though we loose two lines in coverage (if you want, you could check for these) but it is within the limit of our tolerance.

mateuszbaran:
Great! These two lines reported as not covered look like false positives to me so I'd ignore that.

mateuszbaran:
Julia 1.4 CI seems broken :confused:

kellertuer:
It seems that the call to `ambigs = Test.detect_ambiguities(Base)` does not finish in the `our_base_ambiguities()` function (tested locally on my machine with Julia 1.4. Commenting that base ambiguities test out, the test here locally run fine on Julia 1.4

mateuszbaran:
Weird, from my tests it looks like `DoubleFloats` alone stops ambiguity detection from completion. I'll move this test to newer versions and bump the number of ambiguities accordingly.

mateuszbaran:
Well, on 1.6 with all the other packages loaded it also takes way too long.

PR #388, opened 2021-06-23
Adapt Manifolds to new Decorator type

kellertuer:
...that was introduced in ManifoldsBase 0.12.0

This PR also 
* [x] Reduces code lines in MetricManifold
* [x] Reduces code lines in Group Manifold
* [x] checks/updates docs


kellertuer:
The main rework is done, however there is more tests failing than I thought, so I might have missed something, hopefully just for the group manifold, since there we do not have explicit checks for transparency.

kellertuer:
The group manifold has some little strange transparency tests that I do not yet understand, maybe @sethaxen you wrote them?
Why are there tests checking for a function with parameters (https://github.com/JuliaManifolds/Manifolds.jl/blob/8a5d654486dd9259936367fa998a2e22100a1735/test/groups/groups_general.jl#L33-L41) that I understand and that are currently covered (or I hope they are) and some that are without parameters (https://github.com/JuliaManifolds/Manifolds.jl/blob/8a5d654486dd9259936367fa998a2e22100a1735/test/groups/groups_general.jl#L141-L156) these I do not understand, especially since they are different to the first ones...

mateuszbaran:
> Edit: Since I seem to have last touched (git blame) the second part of lines, I might propose now to maybe delete those since I do not know what they are supposed to test.

IIRC such tests are there to just notify that transparency changed. If the change is deliberate, you can adjust the test. If it's not, It'd be better to fix the code.

kellertuer:
The main problem is, that the two contradict, so with parameters is different form without, and I think that is not only strange but also that without parameters the testing is not even well defined nor useful?

mateuszbaran:
Ah, OK, so you can just remove these tests then I think.

kellertuer:
Hm, seems that I have introduced some side effects that I did not intent to; I will have to check closer what is going wrong currently. Sorry that it will take a while, since it seems the CompatHelper still gives us an update every day ( I thought that was fixed?)

mateuszbaran:
> Hm, seems that I have introduced some side effects that I did not intent to; I will have to check closer what is going wrong currently.

No problem. I will continue my work on connections after this is done but I have plenty of other things to do :slightly_smiling_face: . Including writing of better documentation for groups (it's very basic right now).

> it seems the CompatHelper still gives us an update every day ( I thought that was fixed?)

Maybe 0.12.1 release triggered it again? We could easily check :slightly_smiling_face:

kellertuer:
I hope I finally fixed the tests.

* There were two kinds of transparencies used. Instead of having a Hierarchy that (as currently) goes first to the decorator then the transparency type, we could maybe do it the other way around, that would reduce code by a few lines (i.e. Having a TransparentDecoratorType as a super type for TransparentEmbedding and TransparentGroup).
* one function had to be reimplemented: `base_manifold`.

kellertuer:
The 13 lines we miss are due to macros not being checked correctly (I shortened code there by replacing a function definition and a `@signature` with the direct `@function`. I will look at the lins closely when investigating the documentation.

kellertuer:
The 13 lines we still miss might be due to an issue like https://github.com/JuliaLang/julia/issues/36283

mateuszbaran:
Yes, all of these failures look like false positives.

docs/src/interface.md
```diff
@@ -242,13 +242,16 @@ Then
 2. implement `f(DM::OurDecoratorManifold, arg1, arg2)`
 
 This makes it possible to extend a manifold or all manifolds with a feature or replace a feature of the original manifold.
+
 The [`MetricManifold`](@ref) is the best example of the second case, since the default metric indicates for which metric the manifold was originally implemented, such that those functions are just passed through.
 This can best be seen in the [`SymmetricPositiveDefinite`](@ref) manifold with its [`LinearAffineMetric`](@ref).
 
+A final techical note – if several manifold have similar transparency rules concerning functions from the interface, the last parameter `T` of the [`AbstractDecoratorManifold`](@ref)`{𝔽,T<:`[`AbstractDecoratorType`](@ref)`}`. Can be used to dispatch on different transparency schemes.
```
mateuszbaran, line 249:
```suggestion
A final technical note – if several manifolds have similar transparency rules concerning functions from the interface, the last parameter `T` of the [`AbstractDecoratorManifold`](@ref)`{𝔽,T<:`[`AbstractDecoratorType`](@ref)`}` can be used to dispatch on different transparency schemes.
```
This is a bit unclear. Do you want to emphasize here that a single manifold type can have several different transparency rules or that several different manifolds can share the same transparency rules?

docs/src/interface.md
```diff
@@ -261,7 +264,8 @@ Order = [:macro, :type, :function]
 
 ## ValidationManifold
 
-[`ValidationManifold`](@ref) is a simple decorator that “decorates” a manifold with tests that all involved arrays are correct. For example involved input and output paratemers are checked before and after running a function, repectively.
+[`ValidationManifold`](@ref) is a simple decorator using the [`AbstractDecoratorManifold`](@ref) that “decorates” a manifold with tests that all involved arrays are correct.
```
mateuszbaran, line 267:
```suggestion
[`ValidationManifold`](@ref) is a simple decorator using the [`AbstractDecoratorManifold`](@ref) that “decorates” a manifold with tests that all involved points and vectors are valid for the wrapped manifold.
```
It also works when points are represented by more complicated things than an array.

docs/src/interface.md
```diff
@@ -278,7 +282,8 @@ Similar to the metric and [`MetricManifold`](@ref), an embedding is often implic
 We introduce the embedded manifolds hence as an [`AbstractDecoratorManifold`](@ref).
 
 This decorator enables to use such an embedding in an transparent way.
-Different types of embeddings can be distinguished using the [`AbstractEmbeddingType`](@ref).
+Different types of embeddings can be distinguished using the [`AbstractEmbeddingType`](@ref),
+which is am [`AbstractDecoratorType`](@ref).
```
mateuszbaran, line 286:
```suggestion
which is an [`AbstractDecoratorType`](@ref).
```

src/groups/group.jl
```diff
@@ -21,6 +21,29 @@ Note that a manifold is connected with an operation by wrapping it with a decora
 """
 abstract type AbstractGroupOperation end
 
+"""
+    abstract type AbstractGroupDecroatorType <: AbstractDecoratorType
+
+    A common decorator type for all group decorators
```
mateuszbaran, line 27:
```suggestion
A common decorator type for all group decorators.
```

src/groups/group.jl
```diff
@@ -21,6 +21,29 @@ Note that a manifold is connected with an operation by wrapping it with a decora
...
+"""
+abstract type AbstractGroupDecoratorType <: AbstractDecoratorType end
+
+"""
+    struct DefaultGroupDecoratorType <: AbstractDecoratorType
+
+A default group decorator type with no special properties
```
mateuszbaran, line 34:
```suggestion
The default group decorator type with no special properties.
```
How exactly does this differ from `TransparentGroupDecoratorType`?

src/groups/group.jl
```diff
@@ -21,6 +21,29 @@ Note that a manifold is connected with an operation by wrapping it with a decora
...
+"""
+struct DefaultGroupDecoratorType <: AbstractGroupDecoratorType end
+"""
+    struct TransparentGroupDecoratorType <: AbstractDecoratorType
+
+A transparent grou pdecorator type that acts transparent similar to
```
mateuszbaran, line 40:
```suggestion
A transparent group decorator type that acts transparently, similar to
```

src/groups/group.jl
```diff
@@ -21,6 +21,29 @@ Note that a manifold is connected with an operation by wrapping it with a decora
...
+the [`TransparentIsometricEmbedding`](@ref), i.e. it inherits the
```
mateuszbaran, line 41:
```suggestion
the [`TransparentIsometricEmbedding`](@ref), i.e. it passes through all metric-related functions such as
```

src/groups/group.jl
```diff
@@ -21,6 +21,29 @@ Note that a manifold is connected with an operation by wrapping it with a decora
...
+logarithmic and exponential map as well as retraction and inverse retractions
+from the manifold it decorates.
```
mateuszbaran, line 43:
```suggestion
to the manifold it decorates.
```

mateuszbaran: (it fixes the sentence if the comment above is applied)

src/groups/group.jl
```diff
@@ -45,7 +68,7 @@ Group manifolds by default forward metric-related operations to the wrapped mani
     GroupManifold(manifold, op)
 """
 struct GroupManifold{𝔽,M<:AbstractManifold{𝔽},O<:AbstractGroupOperation} <:
```
mateuszbaran, line 70: I'd make the transparency type a parameter of `GroupManifold`. And it looks weird that `DefaultGroupDecoratorType` is not the default, probably the name is misleading.

kellertuer: That would mean quite some rework, but sure we can do that. The naming follows that of the `EmbeddedManifold`, and the transparent case is used far more often. With a Parameter in `GroupManifold` there is no default anymore, but we have to add the transparent type then for all but 2 groups I think.

kellertuer: I gave it a try and made it a parameter of the group manifold. That yields about 100+ errors so possibly _quite_ _quite_ a lot of rework and much breaking changes.

kellertuer: For example the constructor will not work anymore and we would have to think about how to model that. That constructor is heavily used, too.

mateuszbaran: Currently `GroupManifold` is used to define `SemidirectProductGroup`, and that is used for `SpecialEuclidean`. And `SpecialEuclidean` needs to have different connections available, and only one of them would be transparent. Similarly for `SpecialOrthogonal`. This isn't needed for the current state of Manifolds.jl but I think I need this for #381 . Anyway, I can also add this parameter in my PR :slightly_smiling_face: .

mateuszbaran: OK, so I'll do this rework in #381.

src/groups/group.jl
```diff
@@ -21,6 +21,29 @@ Note that a manifold is connected with an operation by wrapping it with a decora
 """
 abstract type AbstractGroupOperation end
 
+"""
+    abstract type AbstractGroupDecroatorType <: AbstractDecoratorType
+
+A common decorator type for all group decorators.
```
mateuszbaran, line 27:
```suggestion
A common decorator type for all group decorators.
It is similar to [`DefaultEmbeddingType`](@ref) but for groups.
```

src/groups/group.jl
kellertuer: Ok, great if you could take that over. I tried and for me it seems like at least half a day to rework this main type of group manifolds. And I did this rework already when introducing the transparency type here. It is not even a fun work.

mateuszbaran (approved):
OK, I think this is fine now :slightly_smiling_face:

kellertuer:
Nice! I will wait for the tests and merge maybe in the evening or tomorrow morning.

PR #392, opened 2021-06-25
Fixes a bug, that occured when calling `exp!(Rotations(2), p, p, X)` since `q=p` was modified.

kellertuer:
So the `exp!` returned a wrong result when called as `exp!(M, p, p X)` due to this side effect. Also adds a test for this.

### DIFF
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -165,7 +165,7 @@ function exp!(M::Rotations{2}, q, p, X)
     @assert size(q) == (2, 2)
     θ = get_coordinates(M, p, X, DefaultOrthogonalBasis())[1]
     sinθ, cosθ = sincos(θ)
-    return copyto!(q, p * [cosθ sinθ; -sinθ cosθ])
+    return copyto!(q, p * SA[cosθ sinθ; -sinθ cosθ])
 end
 function exp!(M::Rotations{3}, q, p, X)
     θ = norm(M, p, X) / sqrt(2)
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -165,7 +165,7 @@ function exp!(M::Rotations{2}, q, p, X)
     @assert size(q) == (2, 2)
     θ = get_coordinates(M, p, X, DefaultOrthogonalBasis())[1]
     sinθ, cosθ = sincos(θ)
-    return copyto!(q, p * SA[cosθ sinθ; -sinθ cosθ])
+    return copyto!(q, p * SA[cosθ -sinθ; sinθ cosθ])
 end
 function exp!(M::Rotations{3}, q, p, X)
     θ = norm(M, p, X) / sqrt(2)
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -165,13 +165,7 @@ function exp!(M::Rotations{2}, q, p, X)
     @assert size(q) == (2, 2)
     θ = get_coordinates(M, p, X, DefaultOrthogonalBasis())[1]
     sinθ, cosθ = sincos(θ)
-    @inbounds begin
-        q[1] = cosθ
-        q[2] = sinθ
-        q[3] = -sinθ
-        q[4] = cosθ
-    end
-    return copyto!(q, p * q)
+    return copyto!(q, p * [cosθ sinθ; -sinθ cosθ])
 end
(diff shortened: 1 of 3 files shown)


test/rotations.jl
```diff
@@ -62,6 +62,12 @@ include("utils.jl")
         v = log(M, pts[1], pts[2])
         @test norm(M, pts[1], v) ≈ (angles[2] - angles[1]) * sqrt(2)
 
+        # check that exp! does not have a side effect
+        q = deeocopy(pts[1])
```
mateuszbaran, line 66:
```suggestion
        q = allocate(pts[1])
```
Let's not overuse `deepcopy`.

src/manifolds/Rotations.jl
```diff
@@ -165,13 +165,7 @@ function exp!(M::Rotations{2}, q, p, X)
...
     θ = get_coordinates(M, p, X, DefaultOrthogonalBasis())[1]
     sinθ, cosθ = sincos(θ)
-    @inbounds begin
-        q[1] = cosθ
-        q[2] = sinθ
-        q[3] = -sinθ
-        q[4] = cosθ
-    end
-    return copyto!(q, p * q)
+    return copyto!(q, p * [cosθ sinθ; -sinθ cosθ])
```
mateuszbaran, line 168:
```suggestion
    return copyto!(q, p * SA[cosθ sinθ; -sinθ cosθ])
```
Static array should be faster.

mateuszbaran (commented):
Actually, this may be a way more common problem. IIRC many mutating operations assume that there is no memory aliasing between output and input arguments. We may need to check it more thoroughly (including tests in general manifold testing).

src/manifolds/Rotations.jl
kellertuer: Do we have that as a direct dependence within manifolds?

test/rotations.jl
kellertuer: No, first of all this is a test, so performance should not matter that much, second, I really want the values of `pts[1]` inside, since in `exp!(M, q, q, v)` it is both input and output for `exp!`.

src/manifolds/Rotations.jl
kellertuer: Ah, we do, then, sure, why not.

test/rotations.jl
mateuszbaran: Ah, sorry, it should be `copy(M, pts[1])`, and my concern is more about style than performance (`deepcopy` won't do the right thing when copying a point for some manifolds).

kellertuer: Well seen.

test/rotations.jl
```diff
@@ -62,6 +62,12 @@ include("utils.jl")
         v = log(M, pts[1], pts[2])
         @test norm(M, pts[1], v) ≈ (angles[2] - angles[1]) * sqrt(2)
 
+        # check that exp! does not have a side effect
+        q = copy(M, pts[1])
```
mateuszbaran, line 66: We could actually move this test to general manifold tests. It is something `Manopt.jl` assumes so checking it for all manifolds makes sense IMO.

kellertuer: `Manopt.jl` would be maybe a little generic, actually _I_ always assumed that that is the case ;)

kellertuer: Since you also opened an Issue for the generic case, maybe let‘s solve it in a generic separate PR?

mateuszbaran: OK, sure, I'll try that in a separate PR in the evening today or tomorrow.

mateuszbaran (approved):

test/rotations.jl
kellertuer: Hm, we do not have the copy you proposed to use.

kellertuer:
There seems to be a problem with invariant metrics (maybe transparency?) I am not sure what to do there since there are some values switched in the corresponding approx and I do not see how that could have happened for a larger Rotations that we did not touch.

test/rotations.jl
mateuszbaran: Weird, I think we should have that `copy`.

kellertuer: I agree, we should, since we have `copyto!`, but maybe lets do a new PR (even in ManifoldsBase?) it would just combining existing methods.

mateuszbaran: Sure, it's an issue for `ManifoldsBase.jl`.

PR #394, opened 2021-06-27
Add in-place tests

### LINKED ISSUE #393: Do not assume that there is no memory aliasing between output of mutating functions and their input arguments.
It can lead to hard-to-find bugs, see #392 .

kellertuer:
This PR adds tests for inlace checks for the following mutating variants

* [x] exp!
* [x] vector transports (not for `_along` since we do not have proper tests)
* [x] retractions
* [x] mid point  

For `log!`, `inverse_retract!`, `embed!`, `project!` such tests do not seem too useful, since often the dimension / return type (vector instead of point) changes, so then in-place is anyways not recommended in the generic interface. Am I missing any others where a mutating return might be given as input?

## Todo

* [x] Which manifolds to activate?
* [x] Wait for #388 to be finished 
* [x] Wait for `copy` from JuliaManifolds/ManifoldsBase.jl#77 to be available to replace `deepcopy`s

This resolves #393.

### DIFF
--- a/src/manifolds/VectorBundle.jl
+++ b/src/manifolds/VectorBundle.jl
@@ -274,7 +274,14 @@ function exp!(B::VectorBundle, q, p, X)
     VXM, VXF = submanifold_components(B.manifold, X)
     # this temporary avoids overwriting `p` when `q` and `p` occupy the same memory
     xqt = exp(B.manifold, xp, VXM)
-    vector_transport_to!(B.manifold, Xq, xp, Xp + VXF, xqt, B.vector_transport.method_point)
+    vector_transport_direction!(
+        B.manifold,
+        Xq,
+        xp,
+        Xp + VXF,
+        VXM,
+        B.vector_transport.method_point,
+    )
     copyto!(B.manifold, xq, xqt)
     return q
 end
--- a/src/manifolds/VectorBundle.jl
+++ b/src/manifolds/VectorBundle.jl
@@ -272,8 +272,10 @@ function exp!(B::VectorBundle, q, p, X)
     xp, Xp = submanifold_components(B.manifold, p)
     xq, Xq = submanifold_components(B.manifold, q)
     VXM, VXF = submanifold_components(B.manifold, X)
-    exp!(B.manifold, xq, xp, VXM)
-    vector_transport_to!(B.manifold, Xq, xp, Xp + VXF, xq, B.vector_transport.method_point)
+    # this temporary avoids overwriting `p` when `q` and `p` occupy the same memory
+    xqt = exp(B.manifold, xp, VXM)
+    vector_transport_to!(B.manifold, Xq, xp, Xp + VXF, xqt, B.vector_transport.method_point)
+    copyto!(B.manifold, xq, xqt)
     return q
 end
 function exp!(M::TangentSpaceAtPoint, q, p, X)
--- a/src/tests/tests_general.jl
+++ b/src/tests/tests_general.jl
@@ -51,11 +51,14 @@ that lie on it (contained in `pts`).
     tests (1 use default). This is deactivated if the `exp_log_atol_multiplier` is nonzero.
 - `retraction_methods = []`: retraction methods that will be tested.
 - `test_atlases = []`: Vector or tuple of atlases that should be tested.
-- `test_exp_log = true`: if true, checkthat [`exp`](@ref) is the inverse of [`log`](@ref).
+- `test_exp_log = true`: if true, check that [`exp`](@ref) is the inverse of [`log`](@ref).
 - `test_forward_diff = true`: if true, automatic differentiation using
     ForwardDiff is tested.
 - `test_injectivity_radius = true`: whether implementation of [`injectivity_radius`](@ref)
     should be tested.
+- `test_inplace = false` : if true check if inplace variants work if they are activated,
+   e.g. check that `exp!(M, p, p, X)` work if `test_exp_log = true`.
+   This in general requires `is_mutating` to be true.
 - `test_is_tangent`: if true check that the `default_inverse_retraction_method`
     actually returns valid tangent vectors.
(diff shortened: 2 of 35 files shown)


mateuszbaran:
> [ ] Which manifolds to activate?

I think ideally all manifolds should have this enabled.

src/tests/tests_general.jl
```diff
@@ -459,11 +482,28 @@ function test_manifold(
...
+                                vector_transport_to!(M, X1a, pts[1], X1a, pts32, vtm)
+                                Test.@test X1a == Xt
+                            end
                     end
                     if test_dir
                         v1t2_m = allocate(v1t2)
                         vector_transport_direction!(M, v1t2_m, pts[1], X1, X2, vtm)
                         Test.@test isapprox(M, pts32, v1t2, v1t2_m)
+                        test_inplace &&
+                            Test.@testset "inplace test for vector_transport_to!" begin
```
mateuszbaran, line 498:
```suggestion
                            Test.@testset "inplace test for vector_transport_direction!" begin
```

src/tests/tests_general.jl
```diff
@@ -584,6 +624,15 @@ function test_manifold(
             mpm = allocate(mp)
             mid_point!(M, mpm, pts[1], pts[2])
             Test.@test isapprox(M, mpm, mid_point12; atol=atolp1p2, rtol=rtolp1p2)
+            test_inplace && Test.@testset "inplace test for vector_transport_to!" begin
```
mateuszbaran, line 627:
```suggestion
            test_inplace && Test.@testset "inplace test for midpoint!" begin
```

mateuszbaran:
> Am I missing any others where a mutating return might be given as input?

These functions I think are most likely to have problems with in-place mutation and we can always add more so I wouldn't worry about adding more functions in this PR.

src/tests/tests_general.jl
```diff
@@ -272,6 +276,16 @@ function test_manifold(
 
         Test.@test norm(M, pts[1], X1) isa Real
         Test.@test norm(M, pts[1], X1) ≈ sqrt(inner(M, pts[1], X1, X1))
+
+        (test_inplace && is_mutating) && Test.@testset "inplace test for exp!" begin
+            p = deepcopy(pts[1])
+            X = deepcopy(X1)
+            q = exp(M, p, X)
+            exp!(M, p, p, X)
+            Test.@test p == q
```
mateuszbaran, line 285: Why not `isapprox(M, p, q)`?

kellertuer: Right, I might change that. My first idea was, that they really should be equal.

kellertuer:
> I think ideally all manifolds should have this enabled.

Ok, I will go through that then.

src/tests/tests_general.jl
mateuszbaran: They should but I wouldn't say that they have to be. `exp` and `exp!` could potentially use slightly different code that leads to different numerical errors.

kellertuer:
added most manifolds (have to look at the groups still) and it seems just `ProductManifold` and `TangentBundle` seem to have problems there, the second component of a `M\timesM` seems to be different sometimes.

mateuszbaran:
> added applicable manifolds and it seems just `ProductManifold` and `TangentBundle` seem to have problems there, the second component of a `M\timesM` seems to be different sometimes.

OK, I'll take a look at these errors.

src/manifolds/VectorBundle.jl
```diff
@@ -272,8 +272,10 @@ function exp!(B::VectorBundle, q, p, X)
     xp, Xp = submanifold_components(B.manifold, p)
     xq, Xq = submanifold_components(B.manifold, q)
     VXM, VXF = submanifold_components(B.manifold, X)
-    exp!(B.manifold, xq, xp, VXM)
-    vector_transport_to!(B.manifold, Xq, xp, Xp + VXF, xq, B.vector_transport.method_point)
+    # this temporary avoids overwriting `p` when `q` and `p` occupy the same memory
+    xqt = exp(B.manifold, xp, VXM)
```
kellertuer, line 276: We could also use `vector_transport_direction!` here, instead?
It does exactly what you do here, per default.

mateuszbaran: Yes, sure -- I didn't think about that.

kellertuer: Well, that‘s what reviews are for – and you fixed all errors, thanks for that! (I do not know why I saw errors with Product, maybe I was wrong? Ah, maybe I just saw a ProductRepr...)

mateuszbaran: Done :slightly_smiling_face: . The product manifold looks fine, you probably just saw `ProductRepr` in `VectorBundle` tests.

kellertuer: Perfect, then this one has just to wait for the decorator update (that is the larger one, I want to finish that first) – ah and I can remodel the deepcopies to copy tomorrow.

src/manifolds/VectorBundle.jl
```diff
@@ -274,7 +274,14 @@ function exp!(B::VectorBundle, q, p, X)
     VXM, VXF = submanifold_components(B.manifold, X)
     # this temporary avoids overwriting `p` when `q` and `p` occupy the same memory
     xqt = exp(B.manifold, xp, VXM)
```
kellertuer, line 276: You can do this one I place again, since `xq` is not used below anymore.

mateuszbaran: I also thought so at first but actually I can't because `xp` (which may refer to the same memory) is used below.

kellertuer: Ah, right, I again forgot that they share memory eventually. Then this has to stay as it is.

mateuszbaran (approved):

PR #397, opened 2021-06-29
Fix projection which had a wrong normalisation factor.

### LINKED ISSUE #395: Elliptope manifold: projection via scaling columns of q, not rows?
Hi All,

I have a project that requires finding the mean or median of a collection of correlation matrices; it seems this isn't an option with the Elliptope manifold in the Manifolds.jl package.  However, in my exploration of the package, I think I may have found an error in the Elliptope projection method.

Points ``p`` on the Elliptope are represented as ``p = q*q'``.  This automatically guarantees that matrix ``p`` is positive semi-definite; an additional constraint that the rows of ``q`` have unit 2-norm guarantees that the diagonals of ``p`` all equal 1.  However, the projection method seems to scale the columns, not the rows, of the potential ``q`` point to be projected.

The MWE code below decomposes correlation matrix ``p`` into ``q*q'``, verifies that ``q`` satisfies the row constraint via ``check_point``, and then projects ``q`` so that the corresponding ``p`` lies on the manifold.  Since projection should be idempotent, the result of the projection should be equal to ``q``.  However, projected output matrix does not satisfy the row constraint, and ``check_point`` throws an error on the value.

As currently implemented, the projection operation is: `` r = q ./ sum(abs2, q, dims=1)``.  If I understand correctly, I think the operation should be something like ``r = q ./ sqrt.(sum(abs2, q, dims=2))``.

If I'm completely off the mark, please let me know---I'm definitely way out of my comfort zone...
- Daniel

```
using LinearAlgebra
using Manifolds

## -- generate 2x2 correlation matrix p
p = [1 0.5; 0.5 1];

## decompose p to get q st q*q' = p
function decomp(p) E = eigen(p); q = E.vectors * Diagonal(sqrt.(E.values)); q end
q = decomp(p)
## 2×2 Array{Float64,2}:
##  -0.5  0.866025
##   0.5  0.866025

q * q'
## 2×2 Array{Float64,2}:
##  1.0  0.5
##  0.5  1.0

M = Elliptope(2,2);

## is q such that p=q*q' is on the Elliptope?
## by construction, p is positive semi-definite
## and yes, p has diagonal elements equal to one
check_point(M, q)
## nothing

## project p (via q) onto the Elliptope
## project should be idempotent: once on the manifold it stays on the manifold
r = project(M, q)
## 2×2 Array{Float64,2}:
##  -1.0  0.57735
##   1.0  0.57735

## however, the check fails
check_point(M, r)
## DomainError([1.3333333333333335; 1.3333333333333335], "The point [-1.0 0.577350269189626; 1.0 0.577350269189626] does not represent a point p=qq^T on Elliptope(2, 2) diagonal is not only ones.")

r * r'
## 2×2 Array{Float64,2}:
##   1.33333   -0.666667
##  -0.666667   1.33333

## method project as currently implemented
r = q ./ sum(abs2, q, dims=1)
## should be (I think)
r_new = q ./ sqrt.(sum(abs2, q, dims=2))
## 2×2 Array{Float64,2}:
##  -0.5  0.866025
##   0.5  0.866025
r_new * r_new'
## 2×2 Array{Float64,2}:
##  1.0  0.5
##  0.5  1.0

check_point(M, r_new)
## nothing
```

kellertuer:
Fixes #395.

### DIFF
--- a/src/manifolds/Elliptope.jl
+++ b/src/manifolds/Elliptope.jl
@@ -129,7 +129,7 @@ project `q` onto the manifold [`Elliptope`](@ref) `M`, by normalizing the rows o
 """
 project(::Elliptope, ::Any)
 
-project!(::Elliptope, r, q) = copyto!(r, q ./ sum(abs2, q, dims=1))
+project!(::Elliptope, r, q) = copyto!(r, q ./ (sqrt.(sum(abs2, q, dims=2))) )
 
 """
     project(M::Elliptope, q, Y)
--- a/src/manifolds/Elliptope.jl
+++ b/src/manifolds/Elliptope.jl
@@ -129,7 +129,7 @@ project `q` onto the manifold [`Elliptope`](@ref) `M`, by normalizing the rows o
 """
 project(::Elliptope, ::Any)
 
-project!(::Elliptope, r, q) = copyto!(r, q ./ (sqrt.(sum(abs2, q, dims=2))) )
+project!(::Elliptope, r, q) = copyto!(r, q ./ (sqrt.(sum(abs2, q, dims=2))))
 
 """
     project(M::Elliptope, q, Y)
--- a/test/elliptope.jl
+++ b/test/elliptope.jl
@@ -43,6 +43,7 @@ include("utils.jl")
                 test_reverse_diff=false,
                 test_forward_diff=false,
                 test_project_tangent=true,
+                test_project_point=true,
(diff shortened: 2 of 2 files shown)


mateuszbaran (approved):

kellertuer:
As soon as all tests pass, I will merge this, but since we finish a few PRs soon anyways, wait for those to do a new version.

PR #403, opened 2021-07-28
Tightens tests for points/vectors on product manifolds

### LINKED ISSUE #349: is_manifold_point (check_manifold_point) broken for semi direct product groups
I noticed, that the checking of manifold points does not seem to work for semi direct product groups.

Here's an example.

```
julia> using Manifolds, Manopt

julia> M = SpecialEuclidean(3)
SpecialEuclidean(3)

julia> P = 1
1

julia> is_point(M,P)
true
```

I would have expected an implementation that checks both product parts somewhere? I think when we designed `check_point` it by default returns nothing, such that the is_manifold_point returns true then, so this implementation seems to be missing.

kellertuer:
This resolves #349 – by throwing an error when one enters product manifolds with wrong encapsulating types (i.e. neither ProductRepr nor ProductArray). We can loosen this once we also allow tuples or arrays in there (would require rewriting accessing the components).

### DIFF
--- a/src/manifolds/ProductManifold.jl
+++ b/src/manifolds/ProductManifold.jl
@@ -156,7 +156,7 @@ end
 function check_point(M::ProductManifold, p; kwargs...)
     return DomainError(
         typeof(p),
-        "The point $p is not a point on $M, since currently only ProductRepr and ProductArray are suppored typoes for points on product manifolds",
+        "The point $p is not a point on $M, since currently only ProductRepr and ProductArray are supported types for points on product manifolds",
     )
 end
 
@@ -192,7 +192,7 @@ end
 function check_vector(M::ProductManifold, p, X; kwargs...)
     return DomainError(
         typeof(X),
-        "The vector $X is not a tangent vector to any tangent space on $M, since currently only ProductRepr and ProductArray are suppored typoes for tangent vectors on product manifolds",
+        "The vector $X is not a tangent vector to any tangent space on $M, since currently only ProductRepr and ProductArray are supported types for tangent vectors on product manifolds",
     )
 end
 
--- a/src/manifolds/ProductManifold.jl
+++ b/src/manifolds/ProductManifold.jl
@@ -153,6 +153,12 @@ function check_point(M::ProductManifold, p::Union{ProductRepr,ProductArray}; kwa
     (length(errors) == 1) && return cerr[1]
     return nothing
 end
+function check_point(M::ProductManifold, p; kwargs...)
+    return DomainError(
+        typeof(p),
+        "The point $p is not a point on $M, since currently only ProductRepr and ProductArray are suppored typoes for points on product manifolds",
+    )
+end
 
 """
     check_vector(M::ProductManifold, p, X; kwargs... )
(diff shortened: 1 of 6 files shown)


src/manifolds/ProductManifold.jl
```diff
@@ -153,6 +153,12 @@ function check_point(M::ProductManifold, p::Union{ProductRepr,ProductArray}; kwa
     (length(errors) == 1) && return cerr[1]
     return nothing
 end
+function check_point(M::ProductManifold, p; kwargs...)
+    return DomainError(
+        typeof(p),
+        "The point $p is not a point on $M, since currently only ProductRepr and ProductArray are suppored typoes for points on product manifolds",
```
mateuszbaran, line 159:
```suggestion
        "The point $p is not a point on $M, since currently only ProductRepr and ProductArray are supported types for points on product manifolds",
```

src/manifolds/ProductManifold.jl
```diff
@@ -183,6 +189,12 @@ function check_vector(
     (length(errors) == 1) && return cerr[1]
     return nothing
 end
+function check_vector(M::ProductManifold, p, X; kwargs...)
+    return DomainError(
+        typeof(X),
+        "The vector $X is not a tangent vector to any tangent space on $M, since currently only ProductRepr and ProductArray are suppored typoes for tangent vectors on product manifolds",
```
mateuszbaran, line 195:
```suggestion
        "The vector $X is not a tangent vector to any tangent space on $M, since currently only ProductRepr and ProductArray are supported types for tangent vectors on product manifolds",
```

mateuszbaran (approved):
Good idea :+1:

kellertuer:
Aha! We do not only support those it seems, I have to check what happens with the cases that now fail. We should carefully revise that it seems.

kellertuer:
The problem is, it seems, that some tests (points/vectors) relied on just passing the (non-existing, but optimistically returning true) test for product manifolds. I will have to check how to solve that.

kellertuer:
I found the culprit (still have to check what to do now) – in `special_euclidan` we do

```
            pts = [affine_matrix(G, ProductRepr(tp...)) for tp in tuple_pts]
            X_pts = [screw_matrix(G, ProductRepr(tX...)) for tX in tuple_X]
```

which produces vectors of points/matrices, but 

```
julia> pts[1]
4×4 Matrix{Float64}:
  0.540302  0.540302      2.30723e-314  1.0
  0.841471  2.30723e-314  2.30718e-314  2.0
 -0.841471  3.49656e-314  0.0           2.30723e-314
  0.0       0.0           0.0           1.0
```

So those are points as matrices not as ProductRepr; but just a matrix, so that is where `is_point` now fails. **But** even before there was never a check for these (just the global default of `check_point` returns `nothing` before. So I think we need tests for these (points / vectors) here?

kellertuer:
I already wanted to write that I got stuck (but I just forgot some `;kwargs...` in a signature).

I implemented tests for affine matrices on `SpecialEuclidean`, so that this manifold now also accepts those as a representation (at least concerning check_point/vector). That seems to be the one thing, that was missing (or where the old variant was too lazy and accepted all point types basically).

mateuszbaran:
Right, `SpeicalEuclidean` is an exception where points on a product manifold are represented by matrices. It's good to have a proper check for that now.

kellertuer:
It was easy to do, since one can easily get the components from the affine matrix. But I did not check which other methods actually do work; however since we do a `text_group` with affine matrices, I think all should work as before.

Only that now the product manifold tests are fare more precise :)

kellertuer:
If ok, this could be merged.

mateuszbaran (approved):
Sure, feel free to merge :+1:

PR #407, opened 2021-08-05
Covariance estimation

mateuszbaran:
A small generalization of the estimator written here: https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1336 (thanks @Affie !)

### DIFF
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -142,6 +142,58 @@ function Base.show(io::IO, method::GeodesicInterpolationWithinRadius)
     return print(io, "GeodesicInterpolationWithinRadius($(method.radius))")
 end
 
+"""
+    Statistics.cov(
+        M::AbstractManifold,
+        x::AbstractVector;
+        basis::AbstractBasis=DefaultOrthonormalBasis(),
+        tangent_space_covariance_estimator::CovarianceEstimator=SimpleCovariance(;
+            corrected=false,
+        ),
+        mean_estimation_method::AbstractEstimationMethod=GradientDescentEstimation(),
+        inverse_retraction_method::AbstractInverseRetractionMethod=default_inverse_retraction_method(
+            M,
+        ),
+    )
+
+Estimate the covariance matrix of points `x` on manifold `M`. Since the covariance matrix
+on a manifold is a rank 2 tensor, the function returns its coefficients in basis induced by
+the given tangent space basis. See Section 5 of [^Pennec2006] for details.
+
+The mean is calculated using the specified `mean_estimation_method`, and tangent vectors at
+this mean are calculated using the provided `inverse_retraction_method`. Finally, the
+covariance matrix in the tangent plane is estimated using `tangent_space_covariance_estimator`
+(the type is defined in `StatsBase.jl` and examples of covariance estimation methods can be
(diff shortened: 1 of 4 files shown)


src/statistics.jl
```diff
@@ -142,6 +142,58 @@ function Base.show(io::IO, method::GeodesicInterpolationWithinRadius)
...
+        inverse_retraction_method::AbstractInverseRetractionMethod=default_inverse_retraction_method(
+            M,
+        ),
+    )
+
+Estimate the covariance matrix of points `x` on manifold `M`. Since the covariance matrix
+on a manifold is a rank 2 tensor, the function returns its coefficients in basis induced by
+the given tangent space basis. See Section 5 of [^Pennec2006] for details.
+
+The mean is calculated using the specified `mean_estimation_method`, and tangent vectors at
```
kellertuer, line 163: Can we link here to the corresponding mean function? That way a user could see more about the estimation methods (otherwise link the `AbstractEstimationMethod` but I think linking the mean would be nicer)

src/statistics.jl
```diff
@@ -142,6 +142,58 @@ function Base.show(io::IO, method::GeodesicInterpolationWithinRadius)
...
+this mean are calculated using the provided `inverse_retraction_method`. Finally, the
+covariance matrix in the tangent plane is estimated using `tangent_space_covariance_estimator`
+(the type is defined in `StatsBase.jl` and examples of covariance estimation methods can be
+found in `CovarianceEstimation.jl`).
```
kellertuer, line 167: Would it be reasonable to link both packages to their repositories or documentation urls?

src/statistics.jl
```diff
@@ -142,6 +142,58 @@ function Base.show(io::IO, method::GeodesicInterpolationWithinRadius)
...
+        tangent_space_covariance_estimator::CovarianceEstimator=SimpleCovariance(;
+            corrected=false,
+        ),
+        mean_estimation_method::AbstractEstimationMethod=GradientDescentEstimation(),
+        inverse_retraction_method::AbstractInverseRetractionMethod=default_inverse_retraction_method(
+            M,
+        ),
+    )
+
+Estimate the covariance matrix of points `x` on manifold `M`. Since the covariance matrix
```
kellertuer, line 159:
```suggestion
Estimate the covariance matrix of a set of points `x` on manifold `M`. Since the covariance matrix
```

src/statistics.jl
kellertuer: I also think that this sentence might be misleading: StatsBase/CovarianceEstimation does not define a tangent_space_covariance_estimator. It defines vector space ones and the tangent space is a vector space.

kellertuer (commented):
The code is nice and short, I just have a few remarks on the documentation.

src/statistics.jl
mateuszbaran: OK, I'll rephrase it.

mateuszbaran:
I've updates the docs, does it look better now?

kellertuer (approved):

sethaxen:
I haven't had a chance to look at this closely, but this looks great! Should be useful for estimating parameters of Pennec's Riemannian normal distribution.

Issue #408, opened 2021-08-07
Potential missing identity_element in v0.6.0, SpecialEuclidean(2)?


dehann:
I'm having a little unexpected difficulty around using identity as we had been doing before.  Trying to debug something else, but now wondering if this is the correct way for `SpecialEuclidean(2)` --- i.e. should this work or is the error valid?
```julia
M = SpecialEuclidean(2)
G = Identity(M)

Manifolds.identity_element(G)

ERROR: MethodError: no method matching identity_element(::Identity{Manifolds.SemidirectProductOperation{RotationAction{TranslationGroup{Tuple{2}, ℝ}, SpecialOrthogonal{2}, LeftAction}}})
Closest candidates are:
  identity_element(::ProductGroup{𝔽, T} where {𝔽, T}) at /home/dehann/.julia/packages/Manifolds/EY0G1/src/groups/product_group.jl:37
  identity_element(::CircleGroup) at /home/dehann/.julia/packages/Manifolds/EY0G1/src/groups/circle_group.jl:27
  identity_element(::CircleGroup, ::Number) at /home/dehann/.julia/packages/Manifolds/EY0G1/src/groups/circle_group.jl:28
  ..
```

_Originally posted by @dehann in https://github.com/JuliaManifolds/Manifolds.jl/issues/401#issuecomment-894675542_

mateuszbaran:
I think the recommended thing to do is to call `identity_element(M)` instead of `identity_element(G)`. Would that work for you?
```julia
julia> M = SpecialEuclidean(2)
SpecialEuclidean(2)

julia> identity_element(M)
ProductRepr with 2 submanifold components:
 Component 1 =
  2-element Vector{Float64}:
   0.0
   0.0
 Component 2 =
  2×2 Matrix{Float64}:
   1.0  0.0
   0.0  1.0
```

dehann:
Yes, i that will work thanks!

mateuszbaran:
Great! Let me know if you'd prefer `identity_element` to return something different for the `SpecialEuclidean` group (for example the affine matrix or a `ProductRepr` of `SArray`s).

dehann:
thanks, I prefer ProductRepr, and perhaps in future pass the datatype similar to Base, e.g. zeros(Int,3).  We're currently using Array{Float64}, and likely to move to MArray or SArray as able.  Also a good chance we will use things like Float16 in the future too.

mateuszbaran:
[...]

PR #409, opened 2021-08-09
Make `allocate_result(G::SemidirectProductGroup, ::typeof(identity_element))` return `ProductRepr` instead of `ProductArray`

mateuszbaran:
Problem reported in #408 . `ProductArray` is deprecated so it shouldn't be returned by default.


kellertuer (approved):
Should we document this default somewhere or did we just miss to implement the documented default?

mateuszbaran:
I'm not sure where it could actually be documented?

kellertuer:
Maybe for the corresponding `identity_element(G)` what the default is when from `identity_element(G,p)` no point (type) `p` is given ?

mateuszbaran:
OK, this should be fine I think?

kellertuer:
Yes, perfect.

Issue #415, opened 2021-08-23
Issue with CircleGroup manifold


Affie:
We are updating some code to CircleGroup and I ran into an issue with the group decorator and I thinkg the identity element.  

```julia
julia> using Manifolds

#With CircleGroup 
julia> M = CircleGroup();

# this is ok 
julia>  ϵ = identity_element(M, [1.0im])
1-element Vector{ComplexF64}:
 1.0 + 0.0im

julia> X = hat(M, ϵ, [pi/4])
1-element Vector{ComplexF64}:
 0.0 + 0.7853981633974483im

#but this errors
julia>  ϵ = identity_element(M)
1.0

julia> X = hat(M, ϵ, pi/4)
ERROR: MethodError: no method matching similar(::Float64, ::Type{Float64})
Closest candidates are:
  similar(::Union{LinearAlgebra.Adjoint{T, var"#s832"}, LinearAlgebra.Transpose{T, var"#s832"}} where {T, var"#s832"<:(AbstractVector{T} where T)}, ::Type{T}) where T at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:230
  similar(::Union{LinearAlgebra.Adjoint{T, S}, LinearAlgebra.Transpose{T, S}} where {T, S}, ::Type{T}) where T at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:234
  similar(::Union{LinearAlgebra.Adjoint{T, S}, LinearAlgebra.Transpose{T, S}} where {T, S}, ::Type{T}, ::Tuple{Vararg{Int64, N}}) where {T, N} at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:235
  ...
Stacktrace:
 [1] allocate(a::Float64, T::Type)
   @ ManifoldsBase ~/.julia/packages/ManifoldsBase/MSCem/src/ManifoldsBase.jl:57
 [2] allocate_result
   @ ~/.julia/packages/ManifoldsBase/MSCem/src/ManifoldsBase.jl:76 [inlined]
 [3] get_vector(M::CircleGroup, p::Float64, X::Float64, B::ManifoldsBase.VeeOrthogonalBasis{ℝ})
[...]

PR #416, opened 2021-08-24
Some fixes for vector handling on `CircleGroup`

mateuszbaran:
Fixes the issue reported in https://github.com/JuliaManifolds/Manifolds.jl/issues/415 .

I think `number_of_coordinates` should ideally have transparency (it's overloaded for `Circle`), what do you think @kellertuer ? Also, I'm not sure if the other `parent` transparencies for `AbstractGroupManifold` are actually correct.


kellertuer:
Yes it should have transparency – is that already the fix? That's cool.

For the other, please check, maybe some should be `:transparent`.

mateuszbaran:
Transparency for `number_of_coordinates` needs to be introduced in ManifoldsBase.jl, I've just added manually transparency for `CircleGroup` (that's the only place where it currently matters I think).

mateuszbaran:
Hm, this breaks `SpecialEuclidean` in a really annoying way: product manifold methods can no longer use matrix variants of `submanifold_components`.

mateuszbaran:
I've applied what looks like the simplest fix but I don't really like it.

kellertuer:
Why don't you like it? It seems to use a similar approach to `SpecialEuclidean`?

mateuszbaran:
It feels a bit weird that `submanifold_componenets` needs to be defined for both manifold and the `SpecialEuclidean` group but maybe that's OK.

kellertuer:
I think that is ok.

mateuszbaran:
@kellertuer is it OK to merge this? The coverage failure seems to be unrelated and I think the solution is finally right.

kellertuer (approved):
I was waiting for exactly some comment on the coverage. Then it's ok.

PR #419, opened 2021-09-01
Action of rotation around axis and real circle group

mateuszbaran:
I've added the real version of `CircleGroup` and action of rotation around an axis, as mentioned here: https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/84 .


src/groups/circle_group.jl
```diff
@@ -90,3 +90,57 @@ _log_lie!(G::CircleGroup, X, q) = (X .= log_lie(G, q))
...
+
+@doc raw"""
+    RealCircleGroup <: GroupManifold{Circle{ℝ},AdditionOperation}
+
+The real circle group is the real circle ([`Circle(ℝ)`](@ref)) equipped with
+the group operation of addition ([`AdditionOperation`](@ref)).
+"""
+const RealCircleGroup = GroupManifold{ℝ,Circle{ℝ},AdditionOperation}
+
+RealCircleGroup() = GroupManifold(Circle{ℝ}(), AdditionOperation())
```
kellertuer, line 102: Just as an idea – Could we also do `CircleGroup{F}`?

mateuszbaran: I tried that, and the problem is that the complex circle group is multiplicative while real one is additive, and there is I think no other field that applies here, and we can't compute one type parameter based on other. So we would need two type parameters for `CircleGroup`, and the change would be breaking, so I thought introducing `RealCircleGroup` is a better solution.

kellertuer: Ah ok.

kellertuer: But you could do a conditional in the second argument of the GroupManifold? Or is that too strange?

mateuszbaran: How would I do that?

kellertuer: Sketch without testing (limited access atm):
```
CircleGroup(𝔽=ℝ) = GroulManifold(Circle{𝔽}(), 𝔽==ℝ ? AdditionOperation() : MultiplicationOperation())
```
or is this a problem with type stability or something?

mateuszbaran: Yes, the constructor is not a problem, the type alias `CircleGroup` is (unless we don't care about it?)

mateuszbaran: `CircleGroup` is exported though so we can't just discard the type alias part without making a breaking release. So the type alias has to stay for now, and it would be a bit weird IMO if `CircleGroup` constructor didn't return an instance of the `CircleGroup` type.

kellertuer: then leave it as it is, i.e. to just cover the real-valued case.

mateuszbaran:
I have no idea why the lack of `number_of_coordinates(M::SkewHermitianMatrices{N,ℂ}, ::AbstractBasis{ℂ})` only now started throwing an error... it should be for example 9 for `number_of_coordinates(SkewHermitianMatrices(3, ℂ), DefaultOrthonormalBasis(ℂ))` but it was 8 before.

PR #421, opened 2021-09-04
`ArrayPartition` as a product manifold representation

mateuszbaran:
This should make `ArrayPartition` work with `ProductManifold`. Some random points:
1. `println(io, "$(T()) for a product manifold")` doesn't work for `DiagonalizingOrthonormalBasis`, so I changed that.
2. I have no idea why `Ξ = [convert(T, V[:, i]) for i in 1:n]` worked without a conversion before but it doesn't now (specifically, without conversion `push!`ing vectors in `get_vectors` of `ProductManifold` fails:
```julia
function get_vectors(
    M::ProductManifold,
    p::ArrayPartition,
    B::CachedBasis{𝔽,<:AbstractBasis{𝔽},<:ProductBasisData},
) where {𝔽}
```

### DIFF
--- a/src/groups/semidirect_product_group.jl
+++ b/src/groups/semidirect_product_group.jl
@@ -177,7 +177,7 @@ end
 # and applies proper padding to the result if `X` happens to be a matrix.
 # Otherwise rare random bugs happen where the padding is not applied.
 function get_vector(G::SemidirectProductGroup, p, X, B::VeeOrthogonalBasis)
-    Y = allocate_result(M, get_vector, p, X)
+    Y = allocate_result(G, get_vector, p, X)
     return get_vector!(G, Y, p, X, B)
 end
 
--- a/src/groups/semidirect_product_group.jl
+++ b/src/groups/semidirect_product_group.jl
@@ -173,6 +173,14 @@ function translate_diff!(G::SemidirectProductGroup, Y, p, q, X, conX::LeftAction
     return Y
 end
 
+# We need to prevent decorator unwrapping so that the correct `get_vector!` gets called
+# and applies proper padding to the result if `X` happens to be a matrix.
+# Otherwise rare random bugs happen where the padding is not applied.
+function get_vector(G::SemidirectProductGroup, p, X, B::VeeOrthogonalBasis)
+    Y = allocate_result(M, get_vector, p, X)
+    return get_vector!(G, Y, p, X, B)
+end
+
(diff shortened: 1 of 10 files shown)


kellertuer:
That would make `ProductRepr` obsolete, but we keep it for now since that would be breaking?

mateuszbaran:
This should replace both `ProductRepr` and `ProductArray` but removing them would be breaking so I'm going to keep all of them for now. I'm also not 100% sure yet about performance of `ArrayPartition` -- it should be fine but we'll see.

kellertuer:
Ah even both.

I think this approach is well thought through, we can drop the old types in one of the next breaking changes, when we are sure the new type is performant (enough).

mateuszbaran:
That's a bit bigger then I thought it would be but it should be ready now.

src/groups/semidirect_product_group.jl
```diff
@@ -173,6 +173,14 @@ function translate_diff!(G::SemidirectProductGroup, Y, p, q, X, conX::LeftAction
...
 end
 
+# We need to prevent decorator unwrapping so that the correct `get_vector!` gets called
+# and applies proper padding to the result if `X` happens to be a matrix.
+# Otherwise rare random bugs happen where the padding is not applied.
+function get_vector(G::SemidirectProductGroup, p, X, B::VeeOrthogonalBasis)
+    Y = allocate_result(G, get_vector, p, X)
+    return get_vector!(G, Y, p, X, B)
+end
+
```
kellertuer, line 183: Can this also be done setting the transparency right? I think so (if this is too complicated, overwriting it like here is of course also one way to go.

kellertuer (commented):
Looks good, I just have a comment – see below; and is there something bad in the new ambiguities? I saw they got raised by 4.

src/groups/semidirect_product_group.jl
mateuszbaran: It's one of those cases where writing a short method manually is much easier than figuring out transparency.

mateuszbaran:
> is there something bad in the new ambiguities? I saw they got raised by 4.

It's just an effect of using `RecursiveArrayTools` -- nothing worse than other `getindex` and `setindex!` ambiguities.

src/groups/semidirect_product_group.jl
kellertuer: Ok.

kellertuer (approved):

Issue #17, opened 2019-06-11
Gradient of a real-valued function


mateuszbaran:
We are definitely going to support optimization of real-valued functions defined on manifolds, so computing gradients of such functions is quite important. What we have now is:

- Manopt.jl seems to have some set of predefined gradients for certain functions but gradients need to be hand-written.
- pymanopt has some capability to use AD for gradients but I haven't looked into it that much.
- FunManifolds has good support for AD gradient on isometrically embedded manifolds (in R^n). I haven't yet thought about AD for MetricManifolds #12 .

There might also be some need for pushforwards, pullbacks etc.

I'll port some parts of FunManifolds's gradient capabilities here, do you have any specific things I should take into account?

kellertuer:
Great topic. For now in Manopt.jl I am trying to help the user by providing necessary functions – but I'd love to extend that to AD capabilities of course.

As far as I read [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl) their main “trick” is to provide basic calculations like a chain rule as well as basic derivatives / gradients.

Using JacobiFields (well on symmetric manifolds available in closed form) we can at least provide a few derivatives/gradients, too (exp,log with respect to both base and argument, distance, norm of course), but we would have to define an extension to the Dual data type of ForwardDiff, i.e. to cover tangent vectors (I don't care much whether `TVector` or just `AbstractArray` but you might know I am in favour of the first).

mateuszbaran:
I've checked pymanopt, they're just using normal AD frameworks and call `egrad2rgrad` to get the correct gradient. We could do a similar thing with ForwardDiff, at least initially. Ultimately, writing a custom, geometry-aware extension to Dual would be great.

For reference, Eric Forgy wrote a very interesting extension to standard AD: https://github.com/EricForgy/StochasticDiff.jl .

kellertuer:
Yes, as a first approach the way pymanopt introduces AD is just fine – though my idea would be to use a more general function name `project` (similar to projecting onto a manifold, just onto a tangent space) instead of the name they use, since one could also project any tangent vector from the embedding onto a tangent plane.

In the long run I was exactly thinking of an extension to standard AD that is geometry aware.

mateuszbaran:
The thing about the `egrad2rgrad` is that for some manifolds they are storing the tangent vector differently (it's not a simple projection). BTW, we already have `project_tangent` for sphere.

kellertuer:
[...]

PR #423, opened 2021-09-10
Sketch change metric.

kellertuer:
This PR is a step towards #17.

the idea here is as follows

`change_metric` (currently just its doctoring, which locally failed to build, let‘s see – I seem to miss something maybe for the docs?) should change the metric, i.e. in a basis compute Z = G_1^-1G_2 X (to switch X from G2 to G1).

Signatures are
```
Z = change_metric(M::AbstractManifold, G2::AbstractMetric, p, X)
```
or
```
Z = change_metric(M::MetricManifold{M,G1}, G2::AbstractMetric, p X)
```

And it would convert _into_ the metric of the first given manifold. From the embedding, e.g. Euclidean with `EuclideanMetric()` one would call for example
```
Z = change_metric(SymmetricPositiveDefinite(n), EuclideanMetric(), p, X)
```

In order to get this into something like the discussed `egrad2rgrad` I would propose a function `convert_tangent(M, N, p, X)` where `N` is the embedding (if `M` itself is not an `EmbeddedManifold`)

1. First call `project(M, p, X)`
2. For the second I am unsure: First change representation? First change metric? Representation info what? For FixedRankMatrices I think we first need a conversion into the 3-factor `TMVVector` – but what if the user wants a different representation? How would we distinguish these? Or would this already be done by project? Maybe with an optional parameter indicating the representation type (e.g. which of the three for `Hyperbolic`)?
3. call `change_metric` as sketched here
4. For some manifold we have to “transfer” this correct tangent vector to the right representation, especially for Lie groups, we would parallel transport in the identity (i.e. Lie algebra)

I think we could provide a default `change_tangent` that just performs steps 1 and 3, and this would be overwritten, e.g. FixedRankMatrices would perform 2 (split into T M V before the metric change) and Lie groups would by default perform 1,3, and 4.

Similarly, with a different `change_metric` for matrix operations on the tangent space (or actions of matrices, so 1-1 and 2-0 tensors in comparison) some `convert_matrix` would be a similar approach to generalise `ehess2rhess`

_Adendum_: So if one has the Euclidean gradient `grad = p -> ...` by some machinery, the Riemannian gradient would – with this sketch be `p -> convert_tangent(M, p, grad(p))` (if `M` has an implicit default embedding). 

_Adendum2_: Of course `convert_tangent_` is not yet the nicest of names, I think, though it converts a Euclidean (or in the embedding) tangent vector to a Riemannian one. Other ideas are welcome.


kellertuer:
For the documentation - I do not understand why it currently complains

```
 LoadError: UndefVarError: change_metric not defined
```

I document the function signature once and export it – as we do for some functions that are first only given in signature. Hm. Maybe I am overlooking something since it is getting late.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,34 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
 
+@doc raw"""
+    change_metric(M::AbstractcManifold, G2::AbstractMetric, p, X)
+
+On the [`AbstractManifold`](@ref) `M` with implicitly given metric ``g_1``
+and a second [`AbstractMetric`](@ref) ``g_2`` this method performs a change of metric in the
+sense that it returns the tangent vector `Z` for which
+
+```
+g_2(X,Y) = g_1(Z,Y) \quad \text{for all} Y ∈ T_p\mathcal M.
```
mateuszbaran, line 62: Could you explain what this condition is useful for? I would've expected `g_2(X,Y) = g_1(change_metric(M, G2, p, X),change_metric(M, G2, p, Y))` to be more important?

mateuszbaran:
> For the documentation - I do not understand why it currently complains
> 
> ```
>  LoadError: UndefVarError: change_metric not defined
> ```
> 
> I document the function signature once and export it – as we do for some functions that are first only given in signature. Hm. Maybe I am overlooking something since it is getting late.

I don't understand it either. Documenter sometimes has weird problems with building PR docs.

src/manifolds/MetricManifold.jl
kellertuer: But that would mean you deine the change metric by using change metric? that is kind of recursive?

This equation is basically what you mentioned with the identity push forward, i.e. X and Z and ”the same“ just with respect to different metrics? Also I think in your equation the two change metrics within the G1 would either cancel (G2*G2^{-1}) or double (G2^2) so this does not seem to be right, I am not sure 100% (its late).

kellertuer:
I can get the same error locally, so this is not only on PR ;)

src/manifolds/MetricManifold.jl
mateuszbaran: No, that's not recursive -- it's the same thing I described in https://github.com/JuliaManifolds/Manifolds.jl/issues/17#issuecomment-915440745 . Here `change_metric` is the function `f` from my previous comment. You get an equation for matrix `B` which defines the transformation.

kellertuer: Or in other words: What is your result, X is the input and it should hold for all Y, then your proposal has no Z to be computed/returned?

mateuszbaran: Maybe my idea is incorrect but still I think it would be nice to explain what exactly do we need `change_metric` for.

kellertuer: > No, that's not recursive -- it's the same thing I described in [#17 (comment)](https://github.com/JuliaManifolds/Manifolds.jl/issues/17#issuecomment-915440745) . Here `change_metric` is the function `f` from my previous comment. You get an equation for matrix `B` which defines the transformation.

my transformation matrix with this is B= G_1^-1G_2

mateuszbaran:
That's a really weird error... maybe environment in which docs are built doesn't have `Manifolds.jl` correctly `dev`ed? I'm not sure.

src/manifolds/MetricManifold.jl
kellertuer: > Maybe my idea is incorrect but still I think it would be nice to explain what exactly do we need `change_metric` for.

If we just take R2, then a metric can be described by a SPD matrix. Let G1 = I be the default metric (that is g_1(x,y) = xTy) and G2 = 2I (and I think I even got the order right).

Assume you have a G2 unit vector with a certain direction (the direction will not change here because both are constant diagonal matrices). In order for it to be g1-unit norm (in general also rotation etc is adopted) you have to G_1^-1G2 the vector (or in other words multiply it by two :D )

kellertuer:
> That's a really weird error... maybe environment in which docs are built doesn't have `Manifolds.jl` correctly `dev`ed? I'm not sure.

I do not know. Locally it just gets weirder

```
ERROR: LoadError: SystemError: opening file "/var/folders/q4/7mgqym9s5nx0kq4jx_z40wj00000gn/T/jl_me64AL.png": No such file or directory
```

whatever. I‘ll lave it for now trying to build to docs, I think its GR again. GR sometimes does _very_ strange things.

src/manifolds/MetricManifold.jl
mateuszbaran: > > No, that's not recursive -- it's the same thing I described in [#17 (comment)](https://github.com/JuliaManifolds/Manifolds.jl/issues/17#issuecomment-915440745) . Here `change_metric` is the function `f` from my previous comment. You get an equation for matrix `B` which defines the transformation.
> 
> my transformation matrix with this is B= G_1^-1G_2

Well, it's not. For example when `g_N` (from my comment) is the identity, then `B` is square root of `g_M`.

> If we just take R2, then a metric can be described by a SPD matrix. Let G1 = I be the default metric (that is g_1(x,y) = xTy) and G2 = 2I (and I think I even got the order right).
> 
> Assume you have a G2 unit vector with a certain direction (the direction will not change here because both are constant diagonal matrices). In order for it to be g1-unit norm (in general also rotation etc is adopted) you have to G_1^-1G2 the vector (or in other words multiply it by two :D )

Maybe I don't understand something but...
```julia
julia> B = [2 0; 0 2]
2×2 Matrix{Int64}:
 2  0
 0  2

julia> X = [1, 0]
2-element Vector{Int64}:
 1
 0

julia> using LinearAlgebra

julia> X' * I * X
1

julia> Xb = B \ X
2-element Vector{Float64}:
 0.5
 0.0

julia> Xb' * B * Xb # norm of Xb in the new basis with your transform
0.5

julia> Xb = sqrt(B) \ X
2-element Vector{Float64}:
 0.7071067811865475
 0.0

julia> Xb' * B * Xb # norm of Xb in the new basis with my transform
0.9999999999999998
```

kellertuer: Or I have a mistake in my derivation – then please correct it. I don‘t see where my mistake is, but I had hoped I got it correct, otherwise I would not have started this PR.

kellertuer: Also what is “your” transformation? As in my first question above, your equation only contained X (the input) and Y (all Y) but no output Z?

mateuszbaran: My transformation is `f: X -> B*X`, where `B` is a matrix such that `G_1 = B^T G_2 B`, where `G_1`  is the original metric and `G_2` is the new metric. Now `<f(X), f(Y)>_{G2} = (BX)^T G_2 (BY) = X^T B^T G_2 B Y = X^T G_1 Y = <X, Y>_{G1}`. Even in the case of norm, notice that the transformed vector is used in both arguments of the metric, not just one: `||X||_{G2} = \sqrt{<X,X>_{G2}}`.

mateuszbaran: If `G_1=L_1 U_1` and `G_2=L_2 U_2` are Cholesky decompositions of metric tensors, then `B=U_2 \ U_1` should solve the equation I think.

kellertuer: Ah I See my mistake at least; I still have to check the last step of yours (the one using LU). But I will then update the docs tomorrow.

kellertuer: (it was that the equation has to hold for all Z, Y to determine B and then return the transformed X, i.e. BX).

kellertuer:
I thought a little bit about steps 2 (changing type) and 4 (adaption to local tangent vector representation).
The order of 3 and 4 is irrelevant I think, so 4 could be after 2.

Then I think both 2 and 4 should be done in project for the following reason: If I project (with the metric from the embedding) onto a manifold, I would expect `is_vector(M, p, project(M, p, X))` to return true. For manifolds like fixed rank this requires the right type (that is project should do 2) and for Lie groups it requires an element from the Lie algebra (that is project should do 4).

I am currently updating the docs to the result from our discussion, but locally I can not build the docs and look why it fails here on PR, since https://github.com/JuliaPlots/Plots.jl/issues/3800 either Plots or GR is just completely broken currently.

kellertuer:
Just when I thought I had understood what we are doing, I think I again do not understand what we are doing. See
https://github.com/NicolasBoumal/manopt/blob/3774b069666a3a1a01a33e10f53062e33495e59e/manopt/manifolds/symfixedrank/sympositivedefinitefactory.m#L102

There the transportation of a tangent vector `X` at `p` (there notation is `\xi` and `X`) gets transformed as `Z = pXp`.
I would have just expected `pX` from the discussion above since the equation reads that we are looking for a transformation B such that (g_E the euclidean we are coming from g_p the linear affine

g_E(X,Y) = tr(XY) = g_p(BX,BY) = tr(p^{-1}BXp^{-1}BY)

Here clearly this yields `B=p` and not the equation above. So _somehow_ the `change_metric` I just updated the documentation to and the linked conversion (the `symm` inside should just be the projection) – are still not the same, so just `change_metric` after projection (type change and parallel transport to I) – is not what Manopt is doing, they somehow more look at

g_E(X,Y) = tr(XY) = g_p(f(X),B) = tr(p^{-1}Xp^{-1}Y)

for _fixed X_ (and all Y) what I started here with, too? because then, sure, `f(X) = pXp`. Or am I missing something?  So what is now the actual conversion we are looking for and why?

kellertuer:
After a short question to the Manopt forum and some further thinking I found the issue at least for the gradient:
We do not want to change the metric (though having https://juliamanifolds.github.io/Manifolds.jl/previews/PR423/manifolds/metric.html#Manifolds.change_metric-Tuple{AbstractManifold,%20AbstractMetric,%20Any,%20Any} is surely nice) but we want to switch the Riesz representer of the differential (since we switch the metric). I documented this at https://juliamanifolds.github.io/Manifolds.jl/previews/PR423/manifolds/metric.html#Manifolds.change_tangent-Tuple{AbstractManifold,%20AbstractMetric,%20Any,%20Any} – where I do not like the name yet.

Note that this requires _one_ representation of tangent vectors, that is the shift to the Lie algebra can happen before or after, since it is not important at which point we switch the metric (for neither of those functions).

What remains is to decide for proper names (one could call the second also `change_representer` or something) and decide where to put the type-conversion and the parallel transport to the identity. I am still for doing both in project.
Nicolas also wrote here https://groups.google.com/g/manopttoolbox/c/eiuq_BfENjU that this method should work for quotient manifolds with total space being Euclidean as well – I think it even works between arbitrary manifolds as long as there is a projection possibility.

mateuszbaran:
I think I'm starting to see how it's supposed to work, in particular why you did `change_metric` that way initially.

> [https://juliamanifolds.github.io/Manifolds.jl/previews/PR423/manifolds/metric.html#Manifolds.change_tangent-Tuple{AbstractManifold,%20AbstractMetric,%20Any,%20Any}](https://juliamanifolds.github.io/Manifolds.jl/previews/PR423/manifolds/metric.html#Manifolds.change_tangent-Tuple%7BAbstractManifold,%20AbstractMetric,%20Any,%20Any%7D) – where I do not like the name yet.

I think a name like `change_gradient_metric` or something like that would be fine.

> Note that this requires _one_ representation of tangent vectors, that is the shift to the Lie algebra can happen before or after, since it is not important at which point we switch the metric (for neither of those functions).

My first thought is to require gradients in the correct representation to be passed to `change_gradient`, so switching representation would happen first. Maybe it's fine to do it the other way too but it's not obvious to me.

> What remains is to decide for proper names (one could call the second also `change_representer` or something) and decide where to put the type-conversion and the parallel transport to the identity. I am still for doing both in project.

It seems reasonable to put both type conversion and parallel transport in `project`. By the way, vector projections of `Rotations` manifold doesn't do parallel transport currently as far as I can tell, it's a different vector transport.

> Nicolas also wrote here https://groups.google.com/g/manopttoolbox/c/eiuq_BfENjU that this method should work for quotient manifolds with total space being Euclidean as well – I think it even works between arbitrary manifolds as long as there is a projection possibility.

Great, that's quite helpful :slightly_smiling_face: .

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
 """
 abstract type RiemannianMetric <: AbstractMetric end
 
+@doc raw"""
+    change_gradient(M::AbstractManifold, G2::AbstractMetric, p, X)
+
+Convert the gradint `X` at `p` on the[`AbstractManifold`](@ref) `M` from one metric to another.
```
mateuszbaran, line 57:
```suggestion
Convert the gradient `X` at `p` on the[`AbstractManifold`](@ref) `M` from one metric to another.
```

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+
+Assume that for a real-valued function ``f: \mathcal M \to ℝ`` we are given the Riesz representer of the differential with respect to the metric ``g_2`` i.e.
```
mateuszbaran, line 59: I think gradient and Riesz representer of the differential are the same thing? Maybe it would be nice to mention it here for clarity?

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+
+    change_gradient(Sphere(2), EuclideanMetric(), p, X)
+
+Since the metric in ``T_p\mathbb S^2`` is the Euclidean metric from the embedding restricted to ``T_p\mathbb S^2``, this just returns `X`
+
+    change_gradient(SymmetricPOsitiveDefinite(3), EuclideanMetric(), p, X)
+
+Here, the default metric in `\mathcal P(3)` is the [`LinearAffineMetric`](@ref) and the transformation can be computed as ``pXp``
+"""
+change_tangent(::AbstractManifold, ::AbstractMetric, ::Any, ::Any)
```
mateuszbaran, line 93:
```suggestion
change_gradient(::AbstractManifold, ::AbstractMetric, ::Any, ::Any)
```
or whatever else we end up calling this :slightly_smiling_face:

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+```
+
+If both metrics are given in their [`local_metric`](@ref) (symmetric positive defintie) matrix
+representations ``G_1`` and ``G_2`` and ``x,y`` are the local coordinates with respect to
+the same basis of the tangent space, the equation reads
+
+```math
+   x^*G_2y = c(x)^*G_1 y \quad \text{for all } y \in ℝ^d,
+```
+where `\cdot^*`` denotes the conjugate transpose.
```
mateuszbaran, line 79: We have conjugate transpose in the notation as `{}^{H}`.

kellertuer:
> > [https://juliamanifolds.github.io/Manifolds.jl/previews/PR423/manifolds/metric.html#Manifolds.change_tangent-Tuple{AbstractManifold,%20AbstractMetric,%20Any,%20Any}](https://juliamanifolds.github.io/Manifolds.jl/previews/PR423/manifolds/metric.html#Manifolds.change_tangent-Tuple%7BAbstractManifold,%20AbstractMetric,%20Any,%20Any%7D) – where I do not like the name yet.
> 
> I think a name like `change_gradient_metric` or something like that would be fine.

That sounds like a reasonable name, I will think whether it can have a more general name even, since this might not only be used for gradients.

> 
> > Note that this requires _one_ representation of tangent vectors, that is the shift to the Lie algebra can happen before or after, since it is not important at which point we switch the metric (for neither of those functions).
> 
> My first thought is to require gradients in the correct representation to be passed to `change_gradient`, so switching representation would happen first. Maybe it's fine to do it the other way too but it's not obvious to me.

That is also what I have in mind, that function should _only_ change the metric and not representation

> > What remains is to decide for proper names (one could call the second also `change_representer` or something) and decide where to put the type-conversion and the parallel transport to the identity. I am still for doing both in project.
> 
> It seems reasonable to put both type conversion and parallel transport in `project`. By the way, vector projections of `Rotations` manifold doesn't do parallel transport currently as far as I can tell, it's a different vector transport.

I will check rotation then, but great that we agree, that both conversion and representation shift (to Lie algebra) should happen in the projection. I will look at that and adopt docs/function implementations.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+
+and we obtain `c(x) = (G_1\backslask G_2)^*x `
```
mateuszbaran, line 81:
```suggestion
and we obtain ``c(X) = (G_1\backslash G_2)^*X``
```
Probably something like that?

src/manifolds/MetricManifold.jl
kellertuer: Sure. The gradient is the Riesz representer of the Differential. I will mention that and maybe even link Riesz to the wikipedia article – good idea.

src/manifolds/MetricManifold.jl
kellertuer: Thanks for noticing – will adopt – for this first sketch I was more focussing on understanding it myself :)

mateuszbaran: Sure, that's perfectly understandable :+1:

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+    z = (G2 \ G1)'x
+    return get_vector(M, p, z, B)
+end
+
+function change_metric(
+    ::MetricManifold{<:M,<:G},
+    ::G,
+    p,
+    X,
+) where {M<:AbstractManifold,G<:AbstractMetric}
```
mateuszbaran, line 113:
```suggestion
function change_metric(
    ::MetricManifold{𝔽,M,G},
    ::G,
    p,
    X,
) where {𝔽,M<:AbstractManifold{𝔽},G<:AbstractMetric}
```
I think this is more correct this way.

kellertuer: Right, I totally forgot about the F.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+Here, the default metric in `\mathcal P(3)` is the [`LinearAffineMetric`](@ref) and the transformation can be computed as ``B=p``
+
+"""
+change_metric(::AbstractManifold, ::AbstractMetric, ::Any, ::Any)
+
+function change_metric(M::AbstractManifold, G::AbstractMetric, p, X)
+    if is_default_metric(M, G)
+        return X
+    end
+    # TODO: For local metric, inverse_local metric, det_local_metric: Introduce a default basis?
```
mateuszbaran, line 157: Default bases is something that may be worth updating but it's a bigger issue. Anyway, here we want the same basis for two different metrics so it has to be more or less the way you wrote it.

kellertuer: Yes, sure, I would like to have B here as an optional parameter, as well as for the local metic (inverse local metric, yet local metric, Jacobean local metric ...)

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,126 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+    B = DefaultOrthogonalBasis()
+    G1 = local_metric(M, p, B)
+    G2 = localMetric(G(M), p, B)
+    x = get_coordinates(M, p, X, B)
+    C1 = cholesky(G1)
+    C2 = cholesky(G2)
+    z = (C1 \ C2)'x
```
mateuszbaran, line 164: I think this is missing some `.U` or `.L`?

kellertuer:
I think, since `change_gradient` is a little more general, maybe even `change_Riesz_representer` would be a good name?

src/manifolds/MetricManifold.jl
kellertuer: I had hopped Cholesky would just give me back a matrix, that is the L basically? I have to check cholesky, since I have not yet started testing on these functions yes; I think `.L` and `.U` are only for LU?

src/manifolds/MetricManifold.jl
mateuszbaran: Yeah, but I wouldn't like to work on that any time soon...

src/manifolds/MetricManifold.jl
kellertuer: Oh, indeed, it‘s always `C.L` here.

src/manifolds/MetricManifold.jl
kellertuer: me neither.Maybe before merging this we open an issue for this to make a few of these metric functions more flexible / provide a default basis for them.

mateuszbaran: Sure, we can definitely have an issue for that.

src/manifolds/MetricManifold.jl
```diff
@@ -179,12 +302,16 @@ flat(::MetricManifold, ::Any...)
...
+    inverse_local_metric(M::AbstractcManifold{𝔽}, p, B::AbstractBasis)
 
-Return the local matrix representation of the inverse metric (cometric) tensor, usually
-written ``g^{ij}``.
+Return the local matrix representation of the inverse metric (cometric) tensor
+of the tangent space at `p` on the [`AbstractManifold`](@ref) `M` with respect
+to the [`AbstractBasis`](@ref) basis `B`.
 
-See also [`local_metric`](@ref)
+The metric tensor (see [`local_metric`](@ref)) os isually denoted by ``G = (g_{ij}) ∈ 𝔽^{d×d}``,
```
mateuszbaran, line 311:
```suggestion
The metric tensor (see [`local_metric`](@ref)) is usually denoted by ``G = (g_{ij}) ∈ 𝔽^{d×d}``,
```
Is there actually a case when the metric tensor wouldn't have all real coefficients?

kellertuer: For a `d` dimensional complex manifold and `d` (real valued) basis vectors, the coefficients would be complex. They would be real if we represent them as `2d` entries, but otherwise complex, I think.

src/manifolds/MetricManifold.jl
kellertuer: See #424.

src/manifolds/MetricManifold.jl
mateuszbaran: Yes, that's probably right.

kellertuer:
I just looked for project a little closer and the docs do not specify the change in representation or the Lie algebra idea – but SpecialLinear for example does a translate diff to get into the Lie algebra. So I will check that thoroughly and we could maybe make the documentation over in Base a little more precise.

kellertuer:
I went through most of the manifolds and for only 4 I was not sure, since there, the Manopt code was unclear (i.e. I could not split their code into projection and change or it was unclear what it does in general):

* [ ] Multinomial double stochastic, since I do not yet understand what this function https://github.com/NicolasBoumal/manopt/blob/c0fe36fd8ebb7cf6bf2ef57f89230cfad9d5411c/manopt/manifolds/multinomial/multinomialdoublystochasticfactory.m#L147 does, but we also do not have a projection for tangents for this manifold yet
* [ ] Multinomial – similarly, the documentation for this https://github.com/NicolasBoumal/manopt/blob/c0fe36fd8ebb7cf6bf2ef57f89230cfad9d5411c/manopt/manifolds/multinomial/multinomialfactory.m#L110 is not that helpful, but it could be that just the point p has to be cholesky-ied - if one compares inner products?
* [ ] MultinoialSymmetric, same as the two above I think https://github.com/NicolasBoumal/manopt/blob/c0fe36fd8ebb7cf6bf2ef57f89230cfad9d5411c/manopt/manifolds/multinomial/multinomialsymmetricfactory.m#L105
* [ ] Spectrahedron https://github.com/NicolasBoumal/manopt/blob/c0fe36fd8ebb7cf6bf2ef57f89230cfad9d5411c/manopt/manifolds/symfixedrank/spectrahedronfactory.m#L107 (Seems to be part of projection, so maybe this is not necessary?)

But in general – projection does 1,2,4 and nónly for the change of the gradient we still need a nice name – and have to check for the 4 mentioned transformations. All others either now have conversions or do not require some (where I wrote the default that from Euclidean to a Manifold with Default Euclidean metric it returns the input X).

mateuszbaran:
> * but we also do not have a projection for tangents for this manifold yet

I think we do have projection for this manifold: https://github.com/JuliaManifolds/Manifolds.jl/blob/8e775d30e01d3e7769dbdb2661b7675e63db644c/src/manifolds/MultinomialDoublyStochastic.jl#L144 . `egrad2rgrad` does left multiplication by `X` and projection if I see right.

> * Multinomial – similarly, the documentation for this https://github.com/NicolasBoumal/manopt/blob/c0fe36fd8ebb7cf6bf2ef57f89230cfad9d5411c/manopt/manifolds/multinomial/multinomialfactory.m#L110 is not that helpful, but it could be that just the point p has to be cholesky-ied - if one compares inner products?

That's also left multiplication by `X` and projection.

> [ ] Spectrahedron https://github.com/NicolasBoumal/manopt/blob/c0fe36fd8ebb7cf6bf2ef57f89230cfad9d5411c/manopt/manifolds/symfixedrank/spectrahedronfactory.m#L107 (Seems to be part of projection, so maybe this is not necessary?)

I don't quite see what it actually does.

> But in general – projection does 1,2,4 and nónly for the change of the gradient we still need a nice name – and have to check for the 4 mentioned transformations. All others either now have conversions or do not require some (where I wrote the default that from Euclidean to a Manifold with Default Euclidean metric it returns the input X).

Yes, that seems right.

kellertuer:
Thanks for the help, I did all now, just spectrahedron seems to just do projection (the projection even does one step more) but the metric is Euclidean anyways, so I think the default implementation should be fine and it should be covered, I think.

So the only thong missing is to
* [ ] carefully reads the docs 
* [ ] write tests

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,129 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+    ::MetricManifold{𝔽,M,G},
+    ::G,
+    p,
+    X,
+) where {𝔽,M<:AbstractManifold{𝔽},G<:AbstractMetric}
+    return X
+end
+
+@doc raw"""
+    change_metric(M::AbstractcManifold, G2::AbstractMetric, p, X)
```
mateuszbaran, line 121: Do we want a mutating variant of this (and `change_tangent`/`change_gradient`)?

src/manifolds/CenteredMatrices.jl
```diff
@@ -119,7 +119,7 @@ where $c_i = \frac{1}{m}\sum_{j=1}^m x_{j,i}$  for $i = 1, \dots, n$.
 """
 project(::CenteredMatrices, ::Any, ::Any)
 
-project!(M::CenteredMatrices, Y, p, X) = (Y .= X .- mean(X, dims=1))
+project!(::CenteredMatrices, Y, p::Any, X) = (Y .= X .- mean(X, dims=1))
```
mateuszbaran, line 122: Do we need `::Any` here?

src/manifolds/GeneralizedGrassmann.jl
```diff
@@ -57,6 +57,28 @@ function GeneralizedGrassmann(
     return GeneralizedGrassmann{n,k,field,typeof(B)}(B)
 end
 
+@doc raw"""
+    change_metric(M::GeneralizedGrassmann, ::EuclideanMetric, p X)
+
+Change `X` to the corresponding representer of the gradient with respect to the scaled metric
+of the [`GeneralizedGrassmann`](@ref) `M`, i.e. `B\X`.
```
mateuszbaran, line 64:
```suggestion
of the [`GeneralizedGrassmann`](@ref) `M`, i.e. `M.B\X`.
```

src/manifolds/GeneralizedGrassmann.jl
```diff
@@ -57,6 +57,28 @@ function GeneralizedGrassmann(
     return GeneralizedGrassmann{n,k,field,typeof(B)}(B)
 end
 
+@doc raw"""
+    change_metric(M::GeneralizedGrassmann, ::EuclideanMetric, p X)
+
+Change `X` to the corresponding representer of the gradient with respect to the scaled metric
```
mateuszbaran, line 63: I think it should either be gradient or representer of the differential.

src/manifolds/GeneralizedGrassmann.jl
```diff
@@ -57,6 +57,28 @@ function GeneralizedGrassmann(
     return GeneralizedGrassmann{n,k,field,typeof(B)}(B)
 end
 
+@doc raw"""
+    change_metric(M::GeneralizedGrassmann, ::EuclideanMetric, p X)
```
mateuszbaran, line 61:
```suggestion
    change_metric(M::GeneralizedGrassmann, ::EuclideanMetric, p, X)
```

src/manifolds/GeneralizedGrassmann.jl
```diff
@@ -57,6 +57,28 @@ function GeneralizedGrassmann(
...
+"""
+function change_gradient(M::GeneralizedGrassmann, ::EuclideanMetric, p, X)
+    return M.B \ X
+end
+
+@doc raw"""
+    change_metric(M::GeneralizedGrassmann, ::EuclideanMetric, p X)
+
+Change `X` to the corresponding vector with respect to the metric of the [`GeneralizedGrassmann`](@ref) `M`,
+i.e. let ``B=LL'`` be the cholesky decomposition of the metrix `B` then the corresponding vector is ``L\X``.
```
mateuszbaran, line 74:
```suggestion
i.e. let ``B=LL'`` be the Cholesky decomposition of the matrix `M.B`, then the corresponding vector is ``L\X``.
```

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -1,3 +1,16 @@
+@doc raw"""
+    change_gradient(M::Hyperbolic, ::EuclideanMetric, p, X)
+
+to change a Euclidena gradient from the embedding that was already projected onto the tangent space at `X`
+we only have to correct for the metric, which means, that the sign of the last entry changes.
```
mateuszbaran, line 5:
```suggestion
Change a Eucliden gradient from the embedding that was already projected onto the tangent space at `p`.
We only have to correct for the metric, which means that the sign of the last entry changes.
```

src/manifolds/HyperbolicPoincareBall.jl
```diff
@@ -1,3 +1,39 @@
...
+function change_metric(
+    ::Hyperbolic,
+    ::EuclideanMetric,
+    p::PoincareBallPoint,
+    X::PoincareBallTVector,
+)
+    α = 2 / (1 - norm(p.value)^2)
+    Y = copy(M, p, X)
+    Y.value ./= α
+    return Y
```
mateuszbaran, line 34: `return PoincareBallTVector(X.value ./ α)` would probably be a bit nicer. Similarly for `change_gradient` above.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,129 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+
+On the [`AbstractManifold`](@ref) `M` with implicitly given metric ``g_1``
+and a second [`AbstractMetric`](@ref) ``g_2`` this method performs a change of metric in the
```
mateuszbaran, line 124:
```suggestion
and a second [`AbstractMetric`](@ref) ``g_2`` this function performs a change of metric in the
```

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,129 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+sense that it returns the tangent vector ``Z=BX`` such that the linear map ``B`` fulfills
+
+````math
+g_2(Y_1,Y_2) = g_1(BY_1,BY_2) \quad \text{for all } Y_1, Y_2 ∈ T_p\mathcal M.
+````
+
+holds.
```
mateuszbaran, line 131: Having both "holds" and "fulfills" seems a bit excessive.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,129 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+    z = (C1 \ C2)'x
+    return get_vector(M, p, z, B)
+end
+
+function change_metric(
+    ::MetricManifold{<:M,<:G},
+    ::G,
+    p,
+    X,
+) where {M<:AbstractManifold,G<:AbstractMetric}
```
mateuszbaran, line 173: Isn't that the same method as  https://github.com/JuliaManifolds/Manifolds.jl/pull/423/files#diff-161f9a28506b35d50f71dd7a22dd0842ee4ade1b2da160994a59110c4902ba0cR111-R118 ?

src/manifolds/MetricManifold.jl
```diff
@@ -111,7 +234,7 @@ Return the [`LeviCivitaConnection`](@ref) for a metric manifold.
 connection(::MetricManifold) = LeviCivitaConnection()
 
 @doc raw"""
-    det_local_metric(M::AbstractManifold, p, B::AbstractBasis)
+    det_local_metric(M::AbstractManifold, p, B::AbstractBasis=DefaultOrthogonalBasis)
```
mateuszbaran, line 237: I'd leave it for https://github.com/JuliaManifolds/Manifolds.jl/issues/424

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -56,6 +56,15 @@ function MultinomialDoubleStochastic(n::Int)
     return MultinomialDoubleStochastic{n}()
 end
 
+"""
+    change_gradient(M::ProbabilitySimplex, ::EuclideanMetric, ü, X)
```
mateuszbaran, line 60:
```suggestion
    change_gradient(M::AbstractMultinomialDoublyStochastic, ::EuclideanMetric, p, X)
```

src/manifolds/PositiveNumbers.jl
```diff
@@ -53,6 +53,23 @@ function check_point(M::PositiveNumbers, p; kwargs...)
     return nothing
 end
 
+@doc raw"""
+    change_metric(M::PositiveNumbers, E::EuclideanMetric, p, X)
+
+Given a tangent vector ``X ∈ T_p\mathcal M``representing a gradient with respect to the [`EuclideanMetric`](@ref) `g_E`,
+this function changes into the positivity metric representation of
+[`PositiveNumbers`](lref) `M` for the same gradient.
```
mateuszbaran, line 61: What's `lref`?

mateuszbaran (commented):
Nice progress! I've checked it a bit, here are some comments.

src/manifolds/PowerManifold.jl
```diff
@@ -78,6 +78,30 @@ end
...
+        Z[i...] = change_gradient(M.manifold, G, p[i...], X[i...])
+    end
+end
+
+"""
+    change_metric(M::AbstractPowerManifold, ::AbstractMetric, p, X)
+
+Since the metric on a power manifold decouples, the change of a representer can be done elementwise
+"""
+function change_gradient(::AbstractPowerManifold, M::AbstractMetric, p, X)
```
mateuszbaran, line 98: You have `change_gradient` twice here.

src/manifolds/MetricManifold.jl
kellertuer: I thought fo that already, too. Not sure yet, but it seems reasonable.

src/manifolds/CenteredMatrices.jl
kellertuer: If we leave out the `p` since the variable is not used, then `::Any` would be nice, sure.

src/manifolds/MetricManifold.jl
mateuszbaran: Anyway, if you don't want to do it in this PR, I'd be perfectly fine with just leaving an issue for that :slightly_smiling_face: .

src/manifolds/GeneralizedGrassmann.jl
kellertuer: I am not yet sure how to generally phrase is, since the generic way would be representer of a linear functional.

src/manifolds/CenteredMatrices.jl
mateuszbaran: I think we generally tend to leave `::Any` and keep variable names for points and tangent vectors, to make the code more descriptive.

src/manifolds/HyperbolicPoincareBall.jl
kellertuer: Yes, but that would basically be the same, right?

src/manifolds/MetricManifold.jl
kellertuer: oh yes, we can leave out the “holds”, sure.

src/manifolds/HyperbolicPoincareBall.jl
mateuszbaran: Not quite, my version also works for immutable array types like `SVector`.

src/manifolds/MetricManifold.jl
kellertuer: Yes, in 111 its a type and it should be change_gradient.

src/manifolds/HyperbolicPoincareBall.jl
kellertuer: ok.

src/manifolds/MetricManifold.jl
kellertuer: Yes sorry, can artefact from when I thought I might do that here and changed the doctoring already.

src/manifolds/PositiveNumbers.jl
kellertuer: alt+l on a German Mac keyboard is the @ - sorry.

src/manifolds/PowerManifold.jl
kellertuer: good old copy paste and forgetting one word... ;)

src/manifolds/CenteredMatrices.jl
kellertuer: ah right, we only did any for doctoring signatures I think. I will switch it to just `p`

src/manifolds/MetricManifold.jl
kellertuer: No, I will do testing anyways, then I can also switch to mutating variants, no problem.

kellertuer: I added it to the todo in the comment above to keep track.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,127 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+```
+
+(It could actually also be the Riesz representer of _any_ linear function, not just ``Df(p)``).
+
+In order to convert this into the gradient with respect to the (implicitly given) metric ``g_1`` of `M`,
+we have to find the conversion function ``c: T_p\mathcal M \to \T_p\mathcal M`` such that
+
+```math
+    g_2(X,Y) = g_1(c(X),Y)
+```
```
mateuszbaran, line 73: Technically the whole discussion about gradient isn't essential to this function, and what it does is described by the formula in line 72. Alternatively it could be described as changing the metric with respect to which the Riesz representer of a cotangent vector is taken.

kellertuer: Exactly! Thats exactly why I am not sure about the name yet. Do you have a good idea for the name of the function then?

mateuszbaran: Maybe something like `representer_in_metric`?

kellertuer: That misses the action of changing the representer. So I was thinking of maybe `change_representer`? The metric ideas would get clear from the signature (because otherwise the name would be too long)?

mateuszbaran: Sure, `change_representer` sounds better and the metric part is indeed clear from the signature :+1: .

kellertuer: Perfect. I will change the name and the docs accordingly. I think the more general name is really better.

kellertuer:
I started testing and polishing the documentation for every function that I add a test for.
So far it looks good. I even got to test, that for the identical (non-Euclidean) metric, it works; even when the other version does not exists, for example Hyperbolic does not have `change_metric` from Euclidean but is still the identity if you enter with Minkowksi

I was wondering if it would make sense to have a function that combines all the things we discussed here (something like `egrad2rgrad`, so that does projection and representer change. But I do not have a good idea for a name. Do you have a good idea?

The signature would involve `somename(M, G, p, X)` where we could make `G` optional to be EuclideanMetric (since that is kind of a default embedding)

Then one could just obtain the Riemannian gradient given the `euclidean_gradient(p) -> ...` (since its Euclidean its without a manifold upfront), as `riemannian_gradient(M,p) = somename(M, p, euclidean_gradient(p))` (or for Hyperbolic `riemannian_gradient(M,p) = somename(M, p, MinkowskiMetric(), embedded gradient())`.

mateuszbaran:
Maybe we could just leave that to our `gradient` function from `riemannian_diff.jl` (`RiemannianProjectionGradientBackend`)?

kellertuer:
That is a very good idea. I will check after docs&test whether that is just a line or two and then even add it to this PR.

kellertuer:
I noticed something interesting when writing my last planned test for the Poincaré ball representation of `Hyperbolic`:

```julia
using Manifolds
M = Hyperbolic(2)
p = convert(PoincareBallPoint, [1.0, 1.0, sqrt(3)])
X = convert(PoincareBallTVector, [1.0, 1.0, sqrt(3)], [1.0, 2.0, sqrt(3)])
Y = change_metric(M, MinkowskiMetric(), p, X)
```

I had hoped, that `change_metric` allocates a `PoincareBallTVector` (since it calls `allocate_result(M, change_metric, X, p)`) but somehow the last call to returns just a vector of length 2 and not a PoincareBallTVector. I do not see why.

mateuszbaran:
That's because `copyto!` of hyperbolic types doesn't return the target value but rather the inner array. Replacing line 136 in Hyperbolic.jl with `Base.copyto!(q::$T, p::$T) = (copyto!(q.value, p.value); q)` should fix this. It might be worth adding a test for that.

kellertuer:
Ah you were even faster I thought it was a decorator problem with copyto. I will  check for your solution and add a test.

kellertuer:
There is now only three cases/manifolds left

* [ ] Sphere (identity test just to be sure)
* [ ] Power
* [ ] Product

Maybe an additional test for the Multinomial or even MultinomialDoublyStochastic to check that they inherit properly – by embedding via Power from Probability simplex – so quite an advanced but neat inheritance that shows the advantage of this system, i.e. that it really avoids reimplementation.

kellertuer:
I also added this step to the Riemannian Diff. I finished all tests and did all documentations. So I think this should be good to go.

kellertuer:
Ah we could still check for a Riemannian diff case where this would not be the identity – and maybe also provide an example (maybe also just over in Manopt), where one would define a cost – the RiemannianDiff Gradient and then do gradient descent for example.

By the way does this resolve/finish #17?

mateuszbaran:
Yes, I think this should resolve #17 . Could you just check that Manopt works correctly with these Riemannian gradients? Maybe just a small test with `ForwardDiff` on `Sphere` or `Hyperbolic`.

kellertuer:
After I fixed the last few test coverage lines, I will start a small tutorial over on Manopt (maybe just Rayleigh again) on how to use a little bit of AD in Manopt.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -1,3 +1,25 @@
...
+We only have to correct for the metric, which means that the sign of the last entry changes, since
+for the result ``Y``  we are looking for a tangent vector such that
+
+```math
+    g_p(Y,Z) = -y_{n+1}z_{n+1} + \sum_{i=1}^n y_iz_i = \sum_{i=1}^{n+1} z_ix_i
+```
+
+holds, which directly yields ``y_i=x_i`` for ``i=1,\ldots,n`` and ``y_{n+1}=-x_{n+1}``.
+"""
+change_representer(::Hyperbolic, ::EuclideanMetric, ::Any, ::Any)
```
mateuszbaran, line 14: Could you add a reference with Riesz representation theorem stated for non-positive-definite inner products? So far I've only seen it for positive-definite inner products.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -1,3 +1,25 @@
...
+
+function change_representer!(M::Hyperbolic, Y, ::EuclideanMetric, p, X)
+    copyto!(M, Y, p, X)
+    Y[end] *= -1
+    return Y
+end
+
+# a metric change does not seem possible here
```
mateuszbaran, line 22: It is indeed not possible due to [Sylvester's law of inertia](https://en.wikipedia.org/wiki/Sylvester%27s_law_of_inertia) . I think it would be fair to throw an error here.

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,164 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
 """
 abstract type RiemannianMetric <: AbstractMetric end
 
+@doc raw"""
+    change_representer(M::AbstractManifold, G2::AbstractMetric, p, X)
+
+Convert the representer `X` of a linear function (in other words a cotangent vector at `p`)
+in the tangent space at `p` on the[`AbstractManifold`](@ref) `M` given with respect to the
```
mateuszbaran, line 58:
```suggestion
in the tangent space at `p` on the [`AbstractManifold`](@ref) `M` given with respect to the
```

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,164 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+[`AbstractMetric`](@ref) `G2` into the representer with respect to the (implicit) metric of `M`.
+
+In order to convert this into the gradient with respect to the (implicitly given) metric ``g_1`` of `M`,
```
mateuszbaran, line 61:
```suggestion
In order to convert `X` into the representer with respect to the (implicitly given) metric ``g_1`` of `M`,
```

src/manifolds/MetricManifold.jl
```diff
@@ -51,6 +51,164 @@ inner product ``g(X, X) > 0`` whenever ``X`` is not the zero vector.
...
+We obtain ``c(X) = (G_1\backslash G_2)^{\mathrm{H}}X``
+
+For example `X` could be the gradient ``\operatorname{grad}f`` of a real-valued function
+``f: \mathcal M \to ℝ``, i.e.
+
+```math
+    g_2(X,Y) = Df(p)[Y] \quad \text{for all } Y ∈ T_p\mathcal M.
+```
+
+and we would change the Riesz representer of the gradient to the representer with respect to the metric ``g_1``.
```
mateuszbaran, line 85:
```suggestion
and we would change the Riesz representer `X` to the representer with respect to the metric ``g_1``.
```

src/manifolds/PositiveNumbers.jl
```diff
@@ -38,6 +38,55 @@ This manifold is modeled as a [`PowerManifold`](@ref) of [`PositiveNumbers`](@re
...
+change_representer(::PositiveNumbers, ::EuclideanMetric, ::Any, ::Any)
+change_representer(::PositiveNumbers, ::EuclideanMetric, p::Real, X::Real) = p * X * p
+
+function change_representer!(::PositiveNumbers, Y, ::EuclideanMetric, p, X)
+    Y .= p .* X .* p
+    return Y
+end
+
+@doc raw"""
+    change_metric(M::SymmetricPositiveDefinite, E::EuclideanMetric, p, X)
```
mateuszbaran, line 67:
```suggestion
    change_metric(M::PositiveNumbers, E::EuclideanMetric, p, X)
```

src/manifolds/PowerManifold.jl
```diff
@@ -78,6 +78,48 @@ end
...
+            _read(M, rep_size, X, i),
+        )
+    end
+    return Y
+end
+
+"""
+    change_metric(M::AbstractPowerManifold, ::AbstractMetric, p, X)
+
+Since the metric on a power manifold decouples, the change of a representer can be done elementwise
```
mateuszbaran, line 105:
```suggestion
Since the metric on a power manifold decouples, the change of metric can be done elementwise.
```

src/manifolds/ProductManifold.jl
```diff
@@ -135,6 +135,42 @@ function ProductVectorTransport(methods::AbstractVectorTransportMethod...)
...
+        submanifold_components(M, p),
+        submanifold_components(M, X),
+    )
+    return Y
+end
+
+"""
+    change_metric(M::ProductManifold, ::AbstractMetric, p, X)
+
+Since the metric on a product manifold decouples, the change of a representer can be done elementwise
```
mateuszbaran, line 159:
```suggestion
Since the metric on a product manifold decouples, the change of metric can be done elementwise.
```

src/manifolds/HyperbolicHyperboloid.jl
kellertuer: Are you sure we need that here?
The `EuclideanMetric` is a metric (i.e. the matrix representation is SPD), but the `MinkowskiMetric` is also a metric (i.e. the inner product positive definite) – _when_ we restrict Minkowski to a tangent space of the Hyperboloid model. That is actually a neat small exercise to do, but that way both Euclidean (from the embedding) and Minkowski are metric (positive definite inner products) on TpM and this can be done here.

src/manifolds/HyperbolicHyperboloid.jl
mateuszbaran: Yes, I think we could have something like
```julia
change_metric(::Hyperbolic, ::EuclideanMetric, ::Any, ::Any) = error("Changing metric from Euclidean to Hyperbolic is not possible (see Sylvester's law of inertia).")
```

src/manifolds/HyperbolicHyperboloid.jl
mateuszbaran: Ah, right, in this case we don't need that.

src/manifolds/HyperbolicHyperboloid.jl
kellertuer: Sure I'll add that.

kellertuer:
While “just shortly adding” a test for the fallback to use the local metric, I am somehow hitting a wall on an ambiguity, where I thin transparency does something and then it does not find the local metric function I implemented. So I will have do take a looooong look into that, where local metric is maybe too transparent or something.

For now I do not yet understand why the exact function I implemented (which I thought would be called for the manifold&metric in the test) is not even mentioned in the ambiguous cases...

mateuszbaran:
Yeah, we definitely don't have `local_metric` transparency tested well enough.

kellertuer:
Hm, I am a little lost on `local_metric`. I never seems to call the function I think it should call and it behaves differently in debugger then when running the current last block of the `metric.jl` tests manually (including the types and function definitions from before).

So I do not know how it was supposed to work nor how it is currently actually non-working, since I can not debug the current error (when the debugger throws an error even earlier).

Any help appreciated. I think `local_metric` currently acts transparently even on `MetricManifolds` which it should not, since without the `MetricManifold` (on a non-default metric) this function does not make sense.

kellertuer:
After about 2 hours fiddling with a lot of transparency, _maybe_ `local_metric` is fixed. But now `get_vector` and `get_coordinates` run into a tremendous amount of ambiguities, too.

I think the decorator pattern and bases being optional is something we should maybe redesign / avoid, since I feel like I can also spent a few weeks without sleep just trying to fix all this stuff and it just breaks even the last braincell after 2 hours.

mateuszbaran:
I think I'd vote for just not using the decorator pattern for metric-related functions. There is just not enough benefit to justify the effort needed to make it work. `get_vector` and `get_coordinates` require a lot of disambiguation but work decently enough now. They should likely be simplified though but I don't think it's an urgent issue.

kellertuer:
Well my current problem is, that I am a little bit stuck here. I do not know how to do it with decorator pattern nor do I knoww how to do it without.

Well I got local_metric running after a few hours – but this now affects `get_vector` and `get_coordinates` that are not working (with metric manifold) - so neither with or without decorator pattern do I know how to resolve the ambiguity here.

kellertuer:
I might have found the way to get the tests running. Note that in the current test manifold I had to

```
function Manifolds.get_vector!(
    M::MetricManifold{ℝ,<:TestEuclidean,<:TestEuclideanMetric},
    X,
    ::Any,
    c,
    ::DefaultOrthogonalBasis{ℝ,<:ManifoldsBase.TangentSpaceType},
)
```
do quite a detailed typing on the basis to resolve the ambiguity.

kellertuer:
It seems that sometimes Format & Documenter Fail because some packages could not be loaded. I just restarted both, since they should work. I also finished code coverage, so this should be good to go.

mateuszbaran:
Great, I'll take a look at it again soon.

src/manifolds/MetricManifold.jl
```diff
@@ -280,13 +428,16 @@ inner(::MetricManifold, ::Any, ::Any, ::Any)
...
 
 Return the local matrix representation at the point `p` of the metric tensor ``g`` with
-respect to the [`AbstractBasis`](@ref) `B` on the [`AbstractManifold`](@ref) `M`, usually written ``g_{ij}``.
-The matrix has the property that ``g(X, Y)=X^\mathrm{T} [g_{ij}] Y = g_{ij} X^i Y^j``,
-where the latter expression uses Einstein summation convention.
-The metric tensor is such that the formula works for the given [`AbstractBasis`](@ref) `B`.
+respect to the [`AbstractBasis`](@ref) `B` on the [`AbstractManifold`](@ref) `M`.
+Let ``d``denote the dimension of the manifold and $b_1,\ldots,b_d$ the basis vectors.
+Then the local matrix representation is a matrix ``G\in 𝔽^{n\times n}`` whose entries are
+given by ``g_{ij} = g_p(b_i,b_j)_p, i,j\in\{1,…,d\}``.
```
mateuszbaran, line 437:
```suggestion
given by ``g_{ij} = g_p(b_i,b_j), i,j\in\{1,…,d\}``.
```

src/manifolds/MetricManifold.jl
```diff
@@ -62,9 +206,9 @@ abstract type RiemannianMetric <: AbstractMetric end
 Compute the Christoffel symbols of the first kind in local coordinates of basis `B`.
 The Christoffel symbols are (in Einstein summation convention)
 
-```math
+````math
```
mateuszbaran, line 209: Is it supposed to have three or four backticks? We are inconsistent about it.

mateuszbaran (commented):
I think with some documentation polishing this should be ready to go :slightly_smiling_face:

src/manifolds/MetricManifold.jl
kellertuer: From Markdown both are fine; I prefer (and use most often) 3, I do not know why exactly. But overall our docs are not consistent in that, that's right.

mateuszbaran: OK.

mateuszbaran (approved):
I think we can merge this when CI passes and finally resolve #17  :tada:

src/manifolds/MetricManifold.jl
kellertuer: I also think we do not have to unify that, Markdown says any number of backpacks is fine and I would mostly use 3 but if I have. code block in my code block I would use 4 and for the inner one 3 :)

kellertuer:
Was a way to go, but I am happy we have that now :)

PR #427, opened 2021-09-20
Support for Zygote and ReverseDiff gradients

mateuszbaran:


mateuszbaran:
The format workflow failure looks bizarre.

mateuszbaran:
`Zygote` caused 4 new ambiguities in `in`.

mateuszbaran:
Now there are unsatisfiable requirements on Julia 1.4... the fact that Plots and the DiffEq ecosystem abandoned it quite a while ago makes me think that we should do so as well.

kellertuer:
Sure, we can got to 1.5; but when you adapt the test anyways maybe also start a test for 1.7 (i.e. that we have 1.5,1.6,1.7-rc1)?

mateuszbaran:
I've added tests on 1.7-rc1. The failure on Stiefel looks like a real bug in the code and I can reproduce it locally, could you take a look at it?

kellertuer:
It might take a few days for me to find time (also to install 1.7rc1 finally), but sure then I can take a look

mateuszbaran:
It seems that the Stiefel failure is the last one now.

kellertuer:
At first I thought – oh no! That is a quite involved part of approximating the log on Stiefel – but luckily I started the right way – two terminals one with 1.6 one with 1.7.

They changed the default random number generator from MersenneTwister to TaskLocalRNG, so with the different (still reproducible) tangent vector, the new tangent vector was too long, i.e. the local approximation was not possible anymore.

mateuszbaran:
Thanks for looking into it!

kellertuer:
Hm it now seems to fail on 1.6? Now I am a little confused, since for smaller vectors id should even less fail than before. I'll have to check.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.6.7"
+version = "0.6.8"
```
kellertuer, line 4:
```suggestion
version = "0.6.9"
```
(the one we merge today will have 8)

mateuszbaran: I think we'll have to make a small merge commit anyway :wink:

kellertuer: Sure, I just noticed, when I checked whether I have to bump further over on the other PR.

kellertuer (approved):
Looks good to me.

Issue #405, opened 2021-08-03
Taking AD seriously


kellertuer:
I think I would like to get into a proper way of handling AD on manifolds. I know we have quite some issues open here (#17, #42, #44, #88, #329) and we also have some support already. But we have not made much changes recently. Maybe this would be a good topic to tackle next.

In general I see two ways to go

1. Embedded manifolds and classical AD
2. “Intrinsic, geometric AD”

The first is, what I think [pymanopt](https://github.com/pymanopt/pymanopt) does, the second is to some extend done in [manopt](https://github.com/NicolasBoumal/manopt/blob/master/manopt/solvers/gradientapproximations/approxgradientFD.m#L86) with finite differences if I understand correctly, and I tried to do something like that for the Hessian [here](https://github.com/JuliaManifolds/Manopt.jl/blob/4ae4946f6ce7f336b75c967b0a556ffaf121b2d0/src/plans/hessian_plan.jl#L283) but did not yet find the time to continue that.

The main challenge with the first point, I think, is the conversion of an Euclidean gradient in the embedding to the Riemannian gradient, sometimes that is just a projection, sometimes that requires an adaption/calculation to adapt the metric. I have not yet understood for example all implications we would require to have something like `ehess2rhess` for the Hessian transform.

The main challenge for the second point is that from the building blocks (several are available in Manopt.jl, i.e. basic differentials, adjoint differentials, and gradients) we have to provide ChainRules.jl? The good thing is that we already have tangent and cotangent vectors.

For me the main questions / open points are
* [ ] what do the current backends support?
* [ ] What is left to do until we have a generic form of `egrad2rgrad`/ `ehess2rhess` to complete approach 1?
* [ ] Which backends would we support?
* [ ] for the second approach: Can we use backends for that or would we mainly need to defile new ChainRules?
* [ ] ...and of course we might have to look more into mutating and non-mutating variants

Finally, would we want to do it here (we can start here for sure) or do we want to to a ManifoldsAD.jl package?

Let's keep this topic as an overview for this topic for now. Feel free to add more points to the list of things to do.

mateuszbaran:
Good idea, incidentally this is quite high on my priority list as well (though I expect to mostly need Jacobians).

[...]

PR #429, opened 2021-09-23
Polish the Differential Interfaces

### LINKED ISSUE #42: Adding Zygote compatibility
It would be nice to add compatibility for Zygote. I know I personally need Zygote for its complex number support, and I'd also like to use `Manifolds` in the same code.

For the most part, we can probably expect Zygote to just work, with one major block: Zygote doesn't support mutation. I don't think this will be a big problem though. Zygote offers [`Zygote.Buffer`](https://fluxml.ai/Zygote.jl/latest/utils/#Zygote.Buffer), which behaves just like `similar`. All we should need to do is add something like this

```julia
_similar(args...) = similar(args...)

function similar_result(M::Manifold, f, x...)
    T = similar_result_type(M, f, x)
    return _similar(x[1], T)
end

finalize_result(M::Manifold, f, y, x...) = y

@require Zygote="e88e6eb3-aa80-5325-afca-941959d7151f" begin
   using Zygote
    Zygote.@adjoint _similar(args...) = Zygote.Buffer(args...), _->nothing
    finalize_result(M::Manifold, f, y::Zygote.Buffer, x...) = copy(y)
end
```

Every non-mutating function that uses `similar_result` then returns the output of `finalize_result`, and Zygote should just work.

Of course, this just brings Zygote support to the same level as ReverseDiff. It doesn't handle the issues raised in #17.

### LINKED ISSUE #88: Add FiniteDiff as a backend for jacobians.
I just read about it and it looks like it would be a very good choice for high-dimensional power manifolds.

https://github.com/JuliaDiff/FiniteDiff.jl

### LINKED ISSUE #428: local_jacobian dimension bug.
I am currently taking a look at our differential stuff – since we nearly got AD in the embedding working already

I stumbled upon our Jacobins and noticed, that they might be wrong. Let me explain why. I am looking at 

https://github.com/JuliaManifolds/Manifolds.jl/blob/9fc772a00e256db7f0ad51dfe3f063849128942c/src/manifolds/MetricManifold.jl#L450-L472

and I think @sethaxen you wrote this? My problem here is I think the dimension. If p is on the sphere (d+1-dimensional unit vector), we should do a d-dimensional Jacobean, but we actually do a d+1 dimensional one?
Usually the Jacobian is expressed in charts and I think this is missing here. But I think I also have a solution.
We can build an implicit chart given a basis in the tangent space (and we have a basis here anyways. The idea is as follows

1. Instead of `q` we could look at a function `X -> exp(M,p,X)`. this would give the same function locally, just in the tangent space.
2. we can now get to the right dimension by “building” `X` with respect to its coefficients in a basis, i.e. as ` c -> exp(M, p, get_vector(M, p, c, B)` (where `B` is our basis). Then the overall function reads
   ```julia
   c -> local_metric(M, exp(M, p, get_vector(M, p, c, B), B))
   ```
   would have the right dimensions (`manifold_dimension(M)` instead of `size(p,1)` - why the 1 anyways?).

Finally: Since we have a `NoneBackend` already, do we really need the dependency on `FiniteDifferences.jl` or can we load that optionally as we do for the other differentiation packages?

When my approach is ok, I can also add it to the code, I am revising that part anyways currently (and it provides me with nice automatic gradients already!) - so I already assign this to myself.

(related ( maybe also solved in that route: #88, since that is also just a small step there).

kellertuer:
This PR fixes #428, should also resolve #88. To some extend it might also resolve #42 together with the last PR? At least in the embedding Zygote can be used easily now; for intrinsic methods, there is even in theory still more work to do.

* [ ] with the introduction of properly solving the Jacobians, there is a few `get_vector!` functions missing, that are now required (yields a few `ErrorExceptions` where `MethodErrors` were exempted and a few missing implementation errors
* [ ] Test Coverage has to be checked.

This PR also makes the dependency on FiniteDifferences optional, since we now have enough to choose from.


kellertuer:
(I think Github might have gotten confused by my merge as well...)

mateuszbaran:
It looks like merging branches can be not only confusing for humans but for github as well :wink: .

I think after this we can mark #42 as resolved; the intrinsic approach is a somewhat separate issue, tracked in #405 IIUC.

src/Manifolds.jl
```diff
@@ -664,9 +668,9 @@ export get_basis,
 export AbstractDiffBackend,
     AbstractRiemannianDiffBackend,
     FiniteDifferencesBackend,
-    RiemannianONBDiffBackend,
-    RiemannianProjectionGradientBackend
-export diff_backend, diff_backend!, diff_backends
+    TangentDiffBackend,
+    RiemannianProjectionBackend
```
mateuszbaran, line 672: The new names are nice but unfortunately this is breaking (so I'd recommend calling it 0.7.0)

kellertuer: I think we even had one more thing we wanted to do for the next break? Maybe we should start having a dev branch of a branch for the next 0.x version to collect small breaking stuff...

But here, you are right. This is a breaking change (though it might not have been used before, still).

mateuszbaran: One thing I want to do by the next breaking release was removing `ProductArray`, I'm not sure at the moment if there is anything else but we could track it somewhere.

kellertuer:
I noticed one challenge when I tried fixing the first errors.

Currently we compute the new local metric jacobian here

https://github.com/JuliaManifolds/Manifolds.jl/blob/b2f267474305d02939598271e9d4fce565654303/src/manifolds/MetricManifold.jl#L471-L480

using a default retraction.

This has a limit for the case where we solve the ODE for the geodesic/exp at

https://github.com/JuliaManifolds/Manifolds.jl/blob/b2f267474305d02939598271e9d4fce565654303/src/manifolds/ConnectionManifold.jl#L164-L172

which internally calls `solve_exp_ode` which internally requires `local_jacobian` which .... uses the default retraction (exp) and the circle repeats finishing in s stack overflow.

We have to avoid, that (for the case the local jacbian is used to compute the exp problem) that the `ExponentialRetraction` is used.

src/Manifolds.jl
kellertuer: Ah, that's what I had in mind. maybe a dev or dev0.7 branch would be great for keeping track of those non.urgent but breaking changes (for a next breaking one that we want to actively do like this one)?

mateuszbaran: I have never done a dev branch like that so I don't really know how much trouble would that be to maintain (compared to a github milestone for example).

kellertuer: I have never used a milestone. My idea would have been that instead of just one PR for a new release like now, we collect a few PRs into a dev0.7 branch (all that are breaking for example) and wait for a few of them to be there – or one to be so interesting that we want to do a breaking PR now. with merging that the branch could also vanish, then it would just be a `dev0.7.0` branch.

mateuszbaran:
Hm, that's an interesting problem. My guess would be that the only way to make this work is working in a chart, that is `local_metric_jacobian` can would only be defined for bases induced by charts, where you don't have to use `retract`.

kellertuer:
uff. I would prefer checking (before calling local_jacobian that the (explicit, non-default) retraction is not the exponential retraction in solve_ode_exp. That would be maybe be the easier fix?

src/Manifolds.jl
mateuszbaran: Well, there is no harm in trying that, at worst we would abandon the branch. Feel free to make that if you want :slightly_smiling_face: .

mateuszbaran:
> uff. I would prefer checking (before calling local_jacobian that the (explicit, non-default) retraction is not the exponential retraction in solve_ode_exp. That would be maybe be the easier fix?

I don't see any retraction in `solve_exp_ode`, it just has `christoffel_symbols_second` which calls the problematic Jacobian calculation. You could check, though, that the retraction `local_metric_jacobian` gets is not the fallback case -- at least after refactoring the code of `exp!(M::AbstractConnectionManifold, q, p, X)` a bit...

mateuszbaran:
That is, `exp!(M::AbstractConnectionManifold, q, p, X)` would be an `ODERetractionMethod`, which would be the default retraction on `AbstractConnectionManifold`, and `local_metric_jacobian` could check if its retraction is different from `ODERetractionMethod`.

kellertuer:
> > uff. I would prefer checking (before calling local_jacobian that the (explicit, non-default) retraction is not the exponential retraction in solve_ode_exp. That would be maybe be the easier fix?
> 
> I don't see any retraction in `solve_exp_ode`, it just has `christoffel_symbols_second` which calls the problematic Jacobian calculation. You could check, though, that the retraction `local_metric_jacobian` gets is not the fallback case -- at least after refactoring the code of `exp!(M::AbstractConnectionManifold, q, p, X)` a bit...

Of course not. Until now it used `q+ ... `, that is what we are changing, so we are introducing this difficulty here. `solve_ode_exp` will have a retraction parameter and yes, we have to refactor here also the `exp!` .

kellertuer:
> That is, `exp!(M::AbstractConnectionManifold, q, p, X)` would be an `ODERetractionMethod`, which would be the default retraction on `AbstractConnectionManifold`, and `local_metric_jacobian` could check if its retraction is different from `ODERetractionMethod`.

That is a very good idea since it only approximates exp, that solves the problem to that extend.

mateuszbaran:
> > > uff. I would prefer checking (before calling local_jacobian that the (explicit, non-default) retraction is not the exponential retraction in solve_ode_exp. That would be maybe be the easier fix?
> > 
> > 
> > I don't see any retraction in `solve_exp_ode`, it just has `christoffel_symbols_second` which calls the problematic Jacobian calculation. You could check, though, that the retraction `local_metric_jacobian` gets is not the fallback case -- at least after refactoring the code of `exp!(M::AbstractConnectionManifold, q, p, X)` a bit...
> 
> Of course not. Until now it used `q+ ... `, that is what we are changing, so we are introducing this difficulty here. `solve_ode_exp` will have a retraction parameter and yes, we have to refactor here also the `exp!` .

OK, I thought that's not the issue we are solving here. It seems that `exp!(M::AbstractConnectionManifold, q, p, X)` could either work in a chart (and then no retraction is needed as it just does `+`) or use ManifoldDiffEq.jl, and both method could even coexist. Working in charts means a wider selection of ODE solvers, while using ManifoldDiffEq.jl means no problems caused by charts.

kellertuer:
> > > > uff. I would prefer checking (before calling local_jacobian that the (explicit, non-default) retraction is not the exponential retraction in solve_ode_exp. That would be maybe be the easier fix?
> > > 
> > > 
> > > I don't see any retraction in `solve_exp_ode`, it just has `christoffel_symbols_second` which calls the problematic Jacobian calculation. You could check, though, that the retraction `local_metric_jacobian` gets is not the fallback case -- at least after refactoring the code of `exp!(M::AbstractConnectionManifold, q, p, X)` a bit...
> > 
> > 
> > Of course not. Until now it used `q+ ... `, that is what we are changing, so we are introducing this difficulty here. `solve_ode_exp` will have a retraction parameter and yes, we have to refactor here also the `exp!` .
> 
> OK, I thought that's not the issue we are solving here. It seems that `exp!(M::AbstractConnectionManifold, q, p, X)` could either work in a chart (and then no retraction is needed as it just does `+`) or use ManifoldDiffEq.jl, and both method could even coexist. Working in charts means a wider selection of ODE solvers, while using ManifoldDiffEq.jl means no problems caused by charts.

Hm, then I though about a completely different way of solving this. Remove the`exp!` you mentioned and do that on a Retraction with an `ExponentialODEApproximationRetraction` that has a retraction internally (and with the right basis that retraction could be +).

but for your two approaches I have not yet understood how you mean to solve those, neither in theory for the charts with + not in practive for both.

mateuszbaran:
Could you explain how would you like to use a retraction in `ExponentialODEApproximationRetraction`?

kellertuer:
I would remove the default fallback of `exp!` to the ODE. Reason: The ODE is only a numerical approximation and we avoid the ambiguity we are currently running into.

A retraction of type `ExponentialODEApproximationRetraction` would have a field, which retraction (and maybe also with basis?) to use and this would (in the constructor) be checked that it can not be the `ExponentialRetraction` (since that is what is introducing the current infinite loop).

mateuszbaran:
> I would remove the default fallback of `exp!` to the ODE. Reason: The ODE is only a numerical approximation and we avoid the ambiguity we are currently running into.

Yes, I agree that this is a good idea.

> A retraction of type `ExponentialODEApproximationRetraction` would have a field, which retraction (and maybe also with basis?) to use and this would (in the constructor) be checked that it can not be the `ExponentialRetraction` (since that is what is introducing the current infinite loop).

I don't quite see yet how you want to rewrite `solve_exp_ode`. Anyway, I think the in-chart method could be said to use the following retraction:
```julia
struct ChartRetraction{TA<:AbstractAtlas,TI} <: AbstractRetractionMethod
    A::TA
    i::TI
end

function retract(M, p, X, cr::ChartRetraction)
    pc = get_parameters(M, cr.A, cr.i, p)
    Xc = get_coordinates(M, p, X, InducedBasis(TangentSpaceType, A, i))
    return get_point(M, cr.A, cr.i, pc + Xc)
end
```
The problem is that it's currently only implemented for `StereographicAtlas`.

kellertuer:
Ah that looks also nice and would be the inner retraction then, sure.

mateuszbaran:
Notice though that when you write a solver for geodesic using that retraction you can eliminate calling `get_parameters` and `get_point` on every step (and only do so at the very beginning at at the very end), and in doing so you effectively end up with my "geodesic in chart" idea.

mateuszbaran:
Yes, and I think that solver would be the easiest one to write (and the closest to the current implementation).

kellertuer:
Could you try to propose that? Because I think the generic one should be relatively close to what is done now actually (with the retraction inside instead of the +, sure your approach would really just use + in `_jacobian` call).

mateuszbaran:
I think I could write that.

Also, which + do you want to replace with retraction? Currently the addition (for performing a step) is done inside OrdinaryDiffEq.jl.

kellertuer:
Before we started this PR, there was basically a `q+...` (yes within OrdinaryDiffEq), which we now shifted accordingly into the coordinates of a vector; this in principle (since the function we work on in f \circ retr (circ get_vector) is now replaced by the retraction. So I do not want to, I already did that; just due to the loop that the retr causes an inner soilve_ode – again and again – it is just failing horribly, but hour combined solution ideas should fix that :)

src/Manifolds.jl
kellertuer: Would you prefer milestones? My main point is, that I would like to avoid forgetting barking things that we have in mind to add for the next time we do a breaing release – like now with the removal of the type – I already had forgotten those details.

mateuszbaran: I think either a milestone or an issue with a list would be just easier. You can take a look at, for example, `IncrementalInference.jl` to see how milestones work: https://github.com/JuliaRobotics/IncrementalInference.jl/milestones .

mateuszbaran: I've made a milestone for us :slightly_smiling_face: .

kellertuer:
ON my way of fixing the ODE thingy, I noticed that both connection and metric decorator implement the jacobian-things (local metric, first, second), but while Metric just calls an ODE for the local metric jacobian and continues from there, the ConnectionManifold computes the second jacobians with an ODE (but neither first nor local metric jacobina9: Is that an inconsistency or intended?

kellertuer:
I sketched a first variant of what I mean to use to resolve the problem, that it might happen, that exp is used to approximate exp.

Now the default `exp!` on `AbstractConnectionManifolds` uses the `ODEExponentialRetraction` with a default basis and a default retraction – but this errors if the default retraction is a `Exponentialretraction` itself.

I have not yet fully understood
* [ ] why local metric jacobian is calculated that differently in metric and connection manifolds
* [ ] how the `solve_exp_ode` (in ode.jl) works – I also think we should move this over to `differentiation/differentiation.jl` since that is where it fits better. If I have understood this function I also want to document it better since I think it should not just return sol but the actual point it computed (this extraction of the final point is currently done in the new retract).
* more documentation for the methods in this newest commit are definetly missing, but I'd like first some more feedback on the approach

note that since we switched from some `q+X` in the embedding to using a basis and a retraction, the solver now also required a retraction itself, that is also passed on to jacobian.

The main thing it actually solves / divides more clearly is how the local parameters (c -> Y -> q) are determined and compared to previously avoids that Y -> q is done with exp (when exp is what we aim to compute).

mateuszbaran:
> * why local metric jacobian is calculated that differently in metric and connection manifolds

Connection manifolds don't have metric Jacobians. Riemannian metric manifolds are defined by their `local_metric` and further methods work from that. Connection manifolds are defined by `christoffel_symbols_second` .

> * how the `solve_exp_ode` (in ode.jl) works – I also think we should move this over to `differentiation/differentiation.jl` since that is where it fits better. If I have understood this function I also want to document it better since I think it should not just return sol but the actual point it computed (this extraction of the final point is currently done in the new retract).

Well, that's the point I was trying to make in #431 -- I don't see how it works outside of some very specific cases either. Returning `sol` doesn't make much sense to me apart from being able to use it to sample the geodesic at arbitrary time points.

I'll try to take a closer look at your recent changes.

src/manifolds/ConnectionManifold.jl
```diff
@@ -162,13 +197,13 @@ in an embedded space.
...
-    B = induced_basis(M, A, i, TangentSpace)
-    sol = solve_exp_ode(M, p, X, tspan, B; dense=false, saveat=[1.0])
-    n = length(p)
-    return copyto!(q, sol.u[1][(n + 1):end])
+    return retract!(
+        M,
+        q,
+        p,
+        X,
+        ODEXxponentialRetration(ManifoldsBase.default_retraction_method(M)),
```
mateuszbaran, line 205: What do you think about leaving `exp!` unimplemented and instead returning `ODEExponentialRetration` as the default retraction?

kellertuer: Thats also a neat idea :)

kellertuer:
> > * why local metric jacobian is calculated that differently in metric and connection manifolds
> 
> Connection manifolds don't have metric Jacobians. Riemannian metric manifolds are defined by their `local_metric` and further methods work from that. Connection manifolds are defined by `christoffel_symbols_second` .

Thanks for the clarification – this is not yet that well documented I think

> 
> > * how the `solve_exp_ode` (in ode.jl) works – I also think we should move this over to `differentiation/differentiation.jl` since that is where it fits better. If I have understood this function I also want to document it better since I think it should not just return sol but the actual point it computed (this extraction of the final point is currently done in the new retract).
> 
> Well, that's the point I was trying to make in #431 -- I don't see how it works outside of some very specific cases either. Returning `sol` doesn't make much sense to me apart from being able to use it to sample the geodesic at arbitrary time points.

My current approach works for quite a set of basis and retractions I think – as long as the retraction is well defined for tangent vectors that reach up to `X` (i.e. `X` should be in de domain of the retraction), because that gives you a chart.
Exactly, maybe I want to change the return value to directly return the evaluate at value anyways.

src/manifolds/ConnectionManifold.jl
```diff
@@ -106,15 +141,25 @@ function christoffel_symbols_second_jacobian(
...
-    n = size(p, 1)
+    d = manifold_dimension(M)
     ∂Γ = reshape(
-        _jacobian(q -> christoffel_symbols_second(M, q, B; backend=backend), p, backend),
-        n,
-        n,
-        n,
-        n,
+        _jacobian(
+            c -> christoffel_symbols_second(
```
mateuszbaran, line 150: It looks like a good direction but I'm not quite sure what are the necessary conditions for retraction and basis to make this work.

mateuszbaran: I think it would be a good idea to make an example on a sphere that shows correct/incorrect calculations of its curvature for different bases and retractions.

mateuszbaran:
> because that gives you a chart.

One thing I worry about is that the basis `B` at a point calculated by that retraction (here: https://github.com/JuliaManifolds/Manifolds.jl/pull/429/files#diff-ce4a06670d9649fc569a0f1e4f09b9654c590f061a489e380ca1a978d008752fR152 ) may not be the same as the basis induced by that retraction chart.

src/manifolds/ConnectionManifold.jl
kellertuer: Thinking about it, maybe it is not the best idea to make that the default retraction, because when someone _defines_ `exp!` for his/her manifold, we actually want that to be the default.

src/manifolds/ConnectionManifold.jl
kellertuer: I think I do not follow you here. _Any_ Basis and _any_ retraction give you (locally!) a chart here. Any. It might not work for very large X (projection retraction on the sphere and X of length pi/2), but for small enough X all basis and retraction combinations work.

mateuszbaran: I've elaborated a bit on this here: https://github.com/JuliaManifolds/Manifolds.jl/pull/429#issuecomment-927105712 . The problem is that here we use the basis B at more than one point and there are no requirements on bases that relate these two in any way.

kellertuer: To be precise. Let X be a tangent vector (for approximating exp(m,p,X)) lets say on the sphere to have dimension 2. Then in _any_ basis I can write X = b1X1+b2X2 (X1,X2 the basis vectors), so the curve c(t) = t(b1,b2) (in basis parameter space) is the same as c(t) = X(t) (just with get_vector). this only requires _any_ basis. similarly I get a curve on the manifold with _any_ retraction, i.e. d(t) =retr(M,p, get_vetcor(M, p, c(t))). 

Sure if X is too large such that the retraction is not useful anymore (p North Pole, norm of X larger than pi/2, Projectionretraction), this does not work, but locally it always works. But this is the same for _any_ chart you want to get to.

kellertuer:
I either do not see the problem or we speak about completely different things. I just do not see where this should not work (locally) for sure. And we seem to write the same (both) again and again and again.

I am sure this works, you are not. I don't get why you think that – you seem to get my point. What do we do here? We can continue this for quite a while but I feel a lot like repeating myself already :/

mateuszbaran:
OK, show me that `local_metric_jacobian` works for a basis of R^2 such that `B([p1, p2]) = ([p1, p2], [p1, -p2])` -- here I'm specifying basis vectors at a point `[p1, p2]` -- with the basic Euclidean metric. Let's restrict this to the positive quadrant so that I don't have to complicate my example. That is, for example calculate the Jacobian for point `[2, 3]` using that method. If it gives the correct result -- I admit I was in error. If it doesn't -- please explain why it's not a valid basis.

kellertuer:
Hm, so this is programmatically not so easy since EuclideanMetric undecorated a MetricManifold(Euclidean(), EuclideanMetric()), so I would have to first implement a new metric and that seems to be a lot of effort. Can you elaborate why you think that should fail? I don't see why it should.

mateuszbaran:
So we've discussed it on Slack and the conclusion is that I was expecting `local_metric_jacobian` to use induced bases of the retraction chart of `B` at `p` but here it uses `B` at other points, which may or may not agree. It is also fine but it definitely has to be documented :slightly_smiling_face:.

kellertuer:
I fiddled a little bit with the solve_ode_exp and it now returns the final point and has the right dimensions.

I got the first Metric test to pass, but for the second I again got stuck in implementing the right `get_vector!` function. I thought it is the one now implemented – but its not.

kellertuer:
I fixed most of the errors, all but 4. They are get_vector and local_metric ones with the (not really well documented) TestSphere and InducedBasis. They seem (at least for me) not easy to be fixed, since I have to delve into the Test-here and the corresponding induced bases. Might try somewhen.

src/manifolds/ConnectionManifold.jl
kellertuer: Is it ok to keep `ExponentialRetraction` as default? I think if exp! is present we want to use that instead of solving the ODE, so I would like to keep `ExponentialRetration` as a default retraction,

kellertuer:
I tried quite a few things yesterday for the final two `get_vectors!` that seem to be missing for the TestSphere, but the only thing I seem to be able to do is to increase the errors. Can you maybe take a look @mateuszbaran ? Ist something with the bases and dispatch, I fear and a really nasty one.

mateuszbaran:
I'll try to take a look at it today.

mateuszbaran:
I can see the following errors:
```julia
 Expression: isapprox([sph_to_cart(yi...) for yi = geodesic(M, p, X, T)], geodesic(S, pcart, Xcart, T); atol = 0.001, rtol = 0.001)
  MethodError: local_metric(::AbstractManifold, ::Any, ::AbstractBasis) is ambiguous. Candidates:
```
And that makes sense to me, as the required method of `local_metric` isn't actually implemented. It can be seen in the stack trace which one that is:
```julia
    [5] inverse_local_metric(M::MetricManifold{ℝ, TestSphere{2, Float64}, TestSphericalMetric}, p::SVector{2, Float64}, B::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType})
      @ Manifolds ~/.julia/dev/Manifolds/src/manifolds/MetricManifold.jl:341
```
`MetricManifold{ℝ,<:TestSphere,<:TestSphericalMetric}` only has local metric for the `InducedBasis`. I'm not entirely sure why at the moment but note that the last rework of this code was from the point of view of a different usage of these bases. IIRC I was mostly considering usage of metric-related functions with `InducedBasis` primarily because that's the only case where they could (in my mind) be effectively written in the way I planned -- default bases simply didn't provide enough guarantees.

Anyway, I also didn't like the way I used bases of tangent spaces everywhere there and my intention was to, for example, pass a basis of 3-index tensor space at a given point with the right covariance when getting coefficients of Christoffel symbols. But that was never realized as it seemed like over-engineering of code I likely will never need and I just wanted to get the old metric code somewhat make sense with the introduction of charts.

So, the second error is:
```julia
  get_vector! not implemented on MetricManifold(TestSphere{2, Float64}(5.331830160438614), TestSphericalMetric()) for arguments Vector{Float64}, Vector{Float64}, Vector{Float64} and InducedBasis{ℝ, ManifoldsBase.TangentSpaceType, RetractionAtlas{ℝ, ExponentialRetraction, LogarithmicInverseRetraction, DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType}}, Vector{Float64}}.
```
And indeed I don't see the definition of that method. You explicitly call `get_vector` in `local_metric_jacobian` so that doesn't look surprising to me. It feels like it should be possible to provide a default implementation of `get_vector!` for induced bases but that may not be easy.

src/manifolds/ConnectionManifold.jl
mateuszbaran: That's true. So it really makes sense to keep it this way.

kellertuer:
I implemented both – took me 4 approaches each, one of them still reports, it does not exist (?!) I can not declare it more precisely actually.

And there are now more errors because some numbers are wrong, but that is because my implementations where dummy numbers and not the correct ones. Should we make remove this stuff here and just check that the solving of the ode works instead of writing two whole manifolds in a test? This way the test nearly gets its own library/package (and we maybe should even put the base manifold into an own file since it is defined in multiple places for tests currently).

mateuszbaran:
OK, I think after my changes all remaining test failures are incorrect results.

Also, I think it makes sense to separate ODE tests into a new file.

kellertuer:
Could you explain what I missed? Our implementations are not that different, though mine still did not find some functions...

mateuszbaran:
The test was expecting `local_metric` to be defined for `DefaultOrthonormalBasis`. Removing that method of `inverse_local_metric` didn't break anything so I removed it (I would expect the default implementation to work correctly).

mateuszbaran:
> In the version _before_ this version, the ode-parameters were stored as points in the embedding, but now I thought it is necessary to reduce the size to the manifold_dimension?

I'm not sure. I only know how to do this ODE solving either in a chart or through ManifoldDiffEq.jl.

> tried both with less or more parametrised bases – never tries to “find/catch” - I seem generally unable to “track” down the right signature for bases to be found in dispatch on them with all their parameters.

I'm fixing it, the main problem I think was that you forgot `Manifolds.` to add methods to the function instead of making a new function.

mateuszbaran:
There are still errors but much further along (`local_metric_jacobian` is broken).

kellertuer:
Thanks for the help, I think in my many tries I had that variant, too, but I really seem to struggle a lot with things involving metric/basis combined here, especially which parameters to specify for the basis (I feel like bases really have many parameters, though it is also just 3 or something?).

Sure there are more errors down the road, I will slowly work towards that.

Is there a reason I can not ask the sphere for the metric? I had hoped it just has a metric tensor (namely the one from the embedding restricted)?

mateuszbaran:
> Thanks for the help, I think in my many tries I had that variant, too, but I really seem to struggle a lot with things involving metric/basis combined here, especially which parameters to specify for the basis (I feel like bases really have many parameters, though it is also just 3 or something?).

Bases really have only two meaningful parameters: whether it's a real or complex basis and whether it's a basis of the tangent space or some other space. Probably the length of `<:ManifoldsBase.TangentSpaceType` makes it seem like something much more involved and it would be nice to figure out a shorter notation for bases of tangent spaces.

> Is there a reason I can not ask the sphere for the metric? I had hoped it just has a metric tensor (namely the one from the embedding restricted)?

The only reason is that no one has implemented that function yet :smile: . Metric tensor from the embedding can't work because:
1. On an N-sphere it's an NxN matrix, not an (N+1)x(N+1).
2. There is by default no relation established between bases of embedded manifolds and bases of the embedding.
Both problems could be solved but that's relatively complicated. And it's much easier to just write a special case for sphere.

kellertuer:
Actually, we _have_ the local metric already, since the Sphere has the `EuclideanMetric` inherited from its embedding and for any manifold with the Euclidean metric, we already return the diagonal matrix accordingly. So we (both) were just missing to call the local Metric with the `MetricManifold(Sphere(n), EuclideanMetric())`

kellertuer:
Somewhere we still have some dimension mixup (i.e. whether we are in coordinates at a place or in the representation). Currently I somehow end up in a `get_vector` on the sphere S2 with a coordinate vector of length 3, which is definetlz wrong. We might have to check that thoroughly again (until now – also with the testSphere over in Metric coordinate and vector dimensions always agreed I think?)...

kellertuer:
I am a little stuck (maybe I thought about it too much or something) – somehow I can not manage to write the solve_exp_ode correctly, currently I either end up with two things . Let `d` be the manifold dimension and `n` be the representation size (for a matrix represented in vectors, otherwise reshape to vectors so d and n are numbers) Then I am able to

a) write a variant that tries to work in a too large space (n) where the `@einsum` complains (which is totally correct, since that is of size d\times d and the vectors I would then multiply it with in the ensue are n and n

b) write a variant that works in the right space, then I am stuck in how to initialise the p (just zero?) and how to fill that part of the `u` anyways.

And in general I am not sure in which of the variants to work, since we should have a version (i.e. in some chart) where we can do a “Euclidean” ode, i.e. one where we have a plus right?

I also just tried to write the fallback for `exp!` directly on the `AbstractManifold` level, since for the default metric (or default connection, too) the Jacobian would just be implemented for `Sphere` for example, right?

mateuszbaran:
> a) write a variant that tries to work in a too large space (n) where the `@einsum` complains (which is totally correct, since that is of size d\times d and the vectors I would then multiply it with in the ensue are n and n

Yes, in this variant you need a different way of calculating the tangent vector for ODE and, unfortunately, I don't currently see a good way to do this.

> b) write a variant that works in the right space, then I am stuck in how to initialise the p (just zero?) and how to fill that part of the `u` anyways.
> And in general I am not sure in which of the variants to work, since we should have a version (i.e. in some chart) where we can do a “Euclidean” ode, i.e. one where we have a plus right?

The variant I've written in #431 does exactly that: `p` is initialized based on its position in a chart (it's 0 if you take for example a retraction chart at a point but you can take a different one and it's not zero then) and a normal "Euclidean" ODE is solved by OrdinaryDiffEq.jl.

> I also just tried to write the fallback for `exp!` directly on the `AbstractManifold` level, since for the default metric (or default connection, too) the Jacobian would just be implemented for `Sphere` for example, right?

I think we may wait with that, currently the preferred way of working in Manifolds.jl is to implement explicit retractions, and the support for ODE-based retractions is very experimental. I'd wait with making it the default for non-metric manifolds until we have it at least somewhat tested and benchmarked on a simple problem (such as geodesics on an ellipsoid).

kellertuer:
Thanks for the help. I will take the approach (a) back to what we had, just generalise it to (isometrically) embedded sub manifolds where I think I have an idea (and then getting to finish this PR and just rewriting/reorganising the tests a little bit).

For the generic fallback my idea stems more from the fact that for the default metrics we “strip away” the metric part anyways (in most cases), since we assume to have the implicit (first/default) metric to be implemented without the metric decorator.

mateuszbaran:
> Thanks for the help. I will take the approach (a) back to what we had, just generalise it to (isometrically) embedded sub manifolds where I think I have an idea (and then getting to finish this PR and just rewriting/reorganising the tests a little bit).

OK, I'll wait until you try your idea :slightly_smiling_face: .

> For the generic fallback my idea stems more from the fact that for the default metrics we “strip away” the metric part anyways (in most cases), since we assume to have the implicit (first/default) metric to be implemented without the metric decorator.

I think that could work but you'd still need ManifoldDiffEq.jl instead of OrdinaryDiffEq.jl, right?

kellertuer:
Currently 26 tests fail in metric, and I have no clue why, because I do not understand what they test, since there is no explanations around and code without comments is a little hard to reverse-engineer.

So I am not sure how to continue here – revert all the ODE stuff and just keep the renaming of the embedded AD stuff? Omit this PR totally? The last part would be a little sad since I spent really a lot of time on this. Even the first is a little sad, but I fear I do not understand enough of the ODE and how to do it in a basis or the embedding to fix it here.

mateuszbaran:
> Currently 26 tests fail in metric, and I have no clue why, because I do not understand what they test, since there is no explanations around and code without comments is a little hard to reverse-engineer.

Looking at that code, I would be surprised if it made sense to you :sweat_smile: . It has a few things done very much not right.

OK, so this is what these tests supposed to do:
1) We take a scaled sphere with the default atlas (IIRC it's the exponential retraction atlas):
```julia
g = TestSphericalMetric()
M = MetricManifold(Sr, g)
A = Manifolds.get_default_atlas(M)
```
2) We take a point on that sphere and a chart that contains that point:
```julia
p = vtype([θ, ϕ])
chart_p = get_chart_index(M, A, p)
B_p = induced_basis(M, A, chart_p, TangentSpace)
```
This is somewhat incorrect as `p` is not a point in the correct representation, I must've updated it incorrectly. `p` should be just a normal point, like `[1.0, 0.0, 0.0]`. For some reason `local_metric` for the scaled sphere is written in a way to take this into account but that's not how `local_metric` should work. So it needs to be rewritten to a) accept point in the embedding representation and b) we should make sure that `local_metric` actually returns the coefficients of metric in the basis indicated in the argument.

This leads us to reconsider what point (1) does, and thus we should use the chart corresponding to spherical coordinates, not a chart from the default atlas. It's almost as if someone only cared about making these tests pass :sweat_smile: . Sorry.

3) After the chart confusion is cleared, we encounter the failing tests. They are supposed to test if the coefficients of these various tensors are correct (expressed in the spherical coordinates chart).

Perhaps the mix between these two charts breaks your new code but I'm not entirely sure. I can at least fix the old tests so that they make sense but you need to give me a few days.

> So I am not sure how to continue here – revert all the ODE stuff and just keep the renaming of the embedded AD stuff? Omit this PR totally? The last part would be a little sad since I spent really a lot of time on this. Even the first is a little sad, but I fear I do not understand enough of the ODE and how to do it in a basis or the embedding to fix it here.

The renaming is certainly fine. The ODE part is very tricky and I think we should be careful here -- it'd very likely take quite a long time to polish it.

kellertuer:
Thanks for the heads up. What do you think about maybe splitting this into two PRs then?

I could make a new branch, cherry pick the Renaming commits (or take the lines over), and we do a second one to clear the ODE stuff (maybe on a separate PR maybe on your PR for the ODE in charts)?

I would love to have the differentiation part (without ODE) finished to also finish my tutorial on the gradient conversion. I might take a break from the ODE then, though, to first look into the Hessian conversion. But that said, of course there is absolutely no problem if it takes a while :)

mateuszbaran:
Sure, splitting it into two PRs is a good idea. One PR for renaming, then I can continue my PR to have the geodesics in charts, and then you can work on your approach in another PR.

> I would love to have the differentiation part (without ODE) finished to also finish my tutorial on the gradient conversion. I might take a break from the ODE then, though, to first look into the Hessian conversion. But that said, of course there is absolutely no problem if it takes a while :)

Sure, the differentiation part should be simple to finish. Hessians (and, probably more importantly, Hessian-vector products) may be an interesting thing to do after that :slightly_smiling_face: .

kellertuer:
The dimensions problems for the solve_ode_exp we will have to consider in another PR (will only do that after the Hessians).

PR #430, opened 2021-09-24
Introduces vector transport by projection on FixedRankMatrices.

kellertuer:
This is just a small PR introducing vector transport by projection on fixed Rank matrices. Note that it embeds the UMTTVector in the embedding to project, this is necessary since UMTTVector is stored in a reduced form (i.e. for X at p the data from p is required to get the original matrix to represent X).


kellertuer:
I am confused, there is code failing that I did not touch?

kellertuer:
Ah. There was no (automated) test, since the automatic vector transport test requires an inverse retraction that we do not have. Now the twi missing lines should be covered.

mateuszbaran (approved):
Nice, looks good :+1:

kellertuer:
Perfect, I already fixed the version number – so if the tests pass this can be merged.

PR #432, opened 2021-10-03
Polish Diff Interfaces (V2)

### LINKED ISSUE #88: Add FiniteDiff as a backend for jacobians.

kellertuer:
resolves #88, #42. Supersedes the #429, were I mixed in a fix for the ODE solver for geodesics; that should be done in a new PR. Not this one just introduces an `ODEExponentialRetraction` to properly distinguish between exponential maps in closed form and their approximations.


kellertuer:
This should nearly be done, the only thing I could not figure out is the new `_jacobian!`that passes on to FiniteDiff, since it seems to act differently from passing on the non mutating variant (hence the test is for now marked as broken).

kellertuer:
So the test is still marked as broken because mutating stuff in FiniteDiff is _really_ tricky.

A main thing we missed is, that the function f we want to take the Jacobian of in this case has also to be mutating.
So locally for some functions and some methods I can get it to work. But it does not work with central differences it seems but with forward it did – but not in our framework

For example
```
f1(x) = x[1] + x[2]^2
f2!(Y,X) = ( Y .= x[1] + x[2]^2 )

X1 = FiniteDiff.finite_difference_jacobian(f1, [1.0, -1.0])
X2 = simlar(X1)
FiniteDiff.finite_difference_jacobian!(f2!, X2, [1.0, -1.0])
```

works, but as said I can not write a test that uses our interface and works properly. I also do not want to play around on that more, since it is just guesswork.

So I‘d say this PR is ready to merge though this small point is open (more precisely it is not 100% clear how to use our mutating Jacobian framework, which we anyways only have for FiniteDiff)

mateuszbaran:
I think it would be OK to just not have the mutating Jacobian for now.

Project.toml
```diff
@@ -1,13 +1,12 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.6.9"
+version = "0.6.10"
```
mateuszbaran, line 4: This is breaking so probably 0.7.0?

src/differentiation/differentiation.jl
```diff
@@ -47,98 +49,119 @@ for example [`gradient`](@ref Manifolds.gradient(::AbstractManifold, ::Any, ::An
...
-_gradient(f, p) = _gradient(f, p, diff_backend())
+This retraction approximates the exponential map by solving the correspondig ODE.
+Let ``p\in\mathal M`` and ``X\in T_p\mathcal M`` denote the input for the exponential map
+and ``d`` denote the [`manifold_dimension`](@ref) of `M`.
 
-_gradient!(f, X, p) = _gradient!(f, X, p, diff_backend())
+This the ODE is formulated in a chart constructed using an [`AbstractBasis`](@ref) `B` and
+and using either an embedding or a [`AbstractRetractionMethod`](@ref) `R` as follows.
+Given some coordinates ``c\in ℝ^d`` - these can be used to form a tangent vector
+with restect to th basis `B, i.e. ``c \mapsto Y=``[`get_vector`](@ref)`(M, p, c, B)`.
```
mateuszbaran, line 129: I'd prefer to leave the details regarding how the basis is used out of this PR -- this still needs to be worked out for cases other than the induced basis of a chart.

Project.toml
kellertuer: Ah yes you're right.

src/differentiation/differentiation.jl
kellertuer: I must have missed those lines in my interactive merge (removing parts from the last PR). Will update this, sure.

kellertuer:
> I think it would be OK to just not have the mutating Jacobian for now.

It seems to work to some extend, but I am not yet 100% sure how, since in small examples without our wrapper I got it to work just not yet with the wrapper.

src/differentiation/riemannian_diff.jl
```diff
@@ -38,43 +43,61 @@ function gradient!(M::AbstractManifold, f, X, p, backend::AbstractRiemannianDiff
...
 
-differential(M::AbstractManifold, f, p) = differential(M, f, p, rdifferential_backend())
+@doc raw"""
+    TangentDiffBackend <: AbstractRiemannianDiffBackend
 
-function differential!(M::AbstractManifold, f, X, p)
-    return differential!(M, f, X, p, rdifferential_backend())
-end
+A backend that uses a tangent space and a basis therein to derive an
+(intrinsic, approximate) differentiation scheme.
```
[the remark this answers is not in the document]
kellertuer: HM, you can use any method to compute the Euclidean part, so sure (Zygote for example) we have support for AD.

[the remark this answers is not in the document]
mateuszbaran: So maybe let's remove the word "approximate"?

mateuszbaran:
I've tried fixing that FiniteDiff.jl Jacobian problem but I've also given up.

PR #435, opened 2021-10-07
Remove `ProductArray`

mateuszbaran:
This time it should work


mateuszbaran:
I've made an additional commit with a small change: ManifoldsBase isn't imported with all exported symbols so that from now on we can consider porting things from Manifolds.jl to ManifoldsBase.jl non-breaking (I hope this should work).

kellertuer:
Cool, Then let‘s go for the 0.7.0 :)

Issue #417, opened 2021-08-27
Support projection matrix points for Grassmann


sethaxen:
What would it take to allow representing points on the Grassmann manifold as the unique projection matrix `P = X X'`, where `X` is a point on the Stiefel manifold, instead of `X` as we currently do?

The reason for considering this is that for ManifoldMeasures, we especially want to support measures in PPLs, where identifiability is a major concern. So even though `X` is more efficient for storage, indeed much more efficient when `n>>k`, it may be that the non-identifiability of the `X` representation for a Grassmann variable would cause adaptive sampling strategies like HMC to become even slower.

I imagine to do this via dispatch, we'd need to implement a custom `ProjectionMatrixPoint` or something like that, right?

Or are there more storage efficient unique representations of points on the Grassmann manifold?

kellertuer:
In the draft (see atop) I started implementing the projection representation approach, the question is now – which functions would you like to have – I think that even the metric might first require more work on a structure/interface for quotient manifolds (cf #205), especially horizontal and vertical lift; concerning representations one could think of a quotient manifold even as an `AbstractEmbeddedManifold` subtype.

So for short: Which functions would you like to see included in #443 (for now it is just checks).

mateuszbaran:
Probably exponential map and vector transport would be fine as a start? There are some formulas here: https://arxiv.org/abs/2011.13699 .

kellertuer:
The exponential map (and `inner`) require the horizontal lift. That's why I think these would fit better only after having proper support for quotient manifolds.

mateuszbaran:
We could skip that for this representation but indeed it would be better to do this more generically by introducing horizontal lifts.

kellertuer:
I would maybe finish this issue/PR without those and add them as soon as we have that support? Indeed the simple idea is to have a `AbstractQuotionsmanifold <: AbstractEmbeddedManifold` (again for a default manifold being a quotient already) and a `QuotientManifold{M,N}` to model M as a quotient of N with all the nice functions one can have then (vertical, horizontal lifts, projection...)

mateuszbaran:
I think it's fine to merge #443 without functions requiring lifts but I'm not sure if that would resolve this issue.

kellertuer:
I am also not sure - hence the question. Sure we can continue this after #205 is done and keep this open until then. What do you think @sethaxen ?

kellertuer:
resolved by #443

Issue #205, opened 2020-07-27
Model quotient manifolds.


kellertuer:
I would like to have an abstract embedding type where the manifold is obtained from the embedding by a quotient manifold.

It might be an idea to do this with just an embedding type, but I think it might be worthwhile to even introduce a subtype

`QuotientManifold <: Manifold`,
that for example could be parametrised by two manifolds (whole space, factor) for the case of a submersion.

mateuszbaran:
Actually, what operations on quotient manifolds do you have in mind? Vertical/horizontal decomposition of tangent vectors? Some things that can be constructed using `optimal_alignment`?

kellertuer:
My ideas are

* vertical / horizontal decomposition
* map (usually called pi) to the representations (if that's different, but like project it might be id)
* for all quotients adapt things like `isapprox` to use distance or decomposition above

I came to that idea when I just adapted one `is approx`, but have not yet the full list of features In mind. Still, I feel it might be beneficial similar to EmbeddedManifold.

mateuszbaran:
I think `QuotientManifold <: AbstractEmbeddedManifold` would make sense, right?

About the map pi, would that be different from `project`?

`QuotientManifold` should probably hold the `AbstractGroupAction` that is used to define it. Generic `distance` and `isapprox` could then use `optimal_alignment`.

By the way, landmark manifolds may be a good playground for quotient manifolds.

kellertuer:
For quotient manifolds that are submersions, you are right, they are both embedded and quotient, then `project` is the same as `pi`.
I am not 100% sure whether this is always the case for all quotient manifolds, but for this case `AbstractEmbedddManifold` makes sense and would be useful (using `project` for `pi`.

And yes, in general it should also know its `AbstractGroupAction`.

I haven't worked with landmark manifolds, just stumbled upon this with fixed rand manifolds, or others, where one has to quotient out the orthogonal group.

kellertuer:
In #443 I decided to now call the projection `canonical_project` to avoid confusion with the projection from the embedding.

The `QuotientManifold therein does not yet know its `GroupAction`, but that could also be a trait (i.e. something we again assume to be implicitly known if not specified explicitly).

mateuszbaran:
> _edit_: I will definitely model vertical and horizontal vectors in that PR, but I am not yet sure what to do with optimal alignment, since I have not yet read much about that (in Absil, Mahony, Sepulchre).

[...]

PR #443, opened 2021-10-29
Grassmann manifold with projector points

kellertuer:
This PR is WIP to introduce the representation of points on the Grassmann manifold using projectors of rank k, cf. #417. 

A general question:
* [x] Should the default (Stiefel points) get an own type. too? For Hyperbolic we did this and that that point acts kind of transparently. What do you think?
* [x] If the first point is a yes, what would we call it?
* [x] I copied quite som e functions for broadcast and tangent vectors from Hyperbolic: We could collect (single field / `.value`) types, that are like the three Hyperbolic ones or the new Grassmann an do this definition in one central place at the end for all collected points. What do you think?

_edit:_ this PR is nearly finished, starting a Quotient manifold model as well (cf. / solves #205 ), though not yet that advanced, it does provide again the implicit and explicit quotient manifold things.

Still todo 
* [x] write tests.


mateuszbaran:
> * Should the default (Stiefel points) get an own type. too? For Hyperbolic we did this and that that point acts kind of transparently. What do you think?

I don't think this needs to be done but I wouldn't mind having it as well.

> [ ] I copied quite som e functions for broadcast and tangent vectors from Hyperbolic: We could collect (single field / `.value`) types, that are like the three Hyperbolic ones or the new Grassmann an do this definition in one central place at the end for all collected points. What do you think?

I think a macro that generates that code would be an even better idea.

kellertuer:
I currently plan to do the first with the type – just for completeness (and since we did that for Hyperbolic as well).

kellertuer:
So the first test of proper point/vector types with hyperbolic is fine and much nicer now. Next step is to get the Projection points up and running.

src/manifolds/GrassmannProjector.jl
```diff
@@ -0,0 +1,91 @@
+@doc raw"""
+    ProjectorPoint <: AbstractManifoldPoint
+
+A type to represent points on a manifold that are orthogonal projectors.
```
mateuszbaran, line 4:
```suggestion
A type to represent points on a manifold [`Grassmann`](@ref) that are orthogonal projectors.
```
Is this only for the Grassmann manifold?

src/manifolds/GrassmannProjector.jl
```diff
@@ -0,0 +1,91 @@
...
+must hold, where the `kwargs` can be used to check both for symmetrix of ``X```
+and this equality up to a certain tolerance.
+"""
+function check_vector(
+    M::Grassmann{n,k,𝔽},
+    p::ProjectorPoint,
+    X::ProjectorTVector;
+    kwargs...,
+) where {n,k,𝔽}
+    if !isapprox(norm(X.value - X.value'), 0.0; kwargs...)
```
mateuszbaran, line 75:
```suggestion
    if !isapprox(X.value, X.value'; kwargs...)
```
Maybe that would be a bit more robust?

src/manifolds/GrassmannStiefel.jl
```diff
@@ -0,0 +1,405 @@
...
+a `n`-by-`k` matrix of unitary column vectors and of correct `eltype` with respect to `𝔽`.
+"""
+function check_point(M::Grassmann{n,k,𝔽}, p; kwargs...) where {n,k,𝔽}
+    mpv = invoke(check_point, Tuple{supertype(typeof(M)),typeof(p)}, M, p; kwargs...)
+    mpv === nothing || return mpv
+    c = p' * p
+    if !isapprox(c, one(c); kwargs...)
+        return DomainError(
+            norm(c - one(c)),
+            "The point $(p) does not lie on $(M), because x'x is not the unit matrix.",
```
mateuszbaran, line 55:
```suggestion
            "The point $(p) does not lie on $(M), because p'p is not the unit matrix.",
```

src/manifolds/GrassmannProjector.jl
```diff
@@ -0,0 +1,91 @@
...
+@manifold_vector_forwards ProjectorTVector value
+
+@doc raw"""
+    check_point(::Grassmann{n,k}, p::ProjectorPoint; kwargs...)
+
+Check whether a orthogonal projector is a point from the [`Grassmann`](@ref)`(n,k)` manifold,
+i.e. the [`ProjectorPoint`](@ref) ``p ∈ \mathbb R^{n×n}`` has to fulfill
+``p^{\mathrm{T}} = p``, ``p^2=p``, and ``\operatorname{rank} p = k`.
+"""
+function check_point(M::Grassmann{n,k,𝔽}, p::ProjectorPoint; kwargs...) where {n,k,𝔽}
```
mateuszbaran, line 29: Docstring suggests it only works for the real case but method signature allows complex Grassmann manifold.

kellertuer: You‘re right, I was not careful enough there.

kellertuer:
Ops. this seems to be a little more refactoring than I thought

kellertuer:
Since I screwed up the merge, I started from scratch and even noticed that the Stiefel representation should just be called `StiefelPoint`– and while not useful can now also be used with Stiefel.

kellertuer:
(oh it is not ready for review, I just switched from draft to continue, since with the new interfaces all this will now get a little easier).

kellertuer:
as we discussed in #417 a next step is to consider horizontal lifts, for example to compute the inner product in projector representation. I would propose to first just use a function `horizontal_lift(M::Grassmann, p::ProjectorPoint)` – but for that I have to first see how to best implement this.

mateuszbaran:
> I would propose to first just use a function `horizontal_lift(M::Grassmann, p::ProjectorPoint)` – but for that I have to first see how to best implement this.

I think you missed a `ProjectorTVector` there?

kellertuer:
> > I would propose to first just use a function `horizontal_lift(M::Grassmann, p::ProjectorPoint)` – but for that I have to first see how to best implement this.
> 
> I think you missed a `ProjectorTVector` there?

ah, eh, yes. And I also first got stuck in ambiguities and our macros.

kellertuer:
I started also sketching #205 in here since the projector point model uses quotient structures a lot. This way the naming of new functions is already in line with quotient manifold decorators.

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,88 @@
...
+```math
+    \mathcal M = \mathcal N / ∼ = \bigl\{ [p] : p ∈ \mathcal N \bigr\},
+```
+where ``[p] ≔ \{ q ∈ \matcal N : q ∼ p\}`` denotes the equivalence class containing ``p``.
+For more details see Subsection 3.4.1[^AbsilMahonySepulchre2008].
+
+This manifold type models an explicit quotient structure.
+This should be done if either the default implementation of ``\mathcal M``
+uses another representation different from the quotient structure or if
+it provides a (default) quotient structure that is different from the one tintroduced here.
```
mateuszbaran, line 33:
```suggestion
it provides a (default) quotient structure that is different from the one introduced here.
```

mateuszbaran: By the way, I don't really understand this paragraph right now but it should be more clear later when we have some examples :).

kellertuer: It means the same as for the metric or embedded manifold, we again have implicit cases (just the Trait) or explicit ones. I am not 100% sure whether we maybe even can omit the ProjectionPoints this way (since Stiefel is the default for Grassmann, then the QuotientManifold of Stiefel could again be just build on matrices),

kellertuer:
Besides testing I am missing one thing still, also because I could not figure out how these are computed in 

https://arxiv.org/abs/2011.13699

That is mainly `horizontal_lift` which is similar to an embed (i.e. a tangent vector on the quotient manifold as input yields a tangent vector in the total space). I struggle (a) how to properly design the interface and (b) the formulae in the linked paper for me always have a “there exists a B such that” where I struggle to find the details how to implement this.

Besides that the only thing I currently do not like that much is that `canonical_project(M,p)` currently is not aware of how points are presented on the manifold `M`. To provide some detail: It projects a point `p` (from the total space `N`) onto `M`.
As an example for Grassmann it either

a) is the identity if points on Grassmann are represented by matrices (or analogously `StiefelPoint`)
b) are changed into a projector if represented by `ProjectorPoint`

Now the problem is that `canonical_project` does not know whether we want a or b (`canonical_project!` does). It would be nice for the allocating case to maybe be able to provide the representation type on `M`?

kellertuer:
Maybe one final thing to discuss – I am currently looking for a name to maybe also provide a function that returns the factor maybe? To be clear since

Gr(n,k) = St(n,k) / O(k)

I am looking for a function name that returns O(k) (if that itself is a manifold).

kellertuer:
I was surprised that the tests failed on PowerManifolds, I first thought that merging master might fix it but it seems to persist. Its lines that I did not touch in this PR so I am unsure what is happening there.

mateuszbaran:
> a) is the identity if points on Grassmann are represented by matrices (or analogously `StiefelPoint`) b) are changed into a projector if represented by `ProjectorPoint`
> 
> Now the problem is that `canonical_project` does not know whether we want a or b (`canonical_project!` does). It would be nice for the allocating case to maybe be able to provide the representation type on `M`?

Interesting point. We have a similar issue in other places, for example what should `rand` return. If we want Grassmann to do both then it needs a new type parameter that tells which one is expected.

> Maybe one final thing to discuss – I am currently looking for a name to maybe also provide a function that returns the factor maybe? To be clear since
> 
> Gr(n,k) = St(n,k) / O(k)
> 
> I am looking for a function name that returns O(k) (if that itself is a manifold).

That's a bit of an issue because we don't have O(k) yet :smile: (ref #345 ). Anyway, I think quotients by Lie groups action (like this one) should return the whole action object (something similar to `RotationAction`).

> was surprised that the tests failed on PowerManifolds, I first thought that merging master might fix it but it seems to persist. Its lines that I did not touch in this PR so I am unsure what is happening there.

I will check that.

kellertuer:
> Interesting point. We have a similar issue in other places, for example what should `rand` return. If we want Grassmann to do both then it needs a new type parameter that tells which one is expected.

The one idea I head was to either to a keyword argument or a last optional argument `representation_type` to distinguish both?

> > Maybe one final thing to discuss – I am currently looking for a name to maybe also provide a function that returns the factor maybe? To be clear since
> > Gr(n,k) = St(n,k) / O(k)
> > I am looking for a function name that returns O(k) (if that itself is a manifold).
> 
> That's a bit of an issue because we don't have O(k) yet 😄 (ref #345 ). Anyway, I think quotients by Lie groups action (like this one) should return the whole action object (something similar to `RotationAction`).

Pfft! Details 😄 actually redoing #345 is my next working package (currently thinking whether one can unify rotations, orthogonal and unitary manifolds somehow). For now I am mainly looking for a name. Instead of just the action I would prefer to return the whole Lie group (which includes its action).

mateuszbaran:
> > was surprised that the tests failed on PowerManifolds, I first thought that merging master might fix it but it seems to persist. Its lines that I did not touch in this PR so I am unsure what is happening there.
> 
> I will check that.

#498 will fix this issue, I should be able to finish it tomorrow.



> I have an idea for a name, either
> `get_orbit` or `get_orbit_action` since the group/action? (GL_n is another example by the way).

I like `get_orbit_action` :+1: .

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,165 @@
...
+a [`QuotientManifold`](@ref).
+"""
+get_total_space(::AbstractManifold)
+get_total_space(M::QuotientManifold) = M.total_space
+
+@doc raw"""
+    get_orbit_action(M::AbstractDecoratorManifold)
+
+Return the group action that generates the orbit of an equivalence class, that is for the case
+that
```
mateuszbaran, line 134:
```suggestion
Return the group action that generates the orbit of an equivalence class of the quotient manifold `M`,
that is for the case that
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,165 @@
...
+\mathcal M = \mathcal N / \mathcal O,
+```
+
+where ``\mathcal O`` is a Lie group with its group action generating the orbit.
+"""
+get_orbit_action(::AbstractManifold)
+
+@doc raw"""
+    horizontal_lift(N, q, X)
+    horizontal_lift(QuotientManifold{M,N}, p, X)
```
mateuszbaran, line 146: Why isn't this just `horizontal_lift(N::AbstractManifold, q, X)`?

kellertuer: The general problem here is that q is from N but X is from T_pM (p=pi(q)). So as the first line indicates it is what you ask for, but then we again have a lot of implicit assumptions, i.e. that M and pi (canonical project) are clear from context. If that is not the case M should be enough to know additionally (since then we know domain and range of pi for example). So if that is not the case we need the whole `QuotientManifold{M,N}` instead of just `N`.

But as I said it _is_ just what you say is we have a lot a lot of implicit assumptions.

mateuszbaran:
```suggestion
    horizontal_lift(N::AbstractManifold, q, X)
    horizontal_lift(::QuotientManifold{𝔽,MT<:AbstractManifold{𝔽},NT<:AbstractManifold}, p, X) where {𝔽}
```
OK, so this I think would be more clear?

kellertuer: Sure, that is more detailed, you‘re right.

kellertuer:
Hm, I am a little stuck on why the manifold fallbacks won't work, I would have expected

```
        M = Grassmann(3,2)
        M2 = Stiefel(3,2)
        p = StiefelPoint([1.0 0.0; 0.0 1.0; 0.0 0.0])
        is_point(M,p,true)
       is_point(M2, p, true)
```

to work since we have the manifold fallbacks defined for both – but both report that `StiefelPoint` is not a point on the Euclidean matrix manifold. I am not yet sure why because I do not understand why the decorator trait “acts” first here, I would have supposed it would not (but first call `is_point(M, p.value, true)`.

kellertuer:
But I ran into a similar problem with `ProjectorPoint` which is not a default, but still the Embedded manifold trait kicks in and it is surely not by itself a point on Euclidean. So we somehow have to check in active traits whether one of the arguments is a ProjectorPoint or something?

I am not sure how to get that working, just

```
M = Grassmann(3,2)
p = ProjectorPoint([1.0 0.0 0.0; 0.0 1.0 0.0; 0.0 0.0 0.0])
ispoint(M,p,true)
```

because I also do not want to reimplement `is_point` (`check_point` already is)

kellertuer:
...and if one understand again why we have `embed` - the solution is 4 lines of code ;)

kellertuer:
Hm, now tests are a little strange, they run fine locally but one old test (from before this branch) now fails on Julia 1.6 and I have no clue why.

kellertuer:
Locally I had the error on first run but on no other run after that (even with restarting Julia). Well, let‘s see. There is also still some coverage missing, but I first wanted to see how far I got.

src/manifolds/GrassmannProjector.jl
```diff
@@ -0,0 +1,267 @@
...
+```
+where ``\operatorname{Exp}`` denotes the matrix exponential and ``[A,B] = AB-BA`` denotes the matrix commutator.
+
+For details, see Proposition 3.2 in [^BendokatZimmermannAbsil2020].
+"""
+exp(M::Grassmann, p::ProjectorPoint, X::ProjectorTVector)
+
+function exp!(::Grassmann, q::ProjectorPoint, p::ProjectorPoint, X::ProjectorTVector)
+    xppx = X.value * p.value - p.value * X.value
+    q.value .= exp(xppx) * p.value * exp(-xppx)
```
mateuszbaran, line 209:
```suggestion
    exp_xppx = exp(xppx)
    q.value .= exp_xppx * p.value / exp_xppx
```
since `exp(-xppx) = inv(exp(xppx))`.

src/manifolds/GrassmannProjector.jl
```diff
@@ -0,0 +1,267 @@
...
+
+function parallel_transport_direction!(
+    ::Grassmann,
+    Y::ProjectorTVector,
+    p::ProjectorPoint,
+    X::ProjectorTVector,
+    d::ProjectorTVector,
+)
+    dppd = d.value * p.value - p.value * d.value
+    return Y.value .= exp(dppd) * X.value * exp(-dppd)
```
mateuszbaran, line 263:
```suggestion
    exp_dppd = exp(dppd)
    Y.value .= exp_dppd * X.value / exp_dppd
    return Y
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
+
+"""
+    IsQuotientManifold <: AbstractTrait
+
+Specify that a certain decorated Manifold is a quotient manifold in the sence that it provides
```
mateuszbaran, line 5:
```suggestion
Specify that a certain decorated manifold is a quotient manifold in the sense that it provides
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
...
+    manifold::MT
+    total_space::NT
+end
+
+@inline active_traits(f, ::QuotientManifold, ::Any...) = merge_traits(IsQuotientManifold())
+
+@doc raw"""
+    canonical_project(M, p)
+
+compute the canonical projection ``π`` on a manifold ``\mathcal M`` that
```
mateuszbaran, line 64:
```suggestion
Compute the canonical projection ``π`` on a manifold ``\mathcal M`` that
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
...
+"""
+function canonical_project(M::AbstractManifold, p)
+    q = allocate_result(M, canonical_project, p)
+    return canonical_project!(M, q, p)
+end
+
+@doc raw"""
+    canonical_project!(M, q, p)
+
+compute the canonical projection ``π`` on a manifold ``\mathcal M`` that
```
mateuszbaran, line 84:
```suggestion
Compute the canonical projection ``π`` on a manifold ``\mathcal M`` that
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
...
+See [`canonical_project`](@ref) for more details.
+"""
+canonical_project!(M::AbstractManifold, q, p)
+
+decorated_manifold(M::QuotientManifold) = M.manifold
+
+@doc raw"""
+    differential_canonical_project(M, p, X)
+
+compute the differential of the canonical projection ``π`` on a manifold ``\mathcal M`` that
```
mateuszbaran, line 96:
```suggestion
Compute the differential of the canonical projection ``π`` on a manifold ``\mathcal M`` that
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
...
+"""
+function differential_canonical_project(M::AbstractManifold, p, X)
+    q = allocate_result(M, differential_canonical_project, p, X)
+    return differential_canonical_project!(M, q, p, X)
+end
+
+@doc raw"""
+    differential_canonical_project!(M, Y, p, X)
+
+compute the differential of the canonical projection ``π`` on a manifold ``\mathcal M`` that
```
mateuszbaran, line 116:
```suggestion
Compute the differential of the canonical projection ``π`` on a manifold ``\mathcal M`` that
```

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
...
+a [`QuotientManifold`](@ref).
+"""
+get_total_space(::AbstractManifold)
+get_total_space(M::QuotientManifold) = M.total_space
+
+@doc raw"""
+    get_orbit_action(M::AbstractDecoratorManifold)
+
+Return the group action that generates the orbit of an equivalence class of the quotient manifold `M`,
+that is for the case that
```
mateuszbaran, line 134:
```suggestion
Return the group action that generates the orbit of an equivalence class of the quotient manifold `M`
for which equivalence classes are orbits of an action of a Lie group. For the case that
```
Just a little more details, since it's possible to define a quotient manifold without an action.

src/manifolds/QuotientManifold.jl
```diff
@@ -0,0 +1,169 @@
...
+get_orbit_action(::AbstractManifold)
+
+@doc raw"""
+    horizontal_lift(N::AbstractManifold, q, X)
+    horizontal_lift(::QuotientManifold{𝔽,MT<:AbstractManifold{𝔽},NT<:AbstractManifold}, p, X) where {𝔽}
+
+Given a point `q` such that ``p=π(q)`` is a point on a quotient manifold `M`
+(implicitly given for the first case)
+and a tangent vector `X` this method computes a `Y` on the horizontal space of ``T_q\mathcal N``,
+i.e. that is orthogonal to the kernel of ``Dπ(q)``.
```
mateuszbaran, line 151:
```suggestion
Given a point `q` such that ``p=π(q)`` is a point on a quotient manifold `M`
(implicitly given for the first case) and a tangent vector `X` this method
computes a tangent vector `Y` on the horizontal space of ``T_q\mathcal N``,
i.e. the subspace that is orthogonal to the kernel of ``Dπ(q)``.
```

src/manifolds/SymplecticStiefel.jl
```diff
@@ -67,10 +67,10 @@ It is assumed that the point ``p`` is on ``\operatorname{Sp}(2n, 2n)``.
 """
 function canonical_projection(M::SymplecticStiefel{n,k}, p_Sp) where {n,k}
```
mateuszbaran, line 68:
```suggestion
function canonical_project(M::SymplecticStiefel{n,k}, p_Sp) where {n,k}
```
This name change is technically breaking but I doubt many people use that.

mateuszbaran (commented):
This looks really nice :+1: . These are mostly just documentation improvements.

src/manifolds/Grassmann.jl
```diff
@@ -248,160 +136,46 @@ function default_estimation_method(::Grassmann, ::typeof(mean))
...
+to a projector representation of said subspace, i.e. compute the [`canonical_project!`](@ref)
+for
 
-@doc raw"""
-    zero_vector(M::Grassmann, p)
-
-Return the zero tangent vector from the tangent space at `p` on the [`Grassmann`](@ref) `M`,
-which is given by a zero matrix the same size as `p`.
+```math
+  π^{\mathrm{SG}}(p) = pp^{\mathrm{T)}}.
```
mateuszbaran, line 178:
```suggestion
  π^{\mathrm{SG}}(p) = pp^{\mathrm{T}}.
```

src/manifolds/Grassmann.jl
```diff
@@ -248,160 +136,46 @@ function default_estimation_method(::Grassmann, ::typeof(mean))
...
+for
 
-The implementation is based on Section 2.5.1 in [^Chikuse2003];
-see also Theorem 2.2.2(iii) in [^Chikuse2003].
-
-[^Chikuse2003]:
-    > Y. Chikuse: "Statistics on Special Manifolds", Springer New York, 2003,
-    > doi: [10.1007/978-0-387-21540-2](https://doi.org/10.1007/978-0-387-21540-2).
+```math
+  π^{\mathrm{SG}}(p) = pp^{\mathrm{T)}}.
```
mateuszbaran, line 166:
```suggestion
  π^{\mathrm{SG}}(p) = pp^{\mathrm{T)}.
```

src/manifolds/SymplecticStiefel.jl
kellertuer: yes it is, but we only introduced it here fir the first time and I do not thunk someone used it until now (since we introduced this in January.

mateuszbaran (approved):

kellertuer:
I am getting a little tired of Julia 1.6, this is try number 5 or 6 while on Julia 1.7 everything is fine already...

PR #445, opened 2021-11-23
support specialfunctions 2.x

stevengj:
The only breaking change in SpecialFunctions 2.0 [was deleting](https://github.com/JuliaMath/SpecialFunctions.jl/pull/297) the `factorial(x::Real) = gamma(x+1)` function, and since you don't use `factorial` for non-integer arguments this shouldn't affect you.


mateuszbaran:
Thanks! We already have a CompatHelper PR open: https://github.com/JuliaManifolds/Manifolds.jl/pull/444 . Also, I think I'll wait until our dependencies bump compatibility (most notably Distributions.jl).

stevengj:
Shouldn't be necessary to wait for dependencies — if a dependency requires SpecialFunctions 1, it will hold back SpecialFunctions but remain compatible with this package, since you are still compatible with SpecialFunctions 1

PR #446, opened 2022-01-21
fix two typos in readme

fkastner:


kellertuer:
Thanks for noticing and fixing Felix!

Issue #418, opened 2021-08-31
The Real Symplectic Stiefel Manifold


kellertuer:
Maybe it would be nice to also include some symplectic manifolds, see for example https://arxiv.org/abs/2108.12447 for some recent developments.

They also include some gradient conversion (Euclidean to X) we started (but never finished) in #44 to discuss.

This issue is just to keep track of the Symplectic ideas – might take that as a next project (once I finished my ideas in ManifoldsBase about maybe rewriting the Group Action stuff).

mateuszbaran:
Good idea, though we should probably start with something more simple like the [symplectic group](https://en.wikipedia.org/wiki/Symplectic_group).

I'll go back and look again at that group action stuff probably tomorrow.

kellertuer:
Sure. My idea was to start with the easy ones and slowly work up to the manifold in the paper.

PR #447, opened 2022-01-21
Symplectic Manifolds

johannvk:
Implementation of two manifolds, the symplectic and symplectic Stiefel manifolds, as mentioned in issue #418.  
Based on supervision by Ronny Bergmann during my specialization project at NTNU (Norwegian University of Science and Technology) autumn of 2021.  

Still a work-in-progress when it comes to finalizing the documentation for the symplectic Stiefel manifold, as well as achieving good test coverage for both manifolds.


kellertuer:
For completeness - this addresses #418.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,854 @@
...
+    p_star_X = symplectic_inverse_times(M, p, X)
+
+    divisor = lu(2 * I - p_star_X)
+    q .= p * (divisor \ add_scaled_I!(p_star_X, 2.0))
+    return q
+end
+
+Base.show(io::IO, ::Symplectic{n,𝔽}) where {n,𝔽} = print(io, "Symplectic{$(2n), $(𝔽)}()")
+
+function symplectic_inverse_times(M::Symplectic{n}, p, q) where {n}
```
kellertuer, line 649: This should be documented, I think

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+    @inbounds for i in 1:k, j in 1:n
+        q[i + k, j + n] = p[j, i]
+    end
+    return q
+end
+
+@doc raw"""
+    make_metric_compatible!(::SymplecticStiefel, Y, ::EuclideanMetric, p, X)
+
+TODO: Complete.
```
kellertuer, line 243: can you be more complete here – is it maybe the change<_metric function? Without the details I am not sure what this function does, but we should try to find a better name.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+end
+
+function canonical_projection!(::SymplecticStiefel{n,k}, p, p_Sp) where {n,k}
+    p[:, (1:k)] .= p_Sp[:, (1:k)]
+    p[:, ((k + 1):(2k))] .= p_Sp[:, ((n + 1):(n + k))]
+    return p
+end
+
+# compute p^+q (which is 2kx2k) in place of A
+function symplectic_inverse_times(M::SymplecticStiefel{n,k}, p, q) where {n,k}
```
kellertuer, line 325: This should have a documentation (and be exported maybe)

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+    mul!(A3, p3', q1, -1, 1)    # A3 -= p3'q1
+    mul!(A4, p1', q4)           # A4  = p1'q4
+    mul!(A4, p3', q2, -1, 1)    #A4  -= p3'q2
+    return A
+end
+
+@doc raw"""
+    Compute the exponential mapping from eq. 3.19 of Bendokat-Zimmermann.
+"""
+function exp!(M::SymplecticStiefel{n,k}, q, p, X) where {n,k}
```
kellertuer, line 358: Please add the complete formula here.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+5.2 of [^Bendokat2021].
+
+Let ``A = p^+X`` and ``H = X - pA``. Then an equivalent expression for the Cayley
+retraction defined pointwise above is
+````math
+    \mathcal{R}_p(X) = -p + (H + 2p)(H^+H/4 - A/2 + I_{2k})^{-1}.
+````
+It is this expression we compute inplace of `q`.
+"""
+function retract!(M::SymplecticStiefel{n,k}, q, p, X, ::CayleyRetraction) where {n,k}
```
kellertuer, line 437: We usually document the `retract` variant (which allocates memory and calls this version).

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+\mathcal{L}_p^{\operatorname{Sp}}(q) = 2p\bigl(V - U\bigr) + 2\bigl((p + q)U - p\bigr)
+                                        ∈ T_p\operatorname{Sp}(2n).
+````
+
+[^Bendokat2021]:
+    > Bendokat, Thomas and Zimmermann, Ralf:
+	> The real symplectic Stiefel and Grassmann manifolds: metrics, geodesics and applications
+	> arXiv preprint arXiv:2108.12447, 2021 (https://arxiv.org/abs/2108.12447)
+"""
+function inverse_retract!(M::SymplecticStiefel, X, p, q, ::CayleyInverseRetraction)
```
kellertuer, line 490: We usually document the `inverse_retract` variant.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+    U_inv = lu!(add_scaled_I!(symplectic_inverse_times(M, p, q), 1))
+    V_inv = lu!(add_scaled_I!(symplectic_inverse_times(M, q, p), 1))
+
+    X .= 2 .* ((p / V_inv .- p / U_inv) .+ ((p + q) / U_inv) .- p)
+    return X
+end
+
+function gradient(M::SymplecticStiefel, f, p, backend::RiemannianProjectionBackend)
```
kellertuer, line 498: This should get a documentation.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,554 @@
...
+
+````math
+    \operatorname{min}_{X \in \mathbb{R}^{2n \times 2k}} \frac{1}{2}||X - A||^2, \quad
+    \text{s.t.}\;
+    h(X) = X^T Q p + p^T Q X = 0,
+````
+where ``h : \mathbb{R}^{2n \times 2k} \rightarrow \operatorname{skew}(2k)`` defines
+the restriction of ``X`` onto the tangent space ``T_p\operatorname{SpSt}(2n, 2k)``.
+"""
+function project!(::Union{SymplecticStiefel,Symplectic}, Y, p, A)
```
kellertuer, line 539: We usually document the `project` function.

kellertuer (commented):
This is on a very good way. Here's just a few comments of things still to do :)

src/manifolds/SymplecticStiefel.jl
johannvk: I will expand on the documentation and make it more clear what we're doing here, but it's basically the 'change_representer' function, without the correct Tangent vector space range. I am a bit unsure as to what would be a better name, and I'm considering maybe taking this function out as it doesn't quite do what we would like from a 'change_representer'-function, and might therefore not be so useful.

kellertuer: Ok, so we also do not use this function anywhere? Then it might really be an idea to remove it. I am still not yet sure what it actually does ;)

johannvk: He he yeah that's true. It's a leftover from when I was working on the Euclidean projection-gradient conversion method. 
I managed to find a working 'change_representer'-function on the symplectic manifold, with the convex combination of two 
'metric compatible' functions like this one. (c1 and c2 from the specialization report.) On symplectic Stiefel though, I've not managed to find another Metric compatibility function to attempt the same 'convex combination' trick with.

kellertuer:
I checked test coverage for short – the 4 lines seem to be false-negatives and they should be covered, we can try to trick that maybe (I think one can change elements with one line)

I think one can write
```
tmp = A[i, j + n]
A[i, j + n] = -A[j, i + n]
A[j, i + n] = -tmp
```

also as
```
A[i, j + n], A[j, i+n] = -A[j, i + n], -A[i, j + n]
```
(but please check locally first that the tests run fine then)

src/manifolds/SymplecticStiefel.jl
johannvk: The complete formula has been added to documentation now.

src/manifolds/SymplecticStiefel.jl
johannvk: Done.

src/manifolds/SymplecticStiefel.jl
johannvk: Done.

src/manifolds/SymplecticStiefel.jl
johannvk: Done.

src/manifolds/SymplecticStiefel.jl
kellertuer: Ah, and this documented the mutating exp!` as well, it should document `exp` instead.

kellertuer: Ah now it is ok :)

kellertuer:
Yes, you could try the one line I proposed instead, or I will when I find time.

johannvk:
I tested the change you proposed for the three relevant lines in `inv!(::Symplectic, ...)`, and they lead to identical 
results with zero allocations and perhaps simpler code. So I'm pushing the changes :+1:

src/manifolds/SymplecticStiefel.jl
johannvk: Added draft of documentation.

src/manifolds/SymplecticStiefel.jl
johannvk: I removed the `make_metric_compatible(::Symplectic, ...)` function, along with the relevant tests.

src/manifolds/SymplecticStiefel.jl
johannvk: I've drafted documentation for this function now, and also exported `symplectic_inverse_times(::SymplecticStiefel, p, q)`
and `symplectic_inverse_times!(::SymplecticStiefel, A, p, q)`.

kellertuer:
Perfect – you are just one line away from the 100% code coverage, that would be perfect. And its also just to test that an error is thrown if you alias the gradients (not sure that is the right fix, but it is the current, so you should make sure that aliasing it _does_ yield the error (look for `@test_throws` examples in our tests :)).

johannvk:
I added a `@test_throws` test like suggested, and I the test coverage should be 100% now :)
Additionally I improved the documentation for two of the functions in "Symplectic.jl", which were incomplete.

I think the documentation and tests should be pretty good now, but I'm of course happy for suggestions for improvements :)

kellertuer:
Code coverage looks great! I will check the docs soon then.

docs/src/manifolds/symplectic.md
```diff
@@ -0,0 +1,25 @@
+# Symplectic 
+
+The [`Symplectic`](@ref) manifold, denoted $\operatorname{Sp}(2n)$, represents transformations into symplectic subspaces which keep the
```
mateuszbaran, line 3:
```suggestion
The [`Symplectic`](@ref) manifold, denoted $\operatorname{Sp}(2n, \mathbb{F})$, is a submanifold of $\mathbb{F}^{2n \times 2n}$ that represents transformations into symplectic subspaces which keep the
```
I think it's better to mention the embedding earlier and possibly leave it open for the complex symplectic group even if it isn't implemented yet.

docs/src/manifolds/symplecticstiefel.md
```diff
@@ -0,0 +1,27 @@
+# Symplectic Stiefel
+
+The [`SymplecticStiefel`](@ref) manifold, denoted $\operatorname{SpSt}(2n, 2k)$,
```
mateuszbaran, line 3: Could you maybe reference the symplectic group somewhere here?

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
+@doc raw"""
+    Symplectic{n, ℝ} <: AbstractEmbeddedManifold{ℝ, DefaultIsometricEmbeddingType}
```
mateuszbaran, line 2:
```suggestion
    Symplectic{n, 𝔽} <: AbstractEmbeddedManifold{𝔽, DefaultIsometricEmbeddingType}
```
Corrected to match the definition

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+and indentity matrix in ``ℝ^{n \times n}`` respectively.
+
+Internally the dimensionality of the structure is stored as half of the even dimension
+supplied to the constructor, `Symplectic(2n) -> Symplectic{n}()`,
+as most computations with points on the Real Symplectic manifold takes advantage of the
+natural block structure
+of a matrix ``A ∈ ℝ^{2n × 2n}`` where we consider it as consisting of four
+smaller matrices in ``ℝ^{n × n}``.
+"""
+struct Symplectic{n,𝔽} <: AbstractEmbeddedManifold{𝔽,DefaultIsometricEmbeddingType} end
```
mateuszbaran, line 27: With very little work we could extend it to the symplectic Lie group, what do you think @kellertuer ? This could be done similarly to `GeneralLinear`.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+Q_{2n} =
+\begin{bmatrix}
+0_n & I_n \\
+ -I_n & 0_n
+\end{bmatrix} \quad \in ℝ^{2n \times 2n},
+````
+such that
+````math
+\omega_{2n}(x, y) = x^TQ_{2n}y, \quad x, y \in ℝ^{2n}.
+````
```
mateuszbaran, line 74: Shouldn't λ appear somewhere here?

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+"""
+function change_representer(
+    MetMan::MetricManifold{𝔽,Euclidean{Tuple{m,n},𝔽},ExtendedSymplecticMetric},
+    EucMet::EuclideanMetric,
+    p,
+    X,
+) where {𝔽,m,n}
+    return change_representer!(MetMan, similar(X), EucMet, p, X)
+end
```
mateuszbaran, line 160: Doesn't the generic fallback: https://github.com/JuliaManifolds/Manifolds.jl/blob/f3e1ff5fe567d1df9dac327de22ae913dc8a20f1/src/manifolds/MetricManifold.jl#L166 handle this? We shouldn't have to have this method written.

mateuszbaran: On the other hand, it would make some sense to have `p * p' * X` as the body of the method here, in case we want to work on immutable representations.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+    return Y
+end
+
+@doc raw"""
+    change_representer(MetMan::MetricManifold{𝔽, Euclidean{Tuple{m, n}, 𝔽}, ExtendedSymplecticMetric},
+                       EucMet::EuclideanMetric, p, X)
+    change_representer!(MetMan::MetricManifold{𝔽, Euclidean{Tuple{m, n}, 𝔽}, ExtendedSymplecticMetric},
+                        Y, EucMet::EuclideanMetric, p, X)
+
+Change the representation of a matrix ``χ ∈ \mathbb{R}^{2n \times 2n}``
```
mateuszbaran, line 133: Having the argument named `X` and the matrix `χ` may be a bit misleading, it's very easy to confuse these two symbols.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+    g_p(\text{grad}f(p), X) = \text{D}f(p) \quad\forall\; X \in T_p\operatorname{Sp}(2n).
+````
+
+The default behaviour is to first change the representation of the Euclidean gradient from
+the Euclidean metric to the [`RealSymplecticMetric`](@ref) at ``p``, and then we projecting
+the result onto the correct tangent tangent space ``T_p\operatorname{Sp}(2n, ℝ)``
+w.r.t the Riemannian metric ``g_p`` extended to the entire embedding space.
+
+# Arguments:
+- `extended_metric = true`: If `true`, compute the gradient ``\text{grad}f(p)`` by
```
mateuszbaran, line 361: This should probably be generalized to our generic `gradient` but I need to think about it more.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+@doc raw"""
+    inner(::Symplectic{n, ℝ}, p, X, Y)
+
+Compute the canonical Riemannian inner product [`RealSymplecticMetric`](@ref)
+````math
+    g_p(X, Y) = \operatorname{tr}((p^{-1}X)^T (p^{-1}Y))
+````
+between the two tangent vectors ``X, Y \in T_p\operatorname{Sp}(2n)``.
+"""
+function inner(M::Symplectic{n,ℝ}, p, X, Y)::eltype(p) where {n}
```
mateuszbaran, line 421:
```suggestion
function inner(M::Symplectic{n,ℝ}, p, X, Y) where {n}
```
We generally don't use return type annotations because they don't do much.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+    p_star = inv(M, p)
+    return tr((p_star * X)' * (p_star * Y))
```
mateuszbaran, line 423:
```suggestion
    return dot((p_star * X), (p_star * Y))
```
I didn't check carefully but usually dot(a, b) is the same as tr(a'*b) and the dot version is faster.

mateuszbaran (commented):
Thank you, this is a nice new functionality for Manifolds.jl I haven't looked at everything yet but here are some comments.

src/manifolds/Symplectic.jl
kellertuer: I think it might be a little complicated right now and much more doable once the new interface is there.

mateuszbaran: Yes, this should be easier in the new interface.

src/Manifolds.jl
```diff
@@ -431,7 +434,10 @@ export Euclidean,
     SymmetricPositiveDefinite,
     SymmetricPositiveSemidefiniteFixedRank,
     Torus,
-    Tucker
+    Tucker,
+    Symplectic,
+    SymplecticStiefel,
+    SymplecticMatrix
```
mateuszbaran, line 440: These should be ordered alphabetically (i.e. be inserted above Torus).

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+    Q = SymplecticMatrix(p, X)
+
+    pT_QT_X = p' * Q' * X
+    skew_pT_QT_X = (1 / 2) .* (pT_QT_X .- pT_QT_X')
+
+    Y[:, :] = p * Q * skew_pT_QT_X
+    return Y
+end
+
+function Base.rand(M::Symplectic{n}, hamiltonian_norm::Number=1.0) where {n}
```
mateuszbaran, line 608:
```suggestion
function Base.rand(M::Symplectic{n}; hamiltonian_norm::Number=1.0) where {n}
```
Extending `Base.rand` correctly is quite tricky. `rand(some_manifold)` is not an API we currently have but it should be fine, although any further options should be introduced as keyword arguments.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+end
+
+function Base.rand(::Symplectic{n}, p::AbstractMatrix) where {n}
+    # Generate random symmetric matrix:
+    S = rand(2n, 2n) .- 1 / 2
+    S .= (1 / 2) .* (S + S')
+    Q = SymplecticMatrix(p)
+    lmul!(Q, S)
+    return p * S
+end
```
mateuszbaran, line 622: This API is confusing, maybe let's make this a separate function instead of adding a method to `Base.rand`? For example `random_symmetric_matrix`.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+    mul!(A2, p2', q4, -1, 1) #A2 -= p2'q4
+    mul!(A3, p1', q3) #A3 = p1'q3
+    mul!(A3, p3', q1, -1, 1) # A3 -= p3'q1
+    mul!(A4, p1', q4) # A4 = p1'q4
+    mul!(A4, p3', q2, -1, 1) #A4 -= p3'q2
+    return A
+end
+
+ndims(Q::SymplecticMatrix) = 2
+copy(Q::SymplecticMatrix) = SymplecticMatrix(Q.λ)
```
mateuszbaran, line 715:
```suggestion
copy(Q::SymplecticMatrix) = SymplecticMatrix(copy(Q.λ))
```
Most likely makes no difference in nearly any situation but this seems more appropriate.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+(Base.:+)(Q1::SymplecticMatrix, Q2::SymplecticMatrix) = SymplecticMatrix(Q1.λ + Q2.λ)
+(Base.:-)(Q1::SymplecticMatrix, Q2::SymplecticMatrix) = SymplecticMatrix(Q1.λ - Q2.λ)
+
+(Base.:+)(Q::SymplecticMatrix, p::AbstractMatrix) = p + Q
+function (Base.:+)(p::AbstractMatrix, Q::SymplecticMatrix)
+    # When we are adding, the Matrices must match in size:
+    n, _ = get_half_dims(p, true, false, true)
+
+    # Allocate new memory:
+    TS = Base._return_type(+, Tuple{eltype(p),eltype(Q)})
```
mateuszbaran, line 758:
```suggestion
    TS = typeof(one(eltype(p)) + one(eltype(Q)))
```
This is a more idiomatic element type calculation in Julia. Constant propagation should handle this.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+    )
+
+    return p
+end
+
+function LinearAlgebra.mul!(A::AbstractVecOrMat, Q::SymplecticMatrix, p::AbstractVecOrMat)
+    n, _ = get_half_dims(p, true, false)
+    # Perform left multiply by λ*Q:
+    mul!((@inbounds view(A, 1:n, :)), Q.λ, @inbounds view(p, (n + 1):lastindex(p, 1), :))
+    mul!((@inbounds view(A, (n + 1):lastindex(A, 1), :)), -Q.λ, @inbounds view(p, 1:n, :))
```
mateuszbaran, line 839: If you `@inbounds` access to `A`, you need to check its size first like here: https://github.com/JuliaLang/julia/blob/b486f0dec2f0049a22d263cc0e39f374a1c3b40a/stdlib/LinearAlgebra/src/symmetric.jl#L205-L214 for example.

mateuszbaran (commented):
I've mostly checked the symplectic manifold :slightly_smiling_face: .

src/manifolds/Symplectic.jl
johannvk: :+1:

src/manifolds/Symplectic.jl
johannvk: Good point, I'll change to a different letter like zeta or something for the documentation.

src/manifolds/Symplectic.jl
johannvk: :+1:

src/manifolds/Symplectic.jl
johannvk: Ok good to know. I've mostly used it as an easy way of generating initial points for optimization routines, but it may be cluttering the namespace more than it's worth. Is it worth keeping these functions for the convenience?

src/manifolds/Symplectic.jl
johannvk: :+1:

src/manifolds/Symplectic.jl
johannvk: I've follow the advice of your second comment Mateusz and gone for `return p * p' * X` as the body of the method.

src/manifolds/Symplectic.jl
johannvk: I've tried to clarify the use of the scalar `λ` to keep track of scaling and transpose operations now.

src/manifolds/Symplectic.jl
johannvk: I ported the documentation from the `SymplecticStiefel` case to this one now :+1:

src/manifolds/Symplectic.jl
johannvk: Good catch, I've added checks for the left- and right-multiply functions for `SymplecticMatrix`.

src/manifolds/Symplectic.jl
mateuszbaran: `random_vector(::Symplectic, p)` should be fine I think, it follows our naming for other functions that operate on tangent vectors.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,889 @@
...
+end
+
+(Base.:*)(x::Number, Q::SymplecticMatrix) = SymplecticMatrix(x * Q.λ)
+(Base.:*)(Q::SymplecticMatrix, x::Number) = SymplecticMatrix(x * Q.λ)
+function (Base.:*)(Q1::SymplecticMatrix, Q2::SymplecticMatrix)
+    return LinearAlgebra.UniformScaling(-Q1.λ * Q2.λ)
+end
+
+Base.transpose(Q::SymplecticMatrix) = -Q
+Base.adjoint(Q::SymplecticMatrix) = -SymplecticMatrix(conj(Q.λ))
```
mateuszbaran, line 758:
```suggestion
Base.adjoint(Q::SymplecticMatrix) = SymplecticMatrix(-conj(Q.λ))
```

src/manifolds/Symplectic.jl
mateuszbaran: What I meant is that why
```
Q_{2n} =
\begin{bmatrix}
0_n & I_n \\
 -I_n & 0_n
\end{bmatrix} \quad \in ℝ^{2n \times 2n},
```
isn't something like
```
Q_{2n}(λ) =
\begin{bmatrix}
0_n & λI_n \\
 -λI_n & 0_n
\end{bmatrix} \quad \in ℝ^{2n \times 2n},
```
?

mateuszbaran: Sorry, github showed me the old version for some reason when I was re-checking it.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,664 @@
...
+        Ω ∈ ℝ^{2k × 2k}, Ω^+ = -Ω,
+        p^s ∈ \operatorname{SpSt}(2n, 2(n- k)), B ∈ ℝ^{2(n-k) × 2k}, \},
+````
+where ``Ω \in \mathfrak{sp}(2n,F)`` is Hamiltonian and ``p^s`` means
+the symplectic complement of ``p`` s.t. ``p^{+}p^{s} = 0``.
+
+To then generate random tangent vectors at ``p``, we set ``B = 0`` and generate a random
+Hamiltonian matrix ``Ω``.
+"""
+function random_vector(
```
mateuszbaran, line 556: I've talked with @kellertuer on Slack about random vector generation and we agreed to be a bit more conservative about our interface. So the preferred way to get a random vector is `rand(M::AbstractManifold; vector_at=nothing)`, that is getting a random tangent vector at `p` will be `rand(M; vector_at=p)`. This avoids conflict with the API of `Base.rand` that leaves one argument for the distribution, but we can pass in keyword arguments whatever we like. `random_vector` as written here should stay but be internal (not exported) and called by `rand(M::AbstractManifold)` when `vector_at` is not `nothing`, more or less like this:
```julia
function Base.rand(M::SymplecticStiefel; vector_at=nothing, hamiltonian_norm = (vector_at === nothing ? 1/2 : 1.0))
    if vector_at === nothing
        # code for generation of random points
    else
        random_vector(M, vector_at; hamiltonian_norm = hamiltonian_norm)
    end
end
```

johannvk: Very nice, I also talked with @kellertuer about this today and it now sounds like you've found a good interface for this 
sort of functionality :relaxed:

src/manifolds/Symplectic.jl
mateuszbaran: I think we can leave this as-is here.

kellertuer:
Hm, I wanted to check the docs locally, since they do not build here on PR - but that might take some time, currently my Plots package or GR or PyCall generates a segfault when using a legend. Actually these three packages are the ones that I struggle with nearly every time I have to use them, because they either have outdated artefacts or segfault. I do not like them. I will have to check what best to do to check the docs here ...

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,664 @@
...
+function Base.rand(M::SymplecticStiefel{n,k}; hamiltonian_norm=1 / 2) where {n,k}
+    p_symplectic = rand(Symplectic(2n); hamiltonian_norm=hamiltonian_norm)
+    return canonical_projection(M, p_symplectic)
+end
+
+@doc raw"""
+    random_vector(::SymplecticStiefel{n, k}, p)
+
+The symplectic Stiefel tangent space at ``p`` can be parametrized as [^Bendokat2021]
+````math
```
kellertuer, line 545: I think it would be better do document the tangent space in the SymplecticStiefel type and here just refer to `[`SymplecticStiefel`](@ref)` - that way the form is documented in a central place

If you see a good way to split this focal into multiple lines, that would be great, its quite long).

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
+@doc raw"""
+    Symplectic{n, ℝ} <: AbstractEmbeddedManifold{ℝ, DefaultIsometricEmbeddingType}
+
+Over the real number field ℝ the elements of the symplectic manifold
```
kellertuer, line 4: For completeness we could also document the tangent space here.

kellertuer (commented):
Overall this already looks quite good and its great that the functions have good mathematical documentation. I have a few remarks written below and would like to ask to order the functions within both manifold files alphabetically. That makes them easier to find later on.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,664 @@
...
+
+decorated_manifold(::SymplecticStiefel{n,k,ℝ}) where {n,k} = Euclidean(2n, 2k; field=ℝ)
+
+function ManifoldsBase.default_inverse_retraction_method(::SymplecticStiefel)
+    return CayleyInverseRetraction()
+end
+
+ManifoldsBase.default_retraction_method(::SymplecticStiefel) = CayleyRetraction()
+
+@doc raw"""
```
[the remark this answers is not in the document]
mateuszbaran: Yes, this shouldn't be exported. If we are at it, I think  `grad_euclidean_to_manifold` shouldn't be exported either.

mateuszbaran:
> Hm, I wanted to check the docs locally, since they do not build here on PR - but that might take some time, currently my Plots package or GR or PyCall generates a segfault when using a legend. Actually these three packages are the ones that I struggle with nearly every time I have to use them, because they either have outdated artefacts or segfault. I do not like them. I will have to check what best to do to check the docs here ...

I managed to build docs locally: [docs-PR.zip](https://github.com/JuliaManifolds/Manifolds.jl/files/8053984/docs-PR.zip)

kellertuer:
Thanks, My main problem seems to be the 3d peplos plots in sphere – with an empty sphere.md it worked fine as well. But after 2 hours I did not feel like debugging that further. The comments above are from my rendered version :)

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+Q_{2n} =
+\begin{bmatrix}
+  0_n & I_n \\
+ -I_n & 0_n
+\end{bmatrix},
+````
+with $0_n$ and $I_n$ denoting the $n × n$ zero-matrix
+and indentity matrix in ``ℝ^{n \times n}`` respectively.
+
+Internally the dimensionality of the structure is stored as half of the even dimension
```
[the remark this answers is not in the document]
johannvk: :+1:

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+    Q = SymplecticMatrix(p, X)
+
+    pT_QT_X = p' * Q' * X
+    symmetrized_pT_QT_X = (1 / 2) .* (pT_QT_X + pT_QT_X')
+
+    Y[:, :] = p * Q * (symmetrized_pT_QT_X)
+    return Y
+end
+
+@doc raw"""
```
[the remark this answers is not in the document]
johannvk: Both this function and `project_riemannian` are not exported at the moment, but would it be better to rename them both 
to `project` and `project_normal` and then specialize on the first manifold argument? 

Like you showed as an example with `project(MetricManifold{Extended...Metrix, Symplectic}, Y, p, X)`.

src/manifolds/SymplecticStiefel.jl
johannvk: I've moved the documentation into the `SymplecticStiefel` type now, and removed the documentation of
the `random_vector` function entirely in favor of documenting the new `rand(::SymplecticStiefel; vector_at=...)` function 
more thoroughly.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,664 @@
...
+    amb_grad = _gradient(f, p, backend.diff_backend)
+    return grad_euclidean_to_manifold(M, p, amb_grad)
+end
+
+function gradient!(M::SymplecticStiefel, f, X, p, backend::RiemannianProjectionBackend)
+    _gradient!(f, X, p, backend.diff_backend)
+    return grad_euclidean_to_manifold!(M, X, p, copy(X))
+end
+
+function grad_euclidean_to_manifold(M::SymplecticStiefel, p, ∇f_euc)
```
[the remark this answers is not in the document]
johannvk: We can do that, but how would one then 'manually' transform a known Euclidean gradient to the 
manifold gradient? I've tested and benchmarked a bit the difference between using `ForwardDiff` automatic differentiation and computing the Euclidean gradient "manually" and then transforming it using `grad_euclidean_to_manifold`, and there is quite a big difference when the matrices get bigger.

For example when projecting (200 x 40) matrices onto the symplectic manifold with manual conversion I get
a runtime on the order of 100 milli-seconds, whilst the runtime is closer to ten seconds on my machine when using the `Manifolds.ForwardDiffBackend` with the `gradient!(::SymplecticStiefel, f, p, X, ...)` API.

src/manifolds/Symplectic.jl
johannvk: :+1:

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+    return A
+end
+
+@doc raw"""
+    get_half_dims(p, check_rows=true, check_cols=true, square=false)
+
+Convenience function to check whether or not an abstract matrix is square, with an even
+number (2n, 2n) of rows and columns. Then returns the integer part of the even dimension.
+"""
+function get_half_dims(p::AbstractMatrix, check_rows=true, check_cols=true, square=false)
```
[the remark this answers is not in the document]
johannvk: Yeah this function has grown to a bit of a monstrosity, I've pushed a commit which removes it now :smile:

src/manifolds/SymplecticStiefel.jl
johannvk: I've made the changes to this API-style now.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+Internally the dimensionality of the structure is stored as half of the even dimension
+supplied to the constructor, `Symplectic(2n) -> Symplectic{n}()`,
+as most computations with points on the Real Symplectic manifold takes advantage of the
+natural block structure
+of a matrix ``A ∈ ℝ^{2n × 2n}`` where we consider it as consisting of four
+smaller matrices in ``ℝ^{n × n}``.
+"""
+struct Symplectic{n,𝔽} <: AbstractEmbeddedManifold{𝔽,DefaultIsometricEmbeddingType} end
+
+@doc """
```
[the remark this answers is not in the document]
johannvk: :+1:

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+with $0_n$ and $I_n$ denoting the $n × n$ zero-matrix
+and indentity matrix in ``ℝ^{n \times n}`` respectively.
+
+Internally the dimensionality of the structure is stored as half of the even dimension
+supplied to the constructor, `Symplectic(2n) -> Symplectic{n}()`,
+as most computations with points on the Real Symplectic manifold takes advantage of the
+natural block structure
+of a matrix ``A ∈ ℝ^{2n × 2n}`` where we consider it as consisting of four
+smaller matrices in ``ℝ^{n × n}``.
+"""
```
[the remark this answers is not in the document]
johannvk: I've removed much of the text rambling on a bit about the internals which aren't that important for the users now, 
and combined with the documentation for the constructor.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+@doc raw"""
+The canonical Riemannian metric on the symplectic manifold,
+defined pointwise for ``p \in \operatorname{Sp}(2n)`` by [^FioriSimone2011]
+````math
+\begin{align*}
+    & g_p \colon T_p\operatorname{Sp}(2n) \times T_p\operatorname{Sp}(2n) \rightarrow ℝ, \\
+    & g_p(Z_1, Z_2) = \operatorname{tr}((p^{-1}Z_1)^T (p^{-1}Z_2)).
+\end{align*}
+````
+This metric is also the default metric used within `Maniolds.jl`.
```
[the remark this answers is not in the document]
johannvk: Obs good catch, what's written now doesn't make a lot of sense no :smile:

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,930 @@
...
+"""
+struct RealSymplecticMetric <: RiemannianMetric end
+
+struct ExtendedSymplecticMetric <: AbstractMetric end
```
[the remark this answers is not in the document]
johannvk: :+1:

src/manifolds/Symplectic.jl
johannvk: Sorry I seem to have mixed up some of the commits, I thought I'd already commited this change, but I'll reintroduce it now at least.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,664 @@
...
+with $0_n$ and $I_n$ denoting the $n × n$ zero-matrix
+and indentity matrix in ``ℝ^{n \times n}`` respectively.
+
+Internally the dimensionality of the structure is stored as half of the even dimensions
+supplied to the constructor, `SymplecticStiefel(2n, 2k) -> SymplecticStiefel{n, k, ℝ}()`,
+as most computations with points on the Real symplectic Stiefel manifold takes
+advantage of the natural block structure matrices
+``A ∈ ℝ^{2n × 2k}`` where we consider it as consisting of four
+smaller matrices in ``ℝ^{n × k}``.
+"""
```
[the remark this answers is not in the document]
johannvk: :+1:

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,664 @@
...
+
+@doc raw"""
+    gradient(::SymplecticStiefel, f, p, backend::RiemannianProjectionBackend)
+    gradient!(::SymplecticStiefel, f, X, p, backend::RiemannianProjectionBackend)
+
+Compute the gradient of
+``f\colon \operatorname{SpSt}(2n, 2k) \rightarrow ℝ``
+at ``p \in \operatorname{SpSt}(2n, 2k)``.
+
+We first compute the embedding gradient ``∇f(p) \in ℝ^{2n \times 2k}`` using
```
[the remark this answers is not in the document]
johannvk: :+1:

src/manifolds/Symplectic.jl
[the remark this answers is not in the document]
johannvk: I've clarified what I was trying to say here now.

src/manifolds/SymplecticStiefel.jl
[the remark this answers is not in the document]
kellertuer: gradient should do that – it is just calling this function – so I do not understand where your difference comes from? Ah I know. ForwardDiff does an approximation of the (Euclidean) Gradient with forwardDifference the wy you use it, so that is of course just an approximation to the actual gradient, we can discuss the details in our next meeting as well.

johannvk:
I pushed some more tests to cover the simplified dimensionality checking for operations with the `SymplecticMatrix` struct,
and also refactored `project_riemannian(::Symplectic, ..)` and `project_riemannian_normal(::Symplectic, ...)` to `project(::MetricManifold{𝔽, Euclidean{Tuple{m, n}, 𝔽}, ExtendedSymplecticMetric}, ...)` and `project_normal(::MetricManifold{𝔽, Euclidean{Tuple{m, n}, 𝔽}, ExtendedSymplecticMetric}, ...)`. 

Additionally i removed the documentation for both functions to not export them to the public documentation.

src/manifolds/Symplectic.jl
```diff
@@ -0,0 +1,929 @@
...
+    return gradient!(M, f, Y, p, backend, extended_metric=extended_metric)
+end
+
+function gradient!(
+    M::Symplectic,
+    f,
+    X,
+    p,
+    backend::RiemannianProjectionBackend;
+    extended_metric=true,
```
kellertuer, line 415: I am not yet 100% sure this is the way to go, but this is one way to go.
@mateuszbaran what do you think? The difference here is

With extended `extended_metric = false`the embedding (which is not a parameter here) is just the Euclidean space – we first project and then convert from Euclidean metric to Symplectic (This is actually new, but not fast, still the default gradient conversion way)

With `extended_metric=true` the metric conversion comes first (that is Euclidean from default metric to an Extended Symplectic – which is not a metric, but locally projected is fine) and then the projection (that is w.r.t. symplectic metric).

Since this difference only appears very rarely and for now really only here, I think a kwarg would be fine as it is now implemented. What do you think?

mateuszbaran: Yes, it's usually fine to add new keyword arguments to methods for specific manifolds. We can leave it as is, and change it later when we have a better idea. `gradient` interface isn't particularly stable anyway.

mateuszbaran:
> Additionally i removed the documentation for both functions to not export them to the public documentation.

I think it's fine to have documentation for unexported functions. You can always write in the docstring that it's an internal function to make it clear.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -0,0 +1,676 @@
...
+    \operatorname{dim}(\operatorname{SpSt}(2n, 2k)) = (4n - 2k + 1)k.
+````
+"""
+manifold_dimension(::SymplecticStiefel{n,k}) where {n,k} = (4n - 2k + 1) * k
+
+@doc raw"""
+    project(::Union{SymplecticStiefel,Symplectic}, p, A)
+    project!(::Union{SymplecticStiefel, Symplectic}, Y, p, A)
+
+Given a point ``p \in \operatorname{SpSt}(2n, 2k)``,
```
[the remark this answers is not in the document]
johannvk: Yeah I see what you mean, a couple of university coding courses have left me with an aversion for code duplication, but here I think it makes sense. Easier to document as you say, and people who work on the symplectic manifold will have an easier time finding out what functionality is actually available to them :smile:

johannvk:
I've now completed the last revisions based on the comments of both @mateuszbaran and @kellertuer, and I also looked over the documentation and found some small errors and inaccuracies which I tweaked. 
After talking to @kellertuer I also added an entry for myself under the `Contributors` section in the `.zenodo.json` file, and bumped the version number to `0.7.5`.

kellertuer (approved):
Looks good to me 👍

mateuszbaran (approved):
Looks great, thank you for the contribution :+1: .

PR #448, opened 2022-01-28
Differentials of `log` and `exp`

mateuszbaran:
In this PR I'd like to add functions and some concrete implementations of approximations of differentials of `log` and `exp`. I know Manopt.jl has something similar but only using the Jacobi field method and I won't be adding Jacobi fields here.

### DIFF
--- a/assets/projection_illustration.jl
+++ b/assets/projection_illustration.jl
@@ -43,7 +43,7 @@ tp = @pgf Axis({
     axis_equal,
     xmin = -1.7,
     xmax = 1.7,
-    ymin = -.25,
+    ymin = -0.25,
     ymax = 1.4,
     scale = 1.6,
 })
--- a/assets/retraction_illustration.jl
+++ b/assets/retraction_illustration.jl
@@ -41,7 +41,7 @@ tp = @pgf Axis({
     axis_equal,
     xmin = -1.7,
     xmax = 1.7,
-    ymin = -.25,
+    ymin = -0.25,
     ymax = 1.4,
     scale = 1.6,
 })
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -80,6 +80,7 @@ makedocs(
         ],
         "Features on Manifolds" => [
             "Atlases and charts" => "features/atlases.md",
+            "Differentials" => "features/differentials.md",
             "Differentiation" => "features/differentiation.md",
             "Distributions" => "features/distributions.md",
             "Statistics" => "features/statistics.md",
--- a/docs/src/features/differentials.md
+++ b/docs/src/features/differentials.md
@@ -1,0 +1,7 @@
+# Differentials
+
+```@autodocs
+Modules = [Manifolds,ManifoldsBase]
+Pages = ["differentials.jl"]
+Order = [:type, :function]
+```
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -201,6 +201,7 @@ include("utils.jl")
 include("product_representations.jl")
 include("differentiation/differentiation.jl")
 include("differentiation/riemannian_diff.jl")
+include("differentials.jl")
 
 # Main Meta Manifolds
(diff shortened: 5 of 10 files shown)


kellertuer:
Sure, no problem, if they reach the same completeness (or more?) compared to Manopt we can also move them completely here. The reason we have the DiagonalizingONB is already owing to them being computed with Jacobi fields.

If so, we could also move the other Differentials (distance for example) to here as well. Maybe with nicer names than in Manopt.jl

mateuszbaran:
Sure, at least we will have to update Manopt.jl differentials to the same interface. One point is naming, we already have a few differentials and I was trying to have names closer to what we have in Manifolds.jl already than copy the names from Manopt.jl. What do you think?

Manopt.jl also has adjoint differentials but I don't know what exactly you mean by that. Could you give me a reference for that?

Ideally we would collect here as many differentials as possible but I've started with what seems important to me.

kellertuer:
The way and notation I use is the one described in Section 4.2 (p. 17) https://arxiv.org/pdf/1807.10090.pdf

Concerning the naming, I have not yet understood all you names, but I would prefer a high-level version that is easy to use (and I also do not like the naming scheme in Manopt.jl).

Is there some way we could do a `differential` function? For sure it makes it a little complicated to distinguish the parameters (base point, argument) and on the low level your functions might be fine, but on Layer one maybe something like

```
differential(M, exp, p, X, \xi, ArgumentDifferential())
```
for $D_X \exp_pX [\xi]$ would be neat? (similarly `BasepointDifferential()`for $D_p \exp_pX [Y]$).

mateuszbaran:
> The way and notation I use is the one described in Section 4.2 (p. 17) https://arxiv.org/pdf/1807.10090.pdf

OK, thanks. So, in other words, it's a pullback working on Riesz representers?

> Concerning the naming, I have not yet understood all you names, but I would prefer a high-level version that is easy to use (and I also do not like the naming scheme in Manopt.jl).
> 
> Is there some way we could do a `differential` function? For sure it makes it a little complicated to distinguish the parameters (base point, argument) and on the low level your functions might be fine, but on Layer one maybe something like
> 
> ```
> differential(M, exp, p, X, \xi, ArgumentDifferential())
> ```
> 
> for $D_X \exp_pX [\xi]$ would be neat? (similarly `BasepointDifferential()`for $D_p \exp_pX [Y]$).

Sure, we could add a high-level function like that but low-level functions need to have different names. We could also probably try to use (or at least get some inspiration) from the ChainRules.jl interface? Its main limitation is that it requires differentiating with respect to all arguments at once.

kellertuer:
Ah yes, that is another interpretation one could use.

And suer for layer3 we need own functions.

mateuszbaran:
I didn't find any other notable formulas for differentials, so I think I'll just polish what we have right now.

By the way, I've looked at your Jacobi fields (a really cool approach) and I think we could have some improvements there. Mainly I'm thinking whether we could group projections by eigenvalues of the Jacobi operator? So that instead of projecting separately to each eigenvector, we project to principal subspaces, if I'm getting it right.

kellertuer:
If you see something that can be improved, that would be neat. I mainly developed that for the Bezier paper and could generalise that. If you see improvements – sure :) go ahead. We can also discuss whether that fits better here or there; but what I like currently, is that Manopt.jl is decoupled from Manifolds.jl

I will check soon what you wrote here :)

mateuszbaran:
For reference, there has recently been some work on second order differentials of `exp`: https://mathoverflow.net/a/382882 . The papers referenced there are https://arxiv.org/pdf/2008.02252.pdf and https://arxiv.org/pdf/2008.02517.pdf .

mateuszbaran:
The interface seems basically complete now. One thing I could still do here is adding flat differentials to all flat manifolds but we could also turn it into a "good first issue" thing.

mateuszbaran:
I moved the important parts to ManifoldDiff.jl

Issue #454, opened 2022-02-18
Add a backend RiemannieanProjectedBackend that stores the Euclidean gradient


kellertuer:
If we have a gradient in the embedding it would be cool to have a backend in `RiemannianProjectedBackend` that just evaluates the Euclidean Gradient.

mateuszbaran:
Isn't this already possible by just accessing `diff_backend` field of `RiemannianProjectionBackend` and calling `_gradient` with it?

mateuszbaran:
Yes, if projection and metric change are identity then just using the wrapped `diff_backend` is fine. The only potential use case for something like `EmbeddedIdentityBackend` would be overriding projection and metric change of the manifold but I don't see why would you want to do that.

mateuszbaran:
Ah, that's not currently possible. I'd say that we should have a separate function for that actually. It would be then called in `gradient!(M::AbstractManifold, f, X, p, backend::RiemannianProjectionBackend)`.

kellertuer:
I would have thought to dispatch on another `EmbeddedBackend` within the Reimannian_projection_Backend – because we have the call to the embedding gradient in `gradient!` (just that for now it can only use ForwardDiff etc but not just a function call).

kellertuer:
That is my plan. I know what to do there, I just wanted to make sure I did not miss anything. I already assigned this to myself, because (without documentation) this is about 5 lines of code.

PR #455, opened 2022-02-18
Introduce EmbeddedBackend

kellertuer:
Solves #454. 

The `EmbeddedBackend´ contains the gradient (derivative, jacobian) of a function $f$ defined on a manifold that describe its gradient (derivative, jacobian) as a function extended into the embedding. Similar t Euclidean tools available they hence return Euclidean gradients (derivatives, Jacobins) that have to be converted.

* We should find a way to maybe also introduce mutating variants 
* Is the design ok this way?
* Yes, documentation is missing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.7.6"
+version = "0.7.7"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
--- a/docs/src/features/differentiation.md
+++ b/docs/src/features/differentiation.md
@@ -33,3 +33,9 @@ Modules = [Manifolds]
 Pages = ["differentiation/riemannian_diff.jl"]
 Order = [:type, :function, :constant]
 ```
+
+```@autodocs
+Modules = [Manifolds]
+Pages = ["differentiation/embedded_diff.jl"]
+Order = [:type, :function, :constant]
+```
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -713,6 +713,7 @@ export get_basis,
 # differentiation
 export AbstractDiffBackend,
     AbstractRiemannianDiffBackend,
+    ExplicitEmbeddedBackend,
     FiniteDifferencesBackend,
     TangentDiffBackend,
     RiemannianProjectionBackend
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -201,6 +201,7 @@ include("utils.jl")
 include("product_representations.jl")
 include("differentiation/differentiation.jl")
 include("differentiation/riemannian_diff.jl")
+include("differentiation/embedded_diff.jl")
 
 # Main Meta Manifolds
 include("manifolds/ConnectionManifold.jl")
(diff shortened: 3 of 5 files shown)


mateuszbaran:
Yes, I think the design is essentially correct. A few points to consider:
1. Maybe `ExplicitEmbeddedBackend` would be a better name that would highlight that the main feature of this backend.
2. `struct EmbeddedBackend` could just store a `NamedTuple` with functions so that we don't have to add a new field each time a new function is added, and it could then be even extended without modifying Manifolds.jl.
For example
```julia
struct ExplicitEmbeddedBackend{TF<:NamedTuple} <: AbstractDiffBackend
    funs::TF
end
ExplicitEmbeddedBackend(; kwargs...) = ExplicitEmbeddedBackend{typeof(values(kwargs))}(values(kwargs))
```
3. `ExplicitEmbeddedBackend` could also store the function these gradients, differentials and Jacobians correspond to so that it could for example be checked in specific methods that the function given to, for example, `_derivative` is the same one someone wrote the gradient for.

mateuszbaran:
> That would be the `function` field for example, yes; the only thing that might be misleading then is, that the `_gradient`,... have f as their first argument which – if we store it – is a little double information in the arguments?

That's why I think that storing function in `ExplicitEmbeddedBackend` should be optional, only if someone wants the safety.

kellertuer:
> 3. `ExplicitEmbeddedBackend` could also store the function these gradients, differentials and Jacobians correspond to so that it could for example be checked in specific methods that the function given to, for example, `_derivative` is the same one someone wrote the gradient for.

I have one further question to this one – how would I check (in “specific methods”) that the functions are the same and how would I do that? For now I would just check whether the named tuple has the field required for `gradient` or `gradient!` and throw an error otherwise, but how do I check whether functions are the same and why would I do that? Gradients and Derivatives are different...

kellertuer:
I have not yet done the testing, but the modelling and documentation should be complete.

mateuszbaran:
> I have one further question to this one – how would I check (in “specific methods”) that the functions are the same and how would I do that? For now I would just check whether the named tuple has the field required for `gradient` or `gradient!` and throw an error otherwise, but how do I check whether functions are the same and why would I do that? Gradients and Derivatives are different...

I imagined something like
```julia
function check_expected_function(f, e::ExplicitEmbeddedBackend)
    if haskey(e.functions, :func) && if f !== e.functions.func
        error("Wrong function: expected $(e.functions.func), received $f")
    end
end

function _gradient(f, p, e::ExplicitEmbeddedBackend)
    check_expected_function(f, e)
    g = get(e.functions, :gradient, Missing())
```

kellertuer:
Yes one could do that but in most cases we (up to now) where quite lazy in checking (by default, so only if you activate checking we checked things).

mateuszbaran:
Here you have to activate the check as well :slightly_smiling_face:, otherwise it is just not performed.

kellertuer:
Well you implicitly activate it by providing a cost function in the backend, that would be a little implicit, I think. I will think about how to best realise this.

mateuszbaran:
OK, so let's leave it out of this PR if you don't like it.

mateuszbaran:
I thought it would be a nice quick addition but now I see that it may indeed be confusing unless carefully designed, so let's just forget about it.

kellertuer:
But, hm, this is what Pymanopt and Manopt do with there `egrad`s, so I hope we find a way to cover that as well.

mateuszbaran:
What do you mean by "this"? This PR should work fine for gradients.

kellertuer:
Ah then I misread the last parts of the discussion here. I will check but I think this should nearly be good to go (+- tests).

mateuszbaran:
Yes, I only became skeptical about `check_expected_function`, everything else looks fine :slightly_smiling_face: .

kellertuer:
ok löste just leave that out, I do not feel good about that, either.

kellertuer:
I added tests for the current state based on Rayleigh quotient and I hope this does the code coverage – then this is good to go.

mateuszbaran (approved):
Nice, I think it can be merged when coverage passes.

PR #458, opened 2022-02-25
fix Tangent ONB and DiagONB and add a Test.

### LINKED ISSUE #457: `DefaultOrthonormalBasis` on SPD(3) is not orthonormal
Here is a short-ish reproducer:
```julia
julia> M = SymmetricPositiveDefinite(3)
SymmetricPositiveDefinite(3)

julia> q = [2.0 0.0 0.0; 0.0 2.0 0.0; 0.0 0.0 1]
3×3 Matrix{Float64}:
 2.0  0.0  0.0
 0.0  2.0  0.0
 0.0  0.0  1.0

julia> b = get_basis(M, q, DefaultOrthonormalBasis())
DefaultOrthonormalBasis(ℝ) with 6 basis vectors:
 E1 =
  3×3 Matrix{Float64}:
   1.0  0.0  0.0
   0.0  0.0  0.0
   0.0  0.0  0.0
 E2 =
  3×3 Matrix{Float64}:
   0.0       0.707107  0.0
   0.707107  0.0       0.0
   0.0       0.0       0.0
 ⋮
 E5 =
  3×3 Matrix{Float64}:
   0.0  0.0       0.0
   0.0  0.0       0.707107
   0.0  0.707107  0.0
 E6 =
  3×3 Matrix{Float64}:
   0.0  0.0  0.0
   0.0  0.0  0.0
   0.0  0.0  1.0

julia> norm(M, q, b.data[1])
0.4999999999999999
```

kellertuer:
This will fix #457.

Main Changes: We have to take the Eigenvectors of p^-1/2 V p^-1/2 (if V is the frame direction, V = I for ONB so then just p^-1) and the ONB vectors have to be post multiplied with p^1/2 * Xi * p^1/2

* [x] we have to adapt get_vector
* [x] we have to adapt get_coordinates
* [x] we should document this in the docstrings 

But we still need to adapt get_vector and get_coordinates since they are nor much more complicated.

### DIFF
--- a/src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
+++ b/src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
@@ -151,25 +151,31 @@ end
     [Ξ,κ] = get_basis(M::SymmetricPositiveDefinite, p, B::DefaultOrthonormalBasis)
     [Ξ,κ] = get_basis(M::MetricManifold{SymmetricPositiveDefinite{N},LinearAffineMetric}, p, B::DefaultOrthonormalBasis)
 
-Return a default ONB for the tangent space ``T_p\mathcal P(n)`` of the [`SymmetricPositiveDefinite`](@ref) with respect to the [`LinearAffineMetric`]
+Return a default ONB for the tangent space ``T_p\mathcal P(n)`` of the [`SymmetricPositiveDefinite`](@ref) with respect to the [`LinearAffineMetric`](@ref).
 
-    ```math
-        g_p(X,Y) = \operatorname{tr}(p^{-1} X p^{-1} Y),
-    ```
+```math
+    g_p(X,Y) = \operatorname{tr}(p^{-1} X p^{-1} Y),
+```
 
-    The basis constructed here is based on the ONB for symmetric matrices constructed as follows.
-    Let ``\Delta_{i,j} = (a_{kl}=_{k,l=1}^n`` with ``a_{kl} =
-    \begin{cases}
-      1 & \mbox{ for } k=l \text{ if } i=j\\
-      \frac{1}{\sqrt{2}}} & \mbox{ for } k=i, l=j \text{ or } k=j, l=i\\
-      0 & \text{ else.}
-    \end{cases}$
-    which forms an ONB for the space of symmetric matrices.
+The basis constructed here is based on the ONB for symmetric matrices constructed as follows.
+Let
 
-    We then form the ONB by
-    ```math
-    \Xi_{i,j} = p^{\frac{1}{2}}\Delta_{i,j}p^{\frac{1}{2}},\qquad i=1,\ldots,n, j=i,\ldots,n
-    ```
+```math
+\Delta_{i,j} = (a_{k,l})_{k,l=1}^n \quad \text{ with }
+a_{k,l} =
+\begin{cases}
+  1 & \mbox{ for } k=l \text{ if } i=j\\
+  \frac{1}{\sqrt{2}} & \mbox{ for } k=i, l=j \text{ or } k=j, l=i\\
+  0 & \text{ else.}
+\end{cases}
+```
+
+which forms an ONB for the space of symmetric matrices.
+
+We then form the ONB by
+```math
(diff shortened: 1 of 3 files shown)


kellertuer:
My local pycall broke again, so I have not yet seen the docs, but the vector/coordinates are adapted and documented (just sat down simplified both a little bit). Take a look whether they can be made even faster.

I hope all three docs are understandable.

kellertuer:
HM then this was also never covered by tests? Interesting, I will have to take a much closer look than, what I basically did was (for coordinates) simplify the many inner products and ah, I see, I am not allowed to do what I did for get vector the p^1/2 have to be within the sum (or I can only factor out one of them).

mateuszbaran:
We probably only tested `get_vector` and `get_coordinates` at identity matrix, so the tests were quite easy to pass.

kellertuer:
Sure, I added your example as a test, but I will need some time with pen&paper again to check that both my derivations / simplifications for vector and coordinates are correct.

kellertuer:
The last commit is code that works, but the ad hoc generation of the basis vectors and the `inner` could be simplified still (that is what I tried before but must have made a mistake in rearranging the `tr`.

get_vector is finished and works, we can leave `get_coordinates` as it is currently, but there is also space for more performance.

mateuszbaran:
Thanks! I'll check it soon.

mateuszbaran:
I've significantly improved performance. I get roughly 500 ns for get_coordinates and 200 ns for get_vector, on static arrays. There are no obvious optimizations for these two functions. get_basis may be still worth optimizing in a similar way (mainly updating `(V[:, i] * transpose(V[:, j]) + V[:, j] * transpose(V[:, i]))` in-place).

kellertuer:
Great, thanks for the ideas, not only did they make the code faster, it is also more readable now. I adopted that for the basis as well.

src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
```diff
@@ -117,64 +117,182 @@ Return a orthonormal basis `Ξ` as a vector of tangent vectors (of length
...
+@doc raw"""
+    get_vector(::SymmetricPositiveDefinite, p, c, ::DefaultOrthonormalBasis)
+
+Using the basis from [`get_basis`](@ref  get_basis(M::SymmetricPositiveDefinite,p,B::DefaultOrthonormalBasis{<:Any,ManifoldsBase.TangentSpaceType}))
+the vector reconstruction with respect to this ONB can be simplified to
+
+```math
+   X = p^{\frac{1}{2}} \Biggl( \sum_{i=1,j=i}^n c_k \Delta_{i,j} \Biggr) p^{\frac{1}{2}}
+```
+where $k$ is trhe linearized index of the $i=1,\ldots,n, j=i,\ldots,n$.
```
mateuszbaran, line 262:
```suggestion
where $k$ is the linearized index of the $i=1,\ldots,n, j=i,\ldots,n$.
```

src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
```diff
@@ -117,64 +117,182 @@ Return a orthonormal basis `Ξ` as a vector of tangent vectors (of length
...
-    @assert size(X) == (div(N * (N + 1), 2),)
-    @assert size(Y) == (N, N)
+    @assert size(c) == (div(N * (N + 1), 2),)
+    @assert size(X) == (N, N)
+    e = eigen(Symmetric(p))
+    U = e.vectors
+    S = max.(e.values, floatmin(eltype(e.values)))
+    Ssqrt = Diagonal(sqrt.(S))
+    pSqrt = Symmetric(U * Ssqrt * transpose(U))
+    #V = Matrix{Float64}(I, N, N)
```
mateuszbaran, line 280:
```suggestion
```

mateuszbaran:
Great, I think `DiagonalizingOrthonormalBasis` could be modified in the same way as well?

mateuszbaran:
... or maybe not, `V` doesn't have to be diagonal there so it's less clear.

mateuszbaran (approved):

kellertuer:
no maybe not that easy because we do not necessarily have unit vectors but really have to form the corresponding V'V matrices.

mateuszbaran:
Could we maybe bump version, merge this and tag a new release? There doesn't seem anything left to be done here.

kellertuer:
Sure, no problem.

PR #465, opened 2022-03-02
Fix `allocate_coordinates` for `ArrayPartition`

mateuszbaran:
A small fix so that `get_coordinates` doesn't return `ArrayPartition` back.


mateuszbaran:
@kellertuer can we merge this?

kellertuer:
Yes, looks good to me.

kellertuer (approved):

PR #466, opened 2022-03-15
Riemannian Jacobian

mateuszbaran:
A small PR that extends our FD support to Jacobians.


mateuszbaran:
This is now in https://github.com/JuliaManifolds/ManifoldDiff.jl

PR #468, opened 2022-04-03
Upgrade Manifolds.jl to ManifoldsBase 0.13

kellertuer:
This PR adapts Manifolds to the new interface introduced in ManifoldsBase.jl 0.13.0


mateuszbaran:
Cool! Let's see how much test coverage needs to be improved :slightly_smiling_face: .

kellertuer:
I hope my merge does not break tests – and I want to check Docs as well (start that the next few days), for example the `GroupManifold` has its own file, addition and multiplication group functions as well, we have to remove the interface.md file,...

kellertuer:
Interesting, that locally the tests ran fine.

kellertuer:
Docs render fine now, just on `groups.md` the Pennec reference appears six times, we might want to rephrase a few function docs maybe?

mateuszbaran:
I don't really mind references appearing multiple times but feel free to rephrase docs to avoid it if you want :slightly_smiling_face: .

kellertuer:
I will think about it. I usually don't mind if footnotes show up 2 or 3 times, but 6 is quite a lot.

mateuszbaran:
Probably the easiest idea would be to split group-related tests to a separate CI job like here: https://github.com/mcabbott/Tullio.jl/pull/62 . What do you think?

kellertuer:
Sure, that sounds good to me. One idea was anyways to do the Lie groups separately in their own package at some point.

mateuszbaran:
I wouldn't hurry with splitting groups to a separate package but this may be considered a first step in that direction. `test_manifolds` and `test_lie_groups` sound fine :+1: .

kellertuer:
Same for me, no hurry necessary, sure.
Just overall this package is growing quite large overall, and then getting GroupManifold into Base and doing a LieGroups package is a good general idea to follow at some point

mateuszbaran:
OK, I'll split the tests today.

kellertuer:
This looks already quite nice, we could omit the word `group` in the CI title?

mateuszbaran:
Yes, sure, I'll delete it in the next commit. I want to let the tests finish on the current commit first and verify that it works correctly, then I'll change the name and update nightly CI job.

mateuszbaran:
Hm, Julia 1.5 is failing. Given that 1.6 is the new LTS, and Manopt.jl already requires 1.6, can we just drop 1.5 here?

kellertuer:
I think we can do that.

Sad that we lost so much coverage, but it might be that we still have to move a few functions to level3 or we can even remove them here now (like coordinates basis stuff, which is now more covered in ManifoldsBase)?

mateuszbaran:
> I think we can do that.

OK, great.

> Sad that we lost so much coverage, but it might be that we still have to move a few functions to level3 or we can even remove them here now (like coordinates basis stuff, which is now more covered in ManifoldsBase)?

The biggest issue right now is that coverage from the test_manifolds isn't taken into account yet. So coverage should be much better when it is.

kellertuer:
Ah, then it will really be much better than, cool, I did not see that. But shouldn't it just be enough to push both groups to codecov?

mateuszbaran:
> But shouldn't it just be enough to push both groups to codecov?

That's exactly what should happen. `test_manifolds` group just hasn't passed CI yet.

dehann:
Congrats on a big upgrade!

mateuszbaran:
> Congrats on a big upgrade!

Thanks! I think this is a significant improvement to how things work internally.

mateuszbaran:
We finally have code coverage from the full run. Lots of false negatives though.

kellertuer:
But also a lot of them because we kind of “allocate late” now. I am not yet sure how to best resolve these, but it's good that we have an overview what's left to do (and since the docs are fixed, tests is the last thing to do).

kellertuer:
I am not sure why the tests (2) reported to be killed and whether the codecov was updated.

mateuszbaran:
> I am not yet sure why quite a few mutating layer 3 functions are not called, especially the mutating ones (to which all should default.

I'm not sure either, maybe we deleted some relevant tests?

> I am not sure why the tests (2) reported to be killed and whether the codecov was updated.

Only nightly tests were killed (probably an out-of-memory error). You can check if codecov was updated by, for example, checking the edit history of the post by codecov bot.

kellertuer:
Ah checking for the edit is a good tip.
Then removing the unnecessary size checks worked perfectly fine, the main thing left is quite some mutating variants on layer 3 not being hit and sadly quite a few false negatives for layer 3 function signatures.

mateuszbaran:
> What has happened is that we might switch to mutating variants later now so in more and more cases the allocating (layer-3-)variant is called for the allocating case.

Yes, that's most likely it.

kellertuer:
I worked through the first manifold (and noticed we missed some details on Connection). I mainly noticed two things

1. We have quite a few signatures of functions that are now not recognised as covered, but I could not find a reason why that is the case. Do you have an idea how to avoid that?
2. vector transports seem (for example for the next: Euclidean) not easy to hit, I am not sure why.

kellertuer:
Hm and it seems I broke something in Connections, can you maybe take a look ? I had hoped to only fix the things that were missing, but for me Connections are still a very complex part of our model here with so many side effects.

mateuszbaran:
> Hm and it seems I broke something in Connections, can you maybe take a look ? I had hoped to only fix the things that were missing, but for me Connections are still a very complex part of our model here with so many side effects.

OK, I will fix connections.

> 1. We have quite a few signatures of functions that are now not recognised as covered, but I could not find a reason why that is the case. Do you have an idea how to avoid that?

I don't understand it. If the signature has some default values that are never used, I would understand it not being marked as covered when there are no defaults and the method is called, then I don't see any reason to consider the signature not covered. I would ignore these issues because there is nothing we can do about it -- it may fix itself at some point.

> 2\. vector transports seem (for example for the next: Euclidean) not easy to hit, I am not sure why.

I've spent quite a few hours fixing vector transports in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/96 . I wasn't able to quickly figure out what call invokes what method so I had to use debugger. In most cases it made sense, in one the chain of calls was incorrect. I don't have a better idea than using a debugger on calls that were supposed to call the method reported as not covered.

mateuszbaran:
I've changed one-element vectors to zero-dimensional arrays but it's been quite a bit of work and https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/141 now causes `NLSolveInverseRetraction` to not work for circle. To be honest, I'm not even sure why we added that test, I don't see how it could be useful.

kellertuer:
I tried to get code coverage for induced basis on Euclidean (which are not yet covered), but I am not able to get it working. It even too a deep dive into the code to see that the only way to create the basis is currently (after even loading the Interface explicitly in REPL (`E` is `Euclidean(3)`)

```
induced_basis(E, RetractionAtlas(), 0, ManifoldsBase.TangentSpaceType())
```

but adding that to the list of bases at
https://github.com/JuliaManifolds/Manifolds.jl/blob/96fc568bc4dcd6d5e1022828865056e5905b8f77/test/manifolds/euclidean.jl#L56-L61

fails since `_get_basis` does not yet have the right dispatch. What are we missing here and is there maybe also an easier way to construct the basis (that could be added to some doc strings somewhere)?

kellertuer:
Bases and coordinates are just killing me. I now ran into the rolling bug which I hunted for an hour but could not fix

```
using Manifolds, ManifoldsBase
M = Euclidean(2, field=ManifoldsBase.ℂ)
p = [1.0, 1.0]
X = [1.0, 0.0]

B = DiagonalizingOrthonormalBasis([1.0, 2.0, 3.0], ManifoldsBase.ℂ)

b = get_basis(M,p,B)
get_coordinates(M,p,X,B)
get_coordinates(M,p,X,b)
```
first produced different results, the first coordinates were 4 (but a complex manifold with a complex basis should only need 2) the second is correct, ok that was a bug in Euclidean – but for the first still vectors of length 4 get allocated and I can not track down why. Or did I mix up something in my head again what the complex basis means? In this tempo (0 lines per hour) we are on a quite slow road to get this finished - but I try to find an hour a day now.

kellertuer:
As a last thing for today I started a small test code for parallel transport (such that it is one line in the large test suite) since we have/should test this individually. Maybe this is also a good road to get the monster-test method a little slighter if we step by step write singe test functions for the groups. Will continue with this to finish Euclidean (where I hope only PT is missing for now).
Maybe this modularisation could also be a way to introduce more minimalistic tests for some manifolds (for example in PT to test just the first PT and not all from the for loop as it is now sketched), to still test all lines but more minimalistic.

mateuszbaran:
Sorry for the delay, I took an Easter break. Do you still have an issue with bases?

> Maybe this is also a good road to get the monster-test method a little slighter if we step by step write singe test functions for the groups.

This makes sense :+1: .

kellertuer:
Great! 😎 I hope you had a great easter.

It still seems to be the main point to fix for code coverage (also including the false cases where signatures of functions seem to not be covered). This week I might be a little busy with teaching though - I hope to still continue with a few codecov lines.

For the splitting of the test suite, I will do that slowly from time to time to move blocks (that are activated by kwargs) into their own functions and look whether they can be also provided in a reduced form (for example checking one exp/log instead of for all points in the set).

mateuszbaran:
> Great!  I hope you had a great easter.

Thanks! It was quite nice :slightly_smiling_face: .

> It still seems to be the main point to fix for code coverage (also including the false cases where signatures of functions seem to not be covered). This week I might be a little busy with teaching though - I hope to still continue with a few codecov lines.

OK, let me know if you need help with some weird case. I'm not sure if there is anything we can do about signatures not being covered except waiting until it fixes itself. I will try to add some coverage as well but I won't be splitting the test function myself.

kellertuer:
Oh for this PR I also might not start splitting the test function - it is more like a long time goal, splitting and introducing a minimal (time wise, still covering all lines) testset for a faster CI.

mateuszbaran:
There is an annoyingly large number of lines incorrectly marked as not covered. For example, the entire multinomial manifold, or center of orbit for group operation action.

kellertuer:
For whole manifolds we should carefully check whether we accidentally have a default covering that now or something (with the explicit fallbacks and such, I had a few places where we just now had a good default covering that already).

For the function signatures, I already opened a thread at https://discourse.julialang.org/t/codecov-missing-function-signatures-in-code-coverage/79518

mateuszbaran:
Not allowing one-element vectors as points on Circle has caused significant issues with differentiation (FD and Zygote) so I'd prefer to allow it back.

kellertuer:
> Not allowing one-element vectors as points on Circle has caused significant issues with differentiation (FD and Zygote) so I'd prefer to allow it back.

Well, ok.

kellertuer:
I just noticed one thing that is a little inconsistent: For vector transport by projection we do have the global default (i.e. `vector_transport_to_project!` just calls project while `retract_project!` does not. We could introduce the second default as well?

mateuszbaran:
Yes, sure, the default for `retract_project!` makes sense (I think the default should include `embed!` then?)

kellertuer:
I am never that sure about the embed (it is currently not done for the vector transport for example), but sure, we could add that

mateuszbaran:
> (it is currently not done for the vector transport for example)

Embedding is at least done in `vector_transport_to_project!(M::FixedRankMatrices, Y, p, X, q)`.

mateuszbaran:
I don't really like the implicit assumption that `embed` is usually identity and we need to special-case each time it isn't. IMO generic implementations should assume that it is not.

kellertuer:
I agree; Would it be ok to _explicitly_ set `embed!(M,q,p) = (q .=p)` (and similarly for tangent vectors) ? Otherwise changing/adding embed into  would be quite a breaking change.

mateuszbaran:
I think it should be `embed!(M, q, p) = copyto!(M, q, p)` and yes, I think that's a good default to have.

kellertuer:
You're right, I will prepare a short PR on ManifoldsBase.

kellertuer:
I noticed something interesting: In principle, since Grassmann is now IsometricallyEmbedded the check is done in the embedding  (Stiefel), which would mean we can basically remove the checks in Grassmann - but the error message then says that the point is not on Stiefel, which might be misleading. How to best treat this?

mateuszbaran:
> I noticed something interesting: In principle, since Grassmann is now IsometricallyEmbedded the check is done in the embedding (Stiefel), which would mean we can basically remove the checks in Grassmann - but the error message then says that the point is not on Stiefel, which might be misleading. How to best treat this?

I think ideally the `check_point` in Grassmann could expand the message written by check for Stiefel, for example by adding some comment about using the check in embedding. This could be generalized to any embedded manifold.

kellertuer:
But that check point is actually never called. Let me show the route. When you call `is_point` on Grassmann

we get to 

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/d72f79cc5ae7821ebdbc920b313abdf0cd367963/src/decorator_trait.jl#L325-L329

where the check in the embedding is the same the current `check_point` for Grassmann does (so having it only once is nice). But if the check in the embedding fails (line 325) it would also directly throw an error (and for example the sphere would already expect the point to be a valid point in the embedding so that order is the right one).

So the true/false decision is correct, just that the message might be confusing, and with the check in the embedding in 325, there we can not expand the message, because it is thrown within that call.

mateuszbaran:
> So the true/false decision is correct, just that the message might be confusing, and with the check in the embedding in 325, there we can not expand the message, because it is thrown within that call.

Why can't we use `check_point` in line 325 instead?

kellertuer:
Sure we could do that and then throw it if `te` is true, I am just not sure how to (generically at that point) modify the error message precisely: modify it to what?  Something like : `"$p is not a point on $M since it is not a valid point in the embedding, which reported:\n $messageFromLine325"` ?

mateuszbaran:
I think we would need a new error type actually because we don't know what error would we get exactly.
```julia
struct EmbeddingManifoldError{E} <: Exception where {E<:Exception}
    outer_text::String
    error::E
end

function Base.showerror(io::IO, ex::EmbeddingManifoldError)
        print(io, ex.outer_text)
        print(io, ex.error)
end
```
And then in line 325 we could have something like 
```
mpe = check_point(get_embedding(M, p), embed(M, p); kwargs...)
wrapped_error = EmbeddingManifoldError("$p is not a point on $M because it is not a valid point in its embedding: ", mpe)
```

kellertuer:
That sounds good. Let's add that to ManifoldsBase :) - we could maybe call it `ManifoldDomainError`?

mateuszbaran:
OK, `ManifoldDomainError` sounds fine.

kellertuer:
> OK, `ManifoldDomainError` sounds fine.

Started it https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/101 and removed the checks for Grassmann here already, since the only thing that is misleading is the actual error message text then now.

kellertuer:
I hopefully just finished coverage (and added the same style to check size).

kellertuer:
Currently the inner error is printed as
```
[0, 0, 0] is not a tangent vector to [1, 1, 0] on PosQuadrantManifold() because it is not a valid tangent vector in its embedding: DomainError(0, "X[1] ≤ 0")
```
Can this maybe improved / is there a nicer print for errors in strings?

kellertuer:
I started adapting the tests to the new error messages. It seems that in a few points `embed!` dos not work (yet) properly. Will check these and then continue on my alphabetic list with the second representation (PoincareBall) of Hyperbolic.

kellertuer:
(we are at least back to running tests.)

kellertuer:
Ha. Switching to Julia 1.7 _and_ the codecov v2 upload, we got much closer to out goal (basically we have about 50 lines to cover, still).

mateuszbaran:
That's really great! I didn't even know there is a newer codecov upload action.

kellertuer:
I also just saw that more like by accident.

mateuszbaran (approved):
I've also looked at it and I think this can be merged :+1: .

Issue #453, opened 2022-02-11
Towards differentials on SO(n)


const-ae:
I am working on some problem where I would like to implement geodesic regression on Rotation manifolds (SO(n)). Following the [geodesic regression](https://manoptjl.org/stable/tutorials/GeodesicRegression.html) tutorial in manopt, I dove into learning more about differentials of matrix exponentials and read the paper by [Rentmeesters (2011)](https://ieeexplore.ieee.org/abstract/document/6161280) who explains how to form the orthogonal normalizing basis of the Jacobi operator for SO(3). 

However, trying to implement his approach, I don't seem to be able to get the differential right:

```{julia}
function difficult_problem(S)
    Y = random_point(S)
    p = random_point(S)
    v = random_tangent(S, p)
    dir = random_tangent(S, p)
    M = TangentSpace(S, p)

    cost_fnc(arg) = 0.5 * sum((Y - exp(S, p, arg)).^2)
    diff_fnc(arg, v) = -(Y - exp(S, p, arg)) .* differential_exp_argument(S, p, arg, v)
    return check_diff(M, cost_fnc, diff_fnc, p = dir, v = v)
end

difficult_problem(Euclidean(5))
difficult_problem(Sphere(3))
difficult_problem(Rotations(3))
```

---------------

In the above code example I am using `check_diff` function that tests if the Taylor expansion decreases quadrattically following section 4.8 of Boumal (2021).

```{julia}
using Manifolds, Manopt, Random, LinearAlgebra
import Manifolds.embed!
import Manifolds.mul!
import Manifolds.get_basis
using Plots

function get_basis(M::Rotations{3}, p, B::DiagonalizingOrthonormalBasis{ℝ})
    if manifold_dimension(M) != 3
        error("get_basis is only defined for Rotations(3).")
    end
    decomp = schur(B.frame_direction)
    beta = decomp.Schur[3,2]
    U = decomp.vectors
    e1, e2, e3 = [[1,0,0], [0,1,0], [0,0,1]]
    println(sum((B.frame_direction - U * (beta .* (e3 * e2' - e2 * e3') * U')).^2))
    # eigenvectors
    Ξ = [U * (e3 * e2' - e2 * e3') * U',
         U * (e2 * e1' - e1 * e2') * U',
         U * (e3 * e1' - e1 * e3') * U']
    # corresponding eigenvalues
    κ = [0,beta^2 / 4,beta^2 / 4]
[...]

PR #472, opened 2022-05-12
Diagonalizing ONB for rotations

mateuszbaran:
Implementation from #453 (thanks @const-ae!)


mateuszbaran:
The test with `diag_basis_2` sadly fails for SO(4) and SO(5) -- there is an error when constructing the basis and I'm not sure what is wrong.

mateuszbaran:
OK, should be fixed now.

kellertuer (approved):
Looks good, thanks for transferring it :)

PR #475, opened 2022-05-18
Introduce Heisenberg group

mateuszbaran:
A small PR that introduces the Heisenberg group found in other libraries.


kellertuer:
Could you also activate the docs? I would like to see them rendered and think about the names of the two views; the rest looks fine so far (±CI).

mateuszbaran:
Sure, I need to work a bit on the docs anyway.

src/groups/heisenberg.jl
```diff
@@ -0,0 +1,255 @@
...
+```math
+\begin{bmatrix} 1 & \mathbf{a} & c \\
+\mathbf{0} & I_n & \mathbf{b} \\
+0 & \mathbf{0} & 1 \end{bmatrix}
+```
+
+where ``I_n`` is the ``n×n`` unit matrix, ``\mathbf{a}`` is a row vector of length ``n``,
+``\mathbf{b}`` is a column vector of length ``n`` and ``c`` is a real number.
+The group operation is matrix multiplication.
+
```
kellertuer, line 16: Could we maybe add a reference?

src/groups/heisenberg.jl
```diff
@@ -0,0 +1,255 @@
...
+function active_traits(f, ::HeisenbergGroup, args...)
+    return merge_traits(
+        IsGroupManifold(MultiplicationOperation()),
+        IsEmbeddedManifold(),
+        HasLeftInvariantMetric(),
+    )
+end
+
+function _heisenberg_a_view(::HeisenbergGroup{n}, p) where {n}
```
kellertuer, line 29: I do not like this name much, do we maybe have a more semantic name? It is an internal function so maybe this is als not so urgent.

src/groups/heisenberg.jl
kellertuer: The embedding could be mentioned here.

src/groups/heisenberg.jl
```diff
@@ -0,0 +1,255 @@
...
+) where {n}
+    Xⁱ[1:n] .= _heisenberg_a_view(M, X)
+    Xⁱ[(n + 1):(2 * n)] .= _heisenberg_b_view(M, X)
+    Xⁱ[2 * n + 1] = X[1, n + 2]
+    return Xⁱ
+end
+
+get_embedding(::HeisenbergGroup{n}) where {n} = Euclidean(n + 2, n + 2)
+
+function get_vector_orthonormal(::HeisenbergGroup{n}, p, Xⁱ, ::RealNumbers) where {n}
```
kellertuer, line 131: the coordinates function of this was very nicely documented, could we do the same here?

src/groups/heisenberg.jl
```diff
@@ -0,0 +1,255 @@
...
+    q[2:(n + 1), n + 2] .= b_view
+    q[1, n + 2] = X[1, n + 2] + dot(a_view, b_view) / 2
+    return q
+end
+
+@doc raw"""
+    exp(M::HeisenbergGroup, p, X)
+
+Exponential map on the [`HeisenbergGroup`](@ref) `M` with the left-invariant metric.
+"""
```
kellertuer, line 161: can we maybe add a formula here? I think that would be nice.

src/groups/heisenberg.jl
```diff
@@ -0,0 +1,255 @@
...
+    return dot(X_a_view, Y_a_view) +
+           dot(X_b_view, Y_b_view) +
+           X[1, 2 * n + 1] * Y[1, 2 * n + 1]
+end
+
+@doc raw"""
+    log(G::HeisenbergGroup, p, q)
+
+Compute the logarithmic map on the [`HeisenbergGroup`](@ref) group.
+
```
kellertuer, line 190: could we maybe add a formula here?

src/groups/heisenberg.jl
```diff
@@ -0,0 +1,255 @@
...
+parallel_transport_to!(::HeisenbergGroup, Y, p, X, q) = copyto!(Y, X)
+
+function project(M::HeisenbergGroup{n}, p) where {n}
+    return [
+        1 p[1, 2:(n + 2)]'
+        zeros(n, 1) Matrix(I, n, n) _heisenberg_b_view(M, p)
+        zeros(1, n + 1) 1
+    ]
+end
+function project(M::HeisenbergGroup{n}, p, X) where {n}
```
kellertuer, line 227: can we maybe add doc strings here? I assume this is just projection with the Euclidean metric from the embedding, i.e. setting all necessary things to zero/one?

kellertuer: same for the project above :)

kellertuer (commented):
CI and code in general look very good already, I just had a few comments on the docs.

src/groups/heisenberg.jl
mateuszbaran: OK, I'll add that.

src/groups/heisenberg.jl
mateuszbaran: It's a function that makes a view to the "a" part of Heisenberg point/vector matrix. "a" part as in the notation used in docstring. I can change that if you have a better name.

src/groups/heisenberg.jl
mateuszbaran: I'm not exactly sure what to write here. I took the standard definition for exponential at identity and extended it through left-invariance (hence `p \ q`).

kellertuer: I see that in the code, but also the lie exp is not documented, so I would love to have a formula here to at exp_lie

src/groups/heisenberg.jl
mateuszbaran: Done.

src/groups/heisenberg.jl
mateuszbaran: I've added a formula.

mateuszbaran:
I think I've addressed all issues :slightly_smiling_face: .

kellertuer (approved):

PR #481, opened 2022-05-26
Projections change representation

mateuszbaran:
This PR fixes rotations manifold so that projections actually change the representation of tangent vectors. Also a couple of related changes.

### DIFF
--- a/src/manifolds/Circle.jl
+++ b/src/manifolds/Circle.jl
@@ -106,6 +106,19 @@ distance(::Circle{ℝ}, p, q) = abs(sum(sym_rem.(p - q)))
 distance(::Circle{ℂ}, p, q) = acos(clamp(complex_dot(p, q), -1, 1))
 
 @doc raw"""
+    project(M::Circle, p)
+
+Embed a point `p` on [`Circle`](@ref) `M` in the ambient space. It returns `p`.
+"""
+embed(::Circle, p) = p
+@doc raw"""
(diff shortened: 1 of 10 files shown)


src/manifolds/Circle.jl
```diff
@@ -105,6 +105,19 @@ distance(::Circle{ℝ}, p::Real, q::Real) = abs(sym_rem(p - q))
...
+
+Embed a point `p` on [`Circle`](@ref) `M` in the ambient space. It returns `p`.
+"""
+embed(::Circle, p) = p
+@doc raw"""
+    embed(M::Circle, p, X)
+
+Embed a tangent vector `X` at `p` on [`Circle`](@ref) `M` in the ambient space. It returns `X`.
+"""
+embed(::Circle, p, X) = X
```
kellertuer, line 119: I thought this was the default, but if it is not, then sure this line is correct.

kellertuer (commented):
Besides one small question this looks fine – thanks for spotting my bugs :)

src/manifolds/Circle.jl
mateuszbaran: No, the default is to call the mutating `embed!` and it fails for `Circle` when it uses immutable numbers.

kellertuer: ah, right!

kellertuer (approved):

Issue #483, opened 2022-05-30
DomainError with ExponentialRetraction with `mean` of SE2


Affie:
Hi, I'm upgrading code to v0.8 and get a DomainError, here is a minimum example:
```julia
julia> using Manifolds

julia> p1 = ProductRepr([0.0,0.0], [1.0 0.0; 0.0 1.0])
ProductRepr with 2 submanifold components:
 Component 1 =
  2-element Vector{Float64}:
   0.0
   0.0
 Component 2 =
  2×2 Matrix{Float64}:
   1.0  0.0
   0.0  1.0

julia> p2 = ProductRepr([0.0,0.0], [0.0 -1.0; 1.0 0.0])
ProductRepr with 2 submanifold components:
 Component 1 =
  2-element Vector{Float64}:
   0.0
   0.0
 Component 2 =
  2×2 Matrix{Float64}:
   0.0  -1.0
   1.0   0.0

julia> M = SpecialEuclidean(2)
SpecialEuclidean(2)

julia> mean(M, [p1,p2])
ERROR: DomainError with ExponentialRetraction():
You can not use the exponential map as an inner method to solve the ode for the exponential map.
Stacktrace:
  [1] ODEExponentialRetraction(r::ExponentialRetraction, #unused#::DefaultOrthonormalBasis{ℝ, ManifoldsBase.TangentSpaceType})
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/L10sL/src/retractions.jl:93
  [2] ODEExponentialRetraction(r::ExponentialRetraction)
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/L10sL/src/retractions.jl:82
[...]

PR #486, opened 2022-05-31
Propagation of statistics functions through group decorator

mateuszbaran:
Will fix another part of #483


kellertuer (approved):
Looks good to me.

Issue #489, opened 2022-06-03
Difference in vee for v0.7 and v0.8 on RealCircleGroup


Affie:
Hi, I'm upgrading code in IncrementalInference to v0.8.
We use `RealCircleGroup` in one of the tests and the upgrade broke some code that I traced back to a change in behavior in `vee`.

In v0.7 we get a `Vector` back
```julia
julia> p = [0.0]
1-element Vector{Float64}:
 0.0

julia> X = [0.1]
1-element Vector{Float64}:
 0.1

julia> M = RealCircleGroup()
RealCircleGroup()

julia> vee(M, p, X)
1-element Vector{Float64}:
 0.1
```

while in v0.8 we get an `SVector` back that I suspect leads to the stack trace lower down.
```julia
julia> using Manifolds

julia> p = [0.0]
1-element Vector{Float64}:
 0.0

julia> X = [0.1]
1-element Vector{Float64}:
 0.1

julia> M = RealCircleGroup()
RealCircleGroup()

julia> vee(M, p, X)
1-element SVector{1, Float64} with indices SOneTo(1):
 0.1
```

Unfortunately, we need to mutate the result, we need to upgrade it to support immutable types, but for now, everything is built like that.

Do you have any advice on upgrading?

Here is a part of the stack trace, unfortunately, it didn't help a lot. 
```
 setindex!(::SVector{1, Float64}, value, ::Int) is not defined.
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:33
    [2] setindex!(a::SVector{1, Float64}, value::Float64, i::Int64)
      @ StaticArrays ~/.julia/packages/StaticArrays/58yy1/src/indexing.jl:3
    [3] macro expansion
      @ ~/.julia/packages/StaticArrays/58yy1/src/arraymath.jl:105 [inlined]
    [4] _fill!
      @ ~/.julia/packages/StaticArrays/58yy1/src/arraymath.jl:100 [inlined]
    [5] fill!
      @ ~/.julia/packages/StaticArrays/58yy1/src/arraymath.jl:99 [inlined]
    [6] x_of_nans(x::SVector{1, Float64}, Tf::Type{Float64})
      @ NLSolversBase ~/.julia/packages/NLSolversBase/cfJrN/src/NLSolversBase.jl:60
    [7] alloc_DF(x::SVector{1, Float64}, F::Float64)
[...]

PR #490, opened 2022-06-06
Fix issue #489

mateuszbaran:


test/groups/circle_group.jl
```diff
@@ -1,6 +1,10 @@
 include("../utils.jl")
 include("group_utils.jl")
 
+# TODO: remove after bug in StaticArray is fixed
+@inline Base.copy(a::SizedArray) = __copy(a)
+@inline __copy(a::SizedArray{S,T}) where {S,T} = SizedArray{S,T}(copy(a.data))
```
kellertuer, line 6: just a short question - why the two `_`?

kellertuer (approved):

test/groups/circle_group.jl
mateuszbaran: No real reason, I just felt like it's less likely to extend a function we already have somewhere else (but most likely don't).

kellertuer: Ok, well we could also just use `_copy` - just for clarity (others might also wonder why we use `__` here; but I leave this up to you :)

mateuszbaran: I'm going to delete it when https://github.com/JuliaArrays/StaticArrays.jl/pull/1037 is merged anyway so it's just a temporary fix.

PR #492, opened 2022-06-08
Special case exp_lie and log_lie for SO(2)

mateuszbaran:
Just to make it faster and more robust.

### DIFF
--- a/src/groups/special_orthogonal.jl
+++ b/src/groups/special_orthogonal.jl
@@ -77,6 +77,10 @@ function log_lie!(::SpecialOrthogonal{2}, X, p)
     end
     return X
 end
+function log_lie!(::SpecialOrthogonal{2}, X, ::Identity{MultiplicationOperation})
+    fill!(X, 0)
+    return X
+end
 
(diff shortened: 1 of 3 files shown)


src/groups/special_orthogonal.jl
```diff
@@ -44,6 +44,13 @@ function allocate_result(
     return allocate(q)
 end
 
+function exp_lie!(::SpecialOrthogonal{2}, q, X)
```
kellertuer, line 47: I think we usually have 3 arguments here? `exp!(M, q, p, X)` in order to have a different result variable (`q`) compared to the start `p`? Or did I mix this up and `pis the identity here anyways?

mateuszbaran: In `lie` variants of methods `p` is the identity. `exp_lie!(M, q, X)` is like `exp!(M, q, identity_element(M), X)`.

kellertuer: Ok, thanks for the reminder :)

kellertuer (approved):

PR #493, opened 2022-06-11
Polishing SE(n)

mateuszbaran:
Mostly intended to address https://github.com/JuliaManifolds/Manopt.jl/discussions/90#discussioncomment-2913543 .


src/groups/metric.jl
```diff
@@ -84,51 +84,51 @@ function exp!(::TraitList{HasBiinvariantMetric}, M::MetricManifold, q, p, X)
     return exp!(M.manifold, q, p, X)
 end
 
-function get_coordinates_orthonormal(
+function get_coordinates(
     t::TraitList{IT},
-    M::AbstractDecoratorManifold,
+    M::MetricManifold,
```
kellertuer, line 89: just a small question on this – so this would no longer work for a manifold whose default metric is invariant (so it has an invariance trait), which was covered by the implementation before?
(I am not sure we have that case though ;)). If possible I‘d prefer using `AbstractDecoratormanifold` again.

mateuszbaran: This is very convoluted and I don't have a good idea that wouldn't require days of work. The essence of the issue is that sometimes invariance traits are used as "and by the way, this metric is invariant" and sometimes as "disregard the previous metric and use the general formulas for these specific ideas for invariant metrics". This `MetricManifold` restriction is used to mean that this method is to be used only in the second case. I've also tried making these methods the defaults for the lowest level but then it was skipped for `MetricManifold` due to, IIRC, trait ordering.

Feel free to experiment with it but I would prefer to touch it as little as possible because this default is not even useful right now and I was even thinking about deleting it when it didn't want to work.

kellertuer: You’re right we were maybe not that careful with this trait, I was mainly thinking that it was more general before, but if you narrowed it down to avoid the problems you mentioned, then let’s keep it that way. For now I also do both have a good design to distinguish the two cases you mention (but at some point we should better distinguish both).

kellertuer (approved):
Nice.

mateuszbaran:
Well, I understood that the chained oplus operator that I've implemented as a "metric" does not come from a Riemannian metric (and I'm fairly sure it doesn't come from an affine connection either), so I've removed that part.

mateuszbaran:
Anyway, this should be complete now :).

kellertuer:
Thanks for explaining, that point I missed when I approved this, so great that it is fixed now :)

Issue #497, opened 2022-07-04
Power Manifolds usage and bugs


Affie:
Hello,
As Mateusz sugested in https://github.com/JuliaManifolds/Manifolds.jl/issues/495#issuecomment-1172956048
I experimented with using power manifolds to represent points for optimization.

From the documentation, I figured the `NestedReplacingPowerRepresentation` with an `ArrayPartition` of `<:StaticArrays` looked the most promising. So I tried it and ran into errors, so I tried  `NestedPowerRepresentation` with `ArrayPartition` and again errors. 
I completely understand Manifolds.jl is still under development so I tried to help by writing a simple script for testing which combination to try.

I ended up using  "ProductRepr SA NestedReplacingPowerRepresentation" and I got it to work. 

1. What is your recommendation of what `PowerManifold` to use? So we can focus on implementing it and fixing bugs as they come up.
2. What is your recommendation between `ProductRepr` and `ArrayPartition`? 
    I think you already answered it here (so linking for the next person): [There are almost no drawbacks to ArrayPartition vs ProductRepr](https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1546#issuecomment-1173605410)
3. How do I do group operations such as, `identity_element`,`vee`,`hat`
4. I think a PowerManifold dispatch such as `log(M,  Identity(M), p)` where all points are calculated from the identity element can be quite handy, I currently create a vector of identity elements and use that. 

Thanks again for the support. I tried to fix the errors myself, but couldn't quite figure out the trait macros.

Here is the script in case you find it handy:

```julia
using Manifolds
using RecursiveArrayTools: ArrayPartition
using StaticArrays
using Test
using PreallocationTools

G = SpecialEuclidean(2)

# the different options
options = []
#ProductRepr
o = (name = "ProductRepr NestedPowerRepresentation",
     p1 = ProductRepr([0.0,0.0], [1.0 0.0; 0.0 1.0]),
     p2 = ProductRepr([0.0,0.0], [0.0 -1.0; 1.0 0.0]),
     M = PowerManifold(G, NestedPowerRepresentation(), 2));
push!(options, o);

#ArrayPartition
o = (name = "ArrayPartition NestedPowerRepresentation",
     p1 = ArrayPartition([0.0,0.0], [1.0 0.0; 0.0 1.0]),
     p2 = ArrayPartition([0.0,0.0], [0.0 -1.0; 1.0 0.0]),
[...]

PR #498, opened 2022-07-05
Fixing some issues from #497

mateuszbaran:


kellertuer (commented):
Looks good to me. Do you want to dress the Lie part here as well?

mateuszbaran:
I'm going to do Lie parts separately but here I still would like to finish the power manifold - product manifold interactions (it's not fully working yet).

mateuszbaran:
And I have one significant design issue -- `allocate` will have to get the manifold argument as well, or at least I don't see a way around it. When you have nested non-replacing power manifold with ProductRepr-of-SArray elements, allocate needs to covert it to ProductRepr-of-MArray but for nested replacing it should stay ProductRepr-of-SArray. So that's one more thing to do.

kellertuer:
For me that is fine at least for some allocate cases to have the manifold as first argument :)

mateuszbaran:
Yes, manifold definitely won't be mandatory.

mateuszbaran:
OK, that's all I want for this PR. I will improve allocation in a separate PR to not block the `view` implementation for power manifold added here.

kellertuer (approved):
My only overall remark would be, that in the long run we could extract the `SArray` specialisations into a separate `manifolds_sarray` or `Euclidean_SArray.jl` file (like we do for different metrics or representations as well).

But we do not have to do that in this PR :)

PR #499, opened 2022-07-06
Add unitary and orthogonal manifolds and Lie groups

kellertuer:
super seeds (and adapts) #345).

Besides adapting directlyy to the ManifoldsBase.jl 0.13 interface, this PR has one difference in that it distinguishes between unitary (orthogonal) matrices and the corresponding groups. Similarly to distinguishing Rotations and the Special Orthogonal group.

The rest will slowly be adapted here from the previous PR.

### DIFF
--- a/docs/src/misc/notation.md
+++ b/docs/src/misc/notation.md
@@ -13,6 +13,7 @@ Within the documented functions, the utf8 symbols are used whenever possible, as
 | ``\operatorname{Ad}_p(X)`` | adjoint action of element ``p`` of a Lie group on the element ``X`` of the corresponding Lie algebra | | |
 | ``\times`` | Cartesian product of two manifolds | | see [`ProductManifold`](@ref) |
 | ``^{\wedge}`` | (n-ary) Cartesian power of a manifold | | see [`PowerManifold`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/manifolds.html#ManifoldsBase.PowerManifold) |
+| ``\cdot^\mathrm{H}`` | conjugate/Hermitian transposed | |
 | ``a`` | coordinates of a point in a chart | | see [`get_parameters`](@ref) |
 | ``\frac{\mathrm{D}}{\mathrm{d}t}`` | covariant derivative of a vector field ``X(t)`` | | |
 | ``T^*_p \mathcal M`` | the cotangent space at ``p`` | | |
--- a/src/manifolds/Orthogonal.jl
+++ b/src/manifolds/Orthogonal.jl
@@ -1,0 +1,13 @@
+@doc raw"""
+     OrthogonalMatrices{n}
+
+The manifold of (real) orthogonal matrices ``\mathrm{O}(n)``.
+This is the special case of {`UnitaryMatrices`}(@ref) over the reals.
+
+    OrthogonalMatrices(n)
+
+The constructor is equivalent to calling [`Unitary(n,ℝ)`](@ref).
+"""
+const OrthogonalMatrices{n} = UnitaryMatrices{n,ℝ}
+
+OrthogonalMatrices(n) = OrthogonalMatrices{n}()
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -89,8 +89,7 @@ end
     check_vector(M, p, X; kwargs... )
 
 Check whether `X` is a tangent vector to `p` on the [`Rotations`](@ref)
-space `M`, i.e. after [`check_point`](@ref)`(M,p)`, `X` has to be of same
-dimension and orthogonal to `p`.
+space `M`, i.e. after [`check_point`](@ref)`(M,p)`, `X` has to be skew symmetric.
 The tolerance for the last test can be set using the `kwargs...`.
 """
 function check_vector(M::Rotations{N}, p, X; kwargs...) where {N}
@@ -136,10 +135,9 @@ By convention, the returned values are sorted in increasing order. See
 [`angles_4d_skew_sym_matrix`](@ref).
 """
 function cos_angles_4d_rotation_matrix(R)
-    trR = tr(R)
-    a = trR / 4
-    b = sqrt(clamp(tr((R .- transpose(R))^2) / 16 - a^2 + 1, 0, Inf))
-    return (a + b, a - b)
+    a = tr(R)
+    b = sqrt(clamp(2 * dot(transpose(R), R) - a^2 + 8, 0, Inf))
+    return ((a + b) / 4, (a - b) / 4)
 end
 
 @doc raw"""
--- a/src/manifolds/Unitary.jl
+++ b/src/manifolds/Unitary.jl
@@ -1,0 +1,132 @@
+@doc raw"""
+    UnitaryMatrices{n,𝔽} <: AbstractManifold{𝔽}
+
+The manifold ``U(n)`` of ``n×n`` matrices over the field ``\mathbb F``, ``\mathbb C`` by
+default, such that
+
+``p^{\mathrm{H}p = \mathrm{I}_n,``
+
+where ``\mathrm{I}_n`` is the ``n×n`` identity matrix.
+An alternative characterisation is that ``\lVert \det(p) \rVert = 1``.
+
+The tangent spaces are given by
+
+```math
+    T_pU(n) \coloneqq \bigl\{
+    X \big| pY \text{ where } Y \text{ is skew symmetric, i. e. } Y = -Y^{\mathrm{H}}
+    \bigr\}
+```
+
+But note that tangent vectors are represented in the Lie algebra, i.e. just using ``Y`` in the representation above.
+
+# Constructor
+     Unitary(n, 𝔽=ℂ)
+
+Constructs ``\mathrm{U}(n, 𝔽)``, see also [`OrthogonalMatrices(n)`](@ref) for the real-valued case.
+"""
+struct UnitaryMatrices{n,𝔽} <: AbstractManifold{𝔽} end
+
+UnitaryMatrices(n::Int, field=ℂ) = UnitaryMatrices{n,field}()
+
+function active_traits(f, ::UnitaryMatrices, args...)
+    return merge_traits(IsIsometricEmbeddedManifold(), IsDefaultMetric(EuclideanMetric()))
+end
+
+function allocation_promotion_function(::UnitaryMatrices{n,ℂ}, f, ::Tuple) where {n}
+    return complex
+end
+
+@doc raw"""
+    check_point(M, p; kwargs...)
+
+Check whether `p` is a valid point on the [`UnitaryMatrices`](@ref) `M`,
+i.e. that ``p`` has an determinante of absolute value one, i.e. that ``p^{\mathrm{H}}p``
+
+The tolerance for the last test can be set using the `kwargs...`.
+"""
+function check_point(M::UnitaryMatrices{n,𝔽}, p; kwargs...) where {n,𝔽}
+    if !isapprox(abs(det(p)), 1; kwargs...)
+        return DomainError(
+            det(p),
+            "The absolute value of the determinant of $p has to be 1 but it is $(abs(det(p)))",
+        )
+    end
+    if !isapprox(p' * p, one(p); kwargs...)
+        return DomainError(
+            norm(p' * p - one(p)),
+            "$p must be unitary but it's not at kwargs $kwargs",
+        )
+    end
+    return nothing
+end
+
+@doc raw"""
+    check_vector(M::UnitaryMatrices{n,𝔽}, p, X; kwargs... )
+
+Check whether `X` is a tangent vector to `p` on the [`UnitaryMatrices`](@ref)
+space `M`, i.e. after [`check_point`](@ref)`(M,p)`, `X` has to be skew symmetric (hermitian)
+dimension and orthogonal to `p`.
+
+The tolerance for the last test can be set using the `kwargs...`.
+"""
+function check_vector(M::UnitaryMatrices{n,𝔽}, p, X; kwargs...) where {n,𝔽}
+    return check_point(SkewHermitianMatrices(n, 𝔽), X; kwargs...)
+end
+
+@doc raw"""
+    embed(M::Unitary{n,𝔽}, p, X)
+
+Embed the tangent vector `X` at point `p` in `M` from
+its Lie algebra representation (set of skew matrices) into the
+Riemannian submanifold representation
+
+The formula reads
+```math
+X_{\text{embedded}} = p * X
+```
+"""
+embed(::UnitaryMatrices, p, X)
+
+function embed!(::UnitaryMatrices, Y, p, X)
+    return mul!(Y, p, X)
+end
+
+get_embedding(::UnitaryMatrices{n,𝔽}) where {n,𝔽} = Euclidean(n, n; field=𝔽)
+
+@doc raw"""
+     project(G::UnitaryMatrices{n,𝔽}, p)
+
+Project the point ``p ∈ 𝔽^{n × n}`` to the nearest point in
+``\mathrm{U}(n,𝔽)=``[`Unitary(n,𝔽)`](@ref) under the Frobenius norm.
+If ``p = U S V^\mathrm{H}`` is the singular value decomposition of ``p``, then the projection
+is
+
+````math
+  \operatorname{proj}_{\mathrm{U}(n,𝔽)} \colon p ↦ U V^\mathrm{H}.
+````
+"""
+project(::UnitaryMatrices, p)
+
+function project!(::UnitaryMatrices, q, p)
+    F = svd(p)
+    mul!(q, F.U, F.Vt)
+    return q
+end
+
+@doc raw"""
+     project(G::Unitary{n,𝔽}, p, X)
+
+Orthogonally project the tangent vector ``X ∈ 𝔽^{n × n}`` to the tangent space of
+[`UnitaryMatrices`](@ref)`(n,𝔽)` at ``p``,
+and change the representer to use the Lie algebra ``\mathfrak u(n, \mathbb F)``.
+The projection removes the Hermitian part of ``X``:
+```math
+  \operatorname{proj}_{p}(X) := \frac{1}{2}(X - X^\mathrm{H}).
+```
+"""
+project(::UnitaryMatrices, p, X)
+
+function project!(::UnitaryMatrices{n,𝔽}, Y, p, X) where {n,𝔽}
+    project!(SkewHermitianMatrices(n, 𝔽), Y, X)
+    return Y
+end
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -345,9 +345,11 @@ include("manifolds/SymmetricPositiveDefiniteLinearAffine.jl")
 include("manifolds/SymmetricPositiveDefiniteLogCholesky.jl")
 include("manifolds/SymmetricPositiveDefiniteLogEuclidean.jl")
 include("manifolds/SymmetricPositiveSemidefiniteFixedRank.jl")
-include("manifolds/Tucker.jl")
 include("manifolds/Symplectic.jl")
 include("manifolds/SymplecticStiefel.jl")
+include("manifolds/Tucker.jl")
+include("manifolds/Unitary.jl")
+include("manifolds/Orthogonal.jl")
 
 # Product or power based manifolds
 include("manifolds/Torus.jl")
@@ -374,6 +376,10 @@ include("groups/semidirect_product_group.jl")
 include("groups/general_linear.jl")
 include("groups/special_linear.jl")
 include("groups/translation_group.jl")
+# TODO
+# include("groups/unitary.jl")
+# include("groups/orthogonal.jl")
+# include("groups/special_unitary.jl")
 include("groups/special_orthogonal.jl")
 include("groups/circle_group.jl")
 include("groups/heisenberg.jl")
@@ -467,6 +473,7 @@ export Euclidean,
     MultinomialMatrices,
     MultinomialSymmetric,
     Oblique,
+    OrthogonalMatrices,
     PositiveArrays,
     PositiveMatrices,
     PositiveNumbers,
@@ -695,6 +702,7 @@ export AbstractGroupAction,
     LeftAction,
     LeftInvariantMetric,
     MultiplicationOperation,
+    OrthogonalMatrices,
     ProductGroup,
     ProductOperation,
     RealCircleGroup,
@@ -705,8 +713,8 @@ export AbstractGroupAction,
     SpecialEuclidean,
     SpecialLinear,
     SpecialOrthogonal,
-    TranslationGroup,
-    TranslationAction
+    SpecialUnitary
+TranslationGroup, TranslationAction, UnitaryMatrices
 export AbstractInvarianceTrait
 export IsMetricManifold, IsConnectionManifold
 export IsGroupManifold,
--- a/src/groups/group.jl
+++ b/src/groups/group.jl
@@ -204,6 +204,8 @@ function identity_element(::TraitList{<:IsGroupManifold}, G::AbstractDecoratorMa
     return identity_element!(BG, q)
 end
 
+Base.adjoint(e::Identity) = e
+
 function check_size(
     ::TraitList{<:IsGroupManifold{O}},
     M::AbstractDecoratorManifold,
@@ -541,7 +543,7 @@ function compose!(
     return e
 end
 
-transpose(e::Identity) = e
+Base.transpose(e::Identity) = e
 
 @doc raw"""
     hat(M::AbstractDecoratorManifold{𝔽,O}, ::Identity{O}, Xⁱ) where {𝔽,O<:AbstractGroupOperation}


kellertuer:
I just worked a little on first unifying the code since we basically have 4 scenarios that are very similar, that is on the manifold side
* Orthogonal matrices
* unitary matrices
* orthogonal and determinant +1 also known as rotations
* unitary and determinant +1 (does not have a name)

and similarly on the Group stage (all with multiplication operation) O(n), U(n), SO(n), SU(n).

This does not yet fully work, but I think this way we have mostly one implementation for all four, sometimes two (real/complex) on both the manifold as well as the group level.

Including the group type into the group manifold would be a really major redesign I think, even breaking, I fear.

mateuszbaran:
> Including the group type into the group manifold would be a really major redesign I think, even breaking, I fear.

No, I wouldn't touch the `GroupManifold` type, just change `abstract type AbstractUnitaryMatrices{n,𝔽,S} <: AbstractDecoratorManifold{𝔽} end` to `abstract type AbstractUnitaryMatrices{n,𝔽,S<:AbstractMatrixType,LF<:AbstractPreservedNorm} <: AbstractDecoratorManifold{𝔽} end`
where
```julia
abstract type AbstractPreservedNorm end
struct SymmetricPreservedNorm <: AbstractPreservedNorm end
struct SkewSymmetricPreservedNorm <: AbstractPreservedNorm end
struct HermitianPreservedNorm <: AbstractPreservedNorm end
struct SkewHermitianPreservedNorm <: AbstractPreservedNorm end
```
Then for example `const UnitaryMatrices{n} = AbstractUnitaryMatrices{n,ℂ,AbsoluteDeterminantOneMatrices,SkewSymmetricPreservedNorm}`

kellertuer:
Interesting, I am not yet sure what this parameter is for but the way you propose, we would actually not have any change in my current code, right.

kellertuer:
Ah no, the unitary matrices are currently not a const since the abstract type is actually an abstract type, but I am thinking about maybe changing that.

kellertuer:
> > Including the group type into the group manifold would be a really major redesign I think, even breaking, I fear.
> 
> No, I wouldn't touch the `GroupManifold` type, just change `abstract type AbstractUnitaryMatrices{n,𝔽,S} <: AbstractDecoratorManifold{𝔽} end` to `abstract type AbstractUnitaryMatrices{n,𝔽,S<:AbstractMatrixType,LF<:AbstractPreservedNorm} <: AbstractDecoratorManifold{𝔽} end` where
> 
> ```julia
> abstract type AbstractPreservedNorm end
> struct SymmetricPreservedNorm <: AbstractPreservedNorm end
> struct SkewSymmetricPreservedNorm <: AbstractPreservedNorm end
> struct HermitianPreservedNorm <: AbstractPreservedNorm end
> struct SkewHermitianPreservedNorm <: AbstractPreservedNorm end
> ```
> 
> Then for example `const UnitaryMatrices{n} = AbstractUnitaryMatrices{n,ℂ,AbsoluteDeterminantOneMatrices,SkewSymmetricPreservedNorm}`

For these I am not sure where we see/need/dispatch on them, so I might first transfer what Seth has done before thinking about another additional parameter (and why we would need it).

mateuszbaran:
> For these I am not sure where we see/need/dispatch on them, so I might first transfer what Seth has done before thinking about another additional parameter (and why we would need it).

Maybe that's not the best design (since not all combinations of field and preserved norm make sense) but we need some way to differentiate complex orthogonal and unitary groups, as I've written above.

kellertuer:
> but we need some way to differentiate complex orthogonal and unitary groups, as I've written above.

I do not understand this, where did you write this, where do we have that and how is this solved with the additional parameter?

For me, first transferring Seths implementation (unifying them with the three parameters) is something I can do, while I do not yet understand the fourth nor why and where and how we dispatch on this, for example among the given implementations, where this is not used, so it would be a completely unused parameter for now. Sure we can invent a plethora of parameters, but they might confuse the end user if they are not used (yet).

mateuszbaran:
> I do not understand this, where did you write this, where do we have that and how is this solved with the additional parameter?

Well, I see now that was a bad idea. Here is a better one. Let's have `abstract type AbstractOrthogonalMatrices{n,𝔽,S} <: AbstractDecoratorManifold{𝔽} end` and `abstract type AbstractUnitaryMatrices{n,𝔽,S} <: AbstractDecoratorManifold{𝔽} end`. The first one has a condition `p^Tp = \mathrm{I}_n` (real transposition, regardless of the field) and the second one has `p^{\mathrm{H}p = \mathrm{I}_n,` (Hermitian adjoint). That's not much different from current design and is more future-proof for additional groups.

kellertuer:
> > I do not understand this, where did you write this, where do we have that and how is this solved with the additional parameter?
> 
> Well, I see now that was a bad idea. Here is a better one. Let's have `abstract type AbstractOrthogonalMatrices{n,𝔽,S} <: AbstractDecoratorManifold{𝔽} end` and `abstract type AbstractUnitaryMatrices{n,𝔽,S} <: AbstractDecoratorManifold{𝔽} end`. The first one has a condition `p^Tp = \mathrm{I}_n` (real transposition, regardless of the field) and the second one has `p^{\mathrm{H}p = \mathrm{I}_n,` (Hermitian adjoint). That's not much different from current design and is more future-proof for additional groups.

Hm, I might stop here until we have the type system figured out. Because until now I used consts, but this here is Acton more like towards subtypes. And I am still not sure why we distinguish both, how much that would change what I am currently doing, so to not do it double (its a lot of work trying to unify 4 manifolds into a common code base) - please clarify why we should now again switch from consts to subtypes and which manifolds you aim to do here additionally.

mateuszbaran:
> Hm, I might stop here until we have the type system figured out. Because until now I used consts, but this here is Acton more like towards subtypes. And I am still not sure why we distinguish both, how much that would change what I am currently doing, so to not do it double (its a lot of work trying to unify 4 manifolds into a common code base) - please clarify why we should now again switch from consts to subtypes and which manifolds you aim to do here additionally.

That was a sketch how it would work for the current implementation, and I did not take into account the switch to `const`s. The key is to decouple field and conjugation type to be able to add, for example, complex orthogonal group in the future (see here: https://mathworld.wolfram.com/OrthogonalGroup.html, below Eq. (5)).

kellertuer:
I am not sure (yet) how to include this case. In my current model, this does not fit (which just means one has to implement a few things additionally), but in general it would mean to have two fields F1, F2, where F1 specifies the field type for the matrices and F2 whether its real (transpose) or complex (Hermitian) orthogonality. One could also come up with specific types instead of F2, for sure.

I think this is not too hard to even add later, since the current one would just be F1=F2, or - one would just reimplement a few things for complex matrices with transpose orthogonality (which is the one case currently not covered)

kellertuer:
The manifolds should be unified, the next would be groups (and including (special) unitary) functions from Seths branch.

Thinking about the one case (complex matrices real orthogonality) I think this is really just one case we miss (unless we include Quarternions), so maybe that one as a speparate manifold is for now the easier way to go (compared to a few days pondering about how to include that).

Once we do unitarity on quaternions and have more mixed cases, then we can still remodel this, it is an internal modelling anyways when the user only sees `Orthogonal(n)` and such on the outside.

mateuszbaran:
I will think about it. The meaning of `Orthogonal(n)` and `Unitary(n)` definitely will not change but let's consider `GeneralUnitaryMultiplicationGroup` internal for now.

kellertuer:
I have not yet reduced code duplication on the group level but I just finished going through all code that was on the old PR and transferred them (a few I could leave out, since that was just old-dispatch-overhead).

I will check for
* [ ] duplilcation of code
* [ ] tests
* [ ] documentation

kellertuer:
@mateuszbaran I am not sure what I did wrong, since I did not change the SO(3), but after merging the power group (from master) the tests here do fail.

src/groups/special_unitary.jl
```diff
@@ -0,0 +1,74 @@
...
+        fill!(d, 1)
+        d[n] = conj(detUVt)
+        mul!(q, F.U, Diagonal(d) * F.Vt)
+    else
+        mul!(q, F.U, F.Vt)
+    end
+    return q
+end
+function project!(G::SpecialUnitary{n}, Y, p, X) where {n}
+    inverse_translate_diff!(G, Y, p, p, X, LeftAction())
```
kellertuer, line 67: @sethaxen I am stuck in testing here since `inverse_translate_diff!`passes over to `translate_diff!` and while that has a default implementation for addition group, it does not have that for multiplication groups. Could you help out?

mateuszbaran:
> @mateuszbaran I am not sure what I did wrong, since I did not change the SO(3), but after merging the power group (from master) the tests here do fail.

The rand issue is caused by SO(n) no longer being a GroupManifold. Previously it dispatched into these: https://github.com/JuliaManifolds/Manifolds.jl/pull/500/files#diff-18d2fad9162b22cb1150f7e366667dadae6ff9731da84ac82d8834cef1fa0958R91-R99 and now you need to do something similar for `GeneralUnitaryMultiplicationGroup`.

kellertuer:
Thanks for directly checking,  I will introduce that then (changed a few other places, but used the group manifold trait then, which will not work here).

src/groups/special_unitary.jl
kellertuer: So as far as I read, this should also be the identlty, in general I might also consider writing a fe more details into the doc strings at some point. Not all group functions are documented with their formulae.

src/groups/orthogonal.jl
```diff
@@ -0,0 +1,154 @@
...
+        A² = Symmetric((q - I) ./ 2)
+        P = eigvecs(A²)
+        E = similar(q)
+        fill!(E, 0)
+        α = acos(clamp(cosα, -1, 1))
+        E[2, 1] = -α
+        E[1, 2] = α
+        copyto!(X, P * E * transpose(P))
+    else
+        log_safe!(X, q)
```
kellertuer, line 149: I am not sure what is done here, but it currently fails a test, since points like `Diagonal([-1.0, -1.0, -1.0, 1.0})`are valid here but hit this line _and_ have a determinant of -1. So where is this formula from? Is it correct? What do now? @sethaxen ?

kellertuer:
Similar to the comment above – a dev = -1 matrix on O(3) currently produces NaNs in exp/log test. So I have to pause here until I get a reference / formula for O(n).

src/groups/orthogonal.jl
mateuszbaran: Well, as far as I know there can be no `log_lie` for point with determinant -1. `Diagonal([-1.0, -1.0, -1.0, 1.0})` lies in a different connected component. I guess we should throw an error?

kellertuer: I just discussed this with Seth and you might be right (then one could just add a `is_point(SpecialOrthogonal,...)` there?) Let's see, I will wait for one feedback, since currently on SO(n) and O(n) (which then have the same log/exp lie) we currently have different implementations.

kellertuer:
Hm, and the test on Grassmann that failed ...sometimes... on Julia 1.6 now persistently fails for an `is_vector` that is actually a tangent vector.
Also, we really might want to speed al this up, the manifold tests now run over an hour.

src/groups/orthogonal.jl
kellertuer: right, we would just have to check the determinant, that's right.

mateuszbaran:
I've investigated that Grassmann test and it's weird. `check_size` should (I think) decide that it's not a vector but `embed(Grassman(...), p, X)` "fixes" the size of `X` by allocating result of the correct size. Then it's random based on what the vector `X` gets filled with during the embedding. `copyto!` simply doesn't care about array sizes.

kellertuer:
Ah. I will check that as well, right, if it is filled with arbitrary values then that's where the fail comes from. Interesting!

kellertuer:
> I think the default `embed`
> https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/master/src/decorator_trait.jl#L186-L189
> and
> https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/master/src/decorator_trait.jl#L171-L174
> should check that at least the lengths of copied and destination arrays are equal.

I think the current default (I introduced that again for Grassmann) is that `embed(M, p, X) = X` does not allocate a new one for the identity, but really just returns X (that is what Grassman does for the two types we have now as well actually). Then the size in the embedding is consistenly wrong (and it does not depend on the orthonormality check to fail anymore).

Thanks for noticing.

But sure, the alternative is what you propose (then we could actually delete quite a few `embed(M,p)=p` and `embed(M,p,X)=X` lines).

mateuszbaran:
> Also, we really might want to speed al this up, the manifold tests now run over an hour.

That's a bit problematic. We keep adding functionality and have a ton of mutually interacting features so figuring out which tests are safe to remove is often difficult. We will likely have to split some functionality into separate packages.

kellertuer:
> > Also, we really might want to speed al this up, the manifold tests now run over an hour.
> 
> That's a bit problematic. We keep adding functionality and have a ton of mutually interacting features so figuring out which tests are safe to remove is often difficult. We will likely have to split some functionality into separate packages.

I only see that for Lie groups actually, which have their separate test env already. We could do something similar and do two batches of manifold tests (and hence split those).

By the way I unified the parameters for the tests, all of them are now ENV variables and defined in utils.jl.

src/groups/orthogonal.jl
sethaxen: Computing the determinant is ~1% the cost of computing the matrix logarithm for an orthogonal matrix, so this seems a good way to go.

kellertuer: Perfect, will add that. This function (Orthogonal(4) log lie) seems to be the only test failing for now, doing exp-then-log yields

```
isapprox(Orthogonal(4), Identity(MultiplicationOperation), [0.0 -0.2853981633974483 0.0 0.0; 0.2853981633974483 0.0 0.0 0.0; 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0], [0.0 1.0 0.0 0.0; -1.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0])
```

(where the second one is the one I start with and after exp-then-log I have the first one), I will for now add that as a broken test to check for further test coverage.

sethaxen: Ah, I see now there's an error in the exponential map for `Orthogonal{4}`. I never include the `a₁` term. Just revert to the `Rotations{4}` version for now.

kellertuer: Thanks I will check – the `Orthogonal{4}` function is really just the one from the old branch/PR.

kellertuer: It works fine, I just have no clue how to hit the special if cases in code coverage.

kellertuer: see https://app.codecov.io/gh/JuliaManifolds/Manifolds.jl/compare/499/tree/src/groups/general_unitary_groups.jl

sethaxen: > Thanks I will check – the `Orthogonal{4}` function is really just the one from the old branch/PR.

Which in turn is a slight tweak of `Rotations{4}` to remove an allocation or 2.

> It works fine

No, it is incorrect as noted above. Please revert to the `Rotations{4}` implementation for now.

> I just have no clue how to hit the special if cases in code coverage

See the `Rotations{4}` tests: 
https://github.com/JuliaManifolds/Manifolds.jl/blob/9a76bad003ef14339c1f61cc33e607efd990e198/test/manifolds/rotations.jl#L161-L184

kellertuer: Hm? I just copied for the last commit the code from Rot4.

Sure I will adapt these tests then (but they should be run anyways on that code base now.

kellertuer: so in short - I am unsure why this exp_lie / group_exp has a q https://github.com/JuliaManifolds/Manifolds.jl/blob/38e20ead10af77ae241002fb3abaf309d9c17ea0/src/groups/orthogonal.jl#L169

kellertuer: ah no, I am just stupid.

sethaxen: What was the issue in the end? :sweat_smile:

kellertuer: > Ah, I see now there's an error in the exponential map for `Orthogonal{4}`. I never include the `a₁` term. Just revert to the `Rotations{4}` version for now.

so rot4 for example does not have a group log?

kellertuer: I pushed the current state which fails O(n) still. I am a little lost which is now the right implementation for exp_lie (formerly group_exp) and its log for n=4

sethaxen: Use this for `exp`:
https://github.com/JuliaManifolds/Manifolds.jl/blob/9a76bad003ef14339c1f61cc33e607efd990e198/src/manifolds/Rotations.jl#L203-L243
and this for `log`:
https://github.com/JuliaManifolds/Manifolds.jl/blob/9a76bad003ef14339c1f61cc33e607efd990e198/src/manifolds/Rotations.jl#L604-L624

sethaxen: Just replace `p` with `I`, and you have the `exp_lie`. but you can probably remove an allocation by not multiplying by `I`.

kellertuer: Ok, I will check, my problem was exactly that, together with the exp/log (end exp/log_lie) I had 6 implementations, (Rotations, SpecialOrthogonal, Orthogonal) for each (exp and log) – so I got totally lost which one to take now.

kellertuer: hm, but now line 241 includes a1 again which you said should not be in there? That is also a difference between the one `exp_lie I have, i.e. 

```
    copyto!(q, a₀ * I)
    q .+= a₂ .* X²
    mul!(q, X², X, a₃, true)
```

and the other one which is (in exp, so this is multiplied by p afterwards

```
    pinvq = a₀ * I + a₁ .* X .+ a₂ .* X² .+ a₃ .* X³
```

which one is now correct?

kellertuer: My question was just which one was correct ;) I am further fiddling around (lost all afternoon anyways already to this). It was really tricky with soooooooooooo many implementations basically doing the same.

kellertuer: *phew* I think I finally got it and reduced the code redundancy here, also put all non-group stuff into the matrix/set definitions and the group/lie definitions into the group setting.

kellertuer:
We seem to miss exactly one case `α ≈ π && β ≈ 0` which I also can not see in the comments (or read them wrong).

kellertuer:
I twiddled around a little bit to find such a matrix and I am a little stuck.
* alpha = \pi means cos \alpha = -1
* \beta = 0 means cos \beta = 1

These are the values the magic 4d rotation plane function (`cos_angles_4d_rotation_matrix` ) shall return and as documented they are ordered, so this fits we have
* (a+b)/4 = -1
* (a-b)/4 = -1

Adding both we have 2/4 ^a = 0 so a=0 (-> we need a matrix R with trace 0)
The difference of both provides us with 2/4b = -1 so b=-2. Looking at the code we get that it only produces positive b (clamp and then sqrt). So after this small game I am not sure how to get to this case at all – and we still need a few lines for patch code cov.

So – where is this function from and where is this one case from? That would be neat to know.
Note that even on master (so before this rework here) these 9 lines do not have code coverage: https://app.codecov.io/gh/JuliaManifolds/Manifolds.jl/blob/master/src/manifolds/Rotations.jl (line 610).

mateuszbaran:
Not sure if it would help but section 5.2 of this: https://arxiv.org/pdf/1103.5263.pdf has a similar calculation for rotations in 4 and 5 dimensions.

kellertuer:
I even checked the papers and unterstand the general compuation, also roughly how the alpha and beta come along, just that case is not mentioned, and I think in the current implementation those 9 lines are actually never reachable. At first glance also your link follows the same route but never mentions the case we are currently not reaching (and I think never covered in tests).

mateuszbaran:
Right, I think we could remove this case and fix it when someone opens an issue about it (with an example).

mateuszbaran:
Anyway, equations in docstring for `cos_angles_4d_rotation_matrix` don't seem to match the implementation so it would be nice to fix that.

kellertuer:
I have nt yet fully checked that but removing that case renders the special case for n=4 useless (it is basically again just the log?), so I am unsure about deleting the whole case.

sethaxen:
Thanks for digging into this some more, and I'm sorry for the confusion. I wrote this code a while ago, so it's all a little foggy.

First observation is that `cos_angles_4d_rotation_matrix` and `angles_4d_skew_sym_matrix` are incorrectly documented. They return values in decreasing and increasing order, respectively, while the documentation says they return in increasing and decreasing order, respectively. That means it's impossible for the condition `α ≈ π && β ≈ 0` to be satisfied! The documentation should be fixed, but in `log!` all is well if we simply swap `α` and `β`.

The kind of matrix this check is designed for is `diagm([-1.0, -1, 1, 1])`, for which `exp` is surjective, so real matrix logarithm fails (even on recent Julia versions):

```julia
julia> R = diagm([-1.0, -1, 1, 1])
4×4 Matrix{Float64}:
 -1.0   0.0  0.0  0.0
  0.0  -1.0  0.0  0.0
  0.0   0.0  1.0  0.0
  0.0   0.0  0.0  1.0

julia> log(R)
4×4 Matrix{ComplexF64}:
 0.0+3.14159im  0.0+0.0im      0.0+0.0im  0.0+0.0im
 0.0+0.0im      0.0+3.14159im  0.0+0.0im  0.0+0.0im
 0.0+0.0im      0.0+0.0im      0.0+0.0im  0.0+0.0im
 0.0+0.0im      0.0+0.0im      0.0+0.0im  0.0+0.0im

julia> log(Rotations(4), diagm(ones(4)), R)  # wrong!!
4×4 Matrix{Float64}:
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
```

There are two ways to get this matrix using `exp`:
```julia
julia> X = [[0 -π; π 0] 0*I; 0*I 0*I]
4×4 Matrix{Float64}:
 0.0      -3.14159  0.0  0.0
 3.14159   0.0      0.0  0.0
 0.0       0.0      0.0  0.0
 0.0       0.0      0.0  0.0

julia> exp(X) ≈ exp(-X) ≈ R
true
```

If we repair `log!` to use the correct ordering, then all works:
```julia
julia> function Manifolds.log!(M::Rotations{4}, X, p, q)
           U = transpose(p) * q
           cosα, cosβ = Manifolds.cos_angles_4d_rotation_matrix(U)
           α = acos(clamp(cosα, -1, 1))
           β = acos(clamp(cosβ, -1, 1))
           if α ≈ 0 && β ≈ π
               A² = Symmetric((U - I) ./ 2)
               P = eigvecs(A²)
               E = similar(U)
               fill!(E, 0)
               @inbounds begin
                   E[2, 1] = -β
                   E[1, 2] = β
               end
               copyto!(X, P * E * transpose(P))
           else
               copyto!(X, real(Manifolds.log_safe(U)))
           end
           return Manifolds.project_no_rep_change!(M, X, p, X)
       end

julia> log(Rotations(4), diagm(ones(4)), R)
4×4 Matrix{Float64}:
  0.0      3.14159  0.0  0.0
 -3.14159  0.0      0.0  0.0
  0.0      0.0      0.0  0.0
  0.0      0.0      0.0  0.0
```

The test should also be performed for all cyclic permutations of this diagonal matrix, e.g. `diagm([-1, 1, -1, 1])` and `diagm([1, -1, -1, 1])`.

> So – where is this function from and where is this one case from? That would be neat to know.

IIRC it's novel. When I implemented `exp` I was also trying to implement an efficient `log` that avoided matrix logarithm. I got most of the way there but in the end chose to just use this condition until I had time to finish the derivation. I can share the document with you on Overleaf if you'd like.

kellertuer:
Thanks for the check, I will have to recheck the documentation then, but the other way around (with switched alpha and beta) the case should appear, yes :)

kellertuer:
Hm, for the documentation I am confused because also the formula seems wrong? Why is it both +? and where is it from? I even tried to google some magic stuff like this but could not find it.

```
\begin{aligned}
\cos α + \cos β &= \frac{1}{2} \operatorname{tr}(R)\\
\cos α + \cos β &= \frac{1}{8} \operatorname{tr}(R)^2
                 - \frac{1}{16} \operatorname{tr}((R - R^T)^2) - 1.
\end{aligned}
```

sethaxen:
> do you have an Xf vector for the R that you just wrote

What is `Xf`? Do you mean a tangent vector? I wrote one above.

> I am confused because also the formula seems wrong? Why is it both +?

> \begin{aligned}
\cos α + \cos β &= \frac{1}{2} \operatorname{tr}(R)\\
\cos α + \cos β &= \frac{1}{8} \operatorname{tr}(R)^2
                 - \frac{1}{16} \operatorname{tr}((R - R^T)^2) - 1.
\end{aligned}

You're right, the second equation should be `\cos α \cos β`. See Proposition 3.1 of https://cs.brynmawr.edu/~dxu/206-2550-2.pdf

We have `a = (cos(α) + cos(β))/2` and `b = |cos(α) - cos(β)|/2`, so if `α > 0`, then `(a + b, a-b) == (cos(α), cos(β))`, and the order is flipped when `α < 0`. You can get there from the proposition referenced above and some subsequent algebra.

> edit: I am aware I am asking a lot of questions here, but in understanding all of these details you basically already lost me yesterday, I am just a code monkey here, sorry for that. I really just can't match the second line to what is happening in the code, so any reference or a fix of that line would be very welcome.

No worries! Thanks for finding all of these issues! :sweat_smile:

kellertuer:
> > do you have an Xf vector for the R that you just wrote
> 
> What is `Xf`? Do you mean a tangent vector? I wrote one above.

The old special cases. Nevermind I just wrote the three permutations of R as new cases.

> You're right, the second equation should be `\cos α \cos β`. See Proposition 3.1 of https://cs.brynmawr.edu/~dxu/206-2550-2.pdf

Ah, perfect, I will just remove the plus then

kellertuer:
🥳 we have reached 100% patch coverage and even 0.3% coverage increase (I might check for the two missing lines in Grassmann even here) – so if someone could review/approve this, that would be great :)

docs/src/manifolds/group.md
```diff
@@ -63,67 +63,96 @@ Pages = ["groups/multiplication_operation.jl"]
...
-Order = [:type, :function, :constant]
+Pages = ["groups/heisenberg.jl"]
+Order = [:type,:function]
 ```
 
-### Circle group
+### (Special) Orthogonal and (Special) Unitary group
+
+Since the orthogonal, unitary and special orthogonal and special unitary groups share
+quite some common functions, these are also implemented on a common level
```
mateuszbaran, line 93:
```suggestion
many common functions, these are also implemented on a common level.
```

docs/src/misc/notation.md
```diff
@@ -13,6 +13,7 @@ Within the documented functions, the utf8 symbols are used whenever possible, as
 | ``\operatorname{Ad}_p(X)`` | adjoint action of element ``p`` of a Lie group on the element ``X`` of the corresponding Lie algebra | | |
 | ``\times`` | Cartesian product of two manifolds | | see [`ProductManifold`](@ref) |
 | ``^{\wedge}`` | (n-ary) Cartesian power of a manifold | | see [`PowerManifold`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/manifolds.html#ManifoldsBase.PowerManifold) |
+| ``\cdot^\mathrm{H}`` | conjugate/Hermitian transposed | |
```
mateuszbaran, line 16:
```suggestion
| ``\cdot^\mathrm{H}`` | conjugate/Hermitian transpose | |
```

src/groups/general_unitary_groups.jl
```diff
@@ -0,0 +1,262 @@
...
+    return allocate(X)
+end
+function allocate_result(
+    ::GeneralUnitaryMultiplicationGroup,
+    ::typeof(log),
+    ::Identity{MultiplicationOperation},
+    q,
+)
+    return allocate(q)
+end
```
mateuszbaran, line 43: These methods shouldn't be necessary after #503 .

src/groups/general_unitary_groups.jl
```diff
@@ -0,0 +1,262 @@
...
+
+decorated_manifold(G::GeneralUnitaryMultiplicationGroup) = G.manifold
+
+@doc raw"""
+     exp_lie(G::Orthogonal{2}, X)
+     exp_lie(G::SpecialOrthogonal{2}, X)
+
+Compute the Lie group exponential map on the [`Orthogonal(2)`] group.
```
mateuszbaran, line 51:
```suggestion
Compute the Lie group exponential map on the [`Orthogonal(2)`](@ref) or [`SpecialOrthogonal(2)`](@ref) group.
```
I'm not sure if `@ref` will work with that `(2)`. If not, I'd put `(2)` after the link, like in `exp_lie(::GeneralUnitaryMultiplicationGroup{4,ℝ}, X)` docstring below.

src/groups/general_unitary_groups.jl
```diff
@@ -0,0 +1,262 @@
...
+        fill!(E, 0)
+        @inbounds begin
+            E[2, 1] = -β
+            E[1, 2] = β
+        end
+        copyto!(X, P * E * transpose(P))
+    else
+        det(q) < 0 && throw(
+            DomainError(
+                "The LIe group logarithm is not defined for $q with a negative determinant ($(det(q)) < 0).",
```
mateuszbaran, line 240:
```suggestion
                "The Lie group logarithm is not defined for $q with a negative determinant ($(det(q)) < 0). Point `q` is in a different connected component of the manifold $G",
```
A bit more detailed explanation.

src/groups/group_operation_action.jl
```diff
@@ -26,14 +26,8 @@ function switch_direction(A::GroupOperationAction)
     return GroupOperationAction(A.group, switch_direction(direction(A)))
 end
 
-apply(A::GroupOperationAction, a, p) = translate(A.group, a, p, direction(A))
```
mateuszbaran, line 29: Why did you delete that?

src/groups/special_euclidean.jl
```diff
@@ -4,7 +4,7 @@
 Special Euclidean group $\mathrm{SE}(n)$, the group of rigid motions.
 
 ``\mathrm{SE}(n)`` is the semidirect product of the [`TranslationGroup`](@ref) on $ℝ^n$ and
-[`SpecialOrthogonal(n)`](@ref)
+[`SpecialOrthogonal`](@ref)`(n)
```
mateuszbaran, line 7:
```suggestion
[`SpecialOrthogonal`](@ref)`(n)`
```
Probably?

src/groups/special_unitary.jl
```diff
@@ -0,0 +1,74 @@
+@doc raw"""
+    SpecialUnitary{n} = GeneralUnitaryMultiplicationGroup{n,ℝ,GeneralUnitaryMatrices{n,ℂ,DeterminantOneMatrices}}
+
+The special unitary group ``\mathrm{SU}(n)`` represented by unitary matrices of determinant +1.
+
+The tangent spaces are if the form
```
mateuszbaran, line 6:
```suggestion
The tangent spaces are of the form
```

src/groups/special_unitary.jl
```diff
@@ -0,0 +1,74 @@
...
+
+```math
+T_p\mathrm{SU}(x) = \bigl\{ X \in \mathbb C^{n×n} \big| X = pY \text{ where } Y = -Y^{\mathrm{H}} \bigr\}
+```
+
+and we represent tangent vectors by just storing the [`SkewHermitianMatrices`](@ref) ``Y``,
+or in other words we reoresent the tangent spaces employing the Lie algebra ``\mathfrak{su}(n)``.
```
mateuszbaran, line 13:
```suggestion
or in other words we represent the tangent spaces employing the Lie algebra ``\mathfrak{su}(n)``.
```

src/groups/unitary.jl
```diff
@@ -0,0 +1,83 @@
...
+The group consists of all points ``p ∈ \mathbb C^{n × n}`` where ``p^\mathrm{H}p = pp^\mathrm{H} = I``.
+
+The tangent spaces are if the form
+
+```math
+T_p\mathrm{U}(x) = \bigl\{ X \in \mathbb C^{n×n} \big| X = pY \text{ where } Y = -Y^{\mathrm{H}} \bigr\}
+```
+
+and we represent tangent vectors by just storing the [`SkewHermitianMatrices`](@ref) ``Y``,
+or in other words we reoresent the tangent spaces employing the Lie algebra ``\mathfrak{u}(n)``.
```
mateuszbaran, line 15:
```suggestion
or in other words we represent the tangent spaces employing the Lie algebra ``\mathfrak{u}(n)``.
```

src/manifolds/Orthogonal.jl
```diff
@@ -0,0 +1,14 @@
+@doc raw"""
+     OrthogonalMatrices{n} =  GeneralUnitaryMatrices{n,ℝ,AbsoluteDeterminantOneMatrices}
```
mateuszbaran, line 2:
```suggestion
     OrthogonalMatrices{n} = GeneralUnitaryMatrices{n,ℝ,AbsoluteDeterminantOneMatrices}
```

src/manifolds/GeneralUnitaryMatrices.jl
```diff
@@ -0,0 +1,712 @@
+@doc """
+    AbstractMatrixType
+
+A plain type to distinguish different types of matrices, for example [`DeterminantOneMatrices`](@ref)
+and [`AbsoluteDeterminantOneMatrices`](@ref)
+"""
+abstract type AbstractMatrixType end
+
+@doc """
+    SoecialMatrices <: AbstractMatrixType
```
mateuszbaran, line 10:
```suggestion
    DeterminantOneMatrices <: AbstractMatrixType
```

src/manifolds/GeneralUnitaryMatrices.jl
```diff
@@ -0,0 +1,712 @@
...
+
+A type to indicate that we require special (orthogonal / unitary) matrices, i.e. of determinant 1.
+"""
+struct DeterminantOneMatrices <: AbstractMatrixType end
+
+@doc """
+    AbsoluteDeterminantOneMatrices <: AbstractMatrixType
+
+A type to indicate that we require (orthogonal / unitary) matrices with normed determinant,
+i.e. that the absolute value of the determinant is one of determinant 1.
```
mateuszbaran, line 20:
```suggestion
i.e. that the absolute value of the determinant is 1.
```

src/manifolds/GeneralUnitaryMatrices.jl
```diff
@@ -0,0 +1,712 @@
...
+The tolerance for the last test can be set using the `kwargs...`.
+"""
+function check_point(
+    M::GeneralUnitaryMatrices{n,𝔽,AbsoluteDeterminantOneMatrices},
+    p;
+    kwargs...,
+) where {n,𝔽}
+    if !isapprox(abs(det(p)), 1; kwargs...)
+        return DomainError(
+            det(p),
```
mateuszbaran, line 53:
```suggestion
            abs(det(p)),
```

src/manifolds/GeneralUnitaryMatrices.jl
```diff
@@ -0,0 +1,712 @@
...
+end
+
+@doc raw"""
+    check_vector(M::UnitaryMatrices{n}, p, X; kwargs... )
+    check_vector(M::OrthogonalMatrices{n}, p, X; kwargs... )
+    check_vector(M::Rotations{n}, p, X; kwargs... )
+    check_vector(M::GeneralUnitaryMatrices{n,𝔽}, p, X; kwargs... )
+
+Check whether `X` is a tangent vector to `p` on the [`UnitaryMatrices`](@ref)
+space `M`, i.e. after [`check_point`](@ref)`(M,p)`, `X` has to be skew symmetric (hermitian)
```
mateuszbaran, line 131:
```suggestion
space `M`, i.e. after [`check_point`](@ref)`(M,p)`, `X` has to be skew symmetric (Hermitian)
```

src/manifolds/GeneralUnitaryMatrices.jl
```diff
@@ -0,0 +1,712 @@
...
+dimension and orthogonal to `p`.
```
mateuszbaran, line 132: "dimension" should probably be deleted here?

src/manifolds/Rotations.jl
```diff
@@ -1,17 +1,16 @@
...
-since these matrices represent all rotations of points in $ℝ^n$.
+The manifold of rotation matrices of sice ``n × n``, i.e.
+real-valued orthogonal matrices with determinant ``+1``.
 
 # Constructor
 
     Rotations(n)
 
-Generate the $\mathrm{SO}(n) \subset ℝ^{n × n}$
+Generate the manifold of ``ℝ^{n × n}`` rotation matrices.
```
mateuszbaran, line 11:
```suggestion
Generate the manifold of ``n × n`` rotation matrices.
```

src/manifolds/Unitary.jl
```diff
@@ -0,0 +1,31 @@
+
+@doc raw"""
+    const UnitaryMatrices{n} = AbstarctUnitaryMatrices{n,ℂ,AbsoluteDeterminantOneMatrices}
+
+The manifold ``U(n)`` of ``n×n`` complex matrices
```
mateuszbaran, line 5:
```suggestion
The manifold ``U(n)`` of ``n×n`` complex matrices such that
```

src/manifolds/Unitary.jl
```diff
@@ -0,0 +1,31 @@
...
+
+``p^{\mathrm{H}p = \mathrm{I}_n,``
+
+where ``\mathrm{I}_n`` is the ``n×n`` identity matrix.
+An alternative characterisation is that ``\lVert \det(p) \rVert = 1``.
```
mateuszbaran, line 10:
```suggestion
Such matrices `p` have a property that ``\lVert \det(p) \rVert = 1``.
```
Except for n=1 there are matrices of absolute determinant 1 that are not unitary.

src/manifolds/Unitary.jl
```diff
@@ -0,0 +1,31 @@
...
+    X \big| pY \text{ where } Y \text{ is skew symmetric, i. e. } Y = -Y^{\mathrm{H}}
+    \bigr\}
+```
+
+But note that tangent vectors are represented in the Lie algebra, i.e. just using ``Y`` in the representation above.
+
+# Constructor
+     UnitaryMatrices(n)
+
+see also [`UnitaryMatrices`](@ref) for the real valued case.
```
mateuszbaran, line 25:
```suggestion
see also [`OrthogonalMatrices`](@ref) for the real valued case.
```

test/ambiguities.jl
```diff
@@ -4,6 +4,7 @@
         # Interims solution until we follow what was proposed in
         # https://discourse.julialang.org/t/avoid-ambiguities-with-individual-number-element-identity/62465/2
         fmbs = filter(x -> !any(has_type_in_signature.(x, Identity)), mbs)
+        FMBS_LIMIT = 17
```
mateuszbaran, line 7:
```suggestion
```
Same variable as below.

mateuszbaran (commented):
Just a few things I've noticed.

src/groups/general_unitary_groups.jl
kellertuer: Let's see which one we merge first ;)

src/groups/general_unitary_groups.jl
kellertuer: I think one has to use the parameters then, so probably `[`Orthgonal(2)`](@ref Orthogonal{2})`? But we could try this one as well first.

src/manifolds/GeneralUnitaryMatrices.jl
kellertuer: ah, yes
```suggestion
and orthogonal to `p`.
```

src/groups/group_operation_action.jl
mateuszbaran: I've replaced it with `apply(A::GroupOperationAction, a, p) = error()` and the error is thrown in group_oeration_action.jl tests.

mateuszbaran: And codecov reports it as covered.

mateuszbaran: https://app.codecov.io/gh/JuliaManifolds/Manifolds.jl/blob/master/src/groups/group_operation_action.jl

kellertuer: hm, then I do not know and lets undo this here. I was looking for test coverage I think and it was not covered and removing it, the tests still worked

kellertuer: I restored that line.

src/groups/group_operation_action.jl
```diff
@@ -30,10 +30,6 @@ apply(A::GroupOperationAction, a, p) = translate(A.group, a, p, direction(A))
 
 apply!(A::GroupOperationAction, q, a, p) = translate!(A.group, q, a, p, direction(A))
 
-function inverse_apply(A::GroupOperationAction, a, p)
```
mateuszbaran, line 33: Could you also restore this?

src/groups/group_operation_action.jl
mateuszbaran: I meant `inverse_apply(A::GroupOperationAction, a, p)` in that other comment.

Tests still work because the mutating default is used.

src/groups/general_unitary_groups.jl
kellertuer:
```suggestion
Compute the Lie group exponential map on the [`Orthogonal`](@ref)`(2)` or [`SpecialOrthogonal`](@ref)`(2)` group.
```

what about this variant?

mateuszbaran: Looks fine :+1: .

mateuszbaran (approved):
I think this looks fine :+1: .

src/manifolds/Unitary.jl
```diff
@@ -0,0 +1,31 @@
+
+@doc raw"""
+    const UnitaryMatrices{n} = AbstarctUnitaryMatrices{n,ℂ,AbsoluteDeterminantOneMatrices}
+
+The manifold ``U(n)`` of ``n×n`` complex matrices such that
+
+``p^{\mathrm{H}p = \mathrm{I}_n,``
```
mateuszbaran, line 7: This does render correctly in documentation.

kellertuer:
```suggestion
``p^{\mathrm{H}}p = \mathrm{I}_n,``
```

You forgot a “not” or an “in” (before correct) and I forgot a closing `}` ;)

mateuszbaran: Yes, I meant to write "does not render correctly" :wink:

kellertuer:
Let‘s wait for this final test run.

PR #500, opened 2022-07-07
Power group

mateuszbaran:
Addresses another point raised in #497 .


mateuszbaran:
The new ambiguities are caused by the new `allocate_result` for `PowerGroup` but they shouldn't occur in practice because the group decorator is unwrapped first in those cases.

src/groups/power_group.jl
```diff
@@ -0,0 +1,342 @@
+
+const PowerGroup{𝔽,M<:AbstractManifold{𝔽},TPR<:AbstractPowerRepresentation} =
+    GroupManifold{𝔽,<:AbstractPowerManifold{𝔽,M,TPR},ProductOperation}
+
+const PowerGroupNested{𝔽,M<:AbstractManifold{𝔽}} = PowerGroup{𝔽,M,NestedPowerRepresentation}
+
+const PowerGroupNestedReplacing{𝔽,M<:AbstractManifold{𝔽}} =
+    PowerGroup{𝔽,M,NestedReplacingPowerRepresentation}
+
+"""
```
kellertuer, line 10: I would have added the doctoring to the const – is it better to attach it to the constructor?

src/groups/power_group.jl
```diff
@@ -0,0 +1,342 @@
...
+inv!(::PowerGroup, q::Identity{ProductOperation}, ::Identity{ProductOperation}) = q
+function inv!(
+    ::PowerGroupNestedReplacing,
+    q::Identity{ProductOperation},
+    ::Identity{ProductOperation},
+)
+    return q
+end
+
+_compose!(G::PowerGroup, x, p, q) = _compose!(G.manifold, x, p, q)
```
kellertuer, line 111: `compose` here is dispatched on Level 2 not on level 1 – what is the idea here? For me they look like level 1.

test/ambiguities.jl
```diff
@@ -17,7 +17,7 @@
         # Interims solution until we follow what was proposed in
         # https://discourse.julialang.org/t/avoid-ambiguities-with-individual-number-element-identity/62465/2
         fms = filter(x -> !any(has_type_in_signature.(x, Identity)), ms)
-        FMS_LIMIT = 22
+        FMS_LIMIT = 34
```
kellertuer, line 20: Thats a lot of new ambiguities, are you sure all are necessary?

kellertuer (commented):
Cool we maybe add some documentation for example on the groups.md page?

mateuszbaran:
> And can we maybe add some documentation for example on the `groups.md` page?

Sure, I forgot about it.

test/ambiguities.jl
mateuszbaran: Yes, I don't like it either but the alternative is adding a bunch of methods that will never be used outside of our tests because the are related to metric-dependent functions which dispatch down to the power manifold from power group.

src/groups/power_group.jl
mateuszbaran: Sure, I'll add a docstring to this alias.

src/groups/power_group.jl
mateuszbaran: I copied it from the product group without thinking much. Probably the idea is that this way we don't need to handle `::Identity` disambiguiations.

kellertuer: ah right, that sounds reasonable. Maybe we should add a comment to clarify?

mateuszbaran: Sure.

kellertuer:
> > And can we maybe add some documentation for example on the `groups.md` page?
> 
> Sure, I forgot about it.

No problem, thats what PR and code reviews  are for :)

kellertuer (approved):

Issue #382, opened 2021-06-15
Rigid transforms using unit quaternions to represent rotation


Affie:
Related to the rabbit hole part of https://github.com/JuliaManifolds/Manifolds.jl/issues/358 where `M = Sphere(3,ℍ)` was mentioned to represent unit quaternions.

I was looking at using a semidirect product manifold of `T(3) ⋊ Sphere(3,ℍ)` or similar to compare with `SE(3)`
I cannot find a lot in the documentation on quaternions though. 

Are quaternions implemented? If so how does the representation look? 
Is there a Lie group/algebra implemented for Sphere or similar?

It is not something that is immediately needed at all, I was just curious. 
I also don't know if it will be worth it yet.

sethaxen:
Quaternions are implemented as a field that can underly a manifold (just enter `ℍ`), but because there is not implementation of quaternions in Julia base and there are actually quite a few packages that implement quaternions, we don't commit to or depend on any one package. But there are examples in our tests where we have implemented and tested a quaternion manifold. (see e.g. https://github.com/JuliaManifolds/Manifolds.jl/blob/v0.5.4/test/projective_space.jl#L192-L300)

Semidirect products only make sense between two groups. So you would need to decorate a sphere with a relevant group operation in order to make this work. #25 experimented with doing just that, calling a `Sphere` decorated with a `QuaternionicMultiplication` a `ComplexSymplectic` group.

The right thing to do here is probably to implement quaternions as a group. There are several equivalent ways to do this. One would `U(1, ℍ)`, that is, the group of 1x1 unitary matrices whose elements are quaternions. Such a matrix would contain a single unit quaternion. This would be provided by #345, but I don't plan to finish that PR soon. That PR would also provide `SU(2)`, the special unitary group, which is technically a unit quaternion represented as a complex 2x2 matrix. Both of these suffer from requiring to use matrices instead of just a unit quaternion. Another way is to implement the compact symplectic group `Sp(n,ℍ)` as above, but this would technically be the same as `U(1, ℍ)`, so should be a matrix. 

The next component you would need is an action of one of these points upon a translation vector, similar to our `RotationAction`. With just those two implemented, the semidirect product of this group and `T(3)` should just work as well as `SE(n)`.

sethaxen:
I should note that while this may sound complicated, it really isn't; the complication is just the design decision of which of these isomorphic groups to choose to implement.

Affie:
[...]

PR #506, opened 2022-07-21
Quaternionic unitary group

mateuszbaran:
This should solve #382 . `exp` and `log` are currently only supported for the n=1 case because I couldn't find a generic quaternionic matrix exponential and logarithm.

Also, plain numbers should also work for `QuaternionicUnitary(1)`.


sethaxen:
> `exp` and `log` are currently only supported for the n=1 case because I couldn't find a generic quaternionic matrix exponential and logarithm.

They can be supported by expanding each quaternion into its 2x2 complex matrix representation, which become blocks in a matrix with `nxn` blocks, so an `nxn` quaternionic matrix corresponds to an `2nx2n` complex matrix. This transform preserves the matrix product, so the result is a closed subgroup of `U(2n)`, on which one can compute `exp`/`log` and then map back to quaternions.

But I think it's best to handle this kind of thing at the level of the quaternion implementation, not here. e.g. Quaternions.jl had https://github.com/JuliaGeometry/Quaternions.jl/pull/46 for adding support for matrix exponential, and this could be expanded to support all matrix functions. Are there fallbacks to for `exp_lie` and `log_lie` that call the matrix exponential? If so, then this would already be supported with such a PR to Quaternions.

Why `QuaternionicUnitary(n)` and not something like `Unitary(n, 𝔽=ℂ)`? That would be more consistent with how we allow constructors to take fields elsewhere.

mateuszbaran:
> They can be supported by expanding each quaternion into its 2x2 complex matrix representation, which become blocks in a matrix with `nxn` blocks, so an `nxn` quaternionic matrix corresponds to an `2nx2n` complex matrix. This transform preserves the matrix product, so the result is a closed subgroup of `U(2n)`, on which one can compute `exp`/`log` and then map back to quaternions.

Thanks, I didn't know that.

> But I think it's best to handle this kind of thing at the level of the quaternion implementation, not here. e.g. Quaternions.jl had [JuliaGeometry/Quaternions.jl#46](https://github.com/JuliaGeometry/Quaternions.jl/pull/46) for adding support for matrix exponential, and this could be expanded to support all matrix functions. Are there fallbacks to for `exp_lie` and `log_lie` that call the matrix exponential? If so, then this would already be supported with such a PR to Quaternions.

OK, so I won't work on that here in this PR. Currently there are no such fallbacks but they will be easy to add when quaternionic matrix exp is available.

> Why `QuaternionicUnitary(n)` and not something like `Unitary(n, 𝔽=ℂ)`? That would be more consistent with how we allow constructors to take fields elsewhere.

Mostly to be consistent with the existing `Orthogonal`/`Unitary`. I don't have a strong opinion about it but maybe let's ask @kellertuer what he prefers.

kellertuer:
> > Why `QuaternionicUnitary(n)` and not something like `Unitary(n, 𝔽=ℂ)`? That would be more consistent with how we allow constructors to take fields elsewhere.
> 
> Mostly to be consistent with the existing `Orthogonal`/`Unitary`. I don't have a strong opinion about it but maybe let's ask @kellertuer what he prefers.

I think for Orthogonal and Unitary there is different terms and names, so that's why I preferred to split that, but for complex or quaternionic Unitary the field should be fine.

docs/src/manifolds/generalunitary.md
```diff
@@ -7,21 +7,22 @@ Both [`OrthogonalMatrices`](@ref) and [`UnitaryMatrices`](@ref) are quite simila
...
 ```
 
+
 ## [Common functions](@id generalunitarymatrices)
 
 ```@autodocs
 Modules = [Manifolds]
 Pages = ["manifolds/GeneralUnitaryMatrices.jl"]
-Order = [:type, :function]
+Order = [:type, :function, :constant]
```
kellertuer, line 27: Since the constants here are the Lie groups, we could put those first to always first habe the Lie group definition?

docs/src/misc/notation.md
```diff
@@ -27,7 +27,7 @@ Within the documented functions, the utf8 symbols are used whenever possible, as
 | ``\gamma`` | a geodesic | ``\gamma_{p;q}``, ``\gamma_{p,X}`` | connecting two points ``p,q`` or starting in ``p`` with velocity ``X``. |
 | ``∇ f(p)`` | gradient of function ``f \colon \mathcal{M} \to \mathbb{R}`` at ``p \in \mathcal{M}`` | | |
 | ``\circ`` | a group operation | |
-| ``\cdot^\mathrm{H}`` | Hermitian or conjugate transposed| |
+| ``\cdot^\mathrm{H}`` | Hermitian, quaternionic or conjugate transposed| |
```
kellertuer, line 30: this might be misleading, since Hermitian and conjugate are the same thing, while quaternionic is something different (we can keep the `^H` for sure, just the text is misleading
```suggestion
| ``\cdot^\mathrm{H}`` | Hermitian or conjugate transposed for both complex or quaternion matrices| |
```

docs/src/manifolds/generalunitary.md
mateuszbaran: Sure, I'll change that.

src/groups/unitary.jl
```diff
@@ -1,32 +1,35 @@
...
 and we represent tangent vectors by just storing the [`SkewHermitianMatrices`](@ref) ``Y``,
-or in other words we represent the tangent spaces employing the Lie algebra ``\mathfrak{u}(n)``.
+or in other words we represent the tangent spaces employing the Lie algebra ``\mathfrak{u}(n, 𝔽)``.
+
+Quaternionic unitary group is isomorphic to the compact symplectic group of the same dimension.
 
 # Constructor
 
-    Unitary(n)
+    Unitary(n, 𝔽::AbstractNumbers=ℂ)
```
kellertuer, line 22: is positional the normal way we do? I think so but just to check – For Euclidean we have a keyword, because the number of parameters is not fixed but otherwise it is a positional?

mateuszbaran: Also, `Sphere` has positional and `ArraySphere` has keyword. I thin the reasoning is exactly what you wrote.

docs/src/manifolds/generalunitary.md
mateuszbaran: Should I change all `:constant` to be the first in order in groups.md file?

docs/src/manifolds/group.md
```diff
@@ -40,7 +40,7 @@ As a concrete wrapper for manifolds (e.g. when the manifold per se is a group ma
 ```@autodocs
 Modules = [Manifolds]
 Pages = ["groups/GroupManifold.jl"]
-Order = [:type, :function]
+Order = [:type, :function, :constant]
```
kellertuer, line 43: ...and here ;) and in 55 and in 63, 71, 79, 87, 100, 109... ;)

docs/src/manifolds/generalunitary.md
kellertuer: hehe, See my comment below, yes please.

kellertuer (approved):

PR #507, opened 2022-07-21
Speedup CI :magic_wand: 

kellertuer:
This is an idea to speed up CI,
I did a small benchmark, see #144 and it seems from

https://github.com/JuliaDiff/ChainRules.jl/pull/499/files#diff-3b9314a6f9f2d7eec1d0ef69fa76cfabafdbe6d0df923768f9ec32f27a249c63R27-R32

one can set the compiler flags, though I am not sure this works after running a few (ambiguity & diff) tests. Let's see.


kellertuer:
Ok, this will end up as a looooooong guesswork of randomly combining things from other yaml files without me having any clue what I am doing.

mateuszbaran:
That doesn't sound particularly encouraging :slightly_frowning_face:

kellertuer:
Well I have two examples (the actual runtest action
https://github.com/julia-actions/julia-runtest/blob/master/action.yml
one can try to extract manual test runs from for example like
https://github.com/timholy/SnoopCompile.jl/blob/3d34c0cc5d4ccdb7fc23e1e3627728f86ba55500/.github/workflows/ci.yml
but since nothing is documented I could also try to brew some magic potion.

kellertuer:
So to report some speedups (Just checked 1.7)
* Lie groups from 46 minutes to 24.5
* manifolds from 64 minutes to 30 minutes (approx, maybe the fix for Python/Recipes missing)
* and the additional test is currently at 20 minutes (statistics and some manifolds that were faster on normal compilation=

kellertuer:
So times are good, code coverage reports not yet 🙄

mateuszbaran:
Cool, that's significantly faster.

mateuszbaran:
I wonder if compilation mode affects coverage.

kellertuer:
Hm, all 3 runs generate code coverage all 3 are uploaded but 2 seem to be ignored at codecov and just this newest group is reported. I have (again!) no clue why, not even where to start looking, since the logs report successful upload of all 3 reports.

kellertuer:
> I wonder if compilation mode affects coverage.

If it does then this day of work with magic here was completely useless.

kellertuer:
So there were 3 uploads for the last commit to codecov I am just not sure whether the CI confuses something and reports 3 times the Lie group one, it seemed a little bit like that. Let's try again (I updated nightly to follow the same scheme).

kellertuer:
Sure I am aware of that and as far as I can see it should not affect code cov.

kellertuer:
I am out of ideas why 2 of 3 tests report to code coverage, they reach codecov.io but it does not count (at least there is three reports arriving at codecov) – so. If anyone has ideas left – otherwise this is just faster but useless.

kellertuer:
It seems my manual starts of Julia (compared to the runtest action) does not generate the codecov files (that are collected in the next step), but I have no clue why, since I activate that.

kellertuer:
Last idea: I copied the Julia command from the log file (of the runtests action that reports tests) – now both call the same  stuff, but also this does not seem to work. 

So when even copying the command yields that the action does something different (though there is just an _if_ that now runs two identical commands), I am really at the en of my ideas.

Any help welcome, otherwise, this will just stall for now.

.github/workflows/ci.yml
```diff
@@ -27,6 +28,12 @@ jobs:
         env:
           PYTHON: ""
           MANIFOLDS_TEST_GROUP: ${{ matrix.group }}
+        if: ${{ matrix.group == 'test_on_default_compiler' }}
+      - run: julia --color=yes --depwarn=yes --optimize=0 --compile=min --inline=yes --project=@. -e 'import Pkg;include(joinpath(ENV["GITHUB_ACTION_PATH"], "kwargs.jl"));kwargs = Kwargs.kwargs(;coverage = :(true),force_latest_compatible_version = :(auto), julia_args = ["--check-bounds=yes"]);Pkg.test(; kwargs...)
+        shell: bash
```
SaschaMann, line 33: After this I'm definitely convinced https://github.com/julia-actions/julia-runtest/issues/47 either needs fixing or julia-runtest needs a refactoring and better docs to make it easier to understand what kwargs is even doing.

As a quick workaround until then, the command you're running depends on [`kwargs.jl`](https://github.com/julia-actions/julia-runtest/blob/master/kwargs.jl) which depends on [`autodetect-dependabot.jl`](https://github.com/julia-actions/julia-runtest/blob/master/autodetect-dependabot.jl). My suggestion would be to copy those files into `.github/workflows/` until there's a proper fix (those files are MIT-licensed, so as long as you copy [`LICENSE`](https://github.com/julia-actions/julia-runtest/blob/master/LICENSE) to the top of the files or include it somewhere in the repo, e.g. in a `THIRD_PARTY_NOTICE.md`, copying the files is fine).

If you don't want to do that, an alternative would be to essentially copy their contents and strip out unnecessary stuff like the version check.

---

Unrelated to the above:

GHA let us define custom shells, which lets us treat everything within `run: |` as if it was a Julia file. `shell: julia --color=yes --depwarn=yes --optimize=0 --compile=min --inline=yes --project=@. {0}` creates a file with the content of the `run` block above it and replaces `{0}` with the path to that file. This isn't strictly needed but helps us declutter what's in `run` and hopefully make it easier to debug and follow the script.

```suggestion
      - run: |
          import Pkg
          include(joinpath(ENV["GITHUB_WORKSPACE"], ".github", "workflows", "kwargs.jl"))
          kwargs = Kwargs.kwargs(;coverage = :(true),force_latest_compatible_version = :(auto), julia_args = ["--check-bounds=yes"])
          Pkg.test(; kwargs...)
        shell: julia --color=yes --depwarn=yes --optimize=0 --compile=min --inline=yes --project=@. {0}
```

.github/workflows/ci.yml
```diff
@@ -27,6 +28,12 @@ jobs:
         env:
           PYTHON: ""
           MANIFOLDS_TEST_GROUP: ${{ matrix.group }}
+        if: ${{ matrix.group == 'test_on_default_compiler' }}
+      - run: julia --color=yes --depwarn=yes --optimize=0 --compile=min --inline=yes --project=@. -e 'import Pkg;include(joinpath(ENV["GITHUB_ACTION_PATH"], "kwargs.jl"));kwargs = Kwargs.kwargs(;coverage = :(true),force_latest_compatible_version = :(auto), julia_args = ["--check-bounds=yes"]);Pkg.test(; kwargs...)
```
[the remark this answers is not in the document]
SaschaMann: You can ignore this in favour of the comment/suggestion below, if you want.

SaschaMann (commented):
The suggestions are untested.

Let me know if you have any questions, first step would be to copy those two files and then see if it still errors or works :)

.github/workflows/nightly.yml
```diff
@@ -24,3 +25,9 @@ jobs:
         env:
           PYTHON: ""
           MANIFOLDS_TEST_GROUP: ${{ matrix.group }}
+        if: ${{ matrix.group == 'test_on_default_compiler' }}
+      - run: julia --color=yes --depwarn=yes --optimize=0 --compile=min --inline=yes --project='@.' -e 'import Pkg; ENV["MANIFOLDS_TEST_GROUP"]="${{ matrix.group}}"; Pkg.test(;coverage = true, julia_args = ["--check-bounds=yes"])'
+        shell: bash
+        env:
+          PYTHON: ""
```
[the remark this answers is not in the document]
kellertuer: Tanks, as I said for me this is all more like magic, when which keywords in the yaml have which effect ;)

.github/workflows/ci.yml
kellertuer: Oh I could not get the kwargs to work and then noticed it is extra Julia code you have so I removed that for now, but it seems actually necessaray?

kellertuer: Hm I have to really carefully read this, you seem to really know the stuff and I beryl understand what you ware writing, sorry.

.github/workflows/ci.yml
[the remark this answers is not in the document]
kellertuer: the problem is the idea below with all the files and stuff – I will have to really carefully spend an hour or two on that. Keep in mind that this approach here is 3 days, spending a few hours each and I do not seem to understand much yet in this shell/yaml magic.

.github/workflows/ci.yml
kellertuer: Maybe I'll just wait what happens to the action, because I fear I could also spend a few weeks on this otherwise.

SaschaMann: > Oh I could not get the kwargs to work and then noticed it is extra Julia code you have so I removed that for now, but it seems actually necessaray?

The main thing it does that is also necessary here is figuring out the value of `force_latest_compatible_version`. Its value is different if the PR has been opened by CompatHelper (or Dependabot if that ever receives Julia support). The motivation for that behaviour is outlined here: https://github.com/JuliaLang/Pkg.jl/pull/2439 -- However, you can also choose to ignore that for now and remove `force_latest_compatible_version ` or alternatively have two completely different `run` blocks for the normal case vs the dependabot case with an `if`.

> Maybe I'll just wait what happens to the action, because I fear I could also spend a few weeks on this otherwise.

That works as well, I can't promise anything but it's on my list to look at for this week.

SaschaMann: > Hm I have to really carefully read this, you seem to really know the stuff and I beryl understand what you ware writing, sorry.

No worries at all, that just means I have to explain/write things better

kellertuer: Ah, thanks for the explanation :)

And concerning the update for runtest – do not hurry and all is fine :) I am very happy to have the runtests action – this PR (if it finally works) just would speedup our CI from 60 to 20 minutes (doing most of the tests with `-C0 --compile=min`), so the worst thing is – that our CI stays slow.
And – both this and the runtest action are open source things – so – really nothing of a hurry here. Thanks again for providing the action and for considering the addition of the arguments. And your help even here Really appreciated 👍

kellertuer:
I think I'll close this for now, since there is already one newer PR to work on a new CI.

PR #508, opened 2022-07-22
Quotient action for Grassmann

mateuszbaran:
I've implemented the action that defines the Grassmann manifold as a quotient to complete the implementation.


kellertuer (approved):
Nice!

kellertuer:
and from the last commit you might even have fixed code coverage for the two things I mentioned? That would be very cool :) Slowly working towards more coverage!

mateuszbaran:
I actually don't think this will cover any uncovered lines. And I'm not sure if I should call it `MatrixColumnwiseMultiplicationAction` or `MatrixRowwiseMultiplicationAction`, given that it multiplies transpositions of each row?

kellertuer:
> I actually don't think this will cover any uncovered lines. And I'm not sure if I should call it `MatrixColumnwiseMultiplicationAction` or `MatrixRowwiseMultiplicationAction`, given that it multiplies transpositions of each row?

But isn't it actually (just) the application of the rotation (or the unitary matrix) to the rows of a matrix (due to transpose)? I find the name a little clumsy, why not just ? `RowwiseMultiplicationAction`? The the `MultiplicationAction` we also do not include the matrix.

src/groups/rotation_action.jl
```diff
@@ -129,3 +129,60 @@ end
...
+###
+
+@doc raw"""
+    MatrixColumnwiseMultiplicationAction(
+        M::AbstractManifold,
+        On::GeneralUnitaryMultiplicationGroup,
+        AD::ActionDirection = LeftAction(),
+    )
+
+Space of actions of the (special) unitary or orthogonal group [`GeneralUnitaryMultiplicationGroup`](@ref)
```
kellertuer, line 142: While checking for the name – the type is called action, but the documentation says its the space of the action., I think it is the second? Also the docs could include the subtype info (in line140)

mateuszbaran:
> But isn't it actually (just) the application of the rotation (or the unitary matrix) to the rows of a matrix (due to transpose)? I find the name a little clumsy, why not just ? `RowwiseMultiplicationAction`? The the `MultiplicationAction` we also do not include the matrix.

Yes, `RowwiseMultiplicationAction` is a better name :+1: .

src/groups/rotation_action.jl
mateuszbaran: So, I'm not sure -- do you prefer the "action" or "space of actions" name?

I'll include the subtype info and move the constructor signature below.

kellertuer: I just felt that, when the type is called action, the docs should not say its the space of actions?

mateuszbaran: OK, I see, I'll remove "Space of" then.

mateuszbaran:
So I think this is ready to be merged now?

PR #509, opened 2022-07-23
Sasaki retraction and rework of tangent bundle

mateuszbaran:
This PR adds the Sasaki metric/retraction to tangent bundles. Our current exp/log/distance/PT on tangent bundles are not correct for any reasonable metric, and the Sasaki metric is supposed to be the new "default" metric for tangent bundles -- with the caveat that there is no general closed-form exp or log for the Sasaki metric, so we just have an iterative scheme for exp used in literature. The former exp, log and PT become a retraction, inverse retraction and a vector transport instead.

This still needs some tests and docs explaining the reasoning behind this.

This should be the last significant change to Manifolds.jl before ManifoldDiff.jl -- Sasaki retraction is supposed to be used in the updated geodesic regression example. There are still a couple of smaller changes like `apply_diff_group` but that should be much less work.


src/manifolds/VectorBundle.jl
```diff
@@ -617,23 +529,37 @@ Notation:
     canonical projection of that vector bundle $B$.
     Similarly, $q = (x_q, V_q)$.
 
-The logarithmic map is calculated as
+The inverse retraction is calculated as
 
 $\log_p q = (\log_{x_p}(x_q), V_{\log} - V_p)$
```
kellertuer, line 534: Maybe also adapt the first part here, this make it clearer that the retraction works as element wise things.

```suggestion
$\operatorname{retr}_p q = (\log_{x_p}(x_q), V_{\log} - V_p)$
```

should the retraction maybe have a parameter, which vector transport to use (and the default is `default_vector_transport(M)` when constructing it)?

kellertuer:
Very nice idea; sure I implemented that wrongly until now. We could also do a vector transport based on the same idea as the retraction (this would be what you deleted from `parallel_transport`), just the name is a little complicated, since in your Scheme it would be
`VectorBundleProductVectorTransport`?

src/manifolds/VectorBundle.jl
mateuszbaran: Currently it uses the vector transport stored in the `VectorBundle` object. What it lacks currently is the ability to define the inverse retraction on the manifold that should be used (currently it just does `log`). I will check documentation a bit later.

mateuszbaran:
We actually already have `VectorBundleVectorTransport`, though introduced for a different reason :smiley: .

src/manifolds/VectorBundle.jl
kellertuer: Ah, that sounds also fine, I was not aware, the bundle would store its transport.

kellertuer:
> We actually already have `VectorBundleVectorTransport`, though introduced for a different reason 😃 .

But does it what I was proposing to do? Then its fine :)

mateuszbaran:
It does almost the same thing.

mateuszbaran:
Finally Manopt.jl integration test runs fine :tada:

kellertuer:
🎉 great! I am travelling today but I will try to find time to review this.

docs/src/manifolds/vector_bundle.md
```diff
@@ -23,21 +23,19 @@ It is used for example in musical isomorphisms (the [`flat`](@ref) and [`sharp`]
 
 ## Example
 
-The following code defines two points on a tangent bundle of the sphere $S^2$ and calculates distance between them, distance between their base points and norm of one of these tangent vectors.
+The following code defines a point on a tangent bundle of the sphere $S^2$ and a tangent vector to that point and calculates an approximation of the exponential in the Sasaki metric using 1000 steps.
```
kellertuer, line 26:
```suggestion
The following code defines a point on the tangent bundle of the sphere $S^2$ and a tangent vector to that point and calculates an approximation of the exponential in the Sasaki metric using 1000 steps.
```

kellertuer: That line is also a little long, we could add some new lines in the code maybe.

src/Manifolds.jl
```diff
@@ -247,7 +247,6 @@ using ManifoldsBase:
     @invoke_maker,
     _euclidean_basis_vector,
     combine_allocation_promotion_functions,
-    default_inverse_retraction_method,
```
kellertuer, line 250: where did this line go? I would prefer not to delete that.

src/manifolds/ConnectionManifold.jl
```diff
@@ -335,6 +335,24 @@ function riemann_tensor(
 end
 @trait_function riemann_tensor(M::AbstractDecoratorManifold, p, B::AbstractBasis; kwargs...)
 
+@doc raw"""
```
kellertuer, line 338: I thought we would have that already, but since we do not seems so – Nice to have it :) It would even be a candidate to be moved to `ManifoldsBase.jl` at some point.

src/manifolds/VectorBundle.jl
```diff
@@ -169,6 +201,55 @@ struct VectorBundleBasisData{BBasis<:CachedBasis,TBasis<:CachedBasis}
     vec_basis::TBasis
 end
 
+@doc raw"""
```
kellertuer, line 204: We could add a comment that explains the product in the name? Something like that it works element wise?

src/manifolds/VectorBundle.jl
```diff
@@ -169,6 +201,55 @@ struct VectorBundleBasisData{BBasis<:CachedBasis,TBasis<:CachedBasis}
...
+
+where $V_{\log}$ is the result of vector transport of $V_q$ to the point $x_p$.
+The difference $V_{\log} - V_p$ corresponds to the logarithmic map in the vector space $F$.
+
+See also [`VectorBundleProductRetraction`](@ref).
+"""
+struct VectorBundleInverseProductRetraction <: AbstractInverseRetractionMethod end
+
+@doc raw"""
+    struct VectorBundleProductRetraction <: AbstractRetractionMethod end
```
kellertuer, line 228: Same as above, currently an inexperienced user (or a reviewer with a little bit of jet lag) might wonder why `Product` is in the name.

src/manifolds/VectorBundle.jl
```diff
@@ -943,10 +1073,43 @@ function vector_transport_direction(M::VectorBundle, p, X, d)
     return vector_transport_direction(M, p, X, d, M.vector_transport)
 end
 
+function _vector_transport_direction(
+    M::VectorBundle,
+    p,
+    X,
+    d,
+    m::VectorBundleVectorTransport,
```
kellertuer, line 1081: For consistency this could also get the Product in its name?

kellertuer (commented):
So the VectorBundleProductRetraction is the Sasaki retraction from the PRs title?

Otherwise: Nice, to see some `riemannian_tensor`s to pop up, I think I might need them some time in the future. We could add this to Base also?

src/Manifolds.jl
```diff
@@ -582,7 +581,8 @@ export AbstractRetractionMethod,
     ODEExponentialRetraction,
     PadeRetraction,
     ProductRetraction,
-    PowerRetraction
+    PowerRetraction,
+    SasakiRetraction
```
kellertuer, line 585: Where is this used/implemented?

kellertuer: Sorry I missed that, never mind ;)

src/manifolds/VectorBundle.jl
```diff
@@ -804,6 +817,123 @@ function Random.rand!(rng::AbstractRNG, M::TangentSpaceAtPoint, X; vector_at=not
...
+    SasakiRetraction(L::Int)
+
+In this constructor `L` is the number of integration steps.
+
+[^Muralidharan2012]:
+    > P. Muralidharan and P. T. Fletcher, “Sasaki Metrics for Analysis of Longitudinal Data
+    > on Manifolds,” Proc IEEE Comput Soc Conf Comput Vis Pattern Recognit, vol. 2012,
+    > pp. 1027–1034, Jun. 2012, doi: 10.1109/CVPR.2012.6247780.
+"""
+struct SasakiRetraction <: AbstractRetractionMethod
```
kellertuer, line 881: Could we put types/structs at the beginning of the file?

src/manifolds/VectorBundle.jl
```diff
@@ -1002,6 +1155,16 @@ function vector_transport_to!(
...
 
+@inline function Base.view(x::ArrayPartition, M::VectorBundle, s::Symbol)
+    if s === :point
+        return x.x[1]
+    elseif s === :vector
+        return x.x[2]
+    else
+        throw(DomainError(s, "unknown component $s on $M."))
+    end
+end
```
kellertuer, line 1166:
```suggestion
    (s === :point) && return x.x[1]
    (s === :vector) && return x.x[2]
    throw(DomainError(s, "unknown component $s on $M."))
end
```

this is of course equivalent, I just find this nicer in style, or is there a benefit to use if?

docs/src/manifolds/vector_bundle.md
```diff
@@ -23,21 +23,19 @@ It is used for example in musical isomorphisms (the [`flat`](@ref) and [`sharp`]
...
 M = Sphere(2)
 TB = TangentBundle(M)
 p = ProductRepr([1.0, 0.0, 0.0], [0.0, 1.0, 3.0])
-q = ProductRepr([0.0, 1.0, 0.0], [2.0, 0.0, -1.0])
-println("Distance between p and q: ", distance(TB, p, q))
-println("Distance between base points of p and q: ", distance(M, p[TB, :point], q[TB, :point]))
-println("Norm of p: ", norm(M, p[TB, :point], p[TB, :vector]))
+X = ProductRepr([0.0, 1.0, 0.0], [0.0, 0.0, -2.0])
+println("Approximation of the exponential map: ", exp(TB, p, q, SasakiRetraction(1000)))
 ```
```
kellertuer, line 36: This starts a little spontaneous directly after the example. can we maybe first do these auto docs and then do an extra section for the example?

mateuszbaran:
> So the VectorBundleProductRetraction is the Sasaki retraction from the PRs title?

No, `VectorBundleProductRetraction` is the old `exp`. Sasaki retraction is `SasakiRetraction` :wink: .

> Otherwise: Nice, to see some `riemannian_tensor`s to pop up, I think I might need them some time in the future. We could add this to Base also?

First time they actually turn out to be useful :slightly_smiling_face: . Sure, I can move them to Base later.

kellertuer:
> No, `VectorBundleProductRetraction` is the old `exp`. Sasaki retraction is `SasakiRetraction` 😉 .

Yes, took me a while to see Sasaki in this PR.

> First time they actually turn out to be useful 🙂 . Sure, I can move them to Base later.

An idea could be to add them even now, to avoid a breaking change later on (when removing them here).

src/manifolds/VectorBundle.jl
mateuszbaran: They are essentially equivalent, I just find the if-elseif variant a bit more intuitive but I'm fine with either.

src/manifolds/VectorBundle.jl
mateuszbaran: Sure, I'll move it.

src/manifolds/VectorBundle.jl
kellertuer: I find the other shorter and more readable, but I also do not have a strong opinion on this (we use my proposed style in quite a few places I think).

src/manifolds/VectorBundle.jl
mateuszbaran: Yes, the idea is to apply the retraction, inverse retraction or vector transport element-wise, like on a product manifold. I will add a comment about it.

src/manifolds/VectorBundle.jl
mateuszbaran: We could, though technically it would be breaking. `VectorBundleVectorTransport` has existed for a long time.

mateuszbaran:
> An idea could be to add them even now, to avoid a breaking change later on (when removing them here).

The neat thing is that moving things to ManifoldsBase.jl isn't breaking because we only have explicit imports from ManifoldsBase into Manifolds module. Anyway, I can open a PR to Base now if you prefer.

docs/src/manifolds/vector_bundle.md
mateuszbaran: OK, I'll try to rephrase it.

kellertuer:
> > An idea could be to add them even now, to avoid a breaking change later on (when removing them here).
> 
> The neat thing is that moving things to ManifoldsBase.jl isn't breaking because we only have explicit imports from ManifoldsBase into Manifolds module. Anyway, I can open a PR to Base now if you prefer.

I do not have a strong preference, I just think it is basic enough that it should be in ManifoldsBase, so we could even move it there now :)

src/Manifolds.jl
mateuszbaran: `default_inverse_retraction_method` is actually imported in line 27. This `using` confused me at one point because I needed to add a method to this function but this `using` overrides the `import` statement and disallows adding new methods without specifying module.

kellertuer: oh, I dod not see that. We should be more careful here in the future then that import and using are disjoint (and ordered alphabetically).

docs/src/manifolds/vector_bundle.md
mateuszbaran: OK, I'll put the example lower.

src/manifolds/VectorBundle.jl
mateuszbaran: I've added an explanation.

kellertuer: That is exactly what helps here, great :)

src/manifolds/VectorBundle.jl
kellertuer: We could also introduce an additional `VectorBundleProductVectorTransport` and not break things ;)

mateuszbaran: You mean as an alias?

kellertuer: or as a second type.

mateuszbaran: But would it do anything differently? If not, should the old one be deprecated?

kellertuer: I am not 100% sure what the old one did, since now it is a new implementation, is the old one is now useless, then one could deprecate it, yes. If it has some use we just could keep both?

mateuszbaran: This is the old one, it had `vector_transport_to` defined before. I've just added `vector_transport_direction` here.

kellertuer: Hm, that was chaotic then, because someone (me?) implemented parallel transport (to?) as the component wise wile someone (you?) did the same using a retraction.

Good that it is now unified. But then I would prefer to use `Product` within that for now, unless it makes sense to omit `Product` for all three, retract/inverse retract/vector transport, because it gets clear that the VectorBundleX is the one meant element wise.
If that gets not clear, all should get the Product in their name.

mateuszbaran: OK, I'll add `Product` then, and maybe keep the old name as a deprecated alias?

kellertuer: that would be my preferred way, yes.

mateuszbaran: OK, done. There is a `Base.@deprecate_binding` which I was trying to use instead of a constant but it doesn't seem to do anything, i.e. I could still use the old binding without any warnings in REPL.

kellertuer: at least in my changes here line 1081 still does not have the `Product` inside, is that a bug in GitHub or did you miss to update that?

kellertuer (approved):
I would still prefer to move `riemannian_tensor` to ManifoldsBase already now, but independent of that this PR looks fine.

src/manifolds/VectorBundle.jl
mateuszbaran: Line 1081 is the empty line between two methods of `vector_space_dimension` so probably a Github issue :slightly_smiling_face: .

mateuszbaran:
OK, I'll make a PR to base with riemann_tensor before merging this.

Issue #510, opened 2022-07-25
StackOverflowError in manifold_dimension(Unitary(1,ℍ))


Affie:
Hi, I started experimenting with Unitary for quaternions and came across this error. 
I also wanted to see how `get_coordinates` and `get_vector` worked, but couldn't get them to work.
It can be low priority from my side as we use SO(3), but just wanted to let you know. 
```julia
julia> M = Unitary(1,ℍ)
Unitary(1, ℍ)

julia> manifold_dimension(M)
ERROR: StackOverflowError:
Stacktrace:
 [1] manifold_dimension(M::UnitaryMatrices{1, ℍ}) (repeats 79984 times)
   @ ManifoldsBase ~/.julia/packages/ManifoldsBase/SWdl9/src/decorator_trait.jl:4
```

mateuszbaran:
Sorry, I didn't check that -- I will take a look at it later today.

kellertuer:
Ah that might just be a dispatch missing. Can you maybe post what is not working on the `get_ooordinates`/`get_vector` side?

mateuszbaran:
@kellertuer this seems to be actually not implemented. `get_coordinates` doesn't work at all.

kellertuer:
Maybe, I do not remember, I mainly took over what was in the previous PR.

mateuszbaran:
Manifolds.jl 0.8.23 should have this resolved. Feel free report if there are any further problems :slightly_smiling_face: .

Affie:
Thanks, I tried and thought I'd just mention one thing that caught me with `get_coordinates` (it can perhaps help someone in the future).

I didn't know that the basis also needed the number system, so I tried this:
```julia
get_coordinates(M, p, X, DefaultOrthonormalBasis())
ERROR: MethodError: no method matching similar(::Quaternions.QuaternionF64, ::Type{Quaternions.QuaternionF64}, ::Int64)
Closest candidates are:
  similar(::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{N}}, ::Type{ElType}, ::Any) where {N, ElType} at ~/.julia/juliaup/julia-1.7.3+0.x64/share/julia/base/broadcast.jl:212
[...]

PR #511, opened 2022-07-25
More methods for unitary

mateuszbaran:
Addresses #510 . For now implements `manifold_dimension` and `rand` but maybe more will be needed.


mateuszbaran:
I think we don't have an alias for the manifold of special unitary matrices?

mateuszbaran:
@kellertuer  Just to check: the real coefficient basis of quaternionic unitary should be `DefaultOrthonormalBasis(ℍ)`? And `DefaultOrthonormalBasis(ℝ)` would be the quaternionic coefficient basis?

kellertuer (approved):

PR #513, opened 2022-07-26
Fix an ambiguity from RealCircleGroup

### LINKED ISSUE #512: Possible ambiguous dispatch for identity_element on RealCircleGroup
Hi, the error stack seems suggests an ambiguous dispatch, I'm going to investigate further and find a way to work around this in the mean time.  Will post here again if I learn anything new:

```julia
  MethodError: identity_element(::AbstractManifold, ::Vector{Float64}) is ambiguous. Candidates:
    identity_element(::RealCircleGroup, p::AbstractArray) in Manifolds at /home/dehann/.julia/packages/Manifolds/0EI8E/src/groups/circle_group.jl:170
    identity_element(G::AbstractDecoratorManifold, p) in Manifolds at /home/dehann/.julia/packages/ManifoldsBase/gTRkF/src/nested_trait.jl:274
  To resolve the ambiguity, try making one of the methods more specific, or adding a new method more specific than any of the existing applicable methods.
  Stacktrace:
    [1] identity_element
      @ ~/.julia/packages/ManifoldsBase/gTRkF/src/nested_trait.jl:308 [inlined]
    [2] identity_element (repeats 4 times)
      @ ~/.julia/packages/ManifoldsBase/gTRkF/src/nested_trait.jl:282 [inlined]
    [3] identity_element
      @ ~/.julia/packages/ManifoldsBase/gTRkF/src/nested_trait.jl:275 [inlined]
    [4] #manikde!#51
```

Manifolds v0.8.21
ManifoldsBase v0.13.15

kellertuer:
that was left from the time where we had implicit power manifold groups. 

Resolves #512.

### DIFF
--- a/src/groups/circle_group.jl
+++ b/src/groups/circle_group.jl
@@ -169,6 +169,8 @@ end
 identity_element(G::RealCircleGroup) = 0.0
 identity_element(::RealCircleGroup, p) = zero(p)
 
+Base.inv(G::RealCircleGroup, p::AbstractArray{<:Any,0}) = map(pp -> inv(G, pp), p)
+
 function inverse_translate(
     ::RealCircleGroup,
     p::AbstractArray{<:Any,0},
--- a/src/groups/circle_group.jl
+++ b/src/groups/circle_group.jl
@@ -167,9 +167,7 @@ function compose!(
 end
 
 identity_element(G::RealCircleGroup) = 0.0
-identity_element(::RealCircleGroup, p::AbstractArray) = map(i -> zero(eltype(p)), p)
-
-Base.inv(G::RealCircleGroup, p::AbstractArray{<:Any,0}) = map(pp -> inv(G, pp), p)
+identity_element(::RealCircleGroup, p) = zero(p)
 
 function inverse_translate(
     ::RealCircleGroup,
@@ -192,4 +190,4 @@ function exp_lie(::RealCircleGroup, X)
     return sym_rem(X)
 end
 
-exp_lie!(G::RealCircleGroup, q, X) = (q .= sym_rem(X))
+exp_lie!(::RealCircleGroup, q, X) = (q .= sym_rem(X))
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.8.21"
+version = "0.8.22"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
--- a/test/ambiguities.jl
+++ b/test/ambiguities.jl
@@ -17,7 +17,7 @@
         # Interims solution until we follow what was proposed in
         # https://discourse.julialang.org/t/avoid-ambiguities-with-individual-number-element-identity/62465/2
         fms = filter(x -> !any(has_type_in_signature.(x, Identity)), ms)
-        FMS_LIMIT = 37
+        FMS_LIMIT = 39
         println("Number of Manifolds.jl ambiguities: $(length(fms))")
         if length(fms) > FMS_LIMIT
             for amb in fms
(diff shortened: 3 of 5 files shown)


src/groups/circle_group.jl
```diff
@@ -167,9 +167,7 @@ function compose!(
 end
 
 identity_element(G::RealCircleGroup) = 0.0
-identity_element(::RealCircleGroup, p::AbstractArray) = map(i -> zero(eltype(p)), p)
-
-Base.inv(G::RealCircleGroup, p::AbstractArray{<:Any,0}) = map(pp -> inv(G, pp), p)
+identity_element(::RealCircleGroup, p) = zero(p)
```
mateuszbaran, line 170: I don't remember why we had such an elaborate `identity_element(::RealCircleGroup, p::AbstractArray)` before but this looks fine.

But why did you delete that `inv`?

kellertuer: The reason was that the first implementation was (following the general scheme in Matlab maybe?) both the Circle group a well as implicitly the PowerCirclegroup (implicitly working element-wise when you came along with vectors).

Inv for this group is generically defined (see that it even calls the generic `inv(G, pp)` but also line 172, the inv I deleted, is also from the interpretation of implicitly assuming it might also be its own power manifold.

The reason we changed that to explicitly require a power manifold is a) being more concise and b) exactly avoiding ambiguities like the oner we now had here left ;)

mateuszbaran: Wasn't that `inv` method there to actually support points as zero-index arrays? Note that `inv` doesn't work on zero-index arrays. `inv(fill(1))` throws an error.

kellertuer: Oh, maybe I missed the zero in the code, yes, you might be right.

kellertuer: But we do not have a test case for that then – even more if we introduce this for all other functions (for example also `identity_element` we have a lot of ambiguities to fix.

mateuszbaran: But we don't have any `inv` ambiguities currently as far as I can see and the `inv` method is covered according to Codecov.

kellertuer: you're right, for inv we do not do the trait so we do not have that problem. I already reintroduced that line. We might just run into this if we have further functions where we have to consider special cases.

dehann:
hi, following from the comment above about test coverage.  I have another known (but maybe helpful to repeat) comment.  Test coverage is necessary but not sufficient.  In the sense that when a test passes through a line of code, it is not testing all possible permutations of how one can end up on that line of code.  Hence bugs can slip through.

TLDR; 100% coverage can be misleading.  

e.g. the IIF package we're working on has "poor" advertised coverage at ~75%, but what one doesn't see is that many of the lines of code are tested from many different approaches, so it's more like we have 4x the 70% coverage.

kellertuer:
Our test suite runs about an hour and also tests a lot of things and code from different directions / cases.
Sice it runs for an hour, we for example did not yet add the case of 1-element-vectors for the real circle group (with that we would have noticed your bug for sure). We for example perform 35 667 test on 8651 (of our 8750) lines of code.

Since per cent is percent I find 4 x 70% misleading, but you are right, code coverage is just an indicator that probably there is less bugs, it does not say, bugs are absent.
On the other hand – even in your image of permutations – testing for all possibilities that might ever appear – is, like for permutations, probably exponential in effort (both design of tests as well as running).

What a code coverage _does_ say is that someone thought about 98.86% of the lines in the sense that besides implementation they also considered to check that what they expect (yes only in certain cases) does actually happen.

kellertuer:
...but sure, what I did not do yet but just added is a test that checks for the bug you encountered that it is now fixed. That will not increase code coverage but takes into account that we did not yet check this specific case yet and now do.

mateuszbaran (approved):
This looks fine, I've just added one test that should improve code coverage. Of course, tests don't guarantee that code is correct but it's still nice to have them.

kellertuer:
I would phrase that as: They are a very reasonable step towards more trust in the code, that there are less bugs.

dehann:
> I would phrase that as: They are a very reasonable step towards more trust in the code, that there are less bugs.

yeah, definitely!

> On the other hand – even in your image of permutations – testing for all possibilities that might ever appear – is, like for permutations, probably exponential in effort (both design of tests as well as running).

I should have prefaced better as "random comment on test coverage" :-)  Over by IIF we're definitely feeling that exponential growth in effort as we increase the dispatch and consolidation.  Tests also take around an hour there with a lot of duplication, but each test is slight permutation which was discovered somewhere along the line by usage in the wild.

If it's okay, I'll keep posting these permutation type issues as and when they happen -- might be fine to just close such issues as silly mistake, but will help document for general audience via the issue history?

Thanks for clarifying on Circle group, and the fix!

kellertuer:
Sure, whenever you encounter something where you think it is a bug, please report them here :)

This time, for 1-element arrays you encountered a bug and for the others you learned something – a perfect mix :)

PR #514, opened 2022-07-30
Speeding up `minkowski_metric` and a few minor things

mateuszbaran:
A result of inspecting Manifolds.jl performance for high-dimensional manifolds.


mateuszbaran:
@kellertuer The Manopt.jl test failure looks like a bug in Manopt. It tries to compute distance on `Euclidean(3)` between `[[0.6928694505262161, 2.14681331938508, 2.423689685165601], [0.3972579209670175, 1.8168637330208202, 1.3730336083935892], [0.8031804037077928, 3.2129623034851864, 3.2508546075478995]]` and `[[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]]` which don't look like correct input.

kellertuer:
You open quite a lot of Pas currently ;)

Do you mean the integration test failing? It fails now and then for a few weeks now, I just did not yet have time to investigate why one of the differentials started failing.

kellertuer:
Ah I have an idea for the error but not yet a solution.

For particle swarm, the iterate is on the manifold but since it is modelled by a swarm, the stopping criterion actually is done on the power manifold. With the old implementation, Euclidean did not care much, now it does. But I have no clue how to fix that (for now).

mateuszbaran:
> You open quite a lot of Pas currently ;)

Yes, there are a few things that I would like to get done before I have too many other things to do :slightly_smiling_face: .

> For particle swarm, the iterate is on the manifold but since it is modelled by a swarm, the stopping criterion actually is done on the power manifold. With the old implementation, Euclidean did not care much, now it does. But I have no clue how to fix that (for now). So that is why it actually is the right input.

Maybe particle swarm just shouldn't use `StopWhenChangeLess` but a different, population-oriented criterion? There are other evolutionary algorithms that could use that as well in the future.

kellertuer:
Oh it‘s great that you do a few PRs :)

I do have a solution, and hope I get it done quickly; I am just dispatching for particle swarm on a specific version of the `StopWhenChangeLess` that computes the change of the whole population then.

kellertuer:
I just checked and had to fix one more thing, but now it works again with this far more restrictive Euclidean distance.

mateuszbaran:
Thanks! I think this is still a reasonable implementation, and for high-dimensional Euclidean (I've tested `Eucliden(2^20)`) it's about 10x faster than the previous implementation.

kellertuer:
That‘s fine, it is just a little more restrictive than the old one; but the things it does break were then just too lazy before.

kellertuer (approved):

PR #515, opened 2022-08-02
Introduce SPDPoint

kellertuer:
For linear affine metic (or maybe better called affine in the future?) a lot of functions require the eigenvalues of `p` or even its sqrt and sort_inverse.

This PR introduces a point type to store the eigenvalues and optionally `p.sqrt` and `p.sqrt_inv`. new access functions dispatch on the type of the point and either compute a value if it is missing (or we have a usual matrix) or return the stored value.
If the `exp(M,p,X)` is called with an `SPDPoint` or `exp!(M,q,p,X)` has `q` as an `SPDPoint` its eigenvalues are computed when it is created.

# Todo

* documentation
* test coverage


kellertuer:
It seems a few functions to not like custom point types that much

* `copy(M,p)` – ok, that one might need special handling or some new allocation thing
* `copy(M,p,X)` – also complains in allocation, since we also plug the p into there (necessary?)

Not yet sure how to best fix these two.

src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
```diff
@@ -6,6 +6,146 @@ matrix logarithms and exponentials, which yields a linear and affine metric.
...
+"""
+function get_p_sqrt(p::AbstractMatrix)
+    e = eigen(Symmetric(p))
+    U = e.vectors
+    S = max.(e.values, floatmin(eltype(e.values)))
+    Ssqrt = Diagonal(sqrt.(S))
+    return Symmetric(U * Ssqrt * transpose(U))
+end
+get_p_sqrt(p::SPDPoint) = Symmetric(p.sqrt)
+function get_p_sqrt(p::SPDPoint{P,Missing}) where {P}
```
mateuszbaran, line 92:
```suggestion
function get_p_sqrt(p::SPDPoint{P,Missing}) where {P<:AbstractMatrix}
```
This should remove the ambiguity I think.

mateuszbaran:
How exactly do these `copy` errors look like?

kellertuer:
> How exactly do these `copy` errors look like?

```
using Manifolds, LinearAlgebra
n=3
p = SPDPoint(Matrix{Float64}(I,n,n))
M = SymmetricPositiveDefinite(n)
q = copy(M,p)
```

yields

```
ERROR: MethodError: no method matching one(::Type{Any})
Closest candidates are:
  one(::Type{Union{Missing, T}}) where T at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/base/missing.jl:105
  one(::Union{Type{T}, T}) where T<:AbstractString at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/base/strings/basic.jl:262
  one(::Union{Type{P}, P}) where P<:Dates.Period at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Dates/src/periods.jl:54
  ...
Stacktrace:
  [1] one(#unused#::Type{Any})
    @ Base ./missing.jl:106
  [2] eti_to_one
    @ ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:121 [inlined]
  [3] map
    @ ./tuple.jl:221 [inlined]
  [4] allocate_result_type(#unused#::SymmetricPositiveDefinite{3}, f::typeof(copy), args::Tuple{SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}}})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:122
  [5] allocate_result(M::SymmetricPositiveDefinite{3}, f::Function, x::SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:101
  [6] allocate_result
    @ ~/Repositories/Julia/ManifoldsBase.jl/src/decorator_trait.jl:105 [inlined]
  [7] allocate_result (repeats 3 times)
    @ ~/Repositories/Julia/ManifoldsBase.jl/src/decorator_trait.jl:115 [inlined]
  [8] allocate_result(M::SymmetricPositiveDefinite{3}, f::Function, x::SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/decorator_trait.jl:92
  [9] copy(M::SymmetricPositiveDefinite{3}, p::SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:226
 [10] top-level scope
    @ REPL[10]:1
```

I already implemented a one-line zero vector (basically calling `get_point` before calling zero vector again), so we can do

```
X = zero_vector(M,p)
Y = copy(M, p, X)
```

yields as well

```
ERROR: MethodError: no method matching one(::Type{Any})
Closest candidates are:
  one(::Type{Union{Missing, T}}) where T at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/base/missing.jl:105
  one(::Union{Type{T}, T}) where T<:AbstractString at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/base/strings/basic.jl:262
  one(::Union{Type{P}, P}) where P<:Dates.Period at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Dates/src/periods.jl:54
  ...
Stacktrace:
  [1] one(#unused#::Type{Any})
    @ Base ./missing.jl:106
  [2] eti_to_one
    @ ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:121 [inlined]
  [3] map
    @ ./tuple.jl:222 [inlined]
  [4] allocate_result_type(#unused#::SymmetricPositiveDefinite{3}, f::typeof(copy), args::Tuple{Matrix{Float64}, SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}}})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:122
  [5] allocate_result(::SymmetricPositiveDefinite{3}, ::Function, ::Matrix{Float64}, ::Vararg{Any})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:101
  [6] allocate_result
    @ ~/Repositories/Julia/ManifoldsBase.jl/src/decorator_trait.jl:105 [inlined]
  [7] allocate_result (repeats 3 times)
    @ ~/Repositories/Julia/ManifoldsBase.jl/src/decorator_trait.jl:115 [inlined]
  [8] allocate_result(::SymmetricPositiveDefinite{3}, ::Function, ::Matrix{Float64}, ::SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/decorator_trait.jl:92
  [9] copy(M::SymmetricPositiveDefinite{3}, p::SPDPoint{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}}, X::Matrix{Float64})
    @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/ManifoldsBase.jl:240
 [10] top-level scope
    @ REPL[12]:1
```

So in both cases the problem is that the tricky way to determine the eltype does not work? My idea would be to check what we did for the fixed rank matrices, but I did not yet have time for that.

kellertuer:
That's what I thought and will do hopefully later today.

kellertuer:
Most things should be in the docs, the functions you mentioned are implemented (I did not do the fancy broadcast for points, that we have for example for `UMVTVector`) and I hope the allocation is fine, I was not 10% sure what to do about Eigen there.

mateuszbaran:
Broadcasting for `SVDMPoint` should not be necessary, I will check the rest.

kellertuer:
Hm, I ran the whole test_manifolds suite on `SPDPoints` and the code coverage basically did not change. I had hoped this would have resolved at least 90% of the lines.

mateuszbaran:
That's because your condiditon, `(M != M1 || M != M2)`, is always true. I think it should be `(M != M1 && M != M2)

kellertuer:
Oops ;)

kellertuer:
Now they run and it is “just” 15 left, mainly allocation problems 8calling ^similar` where I had hoped to have fixed allocate) and sometimes when we need a vector still an `SPDPoint` is allocated instead of a matrix for the vector.

kellertuer:
Maybe a little bit of help would be great. For the last hour the functions I added only increased the number of tests and never again decreased them, so I seem to miss which functions to implement (I was hoping just a few `allocate_result` ones, maybe for `log` or `zero_vector` or `get_coordinates`?)

kellertuer:
When writing this I was not so happy with the access functions for the `SPDPoint` sind `get_point(p)`is misleading, `p`_is_ already the point.

So I propose to change that as follows

* `get_point(p)` should be `Matrix(p)`(the identity if `p`_is_ a matrix is automatic)
* `get_p_sqrt(p)` would just be `sqrt`
* `get_p_sqrt_inv`would be `sqrt_inv`

And just the one that does the last two together I do not yet have a good name for. Luckily these are just access functions so it should not interfere with the remaining allocation/similar fixes we need.

kellertuer:
I think the main bugs remaining are

* an interplay of zero vector and tangent bundle, where my specific zero vector method does not get used and hence the allocation fails
* some coordinate function that allocates an SPD point for X

mateuszbaran:
> * `get_point(p)` should be `Matrix(p)`(the identity if `p`_is_ a matrix is automatic)
> * `get_p_sqrt(p)` would just be `sqrt`
> * `get_p_sqrt_inv`would be `sqrt_inv`
> 
> And just the one that does the last two together I do not yet have a good name for. Luckily these are just access functions so it should not interfere with the remaining allocation/similar fixes we need.
> 
> _edit:_ Oh, `sqrt` would have been a really bad type pyracy for `AbstractMatrix`since we assume it is spd, so I used `_sqrt`, `_sqrt_inv` and `_sqrt_and_sqrt_inv` for the access functions but kept the first idea with `Matrix` to materialise the point into a matrix.

The new `Matrix` conversion is a good idea but maybe `eigvals_sqrt` and `eigvals_sqrt_inv` would be more appropriate for the other two? After all they represent square roots (or their inverses) of eigenvalues of `p`.

kellertuer:
sounds good as well, I do not plan to export them anyways, so we can change that, sure.

mateuszbaran:
I'm doing some updates here currently so let me finish them before you change too much :slightly_smiling_face: .

kellertuer:
I am done with renaming, so I will not commit further for now.

kellertuer:
Oh I missed one `eigvals`prefix in line 365 of SymmetricPositiveDefinite.jl.

kellertuer:
Checking the new fix of ManifoldsBase locally, it seems that was the only remaining error? Super cool!

There is then just a few functions left to check like `show`.

mateuszbaran:
Yes, things should now work. I've uploaded my changes but they may not pass until a new version of ManifoldsBase.jl is tagged.

kellertuer:
Hm I maybe accidentally wrote comments to the commit instead of to the PR.

src/manifolds/SymmetricPositiveDefinite.jl
```diff
@@ -30,10 +30,100 @@ struct SymmetricPositiveDefinite{N} <: AbstractDecoratorManifold{ℝ} end
...
+        q,
+        e,
+        p_sqrt,
+        p_sqrt_inv,
+    )
+end
+
+convert(::Type{SPDPoint}, p::AbstractMatrix) = SPDPoint(p)
+
+function Base.:(==)(p::SPDPoint, q::SPDPoint)
```
kellertuer, line 92: What about inconsistent values within the other three fields? Someone might set them (and evil enemy of Manifolds.jl that is).

src/Manifolds.jl
```diff
@@ -151,9 +151,11 @@ import Base:
     inv,
     isempty,
     length,
+    Matrix,
```
kellertuer, line 154: This might be unnecessary if we now only use the convert approach?

src/manifolds/SymmetricPositiveDefinite.jl
```diff
@@ -30,10 +30,100 @@ struct SymmetricPositiveDefinite{N} <: AbstractDecoratorManifold{ℝ} end
...
+    end
+    return SPDPoint{typeof(q),typeof(p_sqrt),typeof(p_sqrt_inv),typeof(e)}(
+        q,
+        e,
+        p_sqrt,
+        p_sqrt_inv,
+    )
+end
+
+convert(::Type{SPDPoint}, p::AbstractMatrix) = SPDPoint(p)
```
kellertuer, line 90: I think I defined a `Matrix` variant here, right? My idea was (similar to `I` that it would “materialize” the matrix into its array form – and you now use a convert (a different one from this) to do that. What is the advantage?

src/manifolds/SymmetricPositiveDefinite.jl
mateuszbaran: In such case -- garbage in, garbage out :wink: . Checking for all possible combinations of missing and non-missing for potential inconsistencies would be quite convoluted.

src/Manifolds.jl
mateuszbaran: Yes, it seems so.

src/manifolds/SymmetricPositiveDefinite.jl
mateuszbaran: The advantage is that `convert` that works on an `AbstractMatrix` doesn't do anything, while `Matrix` constructor would make a copy, and furthermore be problematic when working with other array types like `SMatrix`.

src/manifolds/SymmetricPositiveDefinite.jl
kellertuer: ok.

src/manifolds/SymmetricPositiveDefinite.jl
kellertuer: ok, thanks for the explanation.

kellertuer:
Interesting that the error does not show up on my local Julia 1.7 (or maybe it only shows up if one of the similar initialised values is `NaN`?).

mateuszbaran:
It's a random bug, depends on what value the memory is initialized to.

kellertuer:
That's what I thought. Does your line make it deterministically always work?

mateuszbaran:
It should always work now.

kellertuer:
Cool, I am fixing a few tests (and the ambiguity on 1.6)

kellertuer:
Ah I had a typo so we missed a few lines. Now it should be fine :)

kellertuer:
Hm, for SPDPoint and TangentSpaceAtPoint it seems that depending on where they are used sometimes the links in the docs don‘t work? But I can not figure out why :(

mateuszbaran:
Unfortunately I'm not good at fixing problems with documenter.

kellertuer:
SPDPoint was included twice, I could fix that, but one TangentSpaceAtPoint (just one) and I can not see why, nor which one actually, since the link checker just reports one to be brocken (but all refs are exactly the same).

kellertuer:
Now just one `TangentSpaceAtPoint` link is broken, but I can not even find out which of the two on the vectorbundle.md page.

src/manifolds/SymmetricPositiveDefinite.jl
```diff
@@ -30,10 +30,100 @@ struct SymmetricPositiveDefinite{N} <: AbstractDecoratorManifold{ℝ} end
...
+
+This result only has the result of `eigen` as a mandatory storage, the other three
+can be stored. If they are not stored they are computed and returned (but then still not stored)
+when required.
+
+# Constructor
+
+    SPDPoint(p::AbstractMatrix; store_p=true, store_sqrt=true, store_sqrt_inv=true)
+
+Create an SPD point using an symmetric positive defincite matrix `p`, where you can optionally store `p`, `p_sqrt` and `p_sqrt_inv`
```
mateuszbaran, line 47:
```suggestion
Create an SPD point using an symmetric positive defincite matrix `p`, where you can optionally store `p`, `sqrt` and `sqrt_inv`
```

docs/src/manifolds/symmetricpositivedefinite.md
```diff
@@ -76,4 +78,23 @@ Order = [:function]
 Filter = t -> t === mean
 ```
 
+## Efficient representation
+
+When a point `p` is used in several occasions, it might be beneficial to store the eigenvalues
+and vectors of `p` and eventually/optionally ist square root and the inverse of the square root.
```
mateuszbaran, line 84:
```suggestion
and vectors of `p` and optionally its square root and the inverse of the square root.
```

src/manifolds/SymmetricPositiveDefinite.jl
```diff
@@ -128,6 +276,109 @@ function default_estimation_method(::SymmetricPositiveDefinite, ::typeof(mean))
...
+The matrix is either stored within the [`SPDPoint`](@ref) or reconstructed from `p.eigen`.
+"""
+convert(::Type{AbstractMatrix}, p::SPDPoint)
+convert(::Type{AbstractMatrix}, p::SPDPoint) = p.p
+function convert(::Type{AbstractMatrix}, p::SPDPoint{Missing})
+    return (p.eigen.vectors * Diagonal(p.eigen.values) * p.eigen.vectors')
+end
+
+@doc raw"""
+    eigvals_sqrt(p::AbstractMatrix)
```
mateuszbaran, line 292:
```suggestion
    _sqrt(p::AbstractMatrix)
```
I've taken a closer look at this again and the `eigvals` part actually doesn't make sense, this is actually square root of `p`. Sorry for a poor suggestion.

kellertuer: but then we have to adapt all those occurrences again, right? I liked having a new name here to not have to use `sqrt` nor `_sqrt`, so I would prefer having a prefix still. Maybe `spd_`?

mateuszbaran: Yes, `spd_sqrt` sounds fine.

src/manifolds/SymmetricPositiveDefinite.jl
```diff
@@ -128,6 +276,109 @@ function default_estimation_method(::SymmetricPositiveDefinite, ::typeof(mean))
...
 end
 
+@doc raw"""
+    convert(::Type{AbstractMatrix}, p::SPDPoint)
+
+return the point `p` as a matrix.
+The matrix is either stored within the [`SPDPoint`](@ref) or reconstructed from `p.eigen`.
+"""
+convert(::Type{AbstractMatrix}, p::SPDPoint)
+convert(::Type{AbstractMatrix}, p::SPDPoint) = p.p
```
mateuszbaran, line 286:
```suggestion
"""
convert(::Type{AbstractMatrix}, p::SPDPoint) = p.p
```
That one line isn't useful

src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
```diff
@@ -225,28 +247,21 @@ function get_coordinates_orthonormal!(
...
-    pSqrt = Symmetric(U * sSqrt * transpose(U))
+    p_sqrt = spd_sqrt(p)
     k = 1
-    V = similar(p)
+    V = similar(convert(AbstractMatrix, p))
     fill!(V, zero(eltype(V)))
-
-    F = cholesky(Symmetric(p))
-
+    F = cholesky(Symmetric(convert(AbstractMatrix, p)))
```
mateuszbaran, line 254: You could convert `p` to `AbstractMatrix` only once and assign the result to a variable.

src/manifolds/SymmetricPositiveDefiniteLinearAffine.jl
```diff
@@ -369,25 +383,18 @@ and `log` the logarithmic map on [`SymmetricPositiveDefinite`](@ref)
...
-    U = e.vectors
-    S = max.(e.values, floatmin(eltype(e.values)))
-    Ssqrt = sqrt.(S)
-    SsqrtInv = Diagonal(1 ./ Ssqrt)
-    Ssqrt = Diagonal(Ssqrt)
-    pSqrt = Symmetric(U * Ssqrt * transpose(U)) # p^1/2
-    pSqrtInv = Symmetric(U * SsqrtInv * transpose(U)) # p^(-1/2)
-    tv = Symmetric(pSqrtInv * X * pSqrtInv) # p^(-1/2)Xp^{-1/2}
-    ty = Symmetric(pSqrtInv * q * pSqrtInv) # p^(-1/2)qp^(-1/2)
+    distance(M, p, q) < 2 * eps(eltype(convert(AbstractMatrix, p))) && copyto!(Y, X)
```
mateuszbaran, line 386:
```suggestion
    distance(M, p, q) < 2 * eps(eltype(p)) && copyto!(Y, X)
```
`SPDPoint` supports `eltype`.

mateuszbaran (approved):
The changes look fine, there are just some very minor issues.

PR #516, opened 2022-08-03
Faster minkowski_metric for large vectors.

mateuszbaran:
One more improvement to `minkowski_metric`.


kellertuer (approved):

mateuszbaran:
Updating `q` in-place should be safe in that function.

kellertuer:
Cool, the tests look good. If you have time try to find one line to cover additionally (I will try that from now on to slowly reduce uncovered lines further)

mateuszbaran:
IIRC most of the uncovered lines either need some kind of rework (so the easiest way to increase coverage would be by just deleting them :wink:) or are hard to write tests for.

PR #517, opened 2022-08-04
Changes required for ManifoldDiff AD

mateuszbaran:
This PR is meant to contain changes required to push the AD work to https://github.com/JuliaManifolds/ManifoldDiff.jl .


src/groups/group_action.jl
```diff
@@ -112,6 +123,21 @@ function apply_diff!(A::AbstractGroupAction, Y, a, p, X)
     )
 end
 
+@doc raw"""
+    apply_diff_group(A::AbstractGroupAction, a, X, p)
+
+For a point on manifold ``p ∈ \mathcal M`` and an element `X` of the tangent space at `a`,
```
kellertuer, line 129:
```suggestion
For a point on manifold ``p ∈ \mathcal M`` and an element `X` of the tangent space at `a ∈ A`
```

kellertuer: Do we indicate Lie groups by letters like `A`?

src/groups/group_action.jl
```diff
@@ -32,6 +32,17 @@ Get the direction of the action
 """
 direction(::AbstractGroupAction{AD}) where {AD} = AD()
 
+@doc raw"""
```
kellertuer, line 35: Maybe use a reserved order here p, A, a since it is the adjoint?

src/groups/group_action.jl
mateuszbaran: No, usually `G`, `M` or `H` if there is more than one. I'll rephrase it.

kellertuer: Cool, yeah I also had G in mind, and making sure to know that p is from M and a is from G would be nice as well (or is G=M here?).

src/groups/group_action.jl
mateuszbaran: Do you mean `adjoint_apply_diff_group(A::AbstractGroupAction, p, X, a)`?

src/groups/group_action.jl
mateuszbaran: No, `M` is different as usually when it comes to actions.

kellertuer: Ok, that was also something that did not get that clear (since also `M` is not part of the parameters).

src/groups/group_action.jl
kellertuer: yes, or would that be too irritating?

mateuszbaran: I think we usually try to place tangent vector right after the point it is tangent at and here `X` is tangent at `a` but with so many variants of differentials even to me it's impossible to use them without carefully reading the documentation. So if the `p, X, a` order is more intuitive for you I can use it :slightly_smiling_face:.

kellertuer: Ah maybe your reasoning is better, mine is surely not based on much intuition and your argument sounds quite intuitive. So let‘s keep it as is.

src/groups/group_action.jl
mateuszbaran: So do you think the new text is clear enough?

kellertuer: Nearly. It is still not clear where `M` comes from (it is merely implicitly given by using some `p`?) , since it is not a parameter of the function (and probably should not become that?), so maybe clarifying that a little would even improve the docs, the rest is fine now.

mateuszbaran: I've rewritten it again, hopefully it's better now :slightly_smiling_face: .

src/groups/group_action.jl
```diff
@@ -112,6 +123,29 @@ function apply_diff!(A::AbstractGroupAction, Y, a, p, X)
...
+    apply_diff_group(A::AbstractGroupAction, a, X, p)
+
+Compute the value of differential of action [`AbstractGroupAction`](@ref) `A` on vector `X`,
+where element `a` is acting on `p`, with respect to the group element.
+
+Let ``\mathcal G`` be the group acting on manifold ``\mathcal M`` by the action `A`.
+The action is of element ``g ∈ \mathcal G`` on a point ``p ∈ \mathcal M``.
+The differential transforms vector `X` from the tangent space at `a ∈ \mathcal G`, 
+``X ∈ T_a \mathcal G`` into the tangent space of manifold ``\mathcal M``.
```
kellertuer, line 135: Now its just that X is double in this line and the one before? The rest is fine :) - and which tangent space at M? The one at p?

src/groups/group_action.jl
kellertuer: nearly (just something small at another line) – and sorry for being a little picky/precise here.

src/groups/group_action.jl
mateuszbaran: The result is from `T_{τ_a p} \mathcal M`, as defined below. Feel free to suggest an edit if you have a good idea :slightly_smiling_face: .

kellertuer: ah I missed that the tangent space is given below, so we could just write _a_ tangent space (instead of the) here?

kellertuer, line 135:
```suggestion
``X ∈ T_a \mathcal G`` into a tangent space of the manifold ``\mathcal M``.
```

kellertuer (approved):

Issue #522, opened 2022-08-12
projection onto tangent space to unitary matrices


Nikdwal:
I don't think that the current implementation of `project(::OrthogonalMatrices, p, X)` works as intended.
For any X in the tangent space, the projection should act as the identity operator, but it doesn't do that in this example:
```
using Manifolds
M = OrthogonalMatrices(2)
p = [1 1; -1 1] / √2
@assert is_point(M, p) # test passes
X = [0 1; -1 0]
@assert is_vector(M, p, X) # test passes
X_proj = project(M, p, X)
println(norm(M, p, X - X_proj)) # prints 0.414, not 0 or a small number
```

If the tangent space is `T_p M = { pA | A^T = - A }` and it is identified in code with the skew-symmetric matrices through the isomorphism A -> pA, then I think the correct projection should be A -> (A - A^T)/2.

kellertuer:
The currently implemented – and documented version is 

```julia
   project(M::OrthogonalMatrices{n}, p, X)
   project(M::Rotations{n}, p, X)
   project(M::UnitaryMatrices{n}, p, X)

  Orthogonally project the tangent vector X ∈ 𝔽^{n × n}, \mathbb F ∈ \{\mathbb R, \mathbb C\} to the tangent space of M at p,
  and change the representer to use the corresponding Lie algebra, i.e. we compute

      \operatorname{proj}_p(X) = \frac{pX-(pX)^{\mathrm{T}}}{2},
```

So currently, for OrthogonalMatrices we do not do the isomorphism, but sure your variant would (as the easiest way= be

```
project(SkewHermitianMatrices(2,ℝ), X)
```

which keeps X as is, yes.

I think we should switch to the isomorphism you mentioned – also because thats nicer for the SO(3) – though we then have to adapt exp. We seem to have changed the default when we unified Unitary/Orthogonal/SpecialOrthogonal/SpecialUnitary about a month ago, sorry for that.

I would like to wait for a short feedback from @mateuszbaran ; but the fix is quite easy; for all 4 it is just the one line at 

https://github.com/JuliaManifolds/Manifolds.jl/blob/8274392cc6a1c406f7fe83012e2feefb85e58738/src/manifolds/GeneralUnitaryMatrices.jl#L678

and we just have to remove the `p \` ;)

kellertuer:
[...]

PR #523, opened 2022-08-12
Propose the fix for #522.

kellertuer:


kellertuer:
Maybe an alternative to have both variants as a possibility would be to keep the function as is and do the Lie algebra projection for the group manifold analoga.

kellertuer:
I feel this variant should actually be the best choice: To have both.

For _Manigolds (`Rotations`, `OrthogonalMatrices`, `UnitaryMatrices`) we keep the representation as is (that is `X` is the tangent vector and not necessarily skew, but a rotated skew symmetric), for Lie groups (`Orthogonal`, `SpecialOrthogonal`, `SpecialUnitary`, `Unitary`) we use the Lie group representation. This way it is easy to have either of the representations, depending on what the user wants – in a precisely distinct way.

kellertuer:
Uff. This might be more work than I thought because we have so many special variants of exp/log that now have to be doubled and adapted for the other representation case.

kellertuer:
hm, seems really more complicated than I thought. So I will wait for the decision whether we want to unify it and in which way.

kellertuer:
So if I understand correctly it is not only complicated to finish this but even more to get it work together with differentiation again? I would maybe propose to close this?

mateuszbaran:
This would be a lot of work to finish for not much benefit so I'd suggest closing it.

PR #524, opened 2022-08-17
Fix #520

### LINKED ISSUE #520: Docs, Lie group actions are non-commutative (i.e. non-symmetric)
Hi, I just want to confirm one line in the docs for `exp_lie`.  I have Lie group actions in general as non-commutative?  
https://github.com/JuliaManifolds/Manifolds.jl/blob/8274392cc6a1c406f7fe83012e2feefb85e58738/src/groups/group.jl#L846

My concern is that the composition operation here is not symmetric in the general case, or I am missing something? 

I have in general that the sum of Lie algebra elements (tangent vectors about group identity in this case**?) are commutative up to the Lie bracket.  As in, Baker-Cambell-Hausdorf ([wiki](https://en.wikipedia.org/wiki/Baker%E2%80%93Campbell%E2%80%93Hausdorff_formula#An_important_lemma) and also [this](https://en.wikipedia.org/wiki/Lie_group%E2%80%93Lie_algebra_correspondence)) formula:
```
exp(X)exp(Y) = exp(Z)
```
where
```
Z = X+Y + 0.5[X,Y] + ...
```

for non-commutative, the Lie bracket `[X,Y]` is non-zero.

----

** separate question on `log` for Group (Lie Algebra diagonals), see
- #521

mateuszbaran:


kellertuer (approved):

mateuszbaran:
I've bumped StaticArrays.jl lower bound because we apparently rely on https://github.com/JuliaArrays/StaticArrays.jl/pull/1009 now.

kellertuer:
Cool that you spotted that one :)

mateuszbaran:
We have an unrelated test failure on MacOS :confused: . Something with plotting.

kellertuer:
that might also just be the artefact shortly broken, I currently locally have the same but it vanishes usually in a day or two, we can just recheck soon.

kellertuer:
I am not sure why it still fails, the tests are running fine on my side – we could bump them to 1.8 though?

mateuszbaran:
One sadly fails on Julia 1.6 which we shouldn't bump. I think we need to disable that test on MacOS.

mateuszbaran:
@kellertuer is it fine to merge this PR as it is now?

kellertuer:
Sure.

PR #530, opened 2022-09-09
Update docstring to match implementation

jondeuce:
Hi there, I found what I believe to be a small typo in the doc string of the `project!` method for general unitary matrices, as the implementation projects `p \ X` onto the skew-symmetric matrices, not `p * X`.


kellertuer (approved):
Yes, thanks for noticing.

PR #533, opened 2022-09-21
🍩

mateuszbaran:
Some initial work on implementing a curved torus discussed in https://github.com/JuliaManifolds/Manifolds.jl/discussions/532 .
Currently it's possible to draw a torus:
image
as well as do some basic conversions to/from an atlas.

@FedeClaudi you can take a look how `get_chart_index` is used to select  a chart around a given point, `get_parameters!` and `get_point!` covert to and from parameters in a chart and `get_vector_induced_basis!` converts tangent vectors from a representation in $\mathbb{R}^3$ to an in-chart representation. Does that look somewhat understandable? Higher-genus tori would most likely have to implement the same functions, except the embedding can be skipped if explicit transition maps are provided. Currently by default transition maps are defined using an embedding but that's not required.

I've started work on a fully in-chart geodesic solver but I need to think about it a bit more, so that's not ready for review.


kellertuer:
Just two first feedback points:

1) Should we do a new manifold? Its still the torus? I would switch the points/tangent vector types, since it is another representation (cg. Hyperbolic).
2) The example is nice, if it is a Pluto-Notebook we could even include that into the docs.

FedeClaudi:
First of all, outstanding PR naming game, bravo 💯 !

That looks excellent. I had actually just started working on an atlas for the n-torus here: https://github.com/FedeClaudi/Manifolds.jl/blob/master/torus_wip.jl
I took a different approach and was trying to do something that wasn't dependent on the embedding. I tend to think of manifolds without them being embedded somewhere, but I think `Manifolds.jl` mostly deals with embedded manifolds right?

Anyway, I'll have a look at the code in a second, but it looks great. I wonder how it would generalize to torii of genus > 1 for which there's no nice parametrization? 

Also, I will steal your visualization code if that's okay, that's such a pretty way to visualize surfaces!

mateuszbaran:
> 1. Should we do a new manifold? Its still the torus? I would switch the points/tangent vector types, since it is another representation (cg. Hyperbolic).

Good point, it's topologically the same as the current torus, just with a different metric and point/tangent vector representation.

> 2\. The example is nice, if it is a Pluto-Notebook we could even include that into the docs.

Thanks, I'm going to add a geodesic curve to that picture, then I will ask for some help turning it into a notebook :slightly_smiling_face: . It would be cool to have sliders for selecting tangent vector for geodesic in the docs.

> That looks excellent. I had actually just started working on an atlas for the n-torus here: https://github.com/FedeClaudi/Manifolds.jl/blob/master/torus_wip.jl
> I took a different approach and was trying to do something that wasn't dependent on the embedding. I tend to think of manifolds without them being embedded somewhere, but I think `Manifolds.jl` mostly deals with embedded manifolds right?

Cool, that's also a good approach to making an atlas. Manifolds.jl currently focuses on working in an embedding, yes, but the API allows for working without it. Your `get_parameters` and `get_point` should work fine.

In this PR I will add a way to compute geodesics fully in charts. Maybe with chart switching if there is an easy way to do it in OrdinaryDiffEq.jl. Also maybe with sensitivity analysis for a gradient-descent-based logarithmic map.

> I wonder how it would generalize to torii of genus > 1 for which there's no nice parametrization?

I did a quick search and maybe this would help: https://cs.brown.edu/people/jhughes/papers/Grimm-PNT-2003/paper.pdf ? We would still need to define a metric in some way though, in case of the torus in this PR I simply take the metric from the embedding.

> Also, I will steal your visualization code if that's okay, that's such a pretty way to visualize surfaces!

Sure, I stole most of it from here: https://beautiful.makie.org/dev/examples/generated/2d/surfaces/gabriels_horn/ :smile: .

kellertuer:
> > 1. Should we do a new manifold? Its still the torus? I would switch the points/tangent vector types, since it is another representation (cg. Hyperbolic).
> 
> Good point, it's topologically the same as the current torus, just with a different metric and point/tangent vector representation.

Oh also a different metric, interesting – then you could
a) use a metric manifold and still just arrays
b) use a metric manifold and point/tangent types

but I would avoid c) then (just use points/tangent types) because that might be mistaken of a different representation for the default metric we have.

> 
> > 2. The example is nice, if it is a Pluto-Notebook we could even include that into the docs.
> 
> Thanks, I'm going to add a geodesic curve to that picture, then I will ask for some help turning it into a notebook 🙂 . It would be cool to have sliders for selecting tangent vector for geodesic in the docs.

Sure, no problem, I can help turning that into a Pluto Notebook (and maybe even adding it to the docs, since we should start more tutorials I think).

FedeClaudi:
> since we should start more tutorials I think

As a new user: yep. 

I really like `Manifolds.jl` and appreciate the work you're doing with it and this PR in particular. Let me know if I can help with anything!

mateuszbaran:
Going between charts seems to work in the version I've just uploaded but most likely I have an error with Christoffel symbols for torus (geodesic shouldn't converge to a circle)
image

mateuszbaran:
I've made a nice interactive visualization of geodesics in Makie:
image

mateuszbaran:
To try this you just need to run `plot_thing` from the `generalized_torus.jl` file. Sliders set the initial conditions for the geodesic.

kellertuer:
Nice! That looks great!

test/manifolds/generalized_torus.jl
```diff
@@ -0,0 +1,122 @@
...
+    #p0x = [0.5, -1.2]
+    p0x = [pi / 2, 0.0]
+    X_p0x = [-1.2, 0.4]
+    p = [Manifolds._torus_param(M, p0x...)...]
+    i_p0x = Manifolds.get_chart_index(M, A, p)
+    B = induced_basis(M, A, i_p0x)
+    X = get_vector(M, p, X_p0x, B)
+    @test get_coordinates(M, p, X, B) ≈ X_p0x
+
+    @test_broken norm(X) ≈ norm(M, p0x, TFVector(X_p0x, B))
```
mateuszbaran, line 26: @kellertuer Could you check if this test actually makes sense? This is supposed to test that the vector `X` in embedding of the torus has the same norm as calculated using its coordinates in the induced basis and the metric tensor. I'm not sure where I could have a mistake.

mateuszbaran:
And now we also have parallel transport along geodesics and Gaussian curvature:
image

test/manifolds/generalized_torus.jl
kellertuer: Does this one error? If so – by how much?
Also just `norm(X)`is a little surprising, is that already in the embedding?

mateuszbaran: Yes, `X` is in the embedding and the difference is large:
```
  Expression: norm(X) ≈ norm(M, p0x, TFVector(X_p0x, B))
   Evaluated: 2.6832815729997477 ≈ 1.4422205101855958
```
I'd expect the embedding to be isometric because the metric is actually defined through that embedding.

kellertuer: yes then I would expect that as well. But for the next few days I do not have time to check the code too much I fear.

mateuszbaran: No problem, it can wait.

kellertuer: First idea, did you check that the basis is an ONB? Otherwise you might need two bases (one for decomposition one for composition) to cancel out the non-normal-basis vectors?

mateuszbaran: `B` is not ONB but why would it matter here?

mateuszbaran: `get_vector` and `get_coordinates` are inverses of each other, that's one thing that I've checked.

kellertuer: Hm, ok; then I would have to check how you compute the norm on the right (because then it can not be get_vector&norm in embedding, right? Because if they are inverses that should be the same as X).

mateuszbaran: Yes, there is likely an error in both `get_vector` and `get_coordinates`. I've noticed that I have mixed up notation in a few places so anyway it's better if you wait until I fix that.

kellertuer: Ok, but I hope that in the best case the code works fine after you fixed that 🤓

mateuszbaran: I've found the errors but one of them looks like an issue with API. Currently `local_metric` expects point in the embedding while arguably it would make sense to pass there coordinate in a basis. It would be nice to be able to tell these two cases apart by method signatures. Do you have an idea for that?

kellertuer: Cool, that you found it 

Which `local_metric`do you refer to? The one on the sphere
https://github.com/JuliaManifolds/Manifolds.jl/blob/71d9d8137e6418b4931581ab083da0cc370ac621/src/manifolds/Sphere.jl#L285-L291
Does what I would expect – the point p on the manifold (or would you prefer _that_ in charts/coordinates?) 

Because there `G = local_metric(M, p,B)` returns the metric as a matrix such that for tangent vectors given in coordinates with respect to `B` we have `c1'*G*c2` which is what I would expect.

One could write one that is `g = local_metric(M,p)` such that `X'*g*Y` works for vectors (not coordinates).

So I am not sure what you mean by "point in a basis", since we do not have a basis on the manifold.

mateuszbaran: Sorry, it was a bit late -- I meant passing point as parameters in a chart. So probably something like `local_metric(M::AbstractManifold, A::AbstractAtlas, i, a)` for coefficients of the metric tensor at point with parametrization `a` in chart `i` from atlas `A` in basis induced by that chart at that point.

What I'd like to develop here a bit is working on a manifold without an embedding.

kellertuer: Ah, but then `A` solves your thing here :) I like the signature you propose.

What I also meant was, that there is a (though only subtle and small) difference between point-representation and embedding. But sure in charts is what you meant.

test/manifolds/torus_example.jl
```diff
@@ -0,0 +1,114 @@
+using Revise
```
kellertuer, line 1: this could be reworked into a Pluto notebook, I can help with that (and include it into the docs in the tutorials PR later).

kellertuer (commented):
I think we should document the new torus together with the old torus on the same documentation page, what do you think?

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,215 @@
...
+    Γ¹₂₂ = (M.R + M.r * cosθ) * sinθ / M.r
+    Γ²₁₂ = -M.r * sinθ / (M.R + M.r * cosθ)
+
+    Zc[1] = Xc[2] * Γ¹₂₂ * Yc[2]
+    Zc[2] = Γ²₁₂ * (Xc[1] * Yc[2] + Xc[2] * Yc[1])
+    return Zc
+end
+
+function check_chart_switch(::TorusInR3, A::DefaultTorusAtlas, i, a)
+    return abs(i[1] - a[1]) > 2pi / 3 || abs(i[2] - a[2]) > 2pi / 3
```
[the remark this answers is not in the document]
mateuszbaran: No particular reason, maybe it should be even larger than that. I don't have a source for this threshold. It needs to be less than 2pi - O(integrator step).

I will add documentation and make this function public. It indeed checks whether we need to switch charts when we reach parameter `a` in chart `i`.

test/manifolds/torus_example.jl
mateuszbaran: Yes, it would be great if you could turn it into a Pluto notebook :slightly_smiling_face: .

kellertuer: Sure. My idea would be – finish this PR with the example here as is and I will rework it in the #534 afterwards?

src/manifolds/CurvedTorus.jl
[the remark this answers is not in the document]
mateuszbaran: It may be a good idea but how exactly would you implement it here?

test/manifolds/torus_example.jl
mateuszbaran: Sure, that's a good plan.

src/manifolds/CurvedTorus.jl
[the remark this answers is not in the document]
kellertuer: Well here I would add a fifth argument `\epsilon = ...` (whatever ... is for a reasonable default here) and replace ` > 2pi / 3` by the value for “closer than epsilon to the border” ?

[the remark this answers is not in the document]
mateuszbaran: OK, I have added that.

mateuszbaran:
I've added computing log/distance/geodesic between two points by solving a BVP in a chart. So I would consider this PR feature-complete, and it just needs some docs and a bit of polishing. I'm quite happy that we have now a decent way to work in charts, and what we have here should nicely generalized to higher-genus tori.

FedeClaudi:
looks great, looking forward to playing around with it!

mateuszbaran:
@kellertuer I think this is ready for review. One major point is the new `transition_map_diff` and in-chart `affine_connection`, `inner` and `norm`, this is the beginning of the new API for working (fully) in charts. The new solvers in `ode_callback.jl` and `bvp.jl` are supposed to be somewhat internal and I would prefer to expose them in an API later, maybe after they are validated on higher genus tori.

kellertuer:
Thanks for the summary, I will be travelling the weekend but try to find time to check stuff.

mateuszbaran:
No problem, we can finish this next week :slightly_smiling_face: .

src/atlases.jl
```diff
@@ -55,6 +55,26 @@ function RetractionAtlas(
...
+function affine_connection(M::AbstractManifold, A, i, a, Xc, Yc)
+    Zc = allocate(Xc)
+    return affine_connection!(M, Zc, A, i, a, Xc, Yc)
+end
+
+"""
+    affine_connection!(M::AbstractManifold, Zc, A::AbstractAtlas, i, a, Xc, Yc)
+
+Calculate affine connection on manifold `M` at point with parameters `a` in chart `i` of
+atlas `A` of vectors with coefficients `Xc` and `Yc` in induced basis and save the result
```
kellertuer, line 73:
```suggestion
an [`AbstractAtlas`](@ref) `A` of vectors with coefficients `Zc` and `Yc` in induced basis and save the result
```

src/atlases.jl
```diff
@@ -213,6 +273,34 @@ function transition_map!(M::AbstractManifold, y, A::AbstractAtlas, i_from, i_to,
     return transition_map!(M, y, A, i_from, A, i_to, a)
 end
 
+"""
+    transition_map_diff(M::AbstractManifold, A::AbstractAtlas, i_from, a, c, i_to)
+
+Compute differential of transition map from chart `i_from` to chart `i_to` from atlas `A`
```
kellertuer, line 279:
```suggestion
Compute differential of transition map from chart `i_from` to chart `i_to` from an [`AbstractAtlas`](@ref) `A`
```

src/atlases.jl
```diff
@@ -158,6 +188,36 @@ get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any)
...
+
+Calculate inner product on manifold `M` at point with parameters `a` in chart `i` of
+atlas `A` of vectors with coefficients `Xc` and `Yc` in induced basis.
+"""
+inner(M::AbstractManifold, A::AbstractAtlas, i, a, Xc, Yc)
+
+"""
+    norm(M::AbstractManifold, A::AbstractAtlas, i, a, Xc)
+
+Calculate norm on manifold `M` at point with parameters `a` in chart `i` of
```
kellertuer, line 216:
```suggestion
Calculate norm on manifold `M` at point with parameters `a` in chart `i` of an
```

src/atlases.jl
```diff
@@ -158,6 +188,36 @@ get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any)
...
+# See also
+
+[`get_default_atlas`](@ref)
+"""
+get_chart_index(::AbstractManifold, ::AbstractAtlas, ::Any, ::Any)
+
+"""
+    inner(M::AbstractManifold, A::AbstractAtlas, i, a, Xc, Yc)
+
+Calculate inner product on manifold `M` at point with parameters `a` in chart `i` of
```
kellertuer, line 208:
```suggestion
Calculate inner product on manifold `M` at point with parameters `a` in chart `i` of an
```

src/atlases.jl
```diff
@@ -97,6 +117,16 @@ end
     p,
 )
 
+"""
+    check_chart_switch(M::AbstractManifold, A::AbstractAtlas, i, a)
+
+Determine whether chart should be switched when an operation in chart `i` from atlas `A`
```
kellertuer, line 123:
```suggestion
Determine whether chart should be switched when an operation in chart `i` from an [`AbstractAtlas`](@ref) `A`
```

src/atlases.jl
```diff
@@ -55,6 +55,26 @@ function RetractionAtlas(
...
+"""
+function affine_connection(M::AbstractManifold, A, i, a, Xc, Yc)
+    Zc = allocate(Xc)
+    return affine_connection!(M, Zc, A, i, a, Xc, Yc)
+end
+
+"""
+    affine_connection!(M::AbstractManifold, Zc, A::AbstractAtlas, i, a, Xc, Yc)
+
+Calculate affine connection on manifold `M` at point with parameters `a` in chart `i` of
```
kellertuer, line 72:
```suggestion
Calculate affine connection on manifold `M` at point with parameters `a` in chart `i` of an
```

src/atlases.jl
```diff
@@ -213,6 +273,34 @@ function transition_map!(M::AbstractManifold, y, A::AbstractAtlas, i_from, i_to,
...
+on manifold `M` at point with coordinates `a` on tangent vector with coordinates `c`.
+"""
+function transition_map_diff(M::AbstractManifold, A::AbstractAtlas, i_from, a, c, i_to)
+    old_B = induced_basis(M, A, i_from)
+    new_B = induced_basis(M, A, i_to)
+    p_final = get_point(M, A, i_from, a)
+    return change_basis(M, p_final, c, old_B, new_B)
+end
+
+function transition_map_diff!(
```
kellertuer, line 289: Maybe add some docs to this as well?

src/atlases.jl
```diff
@@ -286,20 +374,31 @@ struct InducedBasis{𝔽,VST<:VectorSpaceType,TA<:AbstractAtlas,TI} <: AbstractB
...
-    VST::VectorSpaceType,
+    VST::VectorSpaceType=TangentSpace,
 ) where {𝔽}
     return InducedBasis{𝔽,typeof(VST),typeof(A),typeof(i)}(VST, A, i)
 end
 
+"""
+    inverse_chart_injectivity_radius(M::AbstractManifold, A::AbstractAtlas, i)
+
+Injectivity radius of `get_point` for chart `i` from atlas `A` of manifold `M`.
```
kellertuer, line 398:
```suggestion
Injectivity radius of `get_point` for chart `i` from an [`AbstractAtlas`](@ref) `A` of a manifold `M`.
```

src/differentiation/bvp.jl
```diff
@@ -0,0 +1,87 @@
...
+        a2,
+        A::AbstractAtlas,
+        i;
+        solver=GeneralMIRK4(),
+        dt=0.05,
+        kwargs...,
+    )
+
+Solve the BVP corresponding to geodesic calculation on [`AbstractManifold`](@ref) M,
+between points with parameters `a1` and `a2` in chart `i` of [`AbstractAtlas`](@ref) `A`
```
kellertuer, line 26:
```suggestion
between points with parameters `a1` and `a2` in a chart `i` of an [`AbstractAtlas`](@ref) `A`
```

src/differentiation/ode_callback.jl
```diff
@@ -0,0 +1,211 @@
+
+"""
+    maybe_switch_chart(u, t, integrator)
+
+Terminate integration when integrator goes too closely to chart boundary.
+"""
+function maybe_switch_chart(u, t, integrator)
```
kellertuer, line 7: this functions name sounds strange compared to what the documentation says

for me a suitable name would be `terminate_integrator_near_chart_boundary`?

Could you maybe also document the parameters? Also the chart switch has a boundary parameter, we could pass that one on?

src/differentiation/ode_callback.jl
```diff
@@ -0,0 +1,211 @@
...
+        # switch charts
+        OrdinaryDiffEq.terminate!(integrator)
+    end
+    return u
+end
+
+"""
+    StitchedChartSolution{TM<:AbstractManifold,TA<:AbstractAtlas,TChart}
+
+Solution of an ODE on manifold `M` in charts of atlas `A`.
```
kellertuer, line 19:
```suggestion
Solution of an ODE on a manifold `M` in charts of an [`AbstractAtlas`](@ref) `A`.
```

src/differentiation/ode_callback.jl
```diff
@@ -0,0 +1,211 @@
...
+        a,
+        Xc,
+        A::AbstractAtlas,
+        i0;
+        solver=AutoVern9(Rodas5()),
+        final_time=1.0,
+        kwargs...,
+    )
+
+Solve geodesic ODE on manifold `M` from point of coordinates `a` in chart `i0` from atlas
```
kellertuer, line 86:
```suggestion
Solve geodesic ODE on a manifold `M` from point of coordinates `a` in chart `i0` from an [`AbstractAtlas`](@ref) `A`.
```

src/differentiation/ode_callback.jl
```diff
@@ -0,0 +1,211 @@
...
+        Xc,
+        A::AbstractAtlas,
+        i0,
+        Yc;
+        solver=AutoVern9(Rodas5()),
+        final_time=1.0,
+        kwargs...,
+    )
+
+Parallel transport vector with coordinates `Yc` along geodesic on manifold `M` from point of
```
kellertuer, line 156:
```suggestion
Parallel transport vector with coordinates `Yc` along geodesic on a manifold `M` from point of
```

src/differentiation/ode_callback.jl
```diff
@@ -0,0 +1,211 @@
...
+coordinates `a` in chart `i0` from atlas `A` in direction of coordinates `Xc` in induced
```
kellertuer, line 157:
```suggestion
coordinates `a` in a chart `i0` from an [`AbstractAtlas`](@ref) `A` in direction of coordinates `Xc` in induced
```

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,223 @@
...
+```
+for θ, φ in $[-π, π)$. It is assumed that $R > r > 0$.
+
+Alternative names include anchor ring, donut and doughnut.
+
+# Constructor
+
+    TorusInR3(R, r)
+"""
+struct TorusInR3{TR<:Real} <: AbstractDecoratorManifold{ℝ}
```
kellertuer, line 19: We could also call this `EmbeddedTorus`? That might be nicer if we extend that to higher dimensions at some point.

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,223 @@
...
+end
+
+aspect_ratio(M::TorusInR3) = M.R / M.r
+
+"""
+    check_point(M::TorusInR3, p; kwargs...)
+
+Check whether `p` is a valid point on the [`TorusInR3`](@ref) `M`.
+The tolerance for the last test can be set using the `kwargs...`.
+"""
```
kellertuer, line 35: We could maybe document the formula that is checked?

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,223 @@
...
+function check_point(M::TorusInR3, p; kwargs...)
+    A = (dot(p, p) + M.R^2 - M.r^2)^2
+    B = 4 * M.R^2 * (p[1]^2 + p[2]^2)
+    if !isapprox(A, B; kwargs...)
+        return DomainError(A - B, "The point $(p) does not lie on the $(M).")
+    end
+    return nothing
+end
+
+function check_vector(M::TorusInR3, p, X; atol=eps(eltype(p)), kwargs...)
```
kellertuer, line 45: This could be documented as well.

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,223 @@
...
+"""
+function gaussian_curvature(M::TorusInR3, p)
+    θ, φ = _torus_theta_phi(M, p)
+    return cos(θ) / (M.r * (M.R + M.r * cos(θ)))
+end
+
+"""
+    inner(M::TorusInR3, ::DefaultTorusAtlas, i, a, Xc, Yc)
+
+Inner product on [`TorusInR3`](@ref) in chart `i` of atlas [`DefaultTorusAtlas`](@ref)
```
kellertuer, line 121:
```suggestion
Inner product on [`TorusInR3`](@ref) in chart `i` in the [`DefaultTorusAtlas`](@ref).
```

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,223 @@
...
+function inner(M::TorusInR3, ::DefaultTorusAtlas, i, a, Xc, Yc)
+    diag_1 = M.r^2
+    diag_2 = (M.R + M.r * cos(a[1] + i[1]))^2
+    return Xc[1] * diag_1 * Yc[1] + Xc[2] * diag_2 * Yc[2]
+end
+
+"""
+    inverse_chart_injectivity_radius(M::AbstractManifold, A::AbstractAtlas, i)
+
+Injectivity radius of `get_point` for chart `i` from atlas `A` of manifold `M`.
```
kellertuer, line 134:
```suggestion
Injectivity radius of `get_point` for chart `i` from the [`DefaultTorusAtlas`](@ref) `A` of the [`TorusInR3`](@ref).
```

src/manifolds/CurvedTorus.jl
```diff
@@ -0,0 +1,223 @@
...
+function _torus_param(M::TorusInR3, θ, φ)
+    sinθ, cosθ = sincos(θ)
+    sinφ, cosφ = sincos(φ)
+    return ((M.R + M.r * cosθ) * cosφ, (M.R + M.r * cosθ) * sinφ, M.r * sinθ)
+end
+
+"""
+    normal_vector(M::TorusInR3, p)
+
+Outward-pointing normal vector to torus at `p`.
```
kellertuer, line 161:
```suggestion
Outward-pointing normal vector on the [`TorusInR3`](@ref) at the point `p`.
```

kellertuer (commented):
Really nice PR; I have a few comments, but nothing too serious.

Could you also check for code coverage a little bit? We are not far away from the required rate, just a few lines shy I think.

mateuszbaran:
> Really nice PR; I have a few comments, but nothing too serious.
> 
> Could you also check for code coverage a little bit? We are not far away from the required rate, just a few lines shy I think

Thanks, I will check the coverage after addressing your comments.

docs/Project.toml
```diff
@@ -1,23 +1,29 @@
...
+BoundaryValueDiffEq = "764a87c0-6b3e-53db-9096-fe964310641d"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 HybridArrays = "1baab800-613f-4b0a-84e4-9cd3431bfbb9"
 Manifolds = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 ManifoldsBase = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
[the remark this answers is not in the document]
mateuszbaran: I don't think I understand, this is a dependency just for building the docs (because without it we can't access methods loaded conditionally by `Requires`).

docs/make.jl
```diff
@@ -1,6 +1,7 @@
 using Plots, RecipesBase, Manifolds, ManifoldsBase, Documenter, PyPlot
-# required for loading the manifold tests functios
+# required for loading the manifold tests functions
 using Test, FiniteDifferences
+using OrdinaryDiffEq, BoundaryValueDiffEq, DiffEqCallbacks
```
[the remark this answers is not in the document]
mateuszbaran: Sure, I will add a comment and move it..

docs/Project.toml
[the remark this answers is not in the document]
kellertuer: Oh, I missed that its the docs file, then ist fine :)

docs/src/manifolds/torus.md
```diff
@@ -24,3 +24,13 @@ Modules = [Manifolds]
 Pages = ["manifolds/Torus.jl"]
 Order = [:type, :function]
 ```
+
+## Curved Torus
```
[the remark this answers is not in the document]
mateuszbaran: That's because the other one actually is flat, and this one is not ;).

src/differentiation/ode_callback.jl
mateuszbaran: `terminate_integrator_near_chart_boundary` sounds better, thanks for the suggestion. I will expand it a bit to include that boundary parameter here.

src/manifolds/CurvedTorus.jl
mateuszbaran: I'm not sure it can actually be reasonably extended, I haven't seen anyone constructing similar curved tori in higher dimensions.

docs/src/manifolds/torus.md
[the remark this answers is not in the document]
kellertuer: Sure as a product of two circles the curvature is zero for the other one – still this title might be misleading and I think the different representations are more helpful to the reader (especially those who have not yet thought about the different metrics and hence curvature).

src/differentiation/ode_callback.jl
kellertuer: oh and it has to be `determinate` I think.

src/manifolds/CurvedTorus.jl
kellertuer: Ok, I still do not like the type name `TorusinR3` – in this case there is no distinction between the one build of two circles (in the name).

docs/src/manifolds/torus.md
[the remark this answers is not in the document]
mateuszbaran: OK, it makes sense -- I will rename it.

src/differentiation/ode_callback.jl
mateuszbaran: What has to be `determinate`?

kellertuer: The function name should be `determinate_integrator_near_chart_boundary` since `terminate` would mean to end (and not to find out).

mateuszbaran: Well, but it actually terminates the internal integrator?

kellertuer: Oh, me re-reading PRs before first coffee 😴

src/manifolds/CurvedTorus.jl
mateuszbaran: `EmbeddedTorus` seems a bit too general to me but you're right that it's better than `TorusinR3`

src/manifolds/CurvedTorus.jl
mateuszbaran: I've added documentation.

src/manifolds/CurvedTorus.jl
mateuszbaran: I've changed the name.

mateuszbaran:
I think I've resolved all issues :slightly_smiling_face: .

kellertuer (approved):

PR #534, opened 2022-09-23
:books:

kellertuer:
This starts a tutorial section, for no just with a dummy notebook, because I am still pondering about a good start.

Maybe something like generating data on the sphere and computing a mean and a second short part (or even notebook) on comparing different metric on SPDs?

One could also adapt the Benchmark and illustrate the use of different array types with that.


mateuszbaran:
Good idea, and also a good PR name :smile:

kellertuer:
@FedeClaudi I started a tutorial, but I am not sure this is too plain to start with, see

https://github.com/JuliaManifolds/Manifolds.jl/blob/kellertuer/start-tutorials/tutorials/getstarted.jl

(can also be opened as a Pluto Notebook and hopefully renders fine also in the docs on the server, it did locally already, but the server is not yet sure).

Let me know what you think and whether this is a nice start (since I am still unsure how and where to start).

kellertuer:
The Pluto notebook finally renders

https://juliamanifolds.github.io/Manifolds.jl/previews/PR534/tutorials/getstarted.html

FedeClaudi:
That's great! I really like this slower approach introducing all concepts one at the time. Great job

kellertuer:
Ok, then I will continue this one, as soon as my teaching (and my language course) allows.

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1051 @@
...
+Since the package heavily depends on [`ManifoldsBase.jl`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/) we will sometimes also link to the interface definition of functions in the interface and mark this with 🔗. When referring to [Wikipedia](https://en.wikipedia.org/), the link is marked with 📖.
+"""
+
+# ╔═╡ b34d2b6c-907e-45b3-9b62-445666413b26
+md"""
+## Contents
+* [Using the library of manifolds](#Using-the-Library-of-Manifolds)
+* [implementing generic functions](#Implementing-generic-Functions)
+* Allocating and in-place computations
+* decorating a manifold
```
mateuszbaran, line 44: I'd suggest adding a short section about differences between chart-based and intrinsic approaches to computations on a manifold. Maybe even one or two paragraphs that would explain that while the standard textbook approach is to have an atlas and do everything in a chart, the package focuses on intrinsic computations based in an embedding. I remember some people reporting being confused by that.

After #533 is merged we can turn that torus example into a section about working in charts.

kellertuer: Uff, until now I would have no clue how to explain that. Even when I started with the second section it not only got far too long (I think), it also is quite technical (and that part is easier than charts I think).

mateuszbaran: OK, I will suggest a text later. I think it's fine that not everyone interested in Manifolds.jl would understand such explanation because it may be important for people familiar with charts.

FedeClaudi: yes absolutely. I came from a differential geometry background and after implementing a couple toy packages that were charts based and I remember it took me a while to realize that Manifolds.jl was working with embedded manifolds. I think a note would clear that up for people interested in it, and it's okay if for some (though I suspect not many) is a bit obscure

kellertuer:
Since the general technique is working, I moved it from draft to open. Feel free to add sections to the first tutorial. I will slowly continue on that as well if I find good ideas.

mateuszbaran:
I've tried converting the torus example into a section of Pluto notebook and within half an hour it managed to annoy me by taking a lot of time to install all packages into a new environment and discarding a paragraph of a description I've written. I'm not sure I want to bother with it further.

Anyway, here is the part of text that wasn't discarded:
```
## Representations with and without charts.

There are two conceptually different approaches to working on a manifold. The first one, widespread in differential geometry textbooks, is based on defining an atlas on the manifold and performing computations in selected charts. This approach, while generic, is not ideally suitable in all circumstances. For example, working in charts that do not cover the entire manifold causes issues with having to switch charts when operating on a manifold. Furthermore, for some manifolds, there are efficient closed-form formulas for standard functions like the exponential map or Riemannian distance that use a chart-free representation. Manifolds.jl supports both approaches, although the chart-free approach is the main focus of the library.

This section of the tutorial focuses on the chart-based computation.

First, we need to load a few additional libraries needed for visualization and solving differential equations that show up in the chart-based approach:
```

mateuszbaran:
Cool, I will leave it to you then.

tutorials/getstarted.jl
kellertuer: I just finished a decorating section and would like to do the charts with the example from Mateusz – I can add a link to that here. I think a “Further Reading” section would be a good final one here.

mateuszbaran: OK, do you think it would make sense to rewrite my example as a docs page or a Pluto notebook?

kellertuer: I will rewrite yours as a Pluto notebook, sure :)

mateuszbaran: Cool, thanks :slightly_smiling_face:

kellertuer:
Ah as a final technical step we have to find out how to best run GLMakie in Documenter on CI (that is without a X11).

kellertuer:
I fixed quite a few typos and turned all DOIs and arXiv numbers into links, as well as unified footnote marks.

I switched for the charts tutorial to WGLMakie, but I fear that (even here, locally, in PlutoStatic) the rendering takes very very long.

kellertuer:
Hm it seems that is only sometimes runs locally, so I am not sure how we can get the plotting to work in PlutoStaticHTML for the tutorial for now.

mateuszbaran:
I think we can just have a PNG of the plot in static documentation with a note that for interactivity the notebook must be run locally.

kellertuer:
Good idea, I can steal that from the Asymptote plots in Manopt.jl where I do the same (basically with a flag ion the beginning of the notebook, that here would be `static=true`.

kellertuer:
I hope I found a good way:

1) `interactive` can be activated or deactivated (showing sliders or setting fixed values)
2) `render_images` is set to render images. If it is false, the images from the hard drive are loaded.

So for the docs both are false, setting interactive to true also loads GLMakie and renders as well. Setting interactive to false and rendering to true updates the images (on interactive mode images are not updated on hard drive).

I also only load GLMakie/Makie if either of the two is true, hoping that this works on CI as well.

mateuszbaran:
Thanks, it sounds like a very good solution :slightly_smiling_face: .

kellertuer:
The only problem is, though I put the `using` into a condition it seems to try to use it even if the condition is false (leading to an error since X11 is not available on CI).

mateuszbaran:
I think you could import `Makie` always, it shouldn't need a display and would let us use things like `Point3f` later in the code.

It's really weird, though, that `GLMakie` is imported, locally it isn't for me.

mateuszbaran:
I wonder if there would be the same error if there was no `using GLMakie` at all?

kellertuer:
I will try but I fear we might need that for the 3D rendering?

mateuszbaran:
Yes, we need GLMakie, I'm mostly curious if that's a PlutoStaticHTML.jl bug.

kellertuer:
I am not sure whether its PlutoStaticHTML or even Pluto itself (as well).

kellertuer:
I started a discussion at https://github.com/fonsp/Pluto.jl/discussions/2345

mateuszbaran:
I think that's also fine, this issue with Pluto/GLMakie shouldn't block this PR.

kellertuer:
Ok, so later today I will remove the second tutorial for now and open an issue for that to keep track :)

docs/src/features/utilities.md
```diff
@@ -7,6 +7,16 @@ in
 TangentSpace
 ````
 
+# Fallback for the exponential map: Solving the corresponding ODE
+
+When additionally loading [`NLSolve.jl`](https://github.com/JuliaNLSolvers/NLsolve.jl) the following fallback for the exponential map is available.
```
mateuszbaran, line 12: NLSolve is for approximate inverse retraction, not the exponential map.

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+"""
+
+# ╔═╡ a02f085b-eaef-4a8f-a8fe-e306e956dd2c
+M₆ = M₂ × M₃
+
+# ╔═╡ 0f135d94-25c8-499d-a9e0-169db76cb901
+md"Since now the representations might differ from element to element, we have to encapsulate these in their own type."
+
+# ╔═╡ fa0baedb-636e-4ac8-9779-039625ca8267
+p₃ = ProductRepr([0, 0, 1], [0, 1, 0])
```
mateuszbaran, line 278:
```suggestion
p₃ = ArrayPartition([0, 0, 1], [0, 1, 0])
```
I think we should generally recommend `ArrayPartition` over `ProductRepr` where both work well.

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+"
+
+# ╔═╡ a0ac32e0-3f41-4bc2-910b-85a0d1f1a4fd
+md"""
+## Allocating and in-place computations
+
+Memory allocation is a [critical performace issue](https://docs.julialang.org/en/v1/manual/performance-tips/#Measure-performance-with-[@time](@ref)-and-pay-attention-to-memory-allocation) when programming in Julia. To take this into account, `Manifolds.jl` provides special functions to reduce the amount of allocations.
+
+We again look at the [📖 expontial map](https://en.wikip edia.org/wiki/Exponential_map_(Riemannian_geometry)). On a manifold `M` the exponential map needs a point `p` (to start from) and a tangent vector `X`, which can be seen as direction to “walk into” as well as the length to walk into this direction. In `Manifolds.jl` the function can then be called with `q = exp(M, p, X)` [🔗](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions.html#Base.exp-Tuple{AbstractManifold,%20Any,%20Any}). This function returns the resulting point `q`, which requires to allocate new memory.
```
mateuszbaran, line 384:
```suggestion
We again look at the [📖 exponential map](https://en.wikip edia.org/wiki/Exponential_map_(Riemannian_geometry)). On a manifold `M` the exponential map needs a point `p` (to start from) and a tangent vector `X`, which can be seen as direction to “walk into” as well as the length to walk into this direction. In `Manifolds.jl` the function can then be called with `q = exp(M, p, X)` [🔗](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions.html#Base.exp-Tuple{AbstractManifold,%20Any,%20Any}). This function returns the resulting point `q`, which requires to allocate new memory.
```

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+end
+
+# ╔═╡ d879ea0d-1bb2-438a-a75e-1c8a89e19f2a
+md"""
+This works fine on the sphere, see [this tutorial](https://manoptjl.org/stable/tutorials/Bezier/) for an optimization task involving Bézier curves.
+"""
+
+# ╔═╡ cac20d84-d5c2-4631-9269-6cd44d1fc8d7
+md"""
+Now on several manifolds the [📖 expontial map](https://en.wikipedia.org/wiki/Exponential_map_(Riemannian_geometry)) and its (locally defined) inverse, the logarithmic map might not be available in an implementation. So one way to generalise this, is the use of a retraction (see [^AbsilMahonySepulchre2008], Def. 4.1.1 for details) and its (local) inverse.
```
mateuszbaran, line 333:
```suggestion
Now on several manifolds the [📖 exponential map](https://en.wikipedia.org/wiki/Exponential_map_(Riemannian_geometry)) and its (locally defined) inverse, the logarithmic map might not be available in an implementation. So one way to generalise this, is the use of a retraction (see [^AbsilMahonySepulchre2008], Def. 4.1.1 for details) and its (local) inverse.
```

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+
+To avoid this allocation, the function `exp!(M, q, p, X)` [🔗](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions.html#ManifoldsBase.exp!-Tuple{AbstractManifold,%20Any,%20Any,%20Any}) can be called. Here `q` is allocated beforehand and is passed as the memory, where the result is returned in.
+It might be used even for interims computations, as long as it does not introduce side effects. Thas means that even with `exp!(M, p, p, X)` the result is correct.
+"""
+
+# ╔═╡ e9e6729f-6818-4f6d-a4ee-2767f8e2b6fa
+md"""
+Let's look at an example.
+
+We take another look at the [Sphere](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/sphere.html#Manifolds.Sphere) but now a high.dimensional one.
```
mateuszbaran, line 394:
```suggestion
We take another look at the [`Sphere`](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/sphere.html#Manifolds.Sphere) but now a high-dimensional one.
```

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+
+# ╔═╡ d4226ae7-d11f-47a1-a2b0-c631945c5fb7
+M₈ = SymmetricPositiveDefinite(3)
+
+# ╔═╡ 82dca58f-52e6-491e-97ec-904bb5b0b36b
+md"""
+which is the manifold of ``3×3`` matrices that are [symmetric and positive definite](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/symmetricpositivedefinite.html#Manifolds.SymmetricPositiveDefinite).
+which has a default as well, the affine invariant [`LinearAffineMetric`](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/symmetricpositivedefinite.html#Default-metric:-the-linear-affine-metric), but also has several different metrics.
+
+Two switch the metric, we use the idea of a [📖 decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern)-like approach. Defining
```
mateuszbaran, line 458:
```suggestion
To switch the metric, we use the idea of a [📖 decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern)-like approach. Defining
```

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+
+[^AbsilMahonySepulchre2008]:
+    > Absil, P.-A., Mahony, R. and Sepulchre R.,
+    > _Optimization Algorithms on Matrix Manifolds_
+    > Princeton University Press, 2008,
+    > doi: [10.1515/9781400830244](https://doi.org/10.1515/9781400830244)
+    > [open access](http://press.princeton.edu/chapters/absil/)
+[^AxenBaranBergmannRzecki2022]:
+	>Axen, S. D., Baran, M., Bergmann, R. and Rzecki, K:
+	> _Manifolds.jl: An Extensible Julia Framework for Data Analysis on Manifolds_,
```
mateuszbaran, line 520:
```suggestion
    >Axen, S. D., Baran, M., Bergmann, R. and Rzecki, K:
    > _Manifolds.jl: An Extensible Julia Framework for Data Analysis on Manifolds_,
```
Just replacing tabs with spaces.

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+    > arXiv preprint, 2022, [2106.08777](https://arxiv.org/abs/2106.08777)
+[^BergmannGousenbourger2018]:
+    > Bergmann, R. and Gousenbourger, P.-Y.:
+	> _A variational model for data fitting on manifolds
```
mateuszbaran, line 524:
```suggestion
    > _A variational model for data fitting on manifolds
```

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1220 @@
...
+"""
+
+# ╔═╡ b34d2b6c-907e-45b3-9b62-445666413b26
+md"""
+## Contents
+* [Using the library of manifolds](#Using-the-Library-of-Manifolds)
+* [implementing generic functions](#Implementing-generic-Functions)
+* [Allocating and in-place computations](#Allocating-and-in-place-computations)
+* [Decorating a manifold](#Decorating-a-manifold)
+* Representations with and without charts.
```
mateuszbaran, line 76:
```suggestion
* Representations with and without charts (see notebook `working-in-charts.jl`, currently not rendered as a part of this documentation).
```

mateuszbaran (commented):
A few quick comments :slightly_smiling_face: .

docs/src/features/utilities.md
kellertuer: Hm? I am confused. It is either approximating Log or exp but approximating a retraction would not make sense (also I think I did not change this, but to fix errors is of course always good).

tutorials/getstarted.jl
kellertuer: This might be also due to Pluto changing the format now and then (I excluded the Pluto notebooks fro the formatter since the default layout of Pluto yields a false in formatter and I do not want them to “fight”).

docs/src/features/utilities.md
mateuszbaran: NLSolve is for approximating Log, for Exp we need OrdinaryDiffEq.

tutorials/getstarted.jl
mateuszbaran: OK, that's a very minor issue.

docs/src/features/utilities.md
kellertuer: That (to me) also sounds a little strange to have two dependencies there, but then ist log probably, yes.

tutorials/getstarted.jl
kellertuer: So this now throws an error...

kellertuer:
The problem was that `ArrayPartition` is currently not exported. Should we export that or do you think that name is too generic for that?

For a beginners tutorial I found `ProductRepr` nicer as a name.

mateuszbaran:
Yes, maybe we should export that. When I use `ArrayPartition` I always just do `using RecursiveArrayTools` so I forgot about that.
Regarding `ProductRepr`, I'm currently not sure that it is needed. I'm considering removing it at some point.

kellertuer:
Ok, will add the export.
Oh I did not say anything about usefulness, just that in the tutorial at that place the _name_ fits better and is easier to understand than ArrayPartition. We could also leave ProductRepr or ProductRepresentation as an alias for ArrayPartition when we remove ProductRepr as its own type?

mateuszbaran:
Yes, we could make `ProductRepr` an alias for `ArrayPartition` in the future. What I meant regarding usefulness is that we probably shouldn't advertise functionality that is not useful in our tutorial :).

kellertuer:
That is why I came up with the idea of the alias (or just slight renaming). Sure we should not advertise deprecated things, just as a novel user working with a product manifold, ProductRepr sounds reasonable and ArrayPartition only after some explanation (ah but I could add that explanation as well why we want to partition the array at that point).

kellertuer:
I added a small remark on the ArrayPartition.

The only problem is that until releasing this version here the Manifold. has to stay upfront, because only after updating the Pluto notebook afterwards, the version used in the notebook is too old.

But besides this, the PR should be finished.

mateuszbaran:
> The only problem is that until releasing this version here the Manifold. has to stay upfront, because only after updating the Pluto notebook afterwards, the version used in the notebook is too old.

I think that's fine, we can remove `Manifold.` in a future update.

tutorials/getstarted.jl
```diff
@@ -0,0 +1,1225 @@
...
+
+# ╔═╡ 0f135d94-25c8-499d-a9e0-169db76cb901
+md"Since now the representations might differ from element to element, we have to encapsulate these in their own type."
+
+# ╔═╡ fa0baedb-636e-4ac8-9779-039625ca8267
+p₃ = Manifolds.ArrayPartition([0, 0, 1], [0, 1, 0])
+
+# ╔═╡ 36140614-7d40-4281-ba75-2b7b7106e637
+md"""
+Here `ArrayPartition` taken from [`RecursiveTools.jl`](https://github.com/SciML/RecursiveArrayTools.jl) to store the point on the product manifold efficiently in one array, still allowing efficient acces to the product elements.
```
mateuszbaran, line 282:
```suggestion
Here `ArrayPartition` taken from [`RecursiveArrayTools.jl`](https://github.com/SciML/RecursiveArrayTools.jl) to store the point on the product manifold efficiently in one array, still allowing efficient access to the product elements.
```

mateuszbaran (approved):

tutorials/getstarted.jl
kellertuer: Oh, yeah, typing and thinking in parallel was difficult here.

PR #535, opened 2022-09-24
Add submersion metric for Stiefel manifold

### LINKED ISSUE #410: Add logarithmic map on the Stiefel manifold
http://arxiv.org/abs/2103.12046 gives an efficient shooting algorithm (Alg 2) for computing the logarithmic map on the `Stiefel(n, k)` manifold for `n ≥ 2k` for a family of metrics that includes both the Euclidean and canonical (also Alg 4) metrics. I haven't check it, but their implementation of the exponential map in Section 3.1 might also improve on ours.

### LINKED ISSUE #463: More metrics on Stiefel
Described here: https://www.aimsciences.org/article/doi/10.3934/jgm.2020031

sethaxen:
resolves #410, resolves #463


sethaxen:
There are a few more edge cases to handle, and docstrings and tests are still needed, but the PR should otherwise be function-complete.

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,290 @@
...
+# Constructor
+
+    StiefelSubmersionMetric(α)
+
+Construct the submersion metric on the Stiefel manifold with the parameter ``α > -1``.
+
+The submersion metric family has two special cases:
+- ``α = -\frac{1}{2}``: [`EuclideanMetric`](@ref)
+- ``α = 0``: [`CanonicalMetric`](@ref)
+"""
```
kellertuer, line 15: Maybe we could add a reference to Zimmermann Hüper here?

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,290 @@
...
+struct StiefelSubmersionMetric{T<:Real} <: RiemannianMetric
+    α::T
+end
+
+struct StiefelShootingInverseRetraction{
```
kellertuer, line 20: We could omit the Stiefel here, since that could be used to arbitrary Shooting methods as well?

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,290 @@
...
+
+    StiefelSubmersionMetric(α)
+
+Construct the submersion metric on the Stiefel manifold with the parameter ``α > -1``.
+
+The submersion metric family has two special cases:
+- ``α = -\frac{1}{2}``: [`EuclideanMetric`](@ref)
+- ``α = 0``: [`CanonicalMetric`](@ref)
+"""
+struct StiefelSubmersionMetric{T<:Real} <: RiemannianMetric
```
kellertuer, line 16: Hm, now that I though about this, the Stiefel is already mentioned in a manifold whenever this is used so maybe SubmersionMetric alone would be fine, too?

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,290 @@
...
+    return StiefelShootingInverseRetraction(
+        max_iterations,
+        tolerance,
+        num_transport_points,
+        retraction,
+        vector_transport,
+    )
+end
+
+struct StiefelKShootingInverseRetraction{T<:Real} <: AbstractInverseRetractionMethod
```
kellertuer, line 47: Same as above + what is the `K` for?

sethaxen: No this one is definitely Stiefel-specific. This is Algorithm 2 of the paper, and it's based on the observation that the solution and all intermediates of the shooting method for Stiefel occupy a `k \times k` matrix subspace, so when `n >> k`, this is much more efficient.

This is also why it's limited to a specific choice of vector transport and retraction. While for shooting we can be generic, here we cannot.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: Maybe! I'll need to look at the original paper. There are many ways to implement shooting. Many of them user derivatives. This one avoids derivatives by using vector transport. To simply call it `ShootingInverseRetraction` we may want to implement something more generic.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: Yup! Documentation is still a WIP.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: I don't think I like that. e.g. for `CanonicalMetric` we created `ApproximateLogarithmicMap`, which is a very generic name, but the implemented fields are just those needed for `CanonicalMetric`, and that object won't necessarily be useful for every approximate logarithmic map one might need. Since we have no use for a submersion metric outside of `Stiefel`, I lean towards keeping it specific.

src/manifolds/StiefelSubmersionMetric.jl
kellertuer: Or even just reusing the type later

src/manifolds/StiefelSubmersionMetric.jl
kellertuer: Ah ok, just in reading this in use (`retract(M::Stiefel, p, X, m::StiefelKShootingInverseretraction)`stifle appears double now (and the `K` reads a little strange in the name.

src/manifolds/StiefelSubmersionMetric.jl
kellertuer: I have two points why I would prefer a generic name

* using the metric we will always use it with Stiefel so in any signature the name Stiefel doubles
* if we have a use for a submersion metric, changing the name here would be a breaking change.

But you are also right, with the alpha it might be quite specific for Stiefel; then the second point is obsolete – still in any function signature it reads a little strange with the Stiefel appearing twice.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: Okay I made it a generic method and then specialized it for `k <= div(n, 2)`, so we could remove the `k`-shooting inverse retraction type.

sethaxen:
The implementation of the generic shooting method is now quite general. It differs in 2 ways from the one in the algorithm in the paper:
1. the initial guess is not re-scaled after projection to have the same length as the gap vector. This is because we have a `ProjectionInverseRetraction` that just computes the projection without rescaling. This is probably not so important.
2. `norm(X)` has been replaced with `norm(M, p, X)`. For `StiefelSubmersionMetric` this will be more expensive, but it probably makes the method more generalizable, since `norm(X)` will not necessarily tell us anything about the magnitude of the vector.

I think the one shooting case the method doesn't cover is the one that minimizes the gradient. This should probably be its own inverse retraction type. Some demos:

```julia
julia> inverse_retraction = ShootingInverseRetraction(ExponentialRetraction(), ProjectionInverseRetraction(), ProjectionTransport(), 4, sqrt(eps()), 1_000);

julia> M = Sphere(5, ℂ);

julia> p = project(M, randn(ComplexF64, representation_size(M)));

julia> X = project(M, p, randn(representation_size(M)));

julia> X .*= inv(norm(M, p, X));

julia> isapprox(M, p, inverse_retract(M, p, exp(M, p, X), inverse_retraction), X)
true
```

kellertuer:
Cool! For the first point we could introduce a `ScaledProjectionRetraction` (which is a retraction as long as in the limit for smaller and smaller vectors the scaling tends to 1) and where one can change/adapt scaling.

I think the second change is, though more expensive, the better choice (in general).

src/shooting.jl
```diff
@@ -0,0 +1,142 @@
...
+        retract!(M, retr_tX, p, X, retraction)
+        inverse_retract!(M, ΔX, retr_tX, q, initial_inverse_retraction)
+        gap = norm(M, retr_tX, ΔX)
+        for t in transport_grid
+            tX .= t .* X
+            retract!(M, retr_tX_new, p, tX, retraction)
+            vector_transport_to!(M, ΔXnew, retr_tX, ΔX, retr_tX_new, vector_transport)
+            # realias storage
+            retr_tX, retr_tX_new, ΔX, ΔXnew, tX = retr_tX_new, retr_tX, ΔXnew, ΔX, ΔX
+        end
```
sethaxen, line 136: This seems like it should be its own generic vector transport method. That is, given `(p, X, q)` and a `retraction_method` such that `q = retr(M, p, X)`, then the method extends a local vector transport to be longer range by incrementally transporting along the discretization given by `t -> retr(M, p, t*X)`.

sethaxen: I think we could to this with something like the following:
```julia

struct VectorTransportAlongRetraction
    M
    p
    X
    ts
    retraction
    vector_transport
end

struct VectorTransportAlong
    M
    c
    vector_transport
end

function vector_transport_to!(
    M::AbstractManifold,
    Y,
    p,
    X,
    q,
    method::VectorTransportAlongRetraction,
)
    c = map(t -> retract(M, p, t * X, method.retraction))
    return vector_transport_to!(M, Y, p, X, q, VectorTransportAlong(M, c, method.vector_transport))
end

function vector_transport_to!(
    M::AbstractManifold,
    Y,
    p,
    X,
    q,
    method::VectorTransportAlong,
)
    c = [method.c..., q]
    return vector_transport_along!(M, Y, p, X, c, method.vector_transport)
end
```

Then we would declare the vector transport used here with
```julia
VectorTransportAlongRetraction(
    M,
    p,
    X,
    range(1, 0; length=num_transport_points)[2:end],
    retraction,
    vector_transport,
)
```
and this loop is just replace by a simple `vector_transport_to!` call.

sethaxen:
> Cool! For the first point we could introduce a `ScaledProjectionRetraction` (which is a retraction as long as in the limit for smaller and smaller vectors the scaling tends to 1) and where one can change/adapt scaling.

I think this should be handled in its own PR, since it seems there would be a number of design decisions, and it doesn't seem to be essential for this PR (shooting works without the scaling).

src/shooting.jl
kellertuer: We have these incremental ideas already in pole ladder and schild‘s ladder – and yes that is more like a vector transport to ;)

kellertuer:
Yes you‘re right, might be too much for this PR.

src/shooting.jl
sethaxen: To be clear, I want to just have `vector_transport_to!` in `shooting!`, and I'm trying to come up with missing generic `struct`s that would allow me to perform this specific vector transport using just a single `vector_transport_to!` call.

Can you give an example of how pole ladder and schild's ladder would accomplish this with the current `struct`s? I could not figure out a way.

kellertuer: ah then I misread the last part of your previous message which I read as: the whole stuff can (instead!) be done by a simple vector transport to, but then I am a little confused that your first function is a `_to` function and not an `along` one.

Both pole and schild are little different approaches to your step wise vector transport, but they are also `_to` methods that discretise the path along the way (given e.g. by an retraction) and then for every “step” perform something (which is not a vector transport but an approximation of that). So what you do can not be done with them since they approximate the single step, but they discretise the same way you do.

sethaxen: The challenge is that when we call `shooting`, we don't know what `X` is, and we don't know where the initial "shot" (result of the retraction for the initial `X`) will be. So we have no way of specifying the discretization. So we can either define the discretization in the method as we have done here, or we can define new vector transport types that only implicitly define a discretization.

And I just realized my above types won't work for this anyways, since they require knowing `X` ahead of time, which we won't. So perhaps defining the discretization as done here is the best way, since setting `num_transport_points=2` ignores the discretization anyways.

kellertuer: yes I think just specifying the discretisation would be the best way to go then

sethaxen:
Should `shooting` and `ShootingInverseRetraction` be immediately upstreamed to ManifoldsBase?

kellertuer:
To me it looks generic enough that it could, @mateuszbaran what do you think?

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,293 @@
...
+        return dot(X, Y) - (T(dot(p'X, p'Y)) * (2α + 1)) / (2 * (α + 1))
+    end
+end
+
+function inverse_retract!(
+    M::MetricManifold{ℝ,Stiefel{n,k,ℝ},<:StiefelSubmersionMetric},
+    X,
+    p,
+    q,
+    method::ShootingInverseRetraction{ExponentialRetraction},
```
sethaxen, line 136: If we defined specialized point and vector types that represented a point and its tangent vectors on `Stiefel` in the `k × k` subspace, then perhaps by implementing the necessary `exp`, `inner`, `project`, and addition methods for these types, we could delete this method entirely. Might be overkill though, since we don't otherwise have a use for these. Perhaps for a future PR if they seem to be useful.

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,293 @@
...
+function exp!(
+    M::MetricManifold{ℝ,Stiefel{n,k,ℝ},<:StiefelSubmersionMetric},
+    q,
+    p,
+    X,
+) where {n,k}
+    α = metric(M).α
+    T = Base.promote_eltype(q, p, X)
+    # TODO:
+    # - dispatch to exp! for α = -1/2 and α = 0 for efficiency
```
sethaxen, line 58: I'm not convinced we should do this. Being able to compare is useful for testing.

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,293 @@
...
+    X,
+    p,
+    q,
+)
+    X .= q .- p
+    project!(M, X, p, X)
+    return X
+end
+
+function _vector_transport_factors!(A, R, S, M, N, gap, ϵ, ::Val{k}) where {k}
```
sethaxen, line 221: This isn't super human-readable unless we have the paper open. Should probably rewrite to explicitly show the points in terms of factors.

sethaxen:
We should probably also use this to implement `log` for `EuclideanMetric`.

mateuszbaran:
> To me it looks generic enough that it could, @mateuszbaran what do you think?

I will take a look at it today.

src/manifolds/StiefelSubmersionMetric.jl
mateuszbaran: I also prefer the name `StiefelSubmersionMetric`.

mateuszbaran (commented):
A few quick comments :slightly_smiling_face: .

mateuszbaran:
> > To me it looks generic enough that it could, @mateuszbaran what do you think?
> 
> I will take a look at it today.

I agree, the shooting method is generic enough that it can be in ManifoldsBase.jl.

sethaxen:
> I agree, the shooting method is generic enough that it can be in ManifoldsBase.jl.

Okay! I'll split it out into a PR directly to ManifoldsBase.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: Yes, but they can I believe both be implemented more efficiently than this general implementation.

src/manifolds/StiefelSubmersionMetric.jl
```diff
@@ -0,0 +1,293 @@
...
+
+The logarithmic map is computed using [`ShootingInverseRetraction`](@ref). For
+``k \le \frac{n}{2}``, this is sped up using the ``k``-shooting method of
+[^ZimmermanHüper2022]. Keyword arguments are forwarded to `ShootingInverseRetraction`; see
+that documentation for details. Their defaults are:
+- `num_transport_points=4`
+- `tolerance=sqrt(eps())`
+- `max_iterations=1_000`
+"""
+function log(
```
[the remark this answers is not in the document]
sethaxen: I'm not certain what you mean. `ApproximateLogarithmicMap` is only used for `Stiefel` with `CanonicalMetric`, and it is indeed used for its `log!` implementation: https://github.com/JuliaManifolds/Manifolds.jl/blob/c5dfdd7648440d07a02758ddd36b394ed2a5d120/src/manifolds/StiefelCanonicalMetric.jl#L142-L152

Its two fields do not cover all of the keyword arguments one would need here. What are you suggesting should change?

src/manifolds/StiefelSubmersionMetric.jl
kellertuer: yes, that is why they already are.

src/manifolds/StiefelSubmersionMetric.jl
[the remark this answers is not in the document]
kellertuer: Oh I missed that one – then we do basically the same here as well with the ShootingInverseRetraction, that's right.

sethaxen:
I opted to create a factorization type and implement only the necessary operations on this type. This is better documented, and it also allows us to still use generic algorithms. e.g. we now use the generic shooting algorithm from ManifoldsBase, but for `k < div(n, 2)` we first compute the factorization and then run shooting with the factorization.

The factorization object is somewhat fragile (e.g. its broadcast implementation), and all implementations assume the same base point (`U`) is used (which we guarantee in our code but cannot guarantee in user code), so for now this is an internal implementation detail.

I'll add unit tests for the factorization object in the coming days.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: I have now documented the types so it is hopefully clear.

src/manifolds/StiefelSubmersionMetric.jl
sethaxen: I have opted to explicitly include the factorizations.

kellertuer:
Cool! Haven‘t checked the rendered docs. but the rest looks good so far. Sure as long as the new type is still a little fragile, lets keep it internal.

kellertuer:
By the way I think this also fixes #463.

kellertuer:
I just noticed that the retraction is only implemented for this new metric. Since retractions are defined just using the differential, they are actually retraction independent of which metric you have in mind.
Could we adapt that and also put the retraction into the first section of the documentation page (common and metric independent functions)? 

I also saw the warning and maybe that early in the doctoring is a little too early and we could only mention that near the end?

sethaxen:
> I just noticed that the retraction is only implemented for this new metric. Since retractions are defined just using the differential, they are actually retraction independent of which metric you have in mind. Could we adapt that and also put the retraction into the first section of the documentation page (common and metric independent functions)?

Which retraction do you mean? The only retraction this PR defines is `exp!`, which is dependent on this metric.

> I also saw the warning and maybe that early in the doctoring is a little too early and we could only mention that near the end?

Which warning? What do you mean by "doctoring"?

kellertuer:
I am speaking about this

https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#ManifoldsBase.inverse_retract-Tuple{MetricManifold{ℝ,%20var%22#s113%22,%20var%22#s112%22}%20where%20{var%22#s113%22%3C:Stiefel,%20var%22#s112%22%3C:StiefelSubmersionMetric},%20Any,%20Any,%20ProjectionInverseRetraction}

retraction (or its inverse) and this

https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#ManifoldsBase.inverse_retract-Tuple{MetricManifold{ℝ,%20var%22#s113%22,%20var%22#s112%22}%20where%20{var%22#s113%22%3C:Stiefel,%20var%22#s112%22%3C:StiefelSubmersionMetric},%20Any,%20Any,%20ShootingInverseRetraction}

Both could be inverse retractions just for Stiefel

For the “doctoring” - sorry that was autocorrect (and before first coffee) and should have been “docstring” I mean

https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#Manifolds.StiefelFactorization

the one in here. We could move that down

kellertuer:
Alternatively we could also do a `Internal Data Structure` Section in the very end of the Stiefel page where we put the doc string?

sethaxen:
> I am speaking about this
> 
> [https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#ManifoldsBase.inverse_retract-Tuple{MetricManifold{ℝ,%20var%22#s113%22,%20var%22#s112%22}%20where%20{var%22#s113%22%3C:Stiefel,%20var%22#s112%22%3C:StiefelSubmersionMetric},%20Any,%20Any,%20ProjectionInverseRetraction}](https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#ManifoldsBase.inverse_retract-Tuple%7BMetricManifold%7B%E2%84%9D,%20var%22#s113%22,%20var%22#s112%22%7D%20where%20%7Bvar%22#s113%22%3C:Stiefel,%20var%22#s112%22%3C:StiefelSubmersionMetric%7D,%20Any,%20Any,%20ProjectionInverseRetraction%7D)
> 
> retraction (or its inverse) and this
> 
> [https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#ManifoldsBase.inverse_retract-Tuple{MetricManifold{ℝ,%20var%22#s113%22,%20var%22#s112%22}%20where%20{var%22#s113%22%3C:Stiefel,%20var%22#s112%22%3C:StiefelSubmersionMetric},%20Any,%20Any,%20ShootingInverseRetraction}](https://juliamanifolds.github.io/Manifolds.jl/previews/PR535/manifolds/stiefel.html#ManifoldsBase.inverse_retract-Tuple%7BMetricManifold%7B%E2%84%9D,%20var%22#s113%22,%20var%22#s112%22%7D%20where%20%7Bvar%22#s113%22%3C:Stiefel,%20var%22#s112%22%3C:StiefelSubmersionMetric%7D,%20Any,%20Any,%20ShootingInverseRetraction%7D)
> 
> Both could be inverse retractions just for Stiefel

Yes, certainly, these can be generalized. Although, depending on how the forwarding works, I may still also need to implement them for this metric.

> Alternatively we could also do a `Internal Data Structure` Section in the very end of the Stiefel page where we put the doc string?

Yes, I think I prefer this approach, since right now it is even listed before the metric itself.

kellertuer:
> Yes, certainly, these can be generalized. Although, depending on how the forwarding works, I may still also need to implement them for this metric.

usually you would implement something like `retract_project` then it works fine.
For the second

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/master/src/shooting.jl#L56

works fine, then the traits system is properly taken into account (no need to reimplement it for the specific metric)

> > Alternatively we could also do a `Internal Data Structure` Section in the very end of the Stiefel page where we put the doc string?
> 
> Yes, I think I prefer this approach, since right now it is even listed before the metric itself.
Perfect that would also be my preferred way here

docs/src/manifolds/stiefel.md
```diff
@@ -44,4 +44,24 @@ Pages = ["manifolds/StiefelCanonicalMetric.jl"]
...
+
+```@autodocs
+Modules = [Manifolds]
+Pages = ["manifolds/StiefelSubmersionMetric.jl"]
+Order = [:type, :function]
+Public = true
+Private = false
+```
+
+# Internal types
```
kellertuer, line 57:
```suggestion
## Internal types
```

I think only the main title should be `#`

docs/src/manifolds/stiefel.md
```diff
@@ -44,4 +44,24 @@ Pages = ["manifolds/StiefelCanonicalMetric.jl"]
 Order = [:function]
 ```
 
+# The submersion or normal metric
```
kellertuer, line 47:
```suggestion
## The submersion or normal metric
```

Similar to the other subsections this should also be a `##` (just the main title is a `#`.)

docs/src/manifolds/stiefel.md
```diff
@@ -44,4 +44,24 @@ Pages = ["manifolds/StiefelCanonicalMetric.jl"]
...
+Public = true
+Private = false
+```
+
+# Internal types
+
+```@autodocs
+Modules = [Manifolds]
+Pages = ["manifolds/StiefelSubmersionMetric.jl"]
+Order = [:type, :function]
```
kellertuer, line 62: Maybe just as a small remark – the section is currently titled to cover internal types but the code might also display internal functions. We could unify that by removing functions here or adding functions in the title.

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -110,6 +110,22 @@ end
...
+
+Compute a projection-based inverse retraction.
+
+The retraction is computed by projecting the logarithm map in the embedding to the tangent
+space at ``p``.
+"""
+inverse_retract(::Stiefel, ::Any, ::Any, ::ProjectionInverseRetraction)
+
+function inverse_retract_project!(M::Stiefel, X, p, q)
+    log!(get_embedding(M), X, p, q)
```
sethaxen, line 124: Maybe this should be using `embed` on the points before calling `log!`?

mateuszbaran: Why can't it just do `X .= p .- q`?

sethaxen: It could be; indeed that's what it evaluates to. But I'm trying to be explicit about how we're performing the inverse retraction in the embedding and then projecting. Seems more general, especially if we define custom points.

kellertuer: Since we are on Stiefel we could really just use `X .= p .- q` here.

For special Stiefel point representations I would expect them to overload this?
Otherwise, for the very general implementation you aim for, you could also fall back to implementing this generically for embedded manifolds?

sethaxen: > Since we are on Stiefel we could really just use `X .= p .- q` here.

For simplicity, I went with this approach. A generic default for embedded manifolds could be implemented in a separate PR.

kellertuer: Yes, I think a generic implementation would be nice as well :) For here the simple approach should be enough :)

kellertuer (approved):

PR #538, opened 2022-09-27
Set default retraction and inverse retraction of Stiefel

mateuszbaran:
Resolves a part of https://github.com/JuliaManifolds/Manopt.jl/discussions/153 .

### DIFF
--- a/src/manifolds/StiefelEuclideanMetric.jl
+++ b/src/manifolds/StiefelEuclideanMetric.jl
@@ -1,6 +1,17 @@
 
+"""
+    default_inverse_retraction_method(::Stiefel)
+
+Return [`PolarInverseRetraction`](@ref) as the default inverse retraction for
+[`Stiefel`] manifold.
+"""
 default_inverse_retraction_method(::Stiefel) = PolarInverseRetraction()
 
+"""
+    default_retraction_method(::Stiefel)
+
+Return [`PolarRetraction`](@ref) as the default retraction for [`Stiefel`] manifold.
+"""
 default_retraction_method(::Stiefel) = PolarRetraction()
 
 @doc raw"""
--- a/src/manifolds/StiefelEuclideanMetric.jl
+++ b/src/manifolds/StiefelEuclideanMetric.jl
@@ -1,19 +1,4 @@
 
-"""
-    default_inverse_retraction_method(::Stiefel)
-
-Return [`PolarInverseRetraction`](@ref) as the default inverse retraction for
-[`Stiefel`](@ref) manifold.
-"""
-default_inverse_retraction_method(::Stiefel) = PolarInverseRetraction()
-
-"""
-    default_retraction_method(::Stiefel)
-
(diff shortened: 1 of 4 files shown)


src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -1,4 +1,8 @@
 
+default_inverse_retraction_method(::Stiefel) = PolarInverseRetraction()
```
kellertuer, line 2: We should maybe document these, so they show up in the docs as well (just in case someone wonders what the default on Stiefel is)?

mateuszbaran: Sure, done.

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -1,4 +1,19 @@
 
+"""
+    default_inverse_retraction_method(::Stiefel)
+
+Return [`PolarInverseRetraction`](@ref) as the default inverse retraction for
+[`Stiefel`] manifold.
```
kellertuer, line 6:
```suggestion
[`Stiefel`](@ref) manifold.
```

src/manifolds/StiefelEuclideanMetric.jl
```diff
@@ -1,4 +1,19 @@
...
+"""
+default_inverse_retraction_method(::Stiefel) = PolarInverseRetraction()
+
+"""
+    default_retraction_method(::Stiefel)
+
+Return [`PolarRetraction`](@ref) as the default retraction for [`Stiefel`] manifold.
```
kellertuer, line 13:
```suggestion
Return [`PolarRetraction`](@ref) as the default retraction for [`Stiefel`](@ref) manifold.
```

mateuszbaran:
We have random unrelated test failures :frowning_face:

kellertuer:
Well at least we can also add a default vector transport here :)

mateuszbaran:
> Well at least we can also add a default vector transport here :)

My only problem with that is that I don't know which one should be default :slightly_smiling_face: .

kellertuer:
That's not so complicated, I will push my idea in a second (something that fits the retraction of course)

kellertuer:
Hm, while it seems that is failed every now and then for the last two years, it seems to have started to fail consistently. If only we had a plot that would not depend on Python. I never got to a stable variant of a python installation on my machine but had hoped the CI would.

mateuszbaran:
GR.jl doesn't need Python I think so maybe it would be a good replacement for our tests?

mateuszbaran:
Well, if it doesn't fix itself I'd probably just disable these tests.

kellertuer:
Would be sad, but I have no good other idea – I for now asked it here https://github.com/JuliaPy/PyPlot.jl/issues/557, let's see. I mean it never has been super stable, but usually a second run fixed it.

mateuszbaran:
CI works again so I will merge this :tada:

kellertuer:
Phew. So it was just a hiccup again (though a white long one)

PR #540, opened 2022-10-04
Fix #539

### LINKED ISSUE #539: norm for MetricManifold calls inner of base manifold
`norm` for `MetricManifold` should call `inner` for that same manifold, but it instead seems to call `inner` from the base manifold. e.g.

```julia
julia> using Manifolds

julia> struct SomeMetric <: RiemannianMetric end

julia> Manifolds.inner(::MetricManifold{ℝ,<:AbstractManifold{ℝ},SomeMetric}, p, X, Y) = 3

julia> M = Sphere(2);

julia> p = project(M, randn(3));

julia> X = project(M, p, randn(3));

julia> inner(MetricManifold(M, SomeMetric()), p, X, X) # expected: 3
3

julia> norm(MetricManifold(M, SomeMetric()), p, X)^2  # expected: 3
3.6649191335512095

julia> norm(X)^2
3.6649191335512095
```

mateuszbaran:
This should resolve #539 .

### DIFF
--- a/src/manifolds/MetricManifold.jl
+++ b/src/manifolds/MetricManifold.jl
@@ -64,11 +64,24 @@ struct MetricManifold{𝔽,M<:AbstractManifold{𝔽},G<:AbstractMetric} <:
     metric::G
 end
 
+function _drop_embedding_type(t::TraitList)
+    return TraitList(t.head, _drop_embedding_type(t.tail))
+end
+function _drop_embedding_type(t::TraitList{IsIsometricEmbeddedManifold})
+    return _drop_embedding_type(t.tail)
+end
+function _drop_embedding_type(t::TraitList{IsEmbeddedSubmanifold})
+    return _drop_embedding_type(t.tail)
+end
+_drop_embedding_type(t::EmptyTrait) = t
+
 function active_traits(f, M::MetricManifold, args...)
+    at = active_traits(f, M.manifold, args...)
+    idm = is_default_metric(M.manifold, M.metric)
     return merge_traits(
-        is_default_metric(M.manifold, M.metric) ? IsDefaultMetric(M.metric) : EmptyTrait(),
+        idm ? IsDefaultMetric(M.metric) : EmptyTrait(),
         IsMetricManifold(),
-        active_traits(f, M.manifold, args...),
+        idm ? at : _drop_embedding_type(at),
         is_metric_function(f) ? EmptyTrait() : IsExplicitDecorator(),
     )
 end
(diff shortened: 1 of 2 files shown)


test/metric.jl
```diff
@@ -640,4 +645,11 @@ end
         Y2 = change_representer(M, G, p, X)
         @test Y2 ≈ 2 .* X #scaled metric has a factor 2, removing introduces this factor
     end
+
+    @testset "issue #539" begin
```
kellertuer, line 649: Can we maybe avoid tests with rand? If they start randomly failing, that is hard to debug/check/reproduce.

src/manifolds/MetricManifold.jl
```diff
@@ -64,11 +64,24 @@ struct MetricManifold{𝔽,M<:AbstractManifold{𝔽},G<:AbstractMetric} <:
...
 function active_traits(f, M::MetricManifold, args...)
+    at = active_traits(f, M.manifold, args...)
+    idm = is_default_metric(M.manifold, M.metric)
     return merge_traits(
-        is_default_metric(M.manifold, M.metric) ? IsDefaultMetric(M.metric) : EmptyTrait(),
+        idm ? IsDefaultMetric(M.metric) : EmptyTrait(),
         IsMetricManifold(),
-        active_traits(f, M.manifold, args...),
+        idm ? at : _drop_embedding_type(at),
         is_metric_function(f) ? EmptyTrait() : IsExplicitDecorator(),
```
kellertuer, line 85: I would have thought the `is_metric_function` would have solved this here that `inner` would work, is `inner` not a `metric_function`?

test/metric.jl
mateuszbaran: Good point. This one should be harmless but many other are not, and it's not good to give a bad example.

kellertuer: I have a few tests in Manopt.jl with rand where at least code coverage is dependent on the rand, but did not yet have the time fixing that; and yes, its then more like annoying.

src/manifolds/MetricManifold.jl
kellertuer: ...but why did that then ever work for exp/log ?

mateuszbaran: It generally works for exp/log because either:
1. The metric is the default one, which doesn't hit this problem.
2. We have a special `exp!`/`log!` method anyway for that metric.
3. We use a generic default like `exp!(::TraitList{IsMetricManifold}, M::AbstractDecoratorManifold, q, p, X)`.

kellertuer: Ah I might have missed the last point, sure.

mateuszbaran:
@kellertuer would it be fine to merge it? I'd like to merge it before #533 as a bugfix.

PR #542, opened 2022-10-12
Repair sign of projection inverse retraction for Stiefel

sethaxen:


mateuszbaran (approved):
Sorry, that looks like an issue with my suggestion here: https://github.com/JuliaManifolds/Manifolds.jl/pull/535#discussion_r991647890 -- thank you for fixing it.

mateuszbaran:
Yes, inner product with a zero vector is always equal to 0. Good catch!

sethaxen:
Should this just check the inner product of the two vectors directly, or just check the norm of the difference?

mateuszbaran:
I guess it was supposed to be `isapprox(norm(M, p, X - Y), 0; kwargs...)`. It might need tweaking the default `atol` though.

sethaxen:
> It might need tweaking the default `atol` though.

We don't do that anywhere else though. So this would probably be more stable as a default: `isapprox(inner(M, p, X, Y), norm(M, p, X) * norm(M, p, Y); kwargs...)`

mateuszbaran:
Good question. Maybe it makes some minor difference for the canonical metric but a simple `isapprox` would probably also be fine. @kellertuer what do you think?

kellertuer:
It depends a little bit, which functions are implemented I think. I am in favour of the `-` version, though for that your tangent vectors have to have a `-` .

src/manifolds/Grassmann.jl
```diff
@@ -99,10 +99,12 @@ injectivity_radius(::Grassmann, p) = π / 2
 injectivity_radius(::Grassmann, ::AbstractRetractionMethod) = π / 2
 injectivity_radius(::Grassmann, p, ::AbstractRetractionMethod) = π / 2
 
-function Base.isapprox(M::Grassmann, p, X, Y; kwargs...)
-    return isapprox(sqrt(inner(M, p, zero_vector(M, p), X - Y)), 0; kwargs...)
+function Base.isapprox(M::Grassmann, p, X, Y; atol=sqrt(max_eps(X, Y)), kwargs...)
```
mateuszbaran, line 102: `GeneralizedGrassmann` has the same issue, it would be worth fixing as well.

kellertuer: Yes, that would be great but the problem should be _exactly_ the same fix :)

sethaxen: Already fixed.

mateuszbaran (approved):

sethaxen:
I had to increase a few tolerances (and use the tolerances in a few places they weren't being used in `test_manifold`), but now all tests for these manifolds should pass, save for one: `vector transport method DifferentiatedRetractionVectorTransport{QRRetraction}(QRRetraction())`. This one seems to fail regardless of the tolerance. Could one of you look into this?

mateuszbaran:
I will try to find some time to look at it in the evening.

sethaxen:
Thanks! Seems to have done the trick!

kellertuer:
This looks great! Since we (for now) do that specific qr in two places (and I'll have to check whether GeneralisedStiefel is affected as well) we could also introduce a function for that in our utils?

sethaxen:
Perhaps (it should be called `qr_unique`), but let's do that in another PR.

mateuszbaran:
Yes, let's not delay this set of bugfixes further than necessary. We can work further on these QR improvements for Stiefel in another PR.

mateuszbaran (approved):

PR #545, opened 2022-10-21
Improvements to log for Stiefel

sethaxen:
This PR makes several improvements to `log` implementations for Stiefel for various metrics, following up after #535:
- Add `log!` for `EuclideanMetric`, which just calls `log!` for `StiefelSubmersionMetric`. This approach is taken because the benchmark in the submersion metric paper (https://arxiv.org/abs/2103.12046) showed that this algorithm was fastest for the Euclidean metric.
- Speed up `log!` for `CanonicalMetric`. This is done by:
  - Replacing the view of the Q factor in the final line with an explicit slice. This drastically speeds up the function, since the view approach elementwise computes the entries of the Q factor, which is much slower than computing them all at once.
  - ~Removing the Procrustes pre-processing. I'm not certain what this does. It's absent from the formulation of the algorithm in Algorithm 4 of the above paper.~
  - Use a `lyap` solve in the inner loop. The paper gives a proof that shows that for short between `p` and `q`, `lyap` ("sylvester solve") outperforms the current approach, and for larger distances, the case is reversed. However, it's not clear how to choose between the two programmatically, and in the benchmarks (Table 1), the `lyap` approach converged faster for all cases considered. 
  - Use the `sylvester_factorization`. This probably doesn't speed up much; it just allows for more code reuse.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.8.35"
+version = "0.8.36"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
--- a/src/manifolds/GrassmannStiefel.jl
+++ b/src/manifolds/GrassmannStiefel.jl
@@ -28,6 +28,7 @@ end
 ManifoldsBase.@manifold_element_forwards StiefelPoint value
 ManifoldsBase.@manifold_vector_forwards StiefelTVector value
 ManifoldsBase.@default_manifold_fallbacks Stiefel StiefelPoint StiefelTVector value value
+ManifoldsBase.@default_manifold_fallbacks (Stiefel{n,k,ℝ} where {n,k}) StiefelPoint StiefelTVector value value
 ManifoldsBase.@default_manifold_fallbacks Grassmann StiefelPoint StiefelTVector value value
 
 @doc raw"""
--- a/src/manifolds/StiefelCanonicalMetric.jl
+++ b/src/manifolds/StiefelCanonicalMetric.jl
@@ -163,17 +163,16 @@ function inverse_retract!(
     LV = allocate(V)
     Zcompl = qr(qfact.Z).Q[1:(2k), (k + 1):(2k)]
     @views begin
-        Vcorner = V[(k + 1):(2k), (k + 1):(2k)] #bottom right corner
         Vpcols = V[1:(2k), (k + 1):(2k)] #second half of the columns
         B = LV[(k + 1):(2k), 1:k]
         C = LV[(k + 1):(2k), (k + 1):(2k)]
         copyto!(V[1:(2k), 1:k], qfact.Z)
+        F = svd(Zcompl[(k + 1):(2k), 1:k]) # preprocessing: Procrustes
     end
-    F = svd(Vcorner) # preprocessing: Procrustes
-    S = allocate(B)
     new_Vpcols = allocate(Vpcols)
     mul!(new_Vpcols, Zcompl, F.U)
     mul!(Vpcols, new_Vpcols, F.V')
+    S = allocate(B)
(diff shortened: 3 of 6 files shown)


kellertuer:
The Preprocessing is I think from Algorithm 1 in https://findresearcher.sdu.dk/ws/portalfiles/portal/123107528/Alg4StiefelLog.pdf

kellertuer:
Ah, I also found the Matlab Code Ralf once sent/linked me

https://github.com/RalfZimmermannSDU/RiemannStiefelLog/blob/main/Stiefel_log_RZ_SIMAX_2017/Matlab/Stiefel_Log.m

He does the preprocessing also in his Maylab code and if I remember correctly I based our implementation on his ideas (paper&code)

sethaxen:
Ah, interesting. Algorithm 4 in the Stiefel submersion metric paper leaves out the Procrustes step (doesn't mention it at all), and so does the corresponding code in the same repo: https://github.com/RalfZimmermannSDU/RiemannStiefelLog/blob/main/Stiefel_log_general_metric/Matlab/Stiefel_Log.m#L54. I wonder why.

sethaxen:
Whoops, spoke to soon. They have the option to perform Procrustes but did not compare use of this option in the paper's benchmark.

kellertuer:
We can ask Ralf, I might meet with him next week anyways.

kellertuer:
> Whoops, spoke to soon. They have the option to perform Procrustes but did not compare use of this option in the paper's benchmark.

Maybe we could make that optional and do a benchmark? I wanted to start a few benchmarks publicly anyways, either here in Pluto Notebooks or in a separate Benchmark Repo.

sethaxen:
> Maybe we could make that optional and do a benchmark? I wanted to start a few benchmarks publicly anyways, either here in Pluto Notebooks or in a separate Benchmark Repo.

I like that idea. e.g. I found that our implementation of the shooting method performs much better in our code than it does in the benchmark in their paper. One issue though is that all options in the canonical metric code come from the generically named `ApproximateLogarithmicMap`. The fields of that object are also somewhat generic, but if we added fields for these options, that would make it less generic.

kellertuer:
Hm, could we maybe do a wrapper with Options for retractions?

`RetractionOptions{R<:AbstractRetraction} <: AbstractRetraction` or would that too much overhead?
And I think we have retractions with explicit (i.e. own) keywords?

mateuszbaran:
One option would be to pass manifold-specific options via keyword arguments to constructor of `ApproximateLogarithmicMap` and store them in a `NamedTuple` in the structure.

kellertuer:
I think I would prefer a generic wrapper for fixing keyword arguments of a retraction for a specific type; I can sketch my code idea the next few days

sethaxen:
I wonder if it would be better for this PR to just use the current approach with `svd`, and then we can address the more complicated design decisions separately.

kellertuer:
Sure, the optional parameters (fixed in a retraction type=, so they can easily be passed to a `retract` is beyond this PR for sure.

sethaxen:
Hm, tests on 1.6 seem to fail due to the following ambiguity:
```julia
(log!(M::Stiefel, X::StiefelTVector, p::StiefelPoint, q::StiefelPoint) in Manifolds at /home/runner/.julia/packages/ManifoldsBase/1x5os/src/point_vector_fallbacks.jl:141, log!(M::Stiefel{n, k, ℝ}, X, p, q) where {n, k} in Manifolds at /home/runner/work/Manifolds.jl/Manifolds.jl/src/manifolds/StiefelEuclideanMetric.jl:129)
```

mateuszbaran:
We probably need to add this then:
```julia
ManifoldsBase.@default_manifold_fallbacks (Stiefel{n, k, ℝ} where {n, k}) StiefelPoint StiefelTVector value value
```

sethaxen:
Seems to have worked! This PR should be ready for final review now.

kellertuer:
Can you maybe check which of the new lines is not covered? It seems we just miss one line to get codecov passed :)
I will take a look at the PR the next few days.

sethaxen:
It's the `@views begin` line, which annoyingly is never covered.

kellertuer:
Oh no! I see; That is sad, well then there is not much we can do about that.

kellertuer (approved):
Looks good to me. Thanks for improving that – I am definetly not an expert in such improvements, but I can sometimes (proudly) tell, that I have super co-devs here, always looking out for these tricks (sometimes means if someone lets me tell).

mateuszbaran (approved):

PR #548, opened 2022-11-01
Rework Test Suite.

kellertuer:
This PR is a slow/ly developing) more structured rework of our test suite.

The general ideas are

1. having a `minimal` keyword that reduces tests to a bare minimum (maybe not covering all lines but most – used for integration tests
2. using the same idea to maybe reduce the number of manifolds/groups tested to something smaller in a minimal integration test
3. modularise tests, that is a single test function per function/group. The whole suite can be shorter in function length and the single functions can maybe also replace a few now longer tests in the large part.
4. check the single functions to avoid double tests (saw a few in the very first test I just sketched)

This might be something longer that I will work on every now and then. One goal is also to make the single manifold test files maybe a little nicer to read and maybe reduce them a little as well.

## More details

We currently have a lot of keywords. We could narrow that down as follows and the same time allow for more modules to be tested:

* Each test gets a symbol (for example `:exp` and `:log`
* we use a dictionary to indicate which tests are run, for example something like `[ :exp => true, :retraction => [ ProjectionRetraction() }. This is the new `tests` keyword
* we have a corresponding dictionary for `atom_multipliers` and `rtol_multipliers` as well as their singular name as a keyword for a global multiplier

This especially allows to add more modules / generic tests into the scheme, since we can do something like
`:representations` with an array of tuples to check representations, e.g. `(M, "Sphere(2)")` as well an points that should error, maybe even again tuples of erroring points and their error message

## Status

Besides restructuring tests and removing duplicates, this also aims to reduce overall time. At the start of this PR the suite ran

```juliarepl
Test Summary: |  Pass  Total      Time
Manifolds.jl  | 36373  36373  23m12.7s
```

## ToDo

Besides rewriting all test sets we currently have into neat own functions, these are things I noticed are missing

* [ ] retractions and inverse retractions interplay not tested (function now already exists to test such tuples of methods)


kellertuer:
As a first note – `Stiefel` is still a little strange to me, that we use the `ApproximateLog` as a `log` and have a keyword argument for the tolerance; this is hard to include into a reasonable generic test.

Can we maybe not do that @sethaxen ? (I would also remove that for other places where we have that). My new tests fail with that white a bit (and I am not sure how the old ones, that I mainly copied over, ever worked).

sethaxen:
> As a first note – `Stiefel` is still a little strange to me, that we use the `ApproximateLog` as a `log` and have a keyword argument for the tolerance; this is hard to include into a reasonable generic test.
> 
> Can we maybe not do that @sethaxen ? (I would also remove that for other places where we have that). My new tests fail with that white a bit (and I am not sure how the old ones, that I mainly copied over, ever worked).

I don't think I implemented `ApproximateLog`. I'm in general for not having keyword arguments, but I'm not certain what the alternative is.

kellertuer:
Oh, I had in mind that was you, sorry for that. You were the last to work with Stiefel.

So currently we have
https://github.com/JuliaManifolds/Manifolds.jl/blob/5bbedadf28b179b011a3d464f9f1a8b4c4b91cfc/src/manifolds/StiefelCanonicalMetric.jl#L142-L152

Which I might remove, because we have a specific type for the approximate log and this way it is too complicated (though with https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/130 it will be able to pass them down, I think this way of combining the log and the approx log is wrong. So I prefer to remove these lines and do the tests separately.

kellertuer:
These lines (in the new submersion part) yield the same problem I fear

https://github.com/JuliaManifolds/Manifolds.jl/blob/5bbedadf28b179b011a3d464f9f1a8b4c4b91cfc/src/manifolds/StiefelSubmersionMetric.jl#L237-L258

kellertuer:
I wrote a small section “More details” in the initial post on how I plan to make the test suite more modular and flexible (such that we do not have so many tests outside of that test suite and that its easier to use with a little less keywords). Let me know what you think.

kellertuer:
One main problem that I currently phase is that I might have introduced a bug somewhere, since copying over the tests should not make them fail, but a few ones now even fail with trying to do logs of negative numbers and such. But well, this will be a long time project anyways.

kellertuer:
The first module nearly works; is just 2 retractions on groups missing and somehow the retraction on the embedded torus using the ODE solver now sails, saying 3 is out of range, while the range is up to 3.

kellertuer:
I had a bit of time today to continue this idea. I will now do at least two dictionaries of features
* `tests` to specify tests (currently a union of functions to tests and features to test, given as symbols)
* `tolerances` (also union of functions and symbols)

I do not like the union so I might split this further into
`test_functions`, `test_features` (with symbols), `tol_functions` and `tol_features` (symbols). But this approach I like, because while there is a global default dictionary for all of these, one can easily define a “global maximal test set” to reduce test time by not testing everything.

Example `exp`, `log` are tests now but so could be their interplay as `:explog`  (a feature to test).

And the remaining long list of kwargs in the test_manifolds function is basically the to do list (as a test I just moved `is_point`).

And for sure currently the tests still fail because I surely missed something along the lines.

kellertuer:
For now the tests are very fast but we nearly test anything longer.

Reason – I was tired of trying to fix all old tests in a new (half-baked) scheme.
So let's do a radical cut and start anew.

I also started designing two structs and a few helpers
1. `ManifoldFeatures` shall collect features on a manifold – which decides the tests to run
2. `find_manifold_features` is more a rough-brute-force way to find out which features to test (might even provide a nice show later or so)
3. `ManifoldExpectedValues` – should contain (oh also the tolerances that just ended up in 1.) what to expect from a manifold – an expected dimension for example

A test for a manifold then should be (for best of cases)
1. provide a point and tangent vector type and a number type for scalars
2. features available are found
3. text expectancies are set
4. the suite to test a manifold is run based on 2. and 3.

this way it is easier to run a minimal test (disabling features) - but also just a few lines to test a manifold on different pint/tangent types.

This might not avoid all things (point conversion come to mind for now) but it might yield a nice way to specify a main-feature-test for most manifolds making the current plethora of code (hence just deleted) irrelevant).

Will continue on this slowly, but let me also know what you think

mateuszbaran:
Hm, that looks like a big change and redoing all tests at once will make this difficult to finish in a finite amount of time.

> So there is now functions to analyse manifolds
> https://github.com/JuliaManifolds/Manifolds.jl/blob/kellertuer/modularize-tests/src/utils/manifold_features.jl
> which find automatically which functions are available.

I think for testing we need to make a distinction between "what we expect to work" and "what throws an error". So IMO `ManifoldFeatures` should be explicitly provided for each manifold instead of that exploratory approach (though for initial filling of features `find_manifold_functions` looks nice).

src/utils/manifold_features.jl
```diff
@@ -0,0 +1,177 @@
...
+# Fields
+
+* `functions`
+* `retractions`
+* `inverseretractions`
+* `vector_transports`
+* `tolerances` – based on functions or symbols provide a tolerance
+"""
+struct ManifoldFeatures{F<:Function}
+    functions::Vector{F}
```
mateuszbaran, line 103:
```suggestion
struct ManifoldFeatures{}
    functions::Vector{Function}
```
This needs to always be abstractly typed.

src/utils/manifold_features.jl
```diff
@@ -0,0 +1,177 @@
...
+
+Collect a set of features available on a manifold.
+
+# Fields
+
+* `functions`
+* `retractions`
+* `inverseretractions`
+* `vector_transports`
+* `tolerances` – based on functions or symbols provide a tolerance
```
mateuszbaran, line 100: There is no such field in the struct.

src/utils/manifold_features.jl
```diff
@@ -0,0 +1,177 @@
...
+"""
+struct ManifoldExpectations
+    values::Dict{Symbol,Int}
+    strings::Dict{Symbol,String}
+    tolerances::Dict{Symbol,Float64}
+end
+function ManifoldExpectations(;
+    values=Dict{Symbol,Int}(),
+    strings=Dict{Symbol,String}(),
+    tolerances=Dict{Symbol,Int}(:default => 1e-14),
```
mateuszbaran, line 154:
```suggestion
    tolerances=Dict(:default => 1e-14),
```
`Int` is not the best type to specify a tolerance, and if you construct a non-empty `Dict` then types are automatically inferred.

src/utils/manifold_features.jl
```diff
@@ -0,0 +1,177 @@
...
+
+For the features from [`ManifoldFeatures`](@ref) this struct provides expected
+values and tolerances
+
+* `values`     a dictionary `:Symbol->value` for expected values, e.g. `:dimension`
+* `strings`    a dictionary `:Symbol->String` for expected strings, e.g. `:repr` for `repr(M)`
+* `tolerances` a dictionary `:Symbol->Float64` for tolerances in checks of functions.
+"""
+struct ManifoldExpectations
+    values::Dict{Symbol,Int}
```
mateuszbaran, line 147: Maybe not necessarily `Int`?

kellertuer:
> Hm, that looks like a big change and redoing all tests at once will make this difficult to finish in a finite amount of time.

This PR started in November last year, so I wouldn't‘t mind if this takes still a while – but also in a normalised way reading the tests will go relatively fast I think, at least that is what I am aiming for with these automated feature detection functions.

> I think for testing we need to make a distinction between "what we expect to work" and "what throws an error". So IMO `ManifoldFeatures` should be explicitly provided for each manifold instead of that exploratory approach (though for initial filling of features `find_manifold_functions` looks nice).

I think a mix might be nice? Currently it just checks for a manifold whether a function errors, then it is not added.
Sure that might need some fine-tuning for one or the other manifold, but it is still – I think – a good starting point/default. One can add/remove functions in the setup phase of the test before calling `test_manifold`.

src/utils/manifold_features.jl
kellertuer: oh yes, that was during design its now in expectations ;)

src/utils/manifold_features.jl
kellertuer: ah that was meant to be float.

src/utils/manifold_features.jl
kellertuer: I was a bit unsure here, but yeah, in principle even vectors would be fine, since it should just collect values we expect – I also was not so sure whether splitting of the strings is necessary, though then the dict is `Any` so that might slow down things – or we maybe also do not care so much.

mateuszbaran: I think `Any` here would be perfectly fine. After all, it's code that would be used very few times so there is not much point in making it type-stable.

kellertuer: Ok, i will check whether then the strings can be added here as well.

mateuszbaran:
> I think a mix might be nice? Currently it just checks for a manifold whether a function errors, then it is not added.
> Sure that might need some fine-tuning for one or the other manifold, but it is still – I think – a good starting point/default. One can add/remove functions in the setup phase of the test before calling `test_manifold`.

I'm mostly afraid that without explicit lists of features an error that makes something (that is expected to work) throw an exception could go quietly unnoticed. For example if I replace distance on `Sphere` with `distance(::Sphere, p, q) = error("TODO")`, this should result in a test failure, not "well, apparently the feature is missing now".

kellertuer:
Hm, that is a valid point, but without some automation, I really do not see any chance in finishing this in finite time an honestly just would give up then.
We are currently running 37k tests – an idea was to categorise and modularise them to be more flexible. But if I go through all these tests _manually_ to categorise them – I‘ll be doing that for a few years. I am annoyingly interested in getting docs and tests right but not _that_ much.

kellertuer:
...hm maybe we could store an (automatically generated, every now and then) config file per manifold what should be tested? That way deleting distance would make the tests fail as long as you do not trigger a regeneration of the config file?

mateuszbaran:
> ...hm maybe we could store an (automatically generated, every now and then) config file per manifold what should be tested? That way deleting distance would make the tests fail as long as you do not trigger a regeneration of the config file?

Yes, I think that's the right solution.

kellertuer:
Nevermind the broken test I will fix that when I design the 2 levels tomorrow.

kellertuer:
I put the prepare part still upfront the test – that way, if the config does not exist, it is generated and saved.

Now it only remains to write the single tests per feature and hope, that everything fits this scheme – and I am not yet 100% sure about error messages, we will see. Maybe I do them extra still.

kellertuer:
I think I'll close this for now. Maybe I'll come back to this, but for now it seems a huge effort, that I miss motivation for, since I do not directly see a use;
It might be nice for upstream tests to have a very minimal check, though, which might have been an outcome here.

mateuszbaran:
It's quite hard to cover everything in such PR and I understand your lack of motivation. I think this might make more sense as a gradual replacement for the current test suite rather than a one grand PR if you wanted to come back to it later?

kellertuer:
Yes, probably step by step in small PRs would be an idea, but then we would need to keep track of that in an issue _and_ first decide whether to do that per function or per manifold.

Also I might do a separate PR for the features function I wrote here, that might be nice to have for the docs for example.

mateuszbaran:
I think it would be easier to work per function (or a couple of functions) than per manifold. We would definitely need to keep track of it in an issue.

> Also I might do a separate PR for the features function I wrote here, that might be nice to have for the docs for example.

Yes, that can be useful on its own.

PR #549, opened 2022-11-06
Remove unused type variables

mateuszbaran:


mateuszbaran:
@kellertuer  I've accidentally discovered that https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/134 causes some failures here. I will pre-emptively add `::Bool` in a few places, and a more complete solution can be done in a separate PR.

kellertuer:
I do not follow completely. Besides eventual ambiguities that have to be resolved here, the other PR should not cause errors. SO maybe we can keep this open for the time the other PR is done to update Manifolds accordingly? (In the end I want to use that in my new test cases anyways).

mateuszbaran:
The failures are caused by ambiguities. Adding `::Bool` in a few places fixes it and makes sense either with or without https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/134 so there is no harm in adding it.

kellertuer:
Ah, ok, sure we need those in a few places, yes.

PR #550, opened 2022-11-07
Introduce Kendall's shape space

### LINKED ISSUE #504: Kendall Shape Space
Hello I have a lot of small shapes that I need to compare and I need invariance to scale and rotation, I consider comparing them in Kendall shape space like for example in [1] and [2]
Can it be done in Julia? are there any metrics describing shape similarity implemented?

[1] https://github.com/geomstats/geomstats/blob/dafcbed9e4148733cdc63ae8ff1ce93fc660e09c/geomstats/geometry/pre_shape.py
[2] https://geomstats.github.io/notebooks/14_real_world_applications__hand_poses_analysis_in_kendall_shape_space.html

Thanks!

mateuszbaran:
I've started working on Kendall's shape space (so this resolves #504 ). I'm not sure how much I'm going to cover here but at least I will make a small example of usage (maybe as a Pluto notebook if @kellertuer could help :smiley: ).


kellertuer:
Interesting!
I just decided with Brynjulf and a master student we supervise together to look into applications with Kendalls shape space to check how to maybe implement that in the future. So get this one as far as you like, she might continue from there.

kellertuer:
Since I saw that often, we should also cover the PreShapeSpace.

mateuszbaran:
> Interesting! I just decided with Brynjulf and a master student we supervise together to look into applications with Kendalls shape space to check how to maybe implement that in the future. So get this one as far as you like, she might continue from there.

Cool! I would be nice if she could make an example using Manifolds.jl :slightly_smiling_face: .

> Since I saw that often, we should also cover the PreShapeSpace.

That's just a special case of `ArraySphere` so I was going to just write in documentation that users should use that when they need preshape space.

> And is maybe `ShapeSpace` too generic? There might be a lot of different shape spaces such that this might be more like a good common abstract thing?

That's indeed a problem, there are many variants depending on the group that is acting (here are some examples: https://digitalcommons.wayne.edu/cgi/viewcontent.cgi?article=1417&context=oa_dissertations ). I doubt, though, that we need any abstraction except quotient manifolds for them. At least I'm not going to add any in this PR. Feel free to suggest a different name, I'm not particularly convinced `ShapeSpace` is the right choice but haven't come up with anything better, maybe something like `KendallShapeSpace`?

kellertuer:
Introducing a new name or even type (that could have ArraySphare as Fallback with ExplicitDecorator) for the `KendallsPreShapeSpace` would be nice I think, also to have a few functions already on that level before “Quotienting out” rotations.

I would maybe go for `KendallsShapeSpace` since usually one has the genitive in the name, but yes tI think that would be better.

Concerning the master student – we will see. We just decided last week to go into that direction, so I will read a bit about that. I hope she will end up either contributing a few functions or a nice example/Tutorial. We will see.

mateuszbaran:
> Introducing a new name or even type (that could have ArraySphare as Fallback with ExplicitDecorator) for the `KendallsPreShapeSpace` would be nice I think, also to have a few functions already on that level before “Quotienting out” rotations.

OK, I think it would be easiest to make it a new `KendallsPreShapeSpace <: AbstractSphere` then.

> I would maybe go for `KendallsShapeSpace` since usually one has the genitive in the name, but yes tI think that would be better.

That sounds fine.

mateuszbaran:
Sorry for a small mistake, preshape space is actually a submanifold of `ArraySphere`, I've fixed it in the recent commit.

kellertuer:
But then even better to have it as its own type as I proposed :)

mateuszbaran:
Yes, exactly, it's definitely necessary :slightly_smiling_face: .

mateuszbaran:
I've rewritten a part of https://geomstats.github.io/notebooks/14_real_world_applications__hand_poses_analysis_in_kendall_shape_space.html in Manifolds.jl :slightly_smiling_face: .

kellertuer:
That sounds very interesting, I was a little ill a few days last week so I now am a little stressed to get back on (semester-)track. But I'll take a look as soon as I find time. Great that the PR directly also includes an example :)

I further hope the Quotient manifolds can show its potential here.

mateuszbaran:
No problem, it will take me a few more days to finish this so you can focus on teaching.

Previous quotient manifold work really simplified the implementation here.

kellertuer:
That was what I was hoping for, since we put quite some thought into the quotient manifold functions.

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,236 @@
...
+
+Return the dimension of the [`KendallsShapeSpace`](@ref) manifold `M`. The dimension is given by
+``nk - 1 - n(n - 1)/2``
+
+"""
+function manifold_dimension(::KendallsShapeSpace{n,k}) where {n,k}
+    return n * (k - 1) - 1 - div(n * (n - 1), 2)
+end
+
+function project!(::KendallsShapeSpace{n,k}, q, p) where {n,k}
```
kellertuer, line 204: This function I do not understand isn't it basically calling itself again?

kellertuer (commented):
Here's a few comments.

src/manifolds/ShapeSpace.jl
mateuszbaran: No, it's calling projection in the embedding, I probably should've written `get_embedding(M)` instead of `KendallsPreShapeSpace(n, k)` (notice that there is `Pre` in the name).

kellertuer: Oh, I missed the `Pre` indeed. Maybe using `get_embedding` is more clear here, yes.

src/groups/rotation_action.jl
```diff
@@ -214,3 +214,76 @@ group_manifold(A::RowwiseMultiplicationAction) = A.manifold
...
+    ColumnwiseMultiplicationAction{
+        TM<:AbstractManifold,
+        TO<:GeneralUnitaryMultiplicationGroup,
+        TAD<:ActionDirection,
+    } <: AbstractGroupAction{TAD}
+
+Action of the (special) unitary or orthogonal group [`GeneralUnitaryMultiplicationGroup`](@ref)
+of type `On` columns of points on a matrix manifold `M`.
+
+# Constructor
```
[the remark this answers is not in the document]
mateuszbaran: A matrix is the standard representation for shape spaces. I think it's fine. Notice that it's not just a power manifold because of additional constraints.

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,120 @@
+
+@doc raw"""
+    ShapeSpace{n,k} <: AbstractDecoratorManifold{ℝ}
+
+Kendall's shape space, defined as quotient of a pre-shape space (equivalent to
+[`ArraySphere`](@ref) represented by n×k matrices) by the action [`ColumnwiseMultiplicationAction`](@ref)
```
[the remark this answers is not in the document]
mateuszbaran: Yes, we have pre shape space now, it seems you're looking at an outdated version here.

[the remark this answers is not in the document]
kellertuer: Oh yeah, sorry was a little confused still when I wrote this and forgot to send it and it was sent with the projection remark review today.
This is clarified already because we renamed this :)

tutorials/hand-gestures.jl
```diff
@@ -0,0 +1,60 @@
...
+
+    scatter3d(hands[1, 1, :], hands[1, 2, :], hands[1, 3, :])
+
+    Mshape = KendallsShapeSpace(3, 22)
+    # projecting hands
+    hands_projected = [project(Mshape, hands[i, :, :]) for i in axes(hands, 1)]
+
+    # doing tangent PCA
+    mean_hand = mean(Mshape, hands_projected)
+    # For tangent PCA, we need coordinates in a basis.
```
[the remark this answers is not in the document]
mateuszbaran: Sure, this tutorial should work well as a Pluto notebook if you could help with that :slightly_smiling_face: .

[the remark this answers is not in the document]
kellertuer: Sure, I can take a look, the only problem is that a version with Kendalls shape space has to be registered before it works in a notebook. So I might do that as a second step after this PR is done.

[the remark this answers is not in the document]
mateuszbaran: OK, sure, then let's turn it into a notebook in a separate PR.

mateuszbaran:
I've looked at other features on shape spaces we could consider but they are complicated and not really essential so I'd prefer to merge this PR without them. #551 tracks those things.

kellertuer:
Could we collect the others you found in there as well?

Besides that – I hope to find some time to review this PR soon then.

mateuszbaran:
That issue already has the most important things. Geomstats also has a bunch of curvature derivatives I don't understand and I'm not sure what exactly they are for but that's something more general than Kendall shape spaces.

docs/src/manifolds/shapespace.md
```diff
@@ -0,0 +1,41 @@
+# Shape spaces
+
+Kendall's pre-shape and shape spaces.
+
+```@example
```
kellertuer, line 5: Should we maybe add a little bit of an explanation / description and a reference to the original paper of Kendall?

Besides that I am not 100% sure whether the example would help more before or after the functions, maybe here before is fine.

src/Manifolds.jl
```diff
@@ -367,6 +367,9 @@ include("manifolds/Unitary.jl")
 include("manifolds/Rotations.jl")
 include("manifolds/Orthogonal.jl")
 
+# shape spaces require Sphere
+include("manifolds/ShapeSpace.jl")
```
kellertuer, line 371: Maybe also call the file `KendallShapeSpaces.jl`? Or we could also split the two manifolds into two files; they can still share one documentation page but it would also make it easier to have two sections for the two manifolds in there.

src/groups/rotation_action.jl
```diff
@@ -214,3 +214,76 @@ group_manifold(A::RowwiseMultiplicationAction) = A.manifold
...
+
+base_group(A::LeftColumnwiseMultiplicationAction) = A.On
+
+group_manifold(A::LeftColumnwiseMultiplicationAction) = A.manifold
+
+function inverse_apply(::LeftColumnwiseMultiplicationAction, a, p)
+    return a \ p
+end
+
+function optimal_alignment(A::LeftColumnwiseMultiplicationAction, p, q)
```
kellertuer, line 279: Can we maybe document this function with a formula and a source paper?

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,241 @@
+
+@doc raw"""
+    KendallsPreShapeSpace{n,k} <: AbstractSphere{ℝ}
```
kellertuer, line 3: Maybe we should document here that the is a quotient manifold (maybe mentioning traits in general would be cool in the manifold docs).

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,241 @@
...
+            )
+        end
+    end
+    return nothing
+end
+
+embed(::KendallsPreShapeSpace, p) = p
+embed(::KendallsPreShapeSpace, p, X) = X
+
+function get_embedding(::KendallsPreShapeSpace{N,K}) where {N,K}
```
kellertuer, line 60: Can we document this, that the pre shape space is embedded into its total space (of the quotient)?

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,241 @@
...
+
+@doc raw"""
+    manifold_dimension(M::KendallsPreShapeSpace)
+
+Return the dimension of the [`KendallsPreShapeSpace`](@ref) manifold `M`. The dimension is
+given by ``n(k - 1) - 1``.
+"""
+manifold_dimension(::KendallsPreShapeSpace{n,k}) where {n,k} = n * (k - 1) - 1
+
+function project!(::KendallsPreShapeSpace, q, p)
```
kellertuer, line 72: Also here mentioning that the projection is exactly the removal of the quotient we get rid of?

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,241 @@
...
+    q .= p .- mean(p, dims=2)
+    q ./= norm(q)
+    return q
+end
+
+function project!(::KendallsPreShapeSpace, Y, p, X)
```
kellertuer, line 78: The same here as for the other `project!`

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,241 @@
...
+    end
+    return pX
+end
+
+###
+
+@doc raw"""
+    KendallsShapeSpace{n,k} <: AbstractDecoratorManifold{ℝ}
+
+Kendall's shape space, defined as quotient of a [`KendallsPreShapeSpace`](@ref)
```
kellertuer, line 114: We could again mention traits, and maybe even mention `QuotientManifold` - to make the connection / helping functions more prominent?

src/manifolds/ShapeSpace.jl
```diff
@@ -0,0 +1,241 @@
...
+@doc raw"""
+    KendallsPreShapeSpace{n,k} <: AbstractSphere{ℝ}
+
+Kendall's pre-shape space of ``k`` landmarks in $ℝ^n$ represented by n×k matrices.
+
+# Constructor 
+
+    KendallsPreShapeSpace(n::Int, k::Int)
+"""
+struct KendallsPreShapeSpace{n,k} <: AbstractSphere{ℝ} end
```
kellertuer, line 11: What about its total space?

kellertuer (commented):
Here's a few comments.

In general it looks very nice, though I do not have an overview which other functions are available as well. I for example would love a tangent space basis (also the diagonalising one including curvature), if that exists. Maybe its even inherited from ArraySphere?

mateuszbaran:
> I for example would love a tangent space basis (also the diagonalising one including curvature), if that exists. Maybe its even inherited from ArraySphere?

For the pre-shape space basically everything can be worked out because it's just a fancy sphere. The only issue is that it can't reuse code for ArraySphere because of an additional constraint, which is which doesn't cause any issues with exp/log/PT but it does for bases. I'm just not sure how much work it would be.

For the shape space bases are much more complicated and I don't really want to work on it.

docs/src/manifolds/shapespace.md
mateuszbaran: I will add some explanation and a reference.

I like adding things at the top of documentation pages because it's more visible that way. Documentation can always be found using the search function.

src/Manifolds.jl
mateuszbaran: It was easier to put it in one file for development but I can split it now.

src/manifolds/ShapeSpace.jl
mateuszbaran: I don't really treat the pre-shape space for the purpose of implementation as a quotient though it is sometimes defined as such, I will mention it here.

src/manifolds/ShapeSpace.jl
mateuszbaran: Should we implement `total_space` if we don't implement it fully as a quotient manifold? Or should we also have `get_orbit_action`? Notice this is for pre-shape space, not shape space.

kellertuer: Yes, I meant exactly for `Pre` here. It is a quotient space (though of point could in R^n). For completeness we could maybe add both these functions? I do not see a use directly now, but maybe someone has a use for these (and for considering the PreShape space as a quotient)?

mateuszbaran: This would be quite some work so maybe let's leave that for a future PR?

kellertuer: Ok :)

src/manifolds/ShapeSpace.jl
mateuszbaran: I've mentioned the `IsQuotientManifold` trait in the docstring.

src/manifolds/ShapeSpace.jl
mateuszbaran: I've added a docstring, do you think the wording is OK?

src/Manifolds.jl
mateuszbaran: I've split the file in the last commit.

docs/src/manifolds/shapespace.md
kellertuer: Ok.

src/manifolds/KendallsPreShapeSpace.jl
```diff
@@ -0,0 +1,149 @@
...
+
+# Constructor 
+
+    KendallsPreShapeSpace(n::Int, k::Int)
+
+# References
+
+[^Kendall1989]:
+    > D. G. Kendall, “A Survey of the Statistical Theory of Shape,” Statist. Sci., vol. 4,
+    > no. 2, pp. 87–99, May 1989, doi: 10.1214/ss/1177012582.
```
kellertuer, line 20: Could you turn the DOIs into links? I went through all of the doc strings to make them links, that renders nicer in the docs and makes the references more accessible..

src/manifolds/KendallsPreShapeSpace.jl
```diff
@@ -0,0 +1,149 @@
...
+"""
+    project(M::KendallsPreShapeSpace, p, X)
+
+Project tangent vector `X` at point `p` from the embedding to [`KendallsPreShapeSpace`](@ref)
+by selecting the right element from the tangent space to orthogonal section representing the
+quotient manifold `M`. See Section 3.7 of [^Srivastava2016] for details.
+
+# References
+
+[^Srivastava2016]:
```
kellertuer, line 115: Oh and maybe one Footnote is enough? Currently this might render as 17-or-so footnotes if it is set in every docstring.

src/manifolds/KendallsPreShapeSpace.jl
mateuszbaran: Sure, no problem.

src/manifolds/ShapeSpace.jl
kellertuer: One could maybe even mention, that the method computes the mean of the points and moves that to zero; afterwards the forbenious norm of the points (as a matrix) is normalised to fix the scaling?

src/manifolds/KendallsPreShapeSpace.jl
```diff
@@ -0,0 +1,149 @@
...
+    project(M::KendallsPreShapeSpace, p)
+
+Project point `p` from the embedding to [`KendallsPreShapeSpace`](@ref) by selecting
+the right element from the orthogonal section representing the quotient manifold `M`.
+See Section 3.7 of [^Srivastava2016] for details.
+
+# References
+
+[^Srivastava2016]:
+    > A. Srivastava and E. P. Klassen, Functional and Shape Data Analysis. Springer New York, 2016.
```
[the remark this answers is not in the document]
mateuszbaran: Yes, this book has both ISBN and DOI, I'll add both.

src/manifolds/KendallsPreShapeSpace.jl
mateuszbaran: I've reduced the number of footnotes but it's a bit unfortunate that then they don't appear in docstrings :(.

src/manifolds/ShapeSpace.jl
mateuszbaran: I've added some docs.

src/manifolds/ShapeSpace.jl
mateuszbaran: Done.

src/manifolds/KendallsPreShapeSpace.jl
kellertuer: well, the only other solution is know would be to switch to DocumenterCite and have one page of references, but then they also would not show up in the individual doc strings (actually in none of them). We could also put the footnote just into the main function where this is needed?

src/manifolds/ShapeSpace.jl
mateuszbaran: Sure, I've added that note.

src/manifolds/KendallsPreShapeSpace.jl
mateuszbaran: Ideally this would get merged: https://github.com/JuliaDocs/Documenter.jl/pull/1464 and solve the root of the problem.

kellertuer: Will check that one, but at first glance my problem would be: Which footnote wins?

mateuszbaran: If they are identical, why would it matter?

kellertuer: right I read the topic and sure if they are identical then one is enough.

src/manifolds/ShapeSpace.jl
kellertuer: Ok, might be useful in the future to add that as well, but maybe too much for this PR.

kellertuer (approved):

PR #553, opened 2022-11-19
Coordinates and random vectors on PositiveNumbers and arrays

mateuszbaran:
A small extension that turned out to be useful for a test of https://github.com/JuliaManifolds/Manopt.jl/pull/167 .


kellertuer (approved):
Maybe a preliminary approve, since we need the tests to pass ;) but the rest looks fine.

PR #556, opened 2022-12-02
Improve readme

mateuszbaran:
The current readme doesn't reflect the current state of Manifolds.jl and its ecosystem well so I think it would be a good idea to update it a bit. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,17 +11,16 @@
 | [![](https://img.shields.io/badge/docs-dev-blue.svg)](https://juliamanifolds.github.io/Manifolds.jl/latest/) | [![CI](https://github.com/JuliaManifolds/Manifolds.jl/workflows/CI/badge.svg)](https://github.com/JuliaManifolds/Manifolds.jl/actions?query=workflow%3ACI+branch%3Amaster) | [![DOI](https://zenodo.org/badge/190447542.svg)](https://zenodo.org/badge/latestdoi/190447542) |
 | | [![codecov.io](http://codecov.io/github/JuliaManifolds/Manifolds.jl/coverage.svg?branch=master)](https://codecov.io/gh/JuliaManifolds/Manifolds.jl/) |
 
-
 Package __Manifolds.jl__ aims to provide both a unified interface to define and
 use manifolds as well as a library of manifolds to use for your projects.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,17 +11,16 @@
...
-This package is under development, and subject to changes as needed.
+This package is mostly stable, see https://github.com/JuliaManifolds/Manifolds.jl/issues/438 for planned upcoming changes.
 
 ## Getting started
 
 To install the package just type
 
 ```julia
-] add Manifolds
+using Pkg; Pkg.add("Manifolds")
```
kellertuer, line 23: I feel the `] add Manifolds` is nicer, but you are right, it always is marked as an error in markdown :/

README.md
```diff
@@ -34,6 +33,20 @@ M = Sphere(2)
 γ(0.5)
 ```
 
+Ecosystem highlights:
```
kellertuer, line 36: Nice overview!

kellertuer (commented):
I was thinking whether it might be reasonable to also link the discussion forum?

Also we could check, whether we find a good place to let users know we would be interested in their use-cases; to link them, cite them or even have more tutorials here.

README.md
mateuszbaran: Thanks!

README.md
mateuszbaran: I think the main advantage of `using Pkg; Pkg.add("Manifolds")` is that it can be directly pasted into the command line but Markdown not marking it as an error is also a nice benefit :slightly_smiling_face: .

mateuszbaran:
> I was thinking whether it might be reasonable to also link the discussion forum?

Good idea, I could also add that we have channels on Slack and Zulip.

> Also we could check, whether we find a good place to let users know we would be interested in their use-cases; to link them, cite them or even have more tutorials here.

I will write a sentence next to links to discussion platforms.

README.md
kellertuer: Ok.

kellertuer (approved):

kellertuer (approved):

kellertuer:
(oh, if you review during a meeting it might be acknowledged twice)

PR #558, opened 2022-12-09
Some improvements to Quaternions usage

sethaxen:
The main thing this PR does is set a type constraint to `Quaternions.Quaternion` wherever that is assumed. It also uses `quat` to construct a `Quaternion` instead of the actual constructor (analogous to `Complex`, where `complex` should usually be used to construct) and fixes an indexing issue.


kellertuer (approved):
Looks good to me – can surely be combined with the version bump to 0.7

mateuszbaran (approved):

kellertuer:
Oh we lost a line of CodeCov; if you have a moment, fixing that would be great.

sethaxen:
> Oh we lost a line of CodeCov; if you have a moment, fixing that would be great.

It seems that neither of these two methods was being hit by the tests before this PR, so the loss of one line in coverage is just because this PR adds one line to the uncovered method. I'm not certain what the methods are there for.

kellertuer:
I see. I can think of something.
Sure, we sometimes miss to do code-cos (or it might be there was some time we did). But I see what those methods are supposed to do and can add something later today (after my meetings).

kellertuer:
Meh. We actually have line covering that, both the infectivity radius and the two functions. Codecov is again doing strange things :(

kellertuer:
I think we have to merge master in here anyways?

kellertuer:
After merging (which went without a merge conflict) we have an error on CI that `quatrand` is not defined, what might be the reason?

sethaxen:
> After merging (which went without a merge conflict) we have an error on CI that `quatrand` is not defined, what might be the reason?

Ah of course, this is because we're no longer bring the exported functions from Quaternions into the namespace. I ended up dropping `quatrand` and using `rand` instead. It's doing the same thing as `rand(QuaternionF64)` but in a less standard way.

kellertuer:
Looks good now. I think we can merge this and #560 and release the next version 0.8.41 :)

PR #560, opened 2022-12-12
propose a `riemannian_gradient` function for easier conversion of gradients

### LINKED ISSUE #559: Easier interface for gradient conversion
I noticed yesterday that it might be complicated to convert a Euclidean gradient to a Riemannian one. The solution I endet up using for a given manifold `M` and a euclidean gradient in the embedding `grad_f_Euclidean` is

```julia
 rpb = Manifolds.RiemannianProjectionBackend(Manifolds.ExplicitEmbeddedBackend(M; gradient=grad_f_Euclidean))
grad_f(M,p) = Manifolds.gradient(M, f, p, rpb)
```

which is a little clumsy. We could introduce a nicer interface for this – I am just not sure whether that should be done here or over in ManifoldDiff.

kellertuer:
Resolves #559.

* [x] Inroduce and document `riemannian_gradient`
* [x] Test coverage
* [x] Check documentation
* [x] Rewrite parts in `Symplectic` and `SymplecticStiefel` to use this easier interface as well.


kellertuer:
I am not sure what CodeCov does again (stating master is at 38%), but I Introduced `riemannian_gradient` now as a nice subset in `gradient`, so that it can be used. In its current form it should be very similar to `egrad2rgrad`, just a little more general, since we can come from a non-Euclidean metric in the embedding as well.

Reasoning for the naming is, that the input should make clear, that the `Y` is from the embedding and what this method then computes/returns is the corresponding _Riemannian gradient_.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -286,40 +286,27 @@ Return the total space of the [`SymplecticStiefel`](@ref) manifold, which is the
...
+function riemannian_gradient(::SymplecticStiefel, p, Y)
+    Q_p = SymplecticMatrix(p, Y) * p
+    return Y * (p' * p) .+ Q_p * (Y' * Q_p)
 end
 
-function gradient!(::SymplecticStiefel, f, X, p, backend::RiemannianProjectionBackend)
-    _gradient!(f, X, p, backend.diff_backend)
-    Q_p = SymplecticMatrix(p, X) * p
-    X .= X * (p' * p) .+ Q_p * (X' * Q_p)
+function riemannian_gradient!(::SymplecticStiefel, X, p, Y)
```
mateuszbaran, line 307: I think this function should accept the embedding metric kwarg if it's Euclidean.
```suggestion
function riemannian_gradient!(::SymplecticStiefel, X, p, Y; embedding_metric::EuclideanMetric=EuclideanMetric())
```

mateuszbaran:
Codecov sometimes needs a few hours to properly calculate coverage.

src/manifolds/SymplecticStiefel.jl
kellertuer: I would prefer to just write `kwargs...` then?
Or maybe not – on the other hand, if the embedding metric is not Euclidean, this function would do something wrong, so it should maybe even error? I am not sure atm.

kellertuer:
Interesting. In a few hours – is that humans doing that? I thought they use computers.

src/manifolds/SymplecticStiefel.jl
mateuszbaran: Well, restricting it to `EuclideanMetric` removes the possibility of passing something else so it's safer though a bit more verbose.

mateuszbaran:
We are using it for free so I guess if their servers have lots of other requests we are quite low on the priority list :slightly_smiling_face: .

src/manifolds/SymplecticStiefel.jl
kellertuer: Hm, maybe I wrote that not precise enough.
Of you pass anything else _than_ EuclideanMetric_ this function does something _wrong_, but that is also because the embedding of Symplectic Stiefel is a euclidean space.

kellertuer: Maybe we have to make the metric a positional argument (that is optional when calling it, defaults to `EuclideanMetric`) such that we can dispatch on that, too?

kellertuer: Oh, indeed!

I am also not sure but if you have an embedding with a non-Euclidean metric, then we have that situation.

But thanks for the note, then this is ok :)

src/differentiation/riemannian_diff.jl
```diff
@@ -205,16 +205,82 @@ struct RiemannianProjectionBackend{TADBackend<:AbstractDiffBackend} <:
...
+``\tilde f`` restricted to the manifold ``M``.
+This can also be done in place of `X`.
+
+By default it uses the following computation:
+Let the projection ``Z = \operatorname{proj}_{T_p\mathcal M}(Y)`` of ``Y`` onto the
+tangent space at ``p`` be given, that is with respect to the inner product in the embedding.
+Then
+
+```math
+⟨Z-Y, W⟩ = 0 \text{ for all } W \in T_p\mathal M,
```
mateuszbaran, line 223:
```suggestion
⟨Z-Y, W⟩ = 0 \text{ for all } W \in T_p\mathcal M,
```

mateuszbaran (approved):
I think this looks fine now :slightly_smiling_face: .

PR #561, opened 2022-12-26
Initial work on flag manifold

### LINKED ISSUE #398: Flag manifolds
I've accidentally found this paper about optimization on flag manifolds: https://www.stat.uchicago.edu/~lekheng/work/flag.pdf . I don't have time to read it right now but it may be interesting.

cc @kellertuer

mateuszbaran:
This will resolve #398 . My goal is to eventually build off of this and Manopt.jl and make a new package with lots of variants of PCA, LDA, MDA, FA, CCA etc., obtained as solutions to optimization problems over either Grassmann or flag manifolds -- I think this could be quite useful for many applications.

TODO:
- [x] tests

Not TODO (for this PR):
- tangent space bases
- figuring out distance formula
- logarithmic map
- Riemannian Hessian
- parallel vector transport
- Verify my idea for QR retraction. Could we have inverse QR retraction? Does it even matter?


kellertuer:
We have ManifoldML for that – at least that was where I wanted to do PCA and such. Feel free to use that then :)

kellertuer:
> * Riemannian Hessian

How do you plan to do that? Similar to the gradient function? Or more generic? (like the backend stuff we have)?

mateuszbaran:
> We have ManifoldML for that – at least that was where I wanted to do PCA and such. Feel free to use that then :)

I'm actually thinking about a package without "Manifold" in its name to not scare people away :wink:. At the moment I'm trying to see what's possible and we can decide on the name later when we know what actually works.

> How do you plan to do that? Similar to the gradient function? Or more generic? (like the backend stuff we have)?
> 
> _edit:_ Oh not this PR – we should still think about how to best do this :)

Yes, that's not for this PR. The paper I'm primarily referencing has it so we may want to remember about the flag manifold when we get later to Riemannian Hessians.

kellertuer:
> > We have ManifoldML for that – at least that was where I wanted to do PCA and such. Feel free to use that then :)
> 
> I'm actually thinking about a package without "Manifold" in its name to not scare people away 😉. At the moment I'm trying to see what's possible and we can decide on the name later when we know what actually works.

Hm, I would still favour something likeManifoldML, but sure we have the same idea for a package, I think, so we just have to agree upon the name some time :)

mateuszbaran:
Tests fail because of https://github.com/JuliaLinearAlgebra/RecursiveFactorization.jl/issues/67 .

mateuszbaran:
Recently in this PR we seem to have random failures. After a quick search it looks like out-of-memory issues are a very likely explanation. What do you think about splitting out another group of tests, for example with
```
include_test("statistics.jl")
include_test("approx_inverse_retraction.jl")

# manifolds requiring ODE solvers
include_test("manifolds/embedded_torus.jl")
# plotting
include_test("recipes.jl")
```
It could be called "integration" (with other libraries).

mateuszbaran:
This PR has the essential things done, so I've just cut the less important features from TODO and will try to get this ready soon. It can be improved in subsequent PRs later.

src/manifolds/Flag.jl
```diff
@@ -0,0 +1,190 @@
...
+    return p.value * X.value[:, 1:k]
+end
+
+"""
+    stiefel_tv_to_orthogonal(M::Flag, p::AbstractMatrix, X::AbstractMatrix)
+
+Convert tangent vector from [`Flag`](@ref) manifold `M` from Stiefel representation to
+orthogonal representation.
+"""
+function stiefel_tv_to_orthogonal(M::Flag, p::AbstractMatrix, X::AbstractMatrix)
```
kellertuer, line 152: same as above, nicer with conversions and speficying one of them to be the array-default-representation cf. Hyperbolic with its three representations and Hyperboloid being the default.

src/manifolds/FlagOrthogonal.jl
```diff
@@ -0,0 +1,132 @@
+
+function check_vector(
```
kellertuer, line 2: Can we document this when a vector is valid? Maybe write a doctoring for a `is_vector`?

src/manifolds/FlagOrthogonal.jl
```diff
@@ -0,0 +1,132 @@
...
+) where {N,dp1}
+    project!(SkewHermitianMatrices(N), Y.value, X.value)
+    for i in 1:dp1
+        Bi = _extract_flag(M, Y.value, i)
+        fill!(Bi, 0)
+    end
+    return Y
+end
+
+function project(M::Flag{N,dp1}, ::OrthogonalPoint, X::OrthogonalTVector) where {N,dp1}
```
kellertuer, line 77: Can we document this as well with its math formula?

src/manifolds/FlagOrthogonal.jl
```diff
@@ -0,0 +1,132 @@
...
+                    Bji = _extract_flag(M, pX.value, j, i)
+                    Bji .= -Bij'
+                end
+            end
+        end
+    end
+    return pX
+end
+
+function retract_qr!(
```
kellertuer, line 119: Similarly, please document.

src/manifolds/FlagStiefel.jl
```diff
@@ -0,0 +1,179 @@
...
+
+"""
+    default_vector_transport_method(M::Flag)
+
+Return the [`ProjectionTransport`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/vector_transports.html#ManifoldsBase.ProjectionTransport)
+as the default vector transport method for the [`Flag`](@ref) manifold.
+"""
+default_vector_transport_method(::Flag) = ProjectionTransport()
+
+function inner(
```
kellertuer, line 26: So opposed to what the docs say, Stiefel is our default ;) ?

src/manifolds/FlagStiefel.jl
```diff
@@ -0,0 +1,179 @@
...
+function _extract_flag_stiefel(M::Flag, pX::AbstractMatrix, i::Int)
+    range = (M.subspace_dimensions[i - 1] + 1):M.subspace_dimensions[i]
+    return view(pX, :, range)
+end
+
+function check_point(M::Flag, p::AbstractMatrix)
+    return check_point(get_embedding(M), p)
+end
+
+function check_vector(
```
kellertuer, line 69: Please document `is_vector` here as well.

kellertuer (changes requested):
Here is a few remarks, just a few docs and a discussion about representation and conversion, please check Grassmann or Hyperbolic for comparison on several representations.

src/manifolds/Flag.jl
```diff
@@ -0,0 +1,190 @@
...
+    else
+        return t.x[i]
+    end
+end
+
+@doc raw"""
+    Flag{N,d} <: AbstractDecoratorManifold{ℝ}
+
+Flag manifold of ``d`` subspaces of ``ℝ^N``[^YeWongLim2022]. By default the manifold uses
+the orthogonal coordinates representation.
```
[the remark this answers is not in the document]
mateuszbaran: Good catch -- at first I was thinking about making orthogonal coordinates the default but then I decided to change the default Stiefel coordinates. But I haven't updated this comment. I will take care of it.

src/manifolds/FlagStiefel.jl
mateuszbaran: Yes, sorry for the confusion.

src/manifolds/Flag.jl
```diff
@@ -0,0 +1,190 @@
...
+    return print(io, ")")
+end
+
+"""
+    orthogonal_tv_to_stiefel(M::Flag, p::OrthogonalPoint, X::OrthogonalTVector)
+
+Convert tangent vector from [`Flag`](@ref) manifold `M` from orthogonal representation to
+Stiefel representation.
+"""
+function orthogonal_tv_to_stiefel(M::Flag, p::OrthogonalPoint, X::OrthogonalTVector)
```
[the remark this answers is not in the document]
kellertuer: Which one makes more sense I do not know – you are the expert, but I felt with the conversion stuff, they interact really nicely.

src/manifolds/FlagOrthogonal.jl
mateuszbaran: Sure, though as far as I can see we document `check_vector`, not `is_vector`.

kellertuer: Oh, well, that is also fine, then.

src/manifolds/Flag.jl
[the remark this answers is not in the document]
kellertuer: Oh, can we not see the `k` somewhere in `p` or `X` (implicitly)?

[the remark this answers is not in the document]
mateuszbaran: I don't see a way to do that unfortunately.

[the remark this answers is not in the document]
kellertuer: Ok, then we could go for that convert, sure.

mateuszbaran:
I think I've addressed all comments :slightly_smiling_face:

src/manifolds/Flag.jl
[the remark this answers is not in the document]
kellertuer: I will check later today or Friday – we could also just do a default fallback for the new convert to the old one?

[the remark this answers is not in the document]
mateuszbaran: I will add a fallback in ManifoldsBase.jl then.

kellertuer (approved):

PR #562, opened 2023-01-01
Moving rand to ManifoldsBase.jl

mateuszbaran:
Adapt to https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/141 .


kellertuer:
We still have to import `rand` and `rand!` from ManifoldsBase now, I think?

mateuszbaran:
That shouldn't be necessary, it's still a function defined in `Random`.

kellertuer:
But it makes it more clearer where it comes from? And do we have the decorator thingy that way here?

kellertuer (approved):
I would prefer to import rand/rand! from ManifoldsBase to make clear it was defined there already and to be sure we have the definitions of the decorator available here... besides that this is good to go, so I‘ll approve it already.

mateuszbaran:
OK, I'll add `import`.

PR #564, opened 2023-01-09
Depend on ManifoldDiff.jl

mateuszbaran:
This should be the last remaining part needed for ManifoldDiff.jl conversion.


kellertuer (approved):
Besides a small bump of the version this should be fine. Nice that we have that on a side-track now :) We could write an announcement on Discourse maybe?

kellertuer:
Sure for 0.4 I will write an announcement and write a short summary of the new things,
but I think the first few steps towards Diff might also be nice to announce, maybe we find people to join in :)

kellertuer:
...or to be precise: We announced ManifoldDiffEq – we could do the same for ManifoldDiff, just for people to know that it exists and there is work done on the topic.

kellertuer:
Sure, but it is maybe a tradeoff between us getting back to that (yes we definetly should do that next!) and letting people know so they can check that out and maybe help? In the end that is a tradeoff, but we have the Euclidean-conversion-AD at least which already works :)

mateuszbaran:
OK, sure, we can always have another announcement when the rules are released :slightly_smiling_face: .

PR #568, opened 2023-01-21
Adapt to new Default methods

kellertuer:
uses `default_X_method(M, t)` where suitable.

I am just not yet sure how to “pass down” the type on the product manifold. That is to provide these three methods also with the type (on Product) and pass that down

https://github.com/JuliaManifolds/Manifolds.jl/blob/cdf9d00c7c8d6c23214141c8bf4932e1b0d4a491/src/manifolds/ProductManifold.jl#L346-L356


mateuszbaran:
I will implement the product manifold myself :slightly_smiling_face:

kellertuer:
Hm,
I am a little lost.
Codecoverage from before my commit suggest, that by then everything was fine.

I added a Grassmann and a metric test. All ran fine. I removed the Metric-default-passdown methods. All tests ran fine (maybe I was confused already).

Then I removed the pass down-parts, because MetricManifold is a decorator and should pass down anyways.
Then tests failed (ODE-exp claiming it can not do with exp, which indicates, some default kicks in and returns `ExponentialRetraction()`, not sure where). So, basically the state that passed tests yesterday now fails in tests. I am confused.

So I readied the Metric-Passdown-stuff (so that the first commit basically only adds tests). And the same tests still fail.
So I am confused now and do not know how to fix this. Maybe you see it @mateuszbaran ?

mateuszbaran:
`ConnectionManifold` pass-through looks incorrect to me unless we restrict it to the default connection?

kellertuer:
We had a passthrough for the retraction already, unjust extended that to the others (similar to the Metric),

Note that the retraction and the vector transport just depend on the differential, which is metric/connection-independent.
In other words: a retraction is a retraction independent of the metric chosen.

(That still leaves the point that in some representations retractions might be hard to compute)

mateuszbaran:
Sure, a retraction in one connection is a retraction in a different connection so this does make some sense. I was just worried about that "default" part. Whether we have this forwarding or not the worst case is that we need some specialization for some wrapped manifold but that's fine.

kellertuer:
We could still have specialized ones for special connections, maybe they are nicer in some sense, yes.

mateuszbaran (approved):

PR #570, opened 2023-02-10
Fused retractions

mateuszbaran:
This adapts Manifolds.jl to changes in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/146 .  It mostly works but I need to polish this a bit.


mateuszbaran:
Tests pass locally for me with the ManifoldsBase.jl changes so it should be more or less ready for review.

kellertuer:
Since Manopt 0.14 is basically good to go, we could register that the next days, though I'd have to check in Manopt how much things would need to be adapted there (just got a little lost in a new feature idea instead).

mateuszbaran:
Sure, I will check Manopt tomorrow and see what needs to be changed.

mateuszbaran:
I've checked Manopt and it allowed me to discover and fix a bug in power manifold. The only change in Manopt that I had to do was, strangely, bumping a little tolerance is `test_bezier.jl`.

kellertuer:
Nice! (Bezier will at some point migrate to the examples, still good to know they still work).

test/statistics.jl
```diff
@@ -734,15 +740,15 @@ end
...
+            X1 = hat(R, p0, [1.0, 0.0, 0.0])
+            X2 = hat(R, p0, [0.0, 1.0, 0.0])
+            x = [geodesic(R, p0, X1, π / 2 * (1:4)); geodesic(R, p0, X2, π / 2 * (1:4))]
             w = pweights([rand(rng) for _ in 1:length(x)])
             m = mean(R, x, w)
             mg = mean(R, x, w, GeodesicInterpolation())
             mf = mean(R, x, w, GradientDescentEstimation(); p0=mg)
-            @test m != mg
-            @test m == mf
+            @test m != mf
```
mateuszbaran, line 750: I have no idea why I had to change this or if it was correct before.

mateuszbaran: Ah, found it.

mateuszbaran:
I've spent hours hunting bugs that show up very weirdly :disappointed: . I was hoping the few uncovered lines wouldn't cause so much trouble.

Interactions of `isapprox` with traits and groups are way too tricky. Implementing the wrong `exp` can lead to bizarre bugs that may or may not appear based on whether `OrdinaryDiffEq` was loaded or not. I don't like it.

kellertuer:
Oh I know these “games”, for me it always helped to check the order “is it first default (here add t) or first dispatch?” so what I expect vs what happened vs what we actually want ;)

kellertuer:
I think we are missing a bit of test coverage, for example the `_isapprox`e in the semi direct product group are never hit. I just fixed two that I saw are easy to fix, but some seem a bit involved to be “easily” fixed (i.e. in the 5 minutes I just had).

kellertuer:
I hopefully found and fixed the last missing line code cov.

kellertuer (approved):

mateuszbaran:
Thanks for fixing the last line!

kellertuer:
No problem :) It was easy to fix. Once this is registered we should check the 3 PRs over on Manopt again and finish those (maybe first my two but then I can do the merge of master to yours) – once we know the tests are still working find on Manopt with this version.

PR #582, opened 2023-03-20
Add horizontal and vertical component calculation to Kendall's shape space

mateuszbaran:
The horizontal and vertical component functions are defined generically. I've fixed dimension calculation in over-dimensioned cases but there might still be some other issues. Norm and inner on Kendall's shape space uses horizontal component calculation to return correct results.


kellertuer (approved):
Besides maybe a bit of documentation, this looks fine to me :)

PR #583, opened 2023-03-25
More accurate distance on sphere

mateuszbaran:
More accurate distance on sphere, and a more reasonable default random point generation on hyperbolic.

This formula for distance on sphere is a bit slower but more accurate for low distances.

The old random point generation on hyperbolic resulted in very distant points being generated on higher-dimensional hyperbolic manifolds. The new one makes them closer.


kellertuer:
Can you check also for the Manopt downstream test for the tolerance? The rest looks fine

mateuszbaran:
Sure, I will make a PR to Manopt relaxing tolerances.

src/manifolds/Sphere.jl
```diff
@@ -139,7 +139,7 @@ both `p` and `q` lie on.
 d_{𝕊}(p,q) = \arccos(\Re(⟨p,q⟩)).
 ````
 """
-distance(::AbstractSphere, p, q) = acos(clamp(real(dot(p, q)), -1, 1))
+distance(::AbstractSphere, p, q) = 2 * abs(atan(norm(p - q), norm(p + q)))
```
sethaxen, line 142: `norm(p + q) == sqrt(norm(p)^2 + norm(q)^2 + 2dot(p, q)) == sqrt(2 * (1 + dot(p, q)))`. Similarly, `norm(p - q) == sqrt(2 * (1 - dot(p, q)))`, so perhaps we could avoid the allocations and just compute this inner product once?

Also, how exactly is this implementation more accurate than the previous one?

mateuszbaran: It's a simplified version of what https://github.com/JeffreySarnoff/AngleBetweenVectors.jl offers. I'm not sure if your version is also as accurate, I think it's the `dot` that causes inaccuracy.

 I'm not entirely sure though whether Manifolds.jl should be more focused on performance or accuracy here but I wanted to put it here for discussion.

kellertuer: I could not find much on accuracy, but atan2 seems to be faster.

sethaxen: > It's a simplified version of what https://github.com/JeffreySarnoff/AngleBetweenVectors.jl offers. I'm not sure if your version is also as accurate, I think it's the `dot` that causes inaccuracy.

Thanks, indeed, https://en.wikipedia.org/wiki/Great-circle_distance has a good discussion of the numerical issues here. The version with `dot` doesn't work since for low angle, we're subtracting something close to 1 from 1, and this doesn't work well with floating point numbers. The version with `acos` doesn't work because `acos` is imprecise for low angle.

However, we can still avoid allocating here:
```suggestion
function distance(::AbstractSphere, p, q)
    # avoid roundoff for p close to q
    num = sqrt(sum(abs2 ∘ Base.splat(-), zip(p, q)))  # norm(p - q) 
    den = sqrt(sum(abs2 ∘ Base.splat(+), zip(p, q)))  # norm(p + q)
    return 2 * abs(atan(num, den))
end
```

But for a 100D sphere, this is 10x more expensive than the implementation using `dot`, presumably because we don't hit BLAS subroutines.

Ideally we would compute `c = dot(p, q)`, and if `abs(c) > some_cutoff`, then we would compute the more expensive but more accurate version. Then we do the fast thing for most cases and the slow thing only when the additional accuracy is needed.

sethaxen: Empirically, near `θ=0`, the relative error of `acos(cos(θ))` for eltype `T` seems to be bounded by `θ^(-2) * eps(T) / 4`:

```julia
julia> using Plots

julia> function relerr(θ)
           e = abs(1 - acos(cos(θ)) / θ)
           return e == 0 ? NaN : e
       end;

julia> p = -7.5:0.01:0;

julia> θ = exp10.(p);

julia> plot(θ, relerr; label="Float64");

julia> plot!(θ, θ.^(-2) .* (eps(Float64) / 4); primary=false);

julia> plot!(Float32.(θ), relerr; label="Float32");

julia> plot!(θ, θ.^(-2) .* (eps(Float32) / 4); primary=false);

julia> plot!(; xlabel="θ", ylabel="relative error", xscale=:log10, yscale=:log10)
```
tmp

Hence, if we want the relative error to be less than `eps(T)`, then we can use `acos(dot(p, q))` if `dot(p, q) > cos(0.5) ≈ 0.88`.
If we want the relative error to be less than `sqrt(eps(T))`, then we can use the fast version when `dot(p, q) > cos(eps(T)^(1//4) / 2)`, which for `Float64` is `6.1e-5`.

Close to `θ = π`, the relative error seems to be bounded by `(π - θ)^(-1) * (eps(T) / 12)`, so with a relative error less than `eps(T)`, we have `dot(p, q) > -cos(1/12) ≈ -0.996`. Relative error is always less than `sqrt(eps(T))`.

Close to `θ = π`, the relative error seems to be bounded by `(π - θ)^(-1) * (eps(T) / 12)`, while the absolute error is larger at `(π - θ)^(-1) * (eps(T) / 4)` so with an absolute error less than `eps(T)`, we have `dot(p, q) > -cos(1/4) ≈ -0.97`. Similarly, for an absolute error less than `sqrt(eps(T))`, we have `dot(p, q) > -cos(sqrt(eps(T))/4)`, which for `Float64` and `Float32` is `-1`.

I propose then that we accept a tolerance of `sqrt(eps(T))` and use the following approach:

```suggestion
function distance(::AbstractSphere, p, q)
    cosθ = dot(p, q)
    T = float(Base.promote_eltype(p, q))
    # relative error of acos is less than sqrt(eps(T))
    -1 < cosθ < cos(eps(T)^(1//4) / 2) && return acos(cosθ)
    # improved accuracy for q close to p or -p
    num = sqrt(sum(abs2 ∘ Base.splat(-), zip(p, q)))  # norm(p - q) 
    den = sqrt(sum(abs2 ∘ Base.splat(+), zip(p, q)))  # norm(p + q)
    return 2 * abs(atan(num, den))
end
```

kellertuer (approved):
Thanks for this great analysis and discussion. For me this looks very good.

sethaxen:
Also we should explicitly test `distance` for low $\theta$ and $\theta=\pi$

mateuszbaran:
I will improve the complex circle part and add an explicit test for low $\theta$ and $\theta \approx \pi$ but I will leave projective spaces for another PR.

src/manifolds/Circle.jl
```diff
@@ -431,6 +438,23 @@ function Random.rand(rng::AbstractRNG, ::Circle{ℝ}; vector_at=nothing, σ::Rea
         return map(_ -> σ * randn(rng), vector_at)
     end
 end
+function Random.rand(M::Circle{ℂ}; vector_at=nothing, σ::Real=1.0)
+    if vector_at === nothing
+        phi = rand() * 2 * π
+        return Complex(cos(phi), sin(phi))
```
sethaxen, line 444: Better to use `sign(randn(ComplexF64))`. Why does this method not accept an `rng` argument?

src/manifolds/Circle.jl
```diff
@@ -431,6 +438,23 @@ function Random.rand(rng::AbstractRNG, ::Circle{ℝ}; vector_at=nothing, σ::Rea
...
+    else
+        # written like that to properly handle `vector_at` being a number or a one-element array
+        return map(p -> project(M, p, σ * rand(typeof(p))), vector_at)
+    end
+end
+function Random.rand(rng::AbstractRNG, M::Circle{ℂ}; vector_at=nothing, σ::Real=1.0)
+    if vector_at === nothing
+        phi = rand(rng) * 2 * π
+        return Complex(cos(phi), sin(phi))
```
sethaxen, line 453: Same here using `sign`

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -423,9 +424,10 @@ function Random.rand!(
     σ=one(eltype(pX)),
 ) where {N}
     if vector_at === nothing
-        a = σ .* randn(rng, N)
-        pX[1:(end - 1)] .= a
-        pX[end] = sqrt(1 + dot(a, a))
+        a = randn(rng, N)
+        f = 1 + σ * abs(randn(rng))
+        pX[1:(end - 1)] .= a .* sqrt(f^2 - 1) / norm(a)
```
sethaxen, line 429: Should this use `firstindex`?
```suggestion
        pX[firstindex(pX):(end - 1)] .= a .* sqrt(f^2 - 1) / norm(a)
```

sethaxen (changes requested):

src/manifolds/HyperbolicHyperboloid.jl
mateuszbaran: Sure, we can use `firstindex`, though we currently don't really care much about supporting non-1-based indexing in Manifolds.jl.

src/manifolds/Circle.jl
mateuszbaran: Probably because I wasn't sure if that's the right thing to do. If you think so then I will refactor at least the methods touched by this PR.

sethaxen: Yes I think this is the right approach. It's the one Random itself takes.

mateuszbaran:
I think algorithm development is a collaborative process -- different people know about different aspects (theory, implementation, applications, etc), and together a nice result can be worked out :slightly_smiling_face: .

mateuszbaran:
I think this should be good to be merged now?

sethaxen (approved):
LGTM, thanks!

PR #584, opened 2023-03-29
More accurate distance for projective spaces

sethaxen:
Follow-up to #583 to use the same approach for projective spaces. The approach is the same, but if we need the more accurate approach, we first perturb `p` to get the element of its equivalence class that is in the same patch of the sphere as `q`, i.e. so that `isreal(dot(p, q))` and `dot(p, q) > 0`.

I've checked that the approach works for complex and quaternionic projective spaces and will add tests later.

### DIFF
--- a/src/manifolds/ProjectiveSpace.jl
+++ b/src/manifolds/ProjectiveSpace.jl
@@ -147,7 +147,16 @@ Note that this definition is similar to that of the [`AbstractSphere`](@ref).
 However, the absolute value ensures that all equivalent `p` and `q` have the same pairwise
 distance.
 """
-distance(::AbstractProjectiveSpace, p, q) = acos(min(abs(dot(p, q)), 1))
+function distance(::AbstractProjectiveSpace, p, q)
+    z = dot(p, q)
+    cosθ = abs(z)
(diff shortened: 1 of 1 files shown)


mateuszbaran:
Cool, thanks for continuing this work :slightly_smiling_face: .

Similar conda failures have happened in the past in other PRs, they seem random and generally fix themselves after some time.

kellertuer:
Compared to closing and reopening the PR I usually click on “rerun failed actions” in the Details of the failed action, but sure the result is the same – Let's see whether its back up and working :)

kellertuer:
Sure, we could; I will not have time to do that for the next, mh, maybe 2-3 weeks, but it would be nice to not need conda/python in the tests.

mateuszbaran:
I didn't know PyPlot is deprecated. Anyway, I can take a look tomorrow if something isn't working with PythonPlot.

sethaxen:
I opened a PR #585; seems to be working. PyPlot seems to work; it's just that Conda.jl itself seems to be not wanting to be set up on CI right now. PythonCall.jl is a little more sane.

mateuszbaran (approved):
LGTM :+1:

PR #585, opened 2023-03-30
Using PythonPlot instead of PyPlot in tests

sethaxen:
Plots.jl has deprecated its PyPlot backend in favor of PythonPlot. This PR switches to using PythonPlot in our tests.


mateuszbaran (approved):
Looks good. The coverage decrease is due to failed Codecov upload for the Lie group tests (in such cases I just re-run the relevant Julia 1.8-ubuntu test).

Codecov uploads keep failing relatively often in recent weeks (it feels like roughly 20%-30% of uploads fail for no reason).

Issue #572, opened 2023-02-14
change_representer for Stiefel (Default/Euclidean metric) and Grassmann (or IsometricEmbeddings in general)


kellertuer:
I just noticed that we could provide `change_representer` “for free” on Isometric Embeddings, like Stiefel (with its Default Metric / Euclidean Metric) or Grassmann (same metric, i.e. also isometric embedding), since there the change is the identity.

Currently
```
using Manifolds
M = Grassmann(5,3)
p = rand(M)
X = rand(M; vector_at=p)
change_representer(M, E p, X)
```
errors (when falling back to MetricManifold somewhere).

And I think this could even be defined for the `IsIsometricEmbedding` case as being the identity.

mateuszbaran:
Sure, that makes sense. Defining it based on `IsIsometricEmbedding` seems appropriate.

kellertuer:
Since we have the EmbeddedManifold and the Metric in ManifoldsBase, it might even be worth checking whether it can be done there – since it does not even require a MetricManifold

mateuszbaran:
Yes, sure, I think `change_representer` can be defined in ManifoldsBase.jl.

kellertuer:
We do not have the isometric embedding there, but it should be doable to define that case here, now. Will check.

kellertuer:
This as default is not as easy is I thought since we have to make sure the metric you come from is the one from the embedding. The special case for Stiefel/Grassman can be easily done, though.

kellertuer:
This was resolved in #587

PR #587, opened 2023-04-01
Performance improvements for Stiefel, Grassmann, and generalized variants

sethaxen:
This PR makes a number of performance improvements for `Stiefel`, `Grassmann`, `GeneralizedStiefel`, and `GeneralizedGrassmann`, which do not change functionality. It also implements point `project`ion for `Grassmann`.


kellertuer:
I noticed on master Documenter is failing (its also still using PyPlot) – and since we are at it, besides project also change_representer would be cool (its the identity, since Grassmann inherits the metric from the embedding), since that would allow automatic gradient conversion.

mateuszbaran (approved):
Thanks, this looks fine except for minor issues in the docstring :+1: .

kellertuer:
Sorry to highjack this just a bit, but since we have `project` now, I also did `change_representer` real quick, since that allows for Euclidean gradients to be converted to Riemannian ones (resolves #572).

src/manifolds/Grassmann.jl
```diff
@@ -88,6 +88,32 @@ function allocation_promotion_function(::Grassmann{n,k,ℂ}, f, args::Tuple) whe
...
+@doc raw"""
+    change_representer(M::Grassmann, ::EuclideanMetric, p, X)
+
+Change `X` to the corresponding representer of a cotangent vector at `p`.
+Since the [`Grassmann`](@ref) manifold `M`, is isometrically embedded, this is the identity
+"""
+change_representer(::Grassmann, ::EuclideanMetric, ::Any, ::Any)
+
+function change_representer!(M::Grassmann, Y, ::EuclideanMetric, p, X)
+    copyto!(M, Y, p, X)
```
sethaxen, line 100: I haven't seen this signature of `copyto!` before. Where is it defined?

kellertuer: The first one at https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions.html#Base.copyto!-Tuple{AbstractManifold,%20Any,%20Any,%20Any} (for vectors).
I just got very used to this, because it also takes care about types and structs; for Grassmann this should still work fine for the `Projector` representation for example.
I started using it when I worked with FixedRank matrices where each point/tangent is stored as 3 arrays and copying a tangent vectors requires the point as well (if I remember correctly).
For manifolds that work on pure arrays this defaults to just `copyto!(Y,X)`.

mateuszbaran: IMO code restricted to a particular manifold should just use the two-arg version of `copyto!` because it's easier for the compiler. These manifold specialisations of `copyto!` are only relevant to certain variants of power manifolds.

sethaxen: I'm fine with this version for now. In the future we could adopt a consistent rule and make sure it is followed throughout the codebase.

kellertuer: Sure it might be a good idea to work a bit on a consistent usage here

PR #589, opened 2023-04-03
Replace PyPlot with PythonPlot

mateuszbaran:
This should fix the recent issues with PyPlot.


kellertuer (approved):
LGTM (± codecov reports not being processed it seems)

mateuszbaran:
Yes, codecov is still failing for no reason. I will merge this and keep re-running tests on master until it works I guess.

PR #592, opened 2023-04-12
Fix defaults on Grassmann

kellertuer:
Main reason – it defaulted to exp/parallel transport, which do not exist in Stiefel representation.


mateuszbaran:
Good idea. Manopt tests are failing because it defines `copy(::NONMUTATINGMANIFOLDS, p) = p` somehow -- it needs to be removed.

kellertuer:
That is also done in the same branch linked above.

PR #593, opened 2023-04-27
Fix check_point for SPD

mateuszbaran:
A small bugfix. `eigvals` returns complex eigenvalues in the provided test case, which are not comparable to 0. This, also, should be faster.

### DIFF
--- a/src/manifolds/SymmetricPositiveDefiniteGeneralizedBuresWasserstein.jl
+++ b/src/manifolds/SymmetricPositiveDefiniteGeneralizedBuresWasserstein.jl
@@ -94,6 +94,8 @@ function exp!(
     m = M.metric.M
     Y = lyapc(p, m, -X) #lyap solves qpM + Mpq - X =0
     q .= p .+ X .+ m * Y * p * Y * m
+    # symmetrizing for better accuracy
+    copyto!(q, (q .+ q') ./ 2)
     return q
 end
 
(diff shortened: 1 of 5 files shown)


kellertuer:
LGTM,
could you also remove those two lines since we are at fixing small things :)

https://github.com/JuliaManifolds/Manifolds.jl/blob/8ccd5d819e72b4d80cb23a55da9d42054370d25f/src/groups/rotation_action.jl#L304-L305

mateuszbaran:
Sure, I will remove them.

kellertuer:
The CI seems to indicate that you probably have to adapt tolerances?

mateuszbaran:
I increased accuracy instead, I hope that's also fine :slightly_smiling_face:

src/manifolds/SymmetricPositiveDefiniteGeneralizedBuresWasserstein.jl
```diff
@@ -94,6 +94,8 @@ function exp!(
     m = M.metric.M
     Y = lyapc(p, m, -X) #lyap solves qpM + Mpq - X =0
     q .= p .+ X .+ m * Y * p * Y * m
+    # symmetrizing for better accuracy
+    copyto!(q, (q .+ q') ./ 2)
```
dkarrasch, line 98: How is this increasing accuracy? How do you know the error in `q` is "symmetric", and reduced by symmetrization? Also, I wonder if this is optimal with respect to memory consumption? Perhaps something like

```julia
temp = m * Y * p * Y * m
temp .= temp .+ p .+ X
copyto!(q, temp)
# or
# copyto!(q, Symmetric(temp))
# or 
# copyto!(q, (temp .+ temp') ./ 2) # which first allocates the result and then copies it to `q`
# or, starting from Julia v1.10
# hermitianpart!(q, temp)
```
?

mateuszbaran: Well, it's more accurate than the previous implementation in the sense that the result is numerically closer to the manifold. I should've made a better comment there.

Maybe taking either the upper or lower part instead of average would be a better idea, I didn't check that. I didn't care much about memory consumption here because previous lines are already quite expensive.

PR #594, opened 2023-04-28
Use GeodesicInterpolationWithinRadius for Circle mean

sethaxen:
- Replace default extrinsic mean with interpolation within radius
- Update docstring
- Increment patch number

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.8.57"
+version = "0.8.58"
 
 [deps]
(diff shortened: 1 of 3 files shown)


sethaxen:
`mean` doesn't like accepting a vector of numbers and instead wants a vector of vectors (see e.g. https://github.com/JuliaManifolds/Manifolds.jl/actions/runs/4829265624/jobs/8604094066?pr=594#step:5:771). Do we have a standard workaround for this for the `Circle`?

mateuszbaran:
Manopt.jl is currently in the process of adding such workaround: https://github.com/JuliaManifolds/Manopt.jl/pull/236 . Essentially it's wrapping points in one-element vectors. It's not ideal for performance, especially in the case of `mean`, but it's the easiest solution.

kellertuer:
HI @sethaxen – do you still plan to continue this? (No pressure, just trying to keep an eye on what PRs we still have open here)

sethaxen:
As if right now, yes, but it's unclear when I will finish it. I have a private lightweight repo efficiently implementing various intrinsic circular statistics that I will eventually finish, and the plan is to then add that as a dependency here to deliver this functionality.

If the plan changes, I'll post here.

kellertuer:
That sounds great! Looking forward to hearing from that :)

PR #598, opened 2023-05-05
Bump allowed ambiguities

mateuszbaran:
It looks like https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/155 caused some new harmless ambiguities in the `convert` function. This PR bumps how many are allowed.


mateuszbaran:
It would be good to merge this before the new version of Manifolds.jl is tagged.

kellertuer (approved):
This looks like quite a bump, I am sure you checked all of them, but should we be worried that this rises so far?

mateuszbaran:
These new ambiguities look fine, they should never be triggered.

PR #603, opened 2023-05-10
Extend and fix `rand`.

### LINKED ISSUE #601: `rand(Sphere(n, ℂ))` only returns real vectors
This should return also complex vectors, I believe:
```
julia> M = Sphere(3, ℂ); rand(M)
4-element Vector{ComplexF64}:
  0.22029993567413428 + 0.0im
   0.8898546674007888 + 0.0im
  -0.3693475326483702 + 0.0im
 -0.15234503396423887 + 0.0im
```

kellertuer:
This PR

* fixes complex sphere random points (resolves #601)
* fixes complex Euclidean random points
* introduces (real) Generalised Grassmann points/vectors
* introduces (real) Generalised Stiefel points/vectors

It seems on `M=Sphere(2, ℍ)` the result is still real and not Quaternion for `rand(M)`, but I could not yet find the culprit in the allocation labyrinth when allocating the solution `p` (before calling `rand!(M, p)`). The following actually works fine

```Julia
b =  [ Quaternions.quat(1.0), Quaternions.quat(1.0), Quaternions.quat(1.0)];
Manifolds.rand!(M,b)
is_point(M,b)
```
yields true and `b` has nonzero entries in all 4 quaternion entries in each quaternion. SO this is really just an allocation problem for quaternion problems.

### DIFF
--- a/src/manifolds/Euclidean.jl
+++ b/src/manifolds/Euclidean.jl
@@ -572,7 +572,7 @@ project(::Euclidean{Tuple{}}, ::Number, X::Number) = X
 project!(::Euclidean, Y, p, X) = copyto!(Y, X)
 
 function Random.rand!(::Euclidean, pX; σ=one(eltype(pX)), vector_at=nothing)
-    pX .= randn(size(pX)) .* σ
+    pX .= randn(eltype(pX), size(pX)) .* σ
     return pX
 end
 function Random.rand!(
@@ -582,7 +582,7 @@ function Random.rand!(
     σ=one(eltype(pX)),
     vector_at=nothing,
 )
-    pX .= randn(rng, size(pX)) .* σ
+    pX .= randn(rng, eltype(pX), size(pX)) .* σ
     return pX
 end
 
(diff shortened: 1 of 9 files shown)


mateuszbaran:
I've fixed the issue with promotion on quaternionic manifolds.

kellertuer:
It seems GitHub has a bit of trouble running CIs recently.

mateuszbaran:
Yes, a little. BTW, one reason it fails is that Julia 1.6 doesn't have `ColumnNorm`.

kellertuer:
Oh, I was just hoping to fix a warning. What do we do there then? On 1.8 and 1.9 we have a deprecation warning.

mateuszbaran:
Probably we just ignore the warning? Alternatively we can make a wrapper that uses different arguments on different Julia versions.

kellertuer:
Yeah we have a few also for the mean I think, as a German I always feel the urge to solve them, but you might be right, we should maybe keep them until we discontinue support for 1.6.

mateuszbaran (approved):

mateuszbaran:
LGTM. One last missing thing is bumping version.

PR #604, opened 2023-05-10
add `rand` for `ProbabilitySimplex`

adienes:
uniformly samples from probability simplex if no point given. otherwise returns a mean-zero vector of gaussians.


adienes:
sorry should have squashed all those first. had to turn off the random tangent test since `isapprox` has `atol` set to `0,` and thus `isapprox(_, 0)` is testing for exact equality.

I think pr is ok now? let me know

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,35 @@ mean(::ProbabilitySimplex, ::Any...)
 
 default_estimation_method(::ProbabilitySimplex, ::typeof(mean)) = GeodesicInterpolation()
```
kellertuer, line 283: We usually document our functions on the non-mutating case, i.e. here `rand()`, see https://github.com/JuliaManifolds/Manifolds.jl/blob/369eef7418089baaf59a6215ed2fd94b23174d8a/src/manifolds/Stiefel.jl#L310-L322 – could you add that here as well? If possible, a formula / name / reference would be great as well

test/manifolds/probability_simplex.jl
```diff
@@ -57,6 +57,8 @@ include("../utils.jl")
                 inverse_retraction_methods=[SoftmaxInverseRetraction()],
                 retraction_methods=[SoftmaxRetraction()],
                 test_inplace=true,
+                test_rand_point=true,
+                #test_rand_tvector=true, # implemented, but turned off until atol can be set locally
```
kellertuer, line 61: Could you elaborate? Where is that missing? In the probability simplex or in the test suite?

You could also write a test outside this test suite: set the seed, call both rand-variants (points / tangent vectors) and test that you got a point / a vector, ,resp.

kellertuer (changes requested):
Thanks for your PR – contributions are always happily welcome :)

I have two small remarks – and never mind several commits – for a PR that is always good to see also a bit the development, we squash-merge PRs anyways.

mateuszbaran:
> sorry should have squashed all those first. had to turn off the random tangent test since `isapprox` has `atol` set to `0,` and thus `isapprox(_, 0)` is testing for exact equality.

The easiest way to handle this is setting `rand_tvector_atol_multiplier=1` keyword argument to `test_manifold`, or a bigger number if needed. I think we should start setting non-zero absolute tolerance for tangent vector tests by default but that's a separate issue :slightly_smiling_face: .

kellertuer:
>  I think we should start setting non-zero absolute tolerance for tangent vector tests by default but that's a separate issue 🙂 .

Good point to keep in mind for #548 .

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,49 @@ mean(::ProbabilitySimplex, ::Any...)
...
+``T_{vector\_at}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ` 
+to have a zero component sum.
+"""
+rand(::ProbabilitySimplex; σ::Real=1.0)
+
+function Random.rand!(M::ProbabilitySimplex, pX; vector_at=nothing, σ=one(eltype(pX)))
+    n = only(representation_size(M))
+    if isnothing(vector_at)
+        _r = sort(vcat(rand(n - 1), [0.0, 1.0]))
+        pX .= (circshift(_r, -1) .- _r)[1:n]
```
sethaxen, line 302: We could make fewer allocations and avoid the `O(n*log(n))` sorting by using the fact that `Dirichlet(alpha=ones(n))` is the uniform distribution on the probability simplex, so the usual Gamma distribution (see https://en.wikipedia.org/wiki/Dirichlet_distribution#Random_variate_generation) approach for drawing from Dirichlet reduces to drawing from `Exponential(1)`:
```suggestion
        Random.randexp!(pX)
        LinearAlgebra.normalize!(pX, 1)
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,49 @@ mean(::ProbabilitySimplex, ::Any...)
...
+    else
+        _r = randn(n) .* σ
+        pX .= _r .- mean(_r)
```
sethaxen, line 305: We don't need the allocation to `r_`.
```suggestion
       Random.randn!(pX)
       pX .= (pX .- mean(pX)) .* σ
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,49 @@ mean(::ProbabilitySimplex, ::Any...)
...
+    end
+    return pX
+end
+function Random.rand!(
```
sethaxen, line 309: I know elsewhere in this package we write separate methods for with and without `rng`, but this was a mistake. In general, we should implement sampling methods with `rng` and then the one without `rng` should just dispatch to this by passing `Random.default_rng()` as the first argument. This is what Random itself does.

sethaxen (changes requested):

src/manifolds/ProbabilitySimplex.jl
kellertuer: Nice! That should also be reflected in the docs.

src/manifolds/ProbabilitySimplex.jl
kellertuer: I was not aware f that we should then slowly improve our code :)

mateuszbaran: I will make a PR that improves this.

kellertuer:
With your current tolerance factor of 1 you are still checking for machine precision zero, which might be a bit tough (that is why 1.6 failed, maybe raise the factor to `10.0`?

test/manifolds/probability_simplex.jl
```diff
@@ -57,6 +57,9 @@ include("../utils.jl")
                 inverse_retraction_methods=[SoftmaxInverseRetraction()],
                 retraction_methods=[SoftmaxRetraction()],
                 test_inplace=true,
+                test_rand_point=true,
+                test_rand_tvector=true,
+                rand_tvector_atol_multiplier=1,
```
kellertuer, line 62:
```suggestion
                rand_tvector_atol_multiplier=10.0,
```
I think I can even trick my remark with a comment and merge it myself, let's try whether this already fixes the tests.

kellertuer, line 62:
```suggestion
                rand_tvector_atol_multiplier=10.0,
```
I think I can even trick my remark with a comment and merge it myself, let's try whether this already fixes the tests.

sethaxen:
Okay, got it. The standard reference is Theorem 2.2, Chapter 5 of

Luc Devroye. Non-Uniform Random Variate Generation. Springer-Verlag, New York, 1986. doi: 10.1007/978-1-4613-8643-8

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,39 @@ mean(::ProbabilitySimplex, ::Any...)
...
+@doc raw"""
+    rand(::ProbabilitySimplex; vector_at=nothing, σ::Real=1.0)
+
+
+When `vector_at` is `nothing`, return a random (uniform) point `x` on the [`ProbabilitySimplex`](@ref)
+manifold `M` by normalizing a sample from an exponential distribution to unit sum.
+
+When `vector_at` is not `nothing`, return a (Gaussian) random vector from the tangent space
+``T_{p}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ` 
+to have a zero component sum.
```
kellertuer, line 293:
```suggestion
to have a zero component sum, see [^ Devroye1986], Theorems 2.1 and 2.2 p. 207, 208.

[^Devroye1986]:
    > Devroye, L.:
    > _Non-Uniform Random Variate Generation_.
    > Springer New York, NY, 1986.
    > doi: [10.1007/978-1-4613-8643-8](https://doi.org/10.1007/978-1-4613-8643-8)
"""
```

adienes:
thank you for adding the reference. we should probably merge one of this PR or #605 first, then rebase the other? I think as-is they are slightly incompatible

kellertuer:
Oh I might have crashed formatting, sorry for that (editing here has its disadvantages). But I‘ll activate Docs rendering as well now :)

mateuszbaran:
> thank you for adding the reference. we should probably merge one of this PR or #605 first, then rebase the other? I think as-is they are slightly incompatible

I think we can merge this one first, and then I will fix incompatibility in #605 .

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,45 @@ mean(::ProbabilitySimplex, ::Any...)
...
+@doc raw"""
+    rand(::ProbabilitySimplex; vector_at=nothing, σ::Real=1.0)
+
+
+When `vector_at` is `nothing`, return a random (uniform) point `x` on the [`ProbabilitySimplex`](@ref)
+manifold `M` by normalizing a sample from an exponential distribution to unit sum.
+
+When `vector_at` is not `nothing`, return a (Gaussian) random vector from the tangent space
+``T_{p}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ` 
+to have a zero component sum, see [^ Devroye1986], Theorems 2.1 and 2.2 p. 207, 208.
```
kellertuer, line 293:
```suggestion
to have a zero component sum, see [^Devroye1986], Theorems 2.1 and 2.2 on p. 207 and 208, respectively.
```

we can not push the docs PR (since the PR is not from this repo) but I saw the error in my type which is fixed here.

kellertuer (dismissed):
My points are clarified. Thanks for the contribution and the good further ideas.

mateuszbaran (approved):

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,45 @@ mean(::ProbabilitySimplex, ::Any...)
 
 default_estimation_method(::ProbabilitySimplex, ::typeof(mean)) = GeodesicInterpolation()
 
+@doc raw"""
+    rand(::ProbabilitySimplex; vector_at=nothing, σ::Real=1.0)
+
+
+When `vector_at` is `nothing`, return a random (uniform) point `x` on the [`ProbabilitySimplex`](@ref)
+manifold `M` by normalizing a sample from an exponential distribution to unit sum.
```
sethaxen, line 289: The reference refers to sampling from the simplex, not its tangent space:
```suggestion
manifold `M` by normalizing independent exponential draws to unit sum, see [^Devroye1986], Theorems 2.1 and 2.2 on p. 207 and 208, respectively.
```

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,45 @@ mean(::ProbabilitySimplex, ::Any...)
...
+
+When `vector_at` is not `nothing`, return a (Gaussian) random vector from the tangent space
+``T_{p}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ` 
+to have a zero component sum, see [^Devroye1986], Theorems 2.1 and 2.2 on p. 207 and 208, respectively.
```
sethaxen, line 293:
```suggestion
to have a zero component sum.
```

sethaxen (changes requested):
Just a minor docstring fix. Also, this should get a patch version bump. Otherwise, LGTM.

mateuszbaran:
I think I will merge this and bump version in #605 -- thank you for your contribution :slightly_smiling_face:

kellertuer:
But we already discussed, that this is w.r.t. the „wrong” metric, i.e. not our default one but the Euclidean (with respect to which the manifold is not complete).

mateuszbaran:
Yes, it's not uniform wrt. our metric but we never claimed to always use uniform distributions (for manifolds of finite volume). And we can change it later.

kellertuer:
Sure, I would still prefer if we could switch this one to be the one with the metric manifold / euclidean metric?

mateuszbaran:
OK, I'll wait then.

kellertuer:
What for? If you feel confident to make a breaking change when we implement the “right” Gaussian Simplex rand, we can merge this.
I think it would be best to just switch it to said metric manifold, then I would feel confident merging this (in the PT issue Seth mentioned how the Fisher-Rao rand should look like, we can also add this here – but my brain is already grilled for today ;) )

mateuszbaran:
> What for? If you feel confident to make a breaking change when we implement the “right” Gaussian Simplex rand, we can merge this.

I was going to wait until this PR has the right distribution? I was just wondering if maybe I should finish #605 first then. Also I do not consider changing the default distribution of `rand!` to be a breaking change.

> I think it would be best to just switch it to said metric manifold, then I would feel confident merging this (in the PT issue Seth mentioned how the Fisher-Rao rand should look like, we can also add this here – but my brain is already grilled for today ;) )

That would also be fine.

sethaxen:
I still think this is the right measure to use. When I pointed out the connections of the metrics to the measures, what I meant was that this measure is uniform wrt the Hausdorff (volume/area) measure on the probability simplex. The Hausdorff measure relies on the metric from an embedding. The other uniform measure comes from taking the Hausdorff measure on the hypersphere and pushing it forward through the elementwise square operation. This is related to the Fisher-Rao metric in the sense that Fisher-Rao on the simplex is equivalent to the Euclidean metric on the hypersphere. There are probably other ways to formulate this measure that doesn't invoke the hypersphere.

My geometric measure theory isn't amazing, but there isn't a direct connection between a measure and a metric like there is between a metric and a geodesic. So having a default metric doesn't necessitate choosing a default measure that is somehow related to that metric. In my opinion, for compact manifolds, the most sensible default metric should be the one that is most widely used as a base measure for density functions. In most cases this is a Haar measure for groups and a Hausdorff measure for manifolds. For the simplex, the Hausdorff measure is the most common one, so I'd recommend using that one.

adienes:
if I am reading it correctly, the only ask for `rand` from `ManifoldsBase.jl` is
>Usually a uniform distribution should be expected for compact manifolds and a Gaussian-like distribution for non-compact manifolds and tangent vectors, although it is not guaranteed. The distribution may change between releases.

As @kellertuer noted in the other thread, since `p_i > 0` is a strict ineq. then probability simplex is not compact, so maybe the uniform distribution is not the right choice either way? Although I'm not sure what a "Gaussian-like" distribution on this space looks like besides just making `α` larger when sampling `Dirichlet(α .* ones(n))`

mateuszbaran:
> My geometric measure theory isn't amazing, but there isn't a direct connection between a measure and a metric like there is between a metric and a geodesic. So having a default metric doesn't necessitate choosing a default measure that is somehow related to that metric.

There definitely is a Riemannian concept of a measure: https://joekmiller.files.wordpress.com/2019/06/geometric-measure-theory-and-volumes-of-manifolds.pdf . I would expect that Hausdorff measure and Riemannian/Lebesgue measure would be the same as the "pushforward from sphere" measure? Do you have a source about measures on the probability simplex that suggest otherwise?

mateuszbaran:
> As @kellertuer noted in the other thread, since `p_i > 0` is a strict ineq. then probability simplex is not compact, so maybe the uniform distribution is not the right choice either way? Although I'm not sure what a "Gaussian-like" distribution on this space looks like besides just making `α` larger when sampling `Dirichlet(α .* ones(n))`

I think one of our problems with simplex is that we are not totally sure if we should consider it without the boundary or add the boundary (which can be done in a consistent way). Then such a simplex with boundary would be a compact. Also, that comment in ManifoldsBase about rand should be considered more like a general suggestion than something that must be always followed.

kellertuer:
I personally prefer the simplex without boundary, but meant exactly like that – I personally ;)

adienes:
it is too bad that such decisions must always come with tradeoffs. I say that, because I think for my selfish purposes, it would be preferable to include the boundary 😂.

for context, my original intent was to solve a version of the optimal transport problem, in which case one might want to add regularization terms for the number of edges taken, i.e. zero-norm of point on simplex, so coordinates would ideally be able to be set to 0

mateuszbaran:
No tradeoffs needed, we can always have both variants :smile: . I would just have to carefully consider what to do on the boundary.

And I would definitely like to support solving optimal transport problems with Manifolds.jl.

kellertuer:
We can do both - either with two manifold types or with a parameter in this one type, but that is one thing we need since even `is_point` is different for both.

But sure having both would also be cool. I think the current functions are more for the open one (where the metric induces that the boundary is infinitely far away).

mateuszbaran:
> (where the metric induces that the boundary is infinitely far away).

Why do you think so? It is at a finite distance. Take a look at Seth's picture, or `injectivity_radius`.

kellertuer:
> > (where the metric induces that the boundary is infinitely far away).
> 
> 
> 
> Why do you think so? It is at a finite distance. Take a look at Seth's picture, or `injectivity_radius`.

I am quite sure I implemented it that way, and that in Seth's picture it is just arbitrarily close.

kellertuer:
But even independent of that, with a small parameter, we can have both and that would be great.

mateuszbaran:
> I am quite sure I implemented it that way, and that in Seth's picture it is just arbitrarily close.

A simple argument to the contrary would be: Fisher-Rao geodesics are pushforwards of geodesics on the sphere. What you see in Seth's picture is actually stitching pushforwards of geodesic arcs from different orthants of the sphere.

kellertuer:
Now I am not so sure anymore.

kellertuer:
As it is so often with manifolds: It depends on the metric (and yesterday on my hike I did not check which one we implement). So I will contribute my favourite one soon, see
https://arxiv.org/pdf/1604.08714.pdf
section 2.3.

With respect to _that_ (with the division by x^2 one can already see the similarity to SPDs in the metric) we do have that the boundary is infinite far away.

mateuszbaran:
> As it is so often with manifolds: It depends on the metric (and yesterday on my hike I did not check which one we implement). So I will contribute my favourite one soon, see
> https://arxiv.org/pdf/1604.08714.pdf
> section 2.3.

That would explain our disagreement then :smile: . Section 2.3 of that manuscript seems to define our `PositiveArrays` manifold though? Do you mean simplex as a submanifold of `PositiveArrays` with arrays that sum to 1?

kellertuer:
Well, you are a bit more correct since you referred to the current implementation.

Yes it is very close to (element wise) positive numbers - with a constraint that they sun to one.

kellertuer:
Let‘s maybe shorten this – we can not solve all things we discussed with the simplex in this PR, so let‘s summarise this in an issue, maybe the one about PT, maybe a new one.

For this one my main point is: Should we

a) keep it as is and „risk“ a breaking change when we move this variant to another metric? Or
b) Move this to the Euclidean metric and add the method Seth already mentioned to the default?

Either should be doable in not so much effort so we can finish this (and sorry @adienes that this PR got longer than I hoped for).

kellertuer:
I would also prefer b).

adienes:
I attempted to apply the suggestions to add both versions of `rand`, as well as rebasing on top of #605 

I wasn't sure what the best way to test `rand(MetricManifold(<:ProbabilitySimplex, <:EuclideanMetric))` was

Also, this is more general Julia thing than `Manifolds.jl` but I'm not sure how exactly which is the "best" way to form the dispatch among the following
```
> rand(MetricManifold(::ProbabilitySimplex, ...)
> rand(MetricManifold(<:ProbabilitySimplex, ...)
> rand(MetricManifold(<:ProbabilitySimplex{n}, ...) where {n}
```

kellertuer:
For the test, just generate `M = MetricManifold(ProbabilitySimplex(3), EuclideanMetric())`.
By default we assume that any manifold (implicitly) has a metric when implemented first – for the probablility simplex that is the Fisher Rao metric. The above one replaces that default by the euclidean matrix. That is what you want.

For your code questions

> >    rand(MetricManifold(::ProbabilitySimplex, ...)
> >   rand(MetricManifold(<:ProbabilitySimplex, ...)
>>    rand(MetricManifold(<:ProbabilitySimplex{n}, ...) where {n}

The first says: When any `ProbabilitySimplex` comes along, the second matches on every subtype (more specific) the matches similarly, just that the parameter is not `Any` (but unkown by name) but gets the name `n` (but since that is not restricted, it is still of type `Any` in general, just now it has a name.

I would propose to do PT and also the change to distinguishing open and closed simpllices in a new PR, to (a) keep this here of a finite length (time and space wise). Also small steps published and tagged is nicer than larger steps (that also might not finish if we get stuck?) – take it step by step (PR by PR).

Thanks for your contribution, looking forward to more then.

I will review your code somewhen later today I hope.

mateuszbaran:
> Also, this is more general Julia thing than `Manifolds.jl` but I'm not sure exactly which is the "best" way to form the dispatch among the following. It does not seem to be entirely consistent among existing manifold implementations.
> 
> ```
> > rand(MetricManifold(::ProbabilitySimplex, ...)
> > rand(MetricManifold(<:ProbabilitySimplex, ...)
> > rand(MetricManifold(<:ProbabilitySimplex{n}, ...) where {n}
> ```

The basic pattern is something like
```julia
function exp!(
    ::MetricManifold{ℝ,SymmetricPositiveDefinite{N},LogCholeskyMetric},
    q,
    p,
    X,
) where {N}
```
We could have `MetricManifold{ℝ,<:SymmetricPositiveDefinite{N},<:LogCholeskyMetric}` there -- adding `<:` is fine, though as far as I understand not having `<:` when it's not necessary makes compiler's work easier, so we drop it sometimes. `<:` is not necessary when all type parameters are explicitly provided (here: for `LogCholeskyMetric` and `SymmetricPositiveDefinite`).

> It sounds like there are more tasks down the line, like adding the parallel transports, possibly adding boundary, etc.; this is an interesting domain to me (and, in self-interest, I'm hoping to use the package for a few projects), so I'm happy to contribute further if the tasks can be broken down / described a little bit :)

I think adding parallel transports would be a good project for start. Adding boundary not so much but I will try to find some time this week to work on it. I will also write in the parallel transport issue how to derive and implement it :slightly_smiling_face: .

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -281,6 +281,44 @@ mean(::ProbabilitySimplex, ::Any...)
...
+rand(::ProbabilitySimplex; σ::Real=1.0)
+
+function Random.rand!(
+    rng::AbstractRNG,
+    M::ProbabilitySimplex,
+    pX;
+    vector_at=nothing,
+    σ=one(eltype(pX)),
+)
+    if isnothing(vector_at)
```
kellertuer, line 311: You transferred this to the rand with the Euclidean Metric which looks nice, but then this should be changed to the rand that Seth described at https://github.com/JuliaManifolds/Manifolds.jl/issues/606#issuecomment-1546869201 - both for point and vector?

kellertuer: Otherwise we would double code, which is not nice – and the whole idea of moving this rand over is to have a second (i.e. different) one here :)

adienes: I think they are different `rand` for point right? one samples uniformly from l1 ball and the other from l2 ball. happy to change the vector as well of course, but not sure I fully understand how the tangent space geometry changes with the two different metrics

mateuszbaran: I think it would be easiest to do `change_metric!(M, pX, EuclideanMetric(), vector_at, pX)` to convert to Fisher-Rao metric.

mateuszbaran:
It looks like the PR stalled a bit. @adienes , if you are OK with it I could just finish these changes myself. I understand that these discussions around distributions may be somewhat confusing.

kellertuer:
I want to add: Thanks for your contribution @adienes ! Sorry that we went a bit off to discuss how to best integrate your great initiative here, please do not feel discouraged by those discussions.
As Mateusz said, we can surely finish these changes required here still ourselves, if you feel confused. Let us know if you need something more than this PR (sure PT is next on the list but we have an issue for that already :) )

adienes:
sorry, busy week at work, didn't mean to stall 😅. let me take one more crack at it; I will make time this week. if you don't hear from me by Saturday feel free to release the hounds

kellertuer:
No worries, we are not in a rush, just careful, if you feel it is too much work or too complicated – sure, take your time :)

adienes:
> @mateuszbaran  change_metric!(M, pX, EuclideanMetric(), vector_at, pX)

I don't entirely follow how this must give back a tangent vector---it won't always be true that `pX` and `sqrt.(vector_at) .* pX` will both be mean-zero, right?

kellertuer:
Change metric should just change X in the following sense that

$\langle X,Y \rangle = \langle c(X), c(Y) \rangle_p$

where the inner product on the left is the Euclidean one (see the third argument, that is the metric we come from), the one on the right is the one on the Simplex (which depends on the point p) and $c$ denotes the change of metric function. But you are right $c$ should map from the tangent space at p into the tangent space at p, if it does not do that, we have to fix said function. We should not need to center afterwards.

mateuszbaran:
I meant changing the metric after centering, like
```julia
pX .= (pX .- mean(pX)) .* σ
change_metric!(M, pX, EuclideanMetric(), vector_at, pX)
```
or just 
```julia
pX .= (pX .- mean(pX)) .* σ .* sqrt.(vector_at)
```
for simplicity.

adienes:
It might not be centered anymore then though right? I thought the tangent vectors had to have 0 l1 norm

mateuszbaran:
Tangent vectors can have any norm, as long as their sum is 0 (i.e. they are centered).

kellertuer:
l1-norm is not completely correct, that would be the sum of the absolute values of the entries. If that is zero you have the zero vector (hence l1-norm is called norm).

The mean – the sum of the entries is 0, but that also holds for both versions, I think.

kellertuer:
No problem ;) I have those sometimes as well.

Concerning the question, I have to think about that a bit, I do not think so, since we are really just adapting the metric on the space i.e. change the elements within the space so the have the same norm and angles as before w.r.t the other metric (hence the name change metric)

mateuszbaran:
Wait... that's actually a good question. I assumed `change_representer` and `change_metric` are fine but when I look at them now they indeed seem to produce vectors that are not tangent. @kellertuer do you remember where you got the formulas from?

kellertuer:
We usually merge, that is easier (for PRs we squash-merge, which makes the master branch relatively clean again).

I will check the change of metric soon.

kellertuer:
Hm, I am now a bit confused, since the documentation of exp for example follows https://arxiv.org/pdf/1603.05285.pdf, Proposition 2.4, but the implementation – at least as far as I understand it – does something completely different.

The metric conversion seems wrong for now, I might have stayed too close to what is done on the positive numbers (element wise), will check for a correction. The change of representer seems right, though.

kellertuer:
HM, that's a tough one to solve, because it seems there are only n degrees of freedom (for the change metric) but I currently seem to have n+1 conditions.

mateuszbaran:
I've discussed with @kellertuer what is wrong with `change_representer` and `change_metric`. I will make a PR fixing them and then you will be able to continue :slightly_smiling_face: . Thank you for your question about validity of `change_metric`, I wouldn't notice that something is wrong without it.

kellertuer:
Yeah, thanks for the discussion it seems there might be a change metric :)

mateuszbaran:
`change_metric` is now fixed on master branch, so you can continue now :slightly_smiling_face:

adienes:
I had tried also to add some tests for the `ProbabilitySimplexEuclideanMetric`, but it was very hard to get any to pass since basically only `rand` is defined. it seems to want `retract` ...

mateuszbaran:
Hm, I think you can just take `exp` from `Euclidean`:
```julia
exp!(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}, q, p, X) = (q .= p .+ X)
exp!(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}, q, p, X, t::Number) = (q .= p .+ t .* X)
```
and it should more or less work.

adienes:
g2g?

test/manifolds/probability_simplex.jl
```diff
@@ -60,6 +62,24 @@ include("../utils.jl")
                 retraction_methods=[SoftmaxRetraction()],
                 test_inplace=true,
                 vector_transport_methods=[ParallelTransport()],
+                test_rand_point=true,
+                test_rand_tvector=true,
+                rand_tvector_atol_multiplier=10.0,
```
mateuszbaran, line 67:
```suggestion
                rand_tvector_atol_multiplier=20.0,
```
Almost G2G, this needs a slightly larger tolerance :slightly_smiling_face:

src/manifolds/ProbabilitySimplexEuclideanMetric.jl
```diff
@@ -0,0 +1,46 @@
...
+    q,
+    p,
+    X,
+    t::Number,
+)
+    return (q .= p .+ t .* X)
+end
+
+@doc raw"""
+    rand(::ProbabilitySimplex; vector_at=nothing, σ::Real=1.0)
```
mateuszbaran, line 13:
```suggestion
    rand(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}; vector_at=nothing, σ::Real=1.0)
```

mateuszbaran (approved):

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -354,6 +354,45 @@ function parallel_transport_to!(M::ProbabilitySimplex{N}, Y, p, X, q) where {N}
     return amplitude_to_simplex_diff!(M, Y, q_s, Ys)
 end
 
+@doc raw"""
+    rand(::ProbabilitySimplex; vector_at=nothing, σ::Real=1.0)
+
+
+When `vector_at` is `nothing`, return a random (uniform over the Fisher-Rao metric)
```
sethaxen, line 361: I'm not certain what the right terminology here is, but I don't think this is it. Maybe something like
```suggestion
When `vector_at` is `nothing`, return a random (uniform over the Fisher-Rao metric; that is, uniform with respect to the `n`-sphere whose positive orthant is mapped to the simplex).
```
If the bijection between the positive orthant and the simplex is somewhere documented already, we could link to that docstring.

But I'll do a literature search to see if there's a more specific name we can give here.

mateuszbaran: I think that can be done in a follow-up PR? This one is already taking way too much time.

kellertuer:
But then we should also register 0.8.71 soon, because I think we were just waiting for this PR.

mateuszbaran:
Yes, I've even forgotten we have a few changes on master without a new registered version.

PR #605, opened 2023-05-11
Simplify random point/tangent generation by using default_rng

mateuszbaran:
See https://github.com/JuliaManifolds/Manifolds.jl/pull/604#discussion_r1191576815 .

There are some one-liners that I haven't converted to this pattern, I'm not sure if I should?

### DIFF
--- a/src/manifolds/FixedRankMatrices.jl
+++ b/src/manifolds/FixedRankMatrices.jl
@@ -444,26 +444,8 @@ and the singular values are sampled uniformly at random.
 If `vector_at` is not `nothing`, generate a random tangent vector in the tangent space of
 the point `vector_at` on the `FixedRankMatrices` manifold `M`.
 """
-function Random.rand(
-    M::FixedRankMatrices{m,n,k};
-    vector_at=nothing,
-    kwargs...,
-) where {m,n,k}
-    if vector_at === nothing
-        p = SVDMPoint(
-            Matrix{Float64}(undef, m, k),
-            Vector{Float64}(undef, k),
-            Matrix{Float64}(undef, k, n),
-        )
-        return rand!(M, p; kwargs...)
-    else
-        X = UMVTVector(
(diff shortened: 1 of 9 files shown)


adienes:
why separate out `vector_at` from the remainder of `kwargs...`?

mateuszbaran:
> why separate out `vector_at` from the remainder of `kwargs...`?

Good point -- default values are handled by the called function so this can be simplified.

> also, do all manifolds subtype `AbstractDecoratorManifold` ? seems like a `rand(M::T, args...; kwargs...) where {T <: AbstractDecoratorManifold} = rand(Random.default_rng(), args...; kwargs...)` would be easier than repeating this pattern every time

No, not all manifolds subtype `AbstractDecoratorManifold` but all subtype `AbstractManifold`. So we could have `rand(M::T, args...; kwargs...) where {T <: AbstractManifold} = rand(Random.default_rng(), M, args...; kwargs...)`. Good idea.

adienes:
I know I have very little history with this package so I apologize if I sound picky or backseat-driving :)

It seems a little funky to put the dispatches for `Base.rand!` and `Base.in` in `src/Manifolds.jl` when the rest of that file is all about just managing `include`, `export`, `requires`, and the like. Maybe these can be factored out into a file like `abstractmanifold_fallbacks.jl` and included into `src/Manifolds.jl` ?

kellertuer:
Sure we can do these fallbacks in a separate file. In the long run they probably could move to base.

and @adienes no, these questions are absolutely ok and fine. Here we were just lazy, I think, but even in general, when we now worked with a few things for quite a while, they might seem normal / obvious for us but maybe they have better alternatives – or actually a good reason; but reflecting/discussing those things (again) is often a good idea.

mateuszbaran:
Maybe let's merge this first since #604 needs a bit more thought?

PR #608, opened 2023-05-18
Add optional boundary to probability simplex

mateuszbaran:
I've checked a few sources, primarily Ay's book, and I think that's all that has to be adjusted. That boundary is still weird because its tangent space effectively doesn't allow for tangent vectors pointing into the simplex. Pushforward of the map between sphere and probability simplex simply zeroes these components.

### DIFF
--- a/src/manifolds/ProbabilitySimplex.jl
+++ b/src/manifolds/ProbabilitySimplex.jl
@@ -8,8 +8,8 @@ The (relative interior of) the probability simplex is the set
 ````
 where $\mathbb{1}=(1,…,1)^{\mathrm{T}}∈ ℝ^{n+1}$ denotes the vector containing only ones.
 
-If `closed` is set to `false`, then the object represents an open simplex. Otherwise,
-that is when `closed` is set to `true`, the boundary is also included:
+If `closed` is set to `:open`, then the object represents an open simplex. Otherwise,
+that is when `closed` is set to `:Closed`, the boundary is also included:
 ````math
 \hat{Δ}^n := \biggl\{ p ∈ ℝ^{n+1}\ \big|\ p_i \geq 0 \text{ for all } i=1,…,n+1,
 \text{ and } ⟨\mathbb{1},p⟩ = \sum_{i=1}^{n+1} p_i = 1\biggr\},
@@ -32,7 +32,7 @@ This implementation follows the notation in [^ÅströmPetraSchmitzerSchnörr2017
 
 # Constructor
 
-    ProbabilitySimplex(n::Int; closed::Bool=false)
+    ProbabilitySimplex(n::Int; closed::Symbol=:open)
 
 [^ÅströmPetraSchmitzerSchnörr2017]:
     > F. Åström, S. Petra, B. Schmitzer, C. Schnörr: “Image Labeling by Assignment”,
@@ -42,7 +42,12 @@ This implementation follows the notation in [^ÅströmPetraSchmitzerSchnörr2017
 """
 struct ProbabilitySimplex{n,closed} <: AbstractDecoratorManifold{ℝ} end
 
-ProbabilitySimplex(n::Int; closed::Bool=false) = ProbabilitySimplex{n,closed}()
+function ProbabilitySimplex(n::Int; closed::Symbol=:open)
+    if closed !== :open && closed !== :Closed
+        throw(ArgumentError("closed can only be set to :open or :Closed; received $closed"))
+    end
+    return ProbabilitySimplex{n,closed}()
+end
 
 """
     FisherRaoMetric <: AbstractMetric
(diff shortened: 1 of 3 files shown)


src/manifolds/ProbabilitySimplex.jl
```diff
@@ -190,18 +207,34 @@ injectivity_radius(M::ProbabilitySimplex) = 0
...
+    > Mathematik und ihrer Grenzgebiete. 3. Folge / A Series of Modern Surveys in
+    > Mathematics. Springer International Publishing, 2017.
+    > doi: [10.1007/978-3-319-56478-4](https://doi.org/10.1007/978-3-319-56478-4)
 """
-function inner(::ProbabilitySimplex, p, X, Y)
+function inner(::ProbabilitySimplex{n,closed}, p, X, Y) where {n,closed}
     d = zero(Base.promote_eltype(p, X, Y))
-    @inbounds for i in eachindex(p, X, Y)
-        d += X[i] * Y[i] / p[i]
+    if closed
```
kellertuer, line 228: Could we maybe do these via dispatch instead of an if? The (small) advantage would be that we could provide two docstrings instead of one.

mateuszbaran: Yes, though then we would have to duplicate initialization of `d` and parts of the docstring so I'm not sure if that's better.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -1,5 +1,5 @@
 @doc raw"""
-    ProbabilitySimplex{n} <: AbstractDecoratorManifold{𝔽}
+    ProbabilitySimplex{n,closed} <: AbstractDecoratorManifold{𝔽}
```
kellertuer, line 2: Maybe just philosophically (also for dispatch) we could also have
* `open` as true false, which does of course he same just on storing `!closed!`instead
* `boundary` which we could do with symbols `:Open` and `:Closed` where one of them ( for historic reasons probably `:Open`) is the default. This might make dispatching code on `ProbabilitySimplex{n,:Open}`more readable then in `ProbablilitySimplex{n, True}` remembering what the `true` stands for?

src/manifolds/ProbabilitySimplex.jl
kellertuer: Duplication in doc strings is not that bad I think, in code a bit more, sure. So overall I am not 100% we should split it into two function definitions, but we should at least think a bit about it.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: Yes, I didn't think about readability. Using symbols is of course nicer. I will change that.

kellertuer: I actually came up with the idea also only _while_ writing the `open=true|false` thing. I think for code-readability the symbols are a bit nicer, yes, though they allow for a bit more undefined cases.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -8,8 +8,8 @@ The (relative interior of) the probability simplex is the set
 ````
 where $\mathbb{1}=(1,…,1)^{\mathrm{T}}∈ ℝ^{n+1}$ denotes the vector containing only ones.
 
-If `closed` is set to `false`, then the object represents an open simplex. Otherwise,
-that is when `closed` is set to `true`, the boundary is also included:
+If `closed` is set to `:open`, then the object represents an open simplex. Otherwise,
+that is when `closed` is set to `:Closed`, the boundary is also included:
```
kellertuer, line 12: should we capitalise the symbols or not?

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -32,7 +32,7 @@ This implementation follows the notation in [^ÅströmPetraSchmitzerSchnörr2017
 
 # Constructor
 
-    ProbabilitySimplex(n::Int; closed::Bool=false)
+    ProbabilitySimplex(n::Int; closed::Symbol=:open)
```
kellertuer, line 35: I would prefer to call the variable then `boundary` and not `closed`?

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: I'm not sure, I think non-capitalized is more common in Julia but I'm fine with either choice.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: Sure, makes sense.

kellertuer (approved):
LGTM. I personally would have dispatched on `boundary` instead of the `if`s, but that is surely just personal favour and style

kellertuer:
Oh, just saw, can you trigger the DomainError in a test for the closed case? That line seems to be not covered.

mateuszbaran:
Well spotted, I will add a test for that.

Issue #609, opened 2023-05-20
Issues integrating on a Lie group


olivierverdier:
Integrating a constant differential equation on a Lie group starting at the identity should give the same as the exponential. I have many problems trying to make sure this is the case. Did I misunderstand how integration is supposed to work? Here is a minimal working example with various configurations.

Here is the code that produces the problem, I have some questions after the code

```julia
using Manifolds
using ManifoldDiffEq
using OrdinaryDiffEq

# necessary for the SpecialEuclidean case to work?
Base.length(x::ProductRepr) = sum(map(Base.length, submanifold_components(x)))

# G = SpecialOrthogonal(4) # -> works
G = SpecialEuclidean(4) # -> wrong answer (@assert fails)
# G =  ProductGroup(ProductManifold(SpecialOrthogonal(4), TranslationGroup(3))) # -> wrong answer (@assert fails)
# G = TranslationGroup(4) # -> right answer, but expected is changed, so assert fails
# G = GeneralLinear(4) # -> doesn't work (missing method)
# G = HeisenbergGroup(4) # -> doesn't work (missing method)

M = G

action = GroupOperationAction(G)

# ---
# the type of vel differs depending on the call,
# it is either ProductRepr, or ArrayPartition (??)
# vel = rand(G; vector_at=Identity(G))
vel = rand(G; vector_at=identity_element(G))
# ---

expected = exp_lie(G, vel)

dyn = ManifoldDiffEq.LieManifoldDiffEqOperator{Float64}() do u, p, t
    return vel
end

# ---
# Same here, depending on how one initializes creates different errors
# u0 = Identity(G)
u0 = identity_element(G)
# ---

dt = .1 # should not matter at all
prob = ManifoldDiffEq.ManifoldODEProblem(dyn, u0, (0, 1.0), M)
alg = ManifoldDiffEq.ManifoldLieEuler(M, ExponentialRetraction(), action)
# higher order method should not help, and don't indeed
# alg = ManifoldDiffEq.RKMK4(M, ExponentialRetraction(), A)

# sometimes, both expected and vel are changed by calling `solve`?
@show expected
@show vel
sol = solve(prob, alg, dt=dt)
@show expected
@show vel

# expected = exp_lie(G, vel)
computed = sol(1)
@assert isapprox(G, expected, computed)
```

[...]

PR #610, opened 2023-05-20
rand for GL and Heisenberg group

mateuszbaran:
Useful for testing here: #609 .

### DIFF
--- a/src/groups/general_linear.jl
+++ b/src/groups/general_linear.jl
@@ -246,6 +246,18 @@ project(::GeneralLinear, p, X) = X
 project!(::GeneralLinear, q, p) = copyto!(q, p)
 project!(::GeneralLinear, Y, p, X) = copyto!(Y, X)
 
+
+@doc raw"""
+    Random.rand(G::GeneralLinear; kwargs...0)
+
+If `vector_at` is `nothing`, return a random point on the [`GeneralLinear`](@ref) group `G`
+by using `rand` in the embedding.
+
+If `vector_at` is not `nothing`, return a random tangent vector from the tangent space of
+the point `vector_at` on the [`GeneralLinear`](@ref) by using by using `rand` in the embedding.
+"""
+rand(G::GeneralLinear; kwargs...)
+
 function Random.rand!(G::GeneralLinear, pX; kwargs...)
     rand!(get_embedding(G), pX; kwargs...)
     return pX
--- a/src/groups/general_linear.jl
+++ b/src/groups/general_linear.jl
@@ -247,7 +247,7 @@ project!(::GeneralLinear, q, p) = copyto!(q, p)
 project!(::GeneralLinear, Y, p, X) = copyto!(Y, X)
 
 @doc raw"""
-    Random.rand(G::GeneralLinear; kwargs...0)
+    Random.rand(G::GeneralLinear; vector_at=nothing, kwargs...)
 
 If `vector_at` is `nothing`, return a random point on the [`GeneralLinear`](@ref) group `G`
 by using `rand` in the embedding.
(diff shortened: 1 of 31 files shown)


src/groups/general_linear.jl
```diff
@@ -246,6 +246,15 @@ project(::GeneralLinear, p, X) = X
 project!(::GeneralLinear, q, p) = copyto!(q, p)
 project!(::GeneralLinear, Y, p, X) = copyto!(Y, X)
 
+function Random.rand!(G::GeneralLinear, pX; kwargs...)
```
kellertuer, line 249: Would that maybe be a generic good idea for certain manifolds (isometrically embedded ones or such?) to pass rand to the embedding?

kellertuer (changes requested):
My only remark for this PR is (the typical Ronny-one), that both new rand methods should be documented.

src/groups/general_linear.jl
mateuszbaran: I would prefer to avoid trait dispatch here to keep the code simple.

kellertuer: Sure, that is also fine :)

mateuszbaran:
I was actually somewhat expecting another long discussion whether this is the right distribution :sweat_smile:

src/groups/general_linear.jl
```diff
@@ -246,6 +246,27 @@ project(::GeneralLinear, p, X) = X
 project!(::GeneralLinear, q, p) = copyto!(q, p)
 project!(::GeneralLinear, Y, p, X) = copyto!(Y, X)
 
+
+@doc raw"""
+    Random.rand(G::GeneralLinear; kwargs...0)
```
kellertuer, line 251:
```suggestion
    Random.rand(G::GeneralLinear; vector_at=nothing, kwargs...)
```

This way the default is mentioned (and I removed a typo)

kellertuer (approved):
Besides the one typo, this looks good to me.

src/groups/semidirect_product_group.jl
```diff
@@ -123,13 +123,15 @@ function _compose!(G::SemidirectProductGroup, x, p, q)
...
+    x_tmp = allocate(x)
     np, hp = submanifold_components(G, p)
     nq, hq = submanifold_components(G, q)
-    nx, hx = submanifold_components(G, x)
+    nx, hx = submanifold_components(G, x_tmp)
     compose!(H, hx, hp, hq)
     nxtmp = apply(A, hp, nq)
     compose!(N, nx, np, nxtmp)
     @inbounds _padpoint!(G, x)
+    copyto!(x, x_tmp)
```
mateuszbaran, line 134: This is the bugfix from https://github.com/JuliaManifolds/Manifolds.jl/issues/609 -- I just didn't want to open yet another PR...

mateuszbaran: Basically `p` and `x` were the same variable and overwriting `x` caused `p` to change.

kellertuer: Oh, indeed that was a side effect that should not happen, well-spotted. Sure we can add it here as well. Though at some point we should maybe also start a Changelog here (to see where we did what also here).

mateuszbaran: Yes, I would really like to have a changelog here but I'm usually less enthusiastic about actually writing one :sweat_smile:

kellertuer: Then let's start one and I take the role (similar to the docs) of reminding everyone (hopefully, forgot that twice in Manopt already as well and did one PR with 3 changelog entries).

I think it's worth it, especially for the really condensed overview of when which topic/feature appeared.

PR #612, opened 2023-05-23
Make ArrayPartition default instead of ProductRepr

mateuszbaran:
Resolves one of the issues raised in #609 .

This should be only slightly breaking so I don't think an actual breaking release is necessary.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.8.66"
+version = "0.8.67"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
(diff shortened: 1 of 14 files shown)


kellertuer:
How breaking this is is maybe hard to estimate?
But we could introduce a default function (yet again) for this as well?

And we could also deprecate ProductRepr?

kellertuer:
You changed quite some code for the switch, so a default data type might be helpful in case we want to switch again?

I think in favour of ArrayPartition we should deprecate ProductRepr (and remove it in 0.9), since I do not see much gain in inventing our own data type if probably better tested and maybe more efficient others do exist.

mateuszbaran:
Most code changes is either adapting tests or adding some missing functionality to `ArrayPartition`. Actually changing the default is something like replacing `ProductRepr` with `ArrayPartition` in a few places.

I guess the only way to switch defaults would be encoding it in the manifold?

`ArrayPartition` is generally a better choice. I don't quite like the way it's printed and the linear indexing seems like a hacky and dangerous feature but it's more compatible thanks to it.

kellertuer:
I just feel it is maybe a bit beyond what Manifolds.jl should provide, if it exists - unless there is a really good reason for ProductRepr.

The printing could be a PR to the original package then?
About the internals, you are more experienced than me.

mateuszbaran:
@kellertuer I think this is ready :slightly_smiling_face:

kellertuer (approved):

mateuszbaran:
I remember reworking Circle some time ago to make it accept one-element vectors as points and tangent vectors but the current implementation is a bit conflicted about it.

kellertuer:
I hope we did not loose any of these in merges? I had in mind Circle works for both, yes.

mateuszbaran:
It was a bit incorrect but I haven't checked git blame to see why.

PR #619, opened 2023-05-27
Deprecate LinearAffineMetric, resolves #527

### LINKED ISSUE #527: Rename `LinearAffineMetric` to `AffineInvariantMetric`
We currently use the type `LinearAffineMetric` at

https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/symmetricpositivedefinite.html#Manifolds.LinearAffineMetric

for the metric that is actually affine invariant. I think this might be an old remedy of earlier code I wrote, where I somehow used this name. We should rename this for the next breaking change.

kellertuer:
use `AffineInvariantMetric` instead. This resolves #527.

I am just not sure the deprecation tests work, it seems one can check for a warning (if tests are run with `depwarn=true` or use `@test_deprecated´ but I can consistently get both to pass and fail. So let's see what the CI likes.

### DIFF
--- a/docs/src/manifolds/symmetricpositivedefinite.md
+++ b/docs/src/manifolds/symmetricpositivedefinite.md
@@ -25,7 +25,7 @@ Filter = t -> t !== mean
 
 ```@autodocs
 Modules = [Manifolds]
-Pages = ["manifolds/SymmetricPositiveDefiniteLinearAffine.jl"]
+Pages = ["manifolds/SymmetricPositiveDefiniteAffineInvariant.jl"]
 Order = [:type]
 ```
 
--- a/docs/src/manifolds/symmetricpositivedefinite.md
+++ b/docs/src/manifolds/symmetricpositivedefinite.md
@@ -21,7 +21,7 @@ Private=false
 Filter = t -> t !== mean
 ```
 
-## Default metric: the linear affine metric
+## Default metric: the affine invariant metric
 
 ```@autodocs
 Modules = [Manifolds]
@@ -29,7 +29,7 @@ Pages = ["manifolds/SymmetricPositiveDefiniteLinearAffine.jl"]
 Order = [:type]
 ```
 
-This metric is also the default metric, i.e. any call of the following functions with `P=SymmetricPositiveDefinite(3)` will result in `MetricManifold(P,LinearAffineMetric())`and hence yield the formulae described in this seciton.
+This metric is also the default metric, i.e. any call of the following functions with `P=SymmetricPositiveDefinite(3)` will result in `MetricManifold(P,AffineInvariantMetric())`and hence yield the formulae described in this seciton.
 
 ```@autodocs
 Modules = [Manifolds]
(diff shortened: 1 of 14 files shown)


tutorials/getstarted.jl
```diff
@@ -458,7 +458,7 @@ M₈ = SymmetricPositiveDefinite(3)
 # ╔═╡ 82dca58f-52e6-491e-97ec-904bb5b0b36b
 md"""
 which is the manifold of ``3×3`` matrices that are [symmetric and positive definite](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/symmetricpositivedefinite.html#Manifolds.SymmetricPositiveDefinite).
-which has a default as well, the affine invariant [`LinearAffineMetric`](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/symmetricpositivedefinite.html#Default-metric:-the-linear-affine-metric), but also has several different metrics.
+which has a default as well, the affine invariant [`AffineInvariantMetric`](https://juliamanifolds.github.io/Manifolds.jl/latest/manifolds/symmetricpositivedefinite.html#Default-metric:-the-linear-affine-metric), but also has several different metrics.
```
mateuszbaran, line 461: The link is probably incorrect now?

kellertuer: Well spotted. I currently can not render the Pluto notebooks with PlutoStaticHTML it seems, but lets see what the CI does.

kellertuer: Zes, I am a bit stuck. PlutoStaticHTML fails because it is either not working with Julia 1.9 completely, or because the Pluto notebooks are 1.8.1 notebooks – I am not sure, but I can also not update the hand gestures notebook, since Pluto says that is not a notebook...

kellertuer: CI says it should be fine now.

docs/src/manifolds/symmetricpositivedefinite.md
```diff
@@ -21,15 +21,15 @@ Private=false
 Filter = t -> t !== mean
 ```
 
-## Default metric: the linear affine metric
+## Default metric: the affine invariant metric
 
 ```@autodocs
 Modules = [Manifolds]
 Pages = ["manifolds/SymmetricPositiveDefiniteLinearAffine.jl"]
```
mateuszbaran, line 28:
```suggestion
Pages = ["manifolds/SymmetricPositiveDefiniteAffineInvariant.jl"]
```
I haven't tested but that looks right.

kellertuer: again, wellSpotten, I had hoped having found all with my search and replaced, but might have missed this.

mateuszbaran: I just checked if the docs get rendered correctly and they weren't, so I looked what might be the reason :slightly_smiling_face: .

mateuszbaran: ...and there was one more reference to update :smile:

mateuszbaran (approved):

kellertuer:
Thanks for the careful revision and help :)

Issue #606, opened 2023-05-14
`parallel_transport_to` for `ProbabilitySimplex`


adienes:
Is this just the identity?

I'm by no means a geometer and am probably way off base, but from what I can gather reading some articles about parallel transport, I think since the simplex is flat everywhere can the method be implemented like
```
parallel_transport_to!(::ProbabilitySimplex, Y, ::Any, X, ::Any) = copyto!(Y, X)
```

kellertuer:
That depends a lot on the metric you impose on that set. Could you provide references? My short search did not bring up that many results that I found helpful concerning parallel transport.
The one reason I think might indicate that this is not the case, is, that the set is open (i.e. you never reach the boundary, which for me feels quite non-Euclidean.

mateuszbaran:
No, `ProbabilitySimplex` is not flat (w.r.t Fisher-Rao metric). As a reference, it looks like https://github.com/geomstats/geomstats/blob/master/geomstats/information_geometry/multinomial.py is the same as our `ProbabilitySimplex` (just with a different scaling) and it says it has positive sectional curvature. I couldn't find any source for PT on that manifold so most likely there is no known closed-form formula.

sethaxen:
I haven't read the literature on the Fisher-Rao metric, but it seems to be closely related to the Euclidean metric on the sphere. Specifically, let $p \in \mathbb{S}^n$ and let $f:  \mathbb{S}^n \to \Delta^n$ be the elementwise square: $q_i = f(p)_i = p_i^2$.

Then the differential of this map $\mathrm{d}f_p: T_p\mathbb{S}^n \to T_{f(p)}\Delta^n$ has the property $(X_q)_i = (\mathrm{d}f_p)(X_p)_i = 2 p_i (X_p)_i$. The Fisher-Rao metric is

$$g: (X_q, Y_q) \mapsto \sum_{i=1}^n \frac{(X_q)_i (Y_q)_i}{q_i} = \langle 2X_p, 2Y_p \rangle_F$$

[...]

PR #621, opened 2023-05-31
Add more functions on ProbabilitySimplex

mateuszbaran:
I've implemented the isomorphism between interior of probability simplex and a scaled sphere, which in turn are used to define an orthonormal basis on the probability simplex. This is essentially what is described in https://github.com/JuliaManifolds/Manifolds.jl/issues/606#issuecomment-1546866436 but with a different scaling. Let me know if this scaling feels right.

### DIFF
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -949,4 +949,14 @@ export get_basis,
 # atlases and charts
 export get_point, get_point!, get_parameters, get_parameters!
 
+# additional functions on Simplex
+export from_probability_amplitude,
+    from_probability_amplitude!,
+    from_probability_amplitude_diff,
+    from_probability_amplitude_diff!,
+    to_probability_amplitude,
+    to_probability_amplitude!,
+    to_probability_amplitude_diff,
+    to_probability_amplitude_diff!
+
 end # module
--- a/src/manifolds/ProbabilitySimplex.jl
+++ b/src/manifolds/ProbabilitySimplex.jl
@@ -211,7 +211,6 @@ function get_coordinates_orthonormal!(
         to_probability_amplitude_diff(M, p, X),
         R,
     )
-    Xc .*= 2
     return Xc
 end
 
@@ -222,8 +221,7 @@ function get_vector_orthonormal!(
     Xc,
     R::RealNumbers,
 ) where {N}
-    Xcs = Xc / 2
-    X = get_vector_orthonormal(Sphere(N), to_probability_amplitude(M, p), Xcs, R)
+    X = get_vector_orthonormal(Sphere(N), to_probability_amplitude(M, p), Xc, R)
     return from_probability_amplitude_diff!(M, Y, to_probability_amplitude(M, p), X)
 end
 
@@ -508,7 +506,7 @@ function to_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
 end
 
 function to_probability_amplitude_diff!(::ProbabilitySimplex, Y, p, X)
-    Y .= X ./ sqrt.(p) ./ 2
+    Y .= X ./ sqrt.(p)
     return Y
 end
 
@@ -516,13 +514,13 @@ end
     from_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
 
 Compute differential of [`from_probability_amplitude`](@ref) at tangent vector `X` from the
-tangent space at `p` from a sphere with a scaled metric.
+tangent space at `p` from a sphere.
 """
 function from_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
     return from_probability_amplitude_diff!(M, similar(X), p, X)
 end
 
 function from_probability_amplitude_diff!(::ProbabilitySimplex, Y, p, X)
-    Y .= 2 .* p .* X
+    Y .= p .* X
     return Y
 end
--- a/src/manifolds/ProbabilitySimplex.jl
+++ b/src/manifolds/ProbabilitySimplex.jl
@@ -197,6 +197,36 @@ function exp!(::ProbabilitySimplex, q, p, X)
     return q
 end
 
+function get_coordinates_orthonormal!(
+    M::ProbabilitySimplex{N},
+    Xc,
+    p,
+    X,
+    R::RealNumbers,
+) where {N}
+    get_coordinates_orthonormal!(
+        Sphere(N),
+        Xc,
+        to_probability_amplitude(M, p),
+        to_probability_amplitude_diff(M, p, X),
+        R,
+    )
+    Xc .*= 2
+    return Xc
+end
+
+function get_vector_orthonormal!(
+    M::ProbabilitySimplex{N},
+    Y,
+    p,
+    Xc,
+    R::RealNumbers,
+) where {N}
+    Xcs = Xc / 2
+    X = get_vector_orthonormal(Sphere(N), to_probability_amplitude(M, p), Xcs, R)
+    return from_probability_amplitude_diff!(M, Y, to_probability_amplitude(M, p), X)
+end
+
 @doc raw"""
     injectivity_radius(M,p)
 
@@ -422,7 +452,7 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
 end
 
 function Base.show(io::IO, ::ProbabilitySimplex{n,boundary}) where {n,boundary}
-    return print(io, "ProbabilitySimplex($(n); boundary=$boundary)")
+    return print(io, "ProbabilitySimplex($(n); boundary=:$boundary)")
 end
 
 @doc raw"""
@@ -434,3 +464,65 @@ returns the zero tangent vector in the tangent space of the point `p`  from the
 zero_vector(::ProbabilitySimplex, ::Any)
 
 zero_vector!(::ProbabilitySimplex, X, p) = fill!(X, 0)
+
+@doc raw"""
+    to_probability_amplitude(M::ProbabilitySimplex{N}, p) where {N}
+
+Convert point `p` on `ProbabilitySimplex` to (real) probability amplitude. The formula reads
+``(\sqrt{p_1}, \sqrt{p_2}, …, \sqrt{p_{N+1}})``. This is an isometry from the interior of
+the probability simplex to the interior of the positive orthant of a sphere.
+"""
+function to_probability_amplitude(M::ProbabilitySimplex, p)
+    return to_probability_amplitude!(M, similar(p), p)
+end
+
+function to_probability_amplitude!(::ProbabilitySimplex, q, p)
+    q .= sqrt.(p)
+    return q
+end
+
+@doc raw"""
+    from_probability_amplitude(M::ProbabilitySimplex{N}, p) where {N}
+
+Convert point (real) probability amplitude `p` on to a point on `ProbabilitySimplex`.
+The formula reads ``(p_1^2, p_2^2, …, p_{N+1}^2)``. This is an isometry from the interior of
+the positive orthant of a sphere to interior of the probability simplex.
+"""
+function from_probability_amplitude(M::ProbabilitySimplex, p)
+    return from_probability_amplitude!(M, similar(p), p)
+end
+
+function from_probability_amplitude!(::ProbabilitySimplex, q, p)
+    q .= p .^ 2
+    return q
+end
+
+@doc raw"""
+    to_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
+
+Compute differential of [`to_probability_amplitude`](@ref) at tangent vector `X` from the
+tangent space at `p` from `M`.
+"""
+function to_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
+    return to_probability_amplitude_diff!(M, similar(X), p, X)
+end
+
+function to_probability_amplitude_diff!(::ProbabilitySimplex, Y, p, X)
+    Y .= X ./ sqrt.(p) ./ 2
+    return Y
+end
+
+@doc raw"""
+    from_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
+
+Compute differential of [`from_probability_amplitude`](@ref) at tangent vector `X` from the
+tangent space at `p` from a sphere with a scaled metric.
+"""
+function from_probability_amplitude_diff(M::ProbabilitySimplex, p, X)
+    return from_probability_amplitude_diff!(M, similar(X), p, X)
+end
+
+function from_probability_amplitude_diff!(::ProbabilitySimplex, Y, p, X)
+    Y .= 2 .* p .* X
+    return Y
+end
--- a/test/manifolds/probability_simplex.jl
+++ b/test/manifolds/probability_simplex.jl
@@ -7,6 +7,7 @@ include("../utils.jl")
     q = [0.3, 0.6, 0.1]
     X = zeros(3)
     Y = [-0.1, 0.05, 0.05]
+    @test repr(M) == "ProbabilitySimplex(2; boundary=:open)"
     @test is_point(M, p)
     @test_throws DomainError is_point(M, p .+ 1, true)
     @test_throws ManifoldDomainError is_point(M, [0], true)
@@ -49,6 +50,7 @@ include("../utils.jl")
             test_manifold(
                 M,
                 pts,
+                basis_types_to_from=(DefaultOrthonormalBasis(),),
                 test_injectivity_radius=false,
                 test_project_tangent=true,
                 test_musical_isomorphisms=true,
@@ -110,5 +112,19 @@ include("../utils.jl")
         @test inner(Mb, p, X, Y) == 8
 
         @test_throws ArgumentError ProbabilitySimplex(2; boundary=:tomato)
+    end
+
+    @testset "Probability amplitudes" begin
+        p = [0.1, 0.7, 0.2]
+        Y = [-0.1, 0.05, 0.05]
+        @test to_probability_amplitude(M, p) ≈
+              [0.31622776601683794, 0.8366600265340756, 0.4472135954999579]
+        @test from_probability_amplitude(
+            M,
+            [0.31622776601683794, 0.8366600265340756, 0.4472135954999579],
+        ) ≈ p
+        @test to_probability_amplitude_diff(M, p, Y) ≈
+              [-0.15811388300841897, 0.02988071523335984, 0.05590169943749475]
+        @test from_probability_amplitude_diff(M, p, Y) ≈ [-0.02, 0.07, 0.02]
     end
 end
--- a/test/manifolds/probability_simplex.jl
+++ b/test/manifolds/probability_simplex.jl
@@ -124,7 +124,7 @@ include("../utils.jl")
             [0.31622776601683794, 0.8366600265340756, 0.4472135954999579],
         ) ≈ p
         @test to_probability_amplitude_diff(M, p, Y) ≈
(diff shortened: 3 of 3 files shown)


src/Manifolds.jl
```diff
@@ -949,4 +949,14 @@ export get_basis,
 # atlases and charts
 export get_point, get_point!, get_parameters, get_parameters!
 
+# additional functions on Simplex
```
kellertuer, line 952: Is it reasonable to export these functions or are they more of an internal part?

mateuszbaran:
> I find the naming a bit unintuitive, if that maps `ProbabilitySimplex` to (the sqrt(2)-radius?) sphere, why not use a convert? or even embed?

No, it maps to radius-1 sphere. I don't see how `convert` could be used but `embed` (with `EmbeddedManifold`) may be a good idea.

src/Manifolds.jl
mateuszbaran: (outside of Manifolds.jl)

kellertuer:
I see, convert might actually be the wrong idea, but embed is maybe an idea then (so embed and project or something?) We could even do the sphere we have in the tutorial for real (that is a sphere with a radius) if that helps here.

mateuszbaran:
I've changed it to use `embed` and `project` through `RealProbabilityAmplitudes`.

mateuszbaran:
And I've also added PT.

kellertuer:
Nice! I am not 100% sure where the name of the new struct comes from and the “some properties carry over” is maybe a bit vague, but the function names are far better :)

mateuszbaran:
> I am not 100% sure where the name of the new struct comes from

This comes from physics. In quantum mechanics we have complex probability amplitudes. If you restrict them to reals, you get this. Maybe it's not the most common name so if you (or someone else) has a better name, I can change it.

> “some properties carry over”

This is a deeper topic that requires an introduction to manifolds with boundaries. I just wanted to indicate that extending to the boundary is not a hopeless endeavor.

kellertuer:
> > I am not 100% sure where the name of the new struct comes from
> 
> This comes from physics. In quantum mechanics we have complex probability amplitudes. If you restrict them to reals, you get this. Maybe it's not the most common name so if you (or someone else) has a better name, I can change it.
> 

I do not have a better name, but maybe providing some insight on the name might help a reader of the docs as well?

> > “some properties carry over”
> 
> This is a deeper topic that requires an introduction to manifolds with boundaries. I just wanted to indicate that extending to the boundary is not a hopeless endeavor.

I even had a short discussion about manifolds with boundary today as well – I was just thinking, being a bit more precise might be nice.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -422,15 +453,93 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
 """
 zero_vector(::ProbabilitySimplex, ::Any)
 
 zero_vector!(::ProbabilitySimplex, X, p) = fill!(X, 0)
+
+@doc raw"""
+    RealProbabilityAmplitudes
+
+An explicit isometric embedding of interior of [`ProbabilitySimplex`] in
+positive quadrant of the [`Sphere`]. Some properties extend to the boundary but not all.
```
sethaxen, line 473: https://en.wikipedia.org/wiki/Orthant
```suggestion
positive orthant of the [`Sphere`]. Some properties extend to the boundary but not all.
```

sethaxen: It would be nice if the docstring explained why it's called this. Because otherwise a name that reflects the two manifolds might be nice.

sethaxen (commented):
Just a question about the naming of the embedding. Also, it would be nice to document somewhere how the Fisher information metric relates to this embedding and the Euclidean metric on the sphere.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: Sure, I will add a comment about it.

mateuszbaran:
I've added some docs about the name and relation between metric on sphere and Fisher-Rao metric on probability simplex.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -421,16 +452,119 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
+
+@doc raw"""
+    project(M::RealProbabilityAmplitudes{N}, p) where {N}
+
+Convert point (real) probability amplitude `p` on to a point on [`ProbabilitySimplex`](@ref).
+The formula reads ``(p_1^2, p_2^2, …, p_{N+1}^2)``. This is an isometry from the interior of
+the positive orthant of a sphere to interior of the probability simplex.
+"""
+function project(M::RealProbabilityAmplitudes, p)
+    return project!(M, similar(p), p)
```
kellertuer, line 534: Why does this not work on the default with allocate and such?
Or phrased differently: I expected project only to be here for the doc-string not for an actual implementation.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -421,16 +452,119 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
+
+@doc raw"""
+    embed(M::RealProbabilityAmplitudes{N}, p) where {N}
+
+Convert point `p` on `ProbabilitySimplex` to (real) probability amplitude. The formula reads
+``(\sqrt{p_1}, \sqrt{p_2}, …, \sqrt{p_{N+1}})``. This is an isometry from the interior of
+the probability simplex to the interior of the positive orthant of a sphere.
+"""
+function embed(M::RealProbabilityAmplitudes, p)
+    return embed!(M, similar(p), p)
```
kellertuer, line 518: Why does this not work on the default with allocate and such?
Or phrased differently: I expected ember only to be here for the doc-string not for an actual implementation.

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -421,16 +452,119 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
+    return q
+end
+
+@doc raw"""
+    embed(M::RealProbabilityAmplitudes, p, X)
+
+Compute differential of [`embed`](@ref) on [`RealProbabilityAmplitudes`](@ref) at tangent
+vector `X` from the tangent space at `p` from sphere.
+"""
+function embed(M::RealProbabilityAmplitudes, p, X)
```
kellertuer, line 548: merely a comment – this got much nicer compared to the previous naming I think :)

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -421,16 +452,119 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
+
+function embed!(::RealProbabilityAmplitudes, Y, p, X)
+    Y .= X ./ sqrt.(p)
+    return Y
+end
+
+@doc raw"""
+    project(M::RealProbabilityAmplitudes, p, X)
+
+Compute differential of [`project`](@ref) on [`RealProbabilityAmplitudes`](@ref) at tangent
```
kellertuer, line 560: This line might require an update?

kellertuer (commented):
Here are just a few minor comments, I think the new EmbeddedManifold improved the code substantially.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: I just really didn't feel like tracking why the trait machinery didn't do the right thing here. It's also faster to compile just this.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: Yes, I agree.

src/manifolds/ProbabilitySimplex.jl
kellertuer: Hm, ok, my only (small) complaint would be a slight inconsistency here.

src/manifolds/ProbabilitySimplex.jl
kellertuer: Or to be more precise – stating that a function that is called project – computes a differential of project (itself?) might be misleading.

mateuszbaran: In this case I think understanding it as a differential of point projection is more appropriate, and understanding it as a projection from embedding is misleading.

kellertuer: Hm.
* Then the double use of the name `project` is misleading
* A function that is not projection but called project is also a bit misleading
* I have no clue what this function does then ;)

kellertuer: Maybe it should indeed be called `diff_project(M, p, X)` then?

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -421,16 +452,119 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
+    return q
+end
+
+@doc raw"""
+    embed(M::RealProbabilityAmplitudes, p, X)
+
+Compute differential of [`embed`](@ref) of a point on `p` [`RealProbabilityAmplitudes`](@ref)
+at tangent vector `X` from the tangent space at `p` from sphere.
+"""
+function embed(M::RealProbabilityAmplitudes, p, X)
```
kellertuer, line 548: Maybe this should then be called `diff_embed` to not confuse it with `embed(M, p, X)` which brings X, a tangent vector at X into the embedding (but is otherwise the identity and not some differential)?

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: It technically is a projection, for a sufficiently broad definition of projection. And AFAIK some vector projections that we have are also differentials of point projections (f.e. sphere). This is a bit weird case because we embed in a space of the same dimension and point projection doesn't return the point closest in the geodesic sense. But otherwise it's technically fine.

kellertuer: Je suis veldig verwirrt now (even in 4 languages).
The vector projection on the sphere actually agrees with projecting the vector onto the tangent space (interpreted as the plane parallel to the tangent plane running through zero of course).

So for now I am a bit lost what this function actually is. The docs say it is the differential of the point-projection (so I would call it `diff_project`).
So for me the most generic definition of a projection (say a function f) is that `f(f(p))=f(p)` or that `||f(p)-p|| is minimal compared to all other values f can attain (here in the tangent space).

Technically – I do not want to discuss and trust you – in confusion level … I would like to de-confuse both me and readers of these docs ;)

src/manifolds/ProbabilitySimplex.jl
kellertuer: Yes, here I see that a bit more compared to the projection, because (besides the change of how the point is stored) embed is the identity, so its differential is as well but is for sure maps from TpM to T_(embed(p)) N.

So here I agree in terms of the mathematics, but for now I think we did not often write that but used the more generic description, that it embeds tangent vectors (without the reference to the differential).

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: I'm definitely interested in clarifying the confusion but I hope to do it correctly :slightly_smiling_face: .

> So for me the most generic definition of a projection (say a function f) is that `f(f(p))=f(p)` or that `||f(p)-p|| is minimal compared to all other values f can attain (here in the tangent space).

It's not the right definition here because because our `f` here has different domain and codomain. So `f(f(p))` is not defined, and you can't compare `p` and `f(p)`. This is already true for our embedding of SE(n) in GL(n+1), it works the same there. But it can indeed be confusing. Maybe using the names `project` and `embed` really isn't right.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: It being the differential is a crucial part of the API though, users should be able to rely that for this particular embedding it is the differential.

src/manifolds/ProbabilitySimplex.jl
kellertuer: For embed (±change of representation in code/variable sense) you convinced me. Here you did not yet, though you are right, since we do have a change of the space (domain-range) my two notions were not that correct here. They are maybe more like (embed=id into the embedding) `f(embed(f(p)) == f(p)` and `embed(f(p))` is the closest point to `p` that embed can reach

src/manifolds/ProbabilitySimplex.jl
kellertuer: Yes, but since we never nowhere (as far as I remember) called it that (and I am not sure we always have that) it might be misleading to use that terminology here (without using the other one the reader is used to).

It's like having a thing with two names A and B, we only call it A but for fun in one place we call it B. Maybe start with A (embed the tangent vector) here first and then mention how that is technically/mathematically achieved?

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: OK, then I think this `project(M, p)` is not a valid projection. It's just a local isometry and we need a different API :slightly_smiling_face: .

kellertuer: I think in the very beginning I already said: What about `convert`? Because for me this looks like a conversion between two representations of the manifold?

kellertuer: ...hm. though that API is maybe also not yet developed to convert between manifolds of same dimension or isomorphically map.

kellertuer: (sorry for so much discussion, just want to make sure I understand it myself as well)

mateuszbaran: `convert` feels weird here because this is actually about establishing a local isometry, not providing a full new representation of the manifold.

kellertuer: Well, in optical terms this is just “bump”, since we deform the plane to the spherical orthant (or back with project) and similarly adapt the metric for tangent vectors (that follow the same bump). But, hm, I am also missing a good name. `bump()` is not a good name ;)

src/manifolds/ProbabilitySimplex.jl
kellertuer: `debump` (cf. the other discussion). I will think about a name as well.

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: Hm, I'm not sure. If we are doing new interface anyway, then I'd prefer either function name or some argument to be specific to this particular local isometry. Something like `bump_local_isometry` or `local_isometry(bump(M, N), p)`

kellertuer: Bump was far from a serious recommendation. I do not have a good idea for the name of the local isometry here.

kellertuer: But why do you want the second variant (and what does bump represent there? An IsometryType?) why then not switch that to domain, range instead of somefunction(domain, range)?

mateuszbaran: `bump` was an example there, a placeholder for the right word. What I'd like there is a way to handle multiple (local) isometries between two manifolds. For example, what if someone wanted an isometry between ProbabilitySimplex and negative orthant of the sphere?

kellertuer: uff, I see. I was already out of creativity when I came up with `local_isometry` as a name, but you could use `AbstractIsometry` as a super type for the first argument and call this one `DefaultIsometry`? So we would do something similar to Embeddings

mateuszbaran: So, something like `local_isometry(DefaultLocalIsometry(ProbabilitySimplex(N), Sphere(N)), p)`? I guess `local_isometry(ProbabilitySimplex(N), Sphere(N), p)` could be a shorter notation for the first one.

kellertuer: That sounds reasonable and mixes both our ideas :)

src/manifolds/ProbabilitySimplex.jl
```diff
@@ -421,16 +452,119 @@ function riemannian_gradient!(M::ProbabilitySimplex, X, p, Y; kwargs...)
...
+
+@doc raw"""
+    RealProbabilityAmplitudes
+
+An explicit isometric embedding of interior of [`ProbabilitySimplex`] in
+positive orthant of the [`Sphere`]. Some properties extend to the boundary but not all.
+
+This embedding isometrically maps the Fisher-Rao metric on the open probability simplex to
+the sphere of radius 1 with Euclidean metric. More details can be found in Section 2.2
+of [^AyJostLeSchwachhöfer2017].
```
mosullivan93, line 494: Just wanted to improve my understanding by asking a question about how this works.

It looks like you're missing the factor of 2 in the differential (of [f](https://github.com/JuliaManifolds/Manifolds.jl/issues/606#issuecomment-1553249494) and [finv](https://github.com/JuliaManifolds/Manifolds.jl/issues/606#issuecomment-1546866436)) when I look at
`embed!(::RealProbabilityAmplitudes, Y, p, X)` and `project!(::RealProbabilityAmplitudes, Y, p, X)` (which I think are supposed to serve this purpose), so I'm not sure this embedding is still isometric?

I only tried the parallel transport with one set of points and it looked correct though so I'm not sure there's a problem in this construction.

kellertuer: The challenge here might be, that we (contrary to theory, but in favour of our iplementation) embed into the 1-sphere and not the 2-sphere. This is done in order to reuse the code we have for the 1-sphere; but our sphere does not have a way do incorporate a radius.

mateuszbaran: The difference is indeed that those `f` and `finv` functions embed in a sphere of radius 2, while the embedding in this PR is in sphere of radius 1. That's why there is no factor of 2.

mosullivan93: I see, thanks both for clarifying and for working on these new functions 👍

src/manifolds/ProbabilitySimplex.jl
mateuszbaran: I've tried working out the `local_isometry` idea but it seems vastly over-complicated for what is needed here, and forces me to consider things that are irrelevant to this PR (how to do `local_isometry` generic enough). Can we just go back to simple functions specialized to this one manifold embedding? Something like `from_probability_amplitude` and `to_probability_amplitude` from the original commit but just with more appropriate names.

kellertuer: I would prefer both domain and range in the name instead of once the domain once the range? `simplex_to_amplitude` and `amplitude_to_simplex` might not be the best names but indicate what I would prefer.

mateuszbaran: Sure, those names sound fine :+1: .

kellertuer: Then let's take those (and refine them later when we have a general scheme). And let's also keep them non-exported for now then.

mateuszbaran: OK, done :slightly_smiling_face:

kellertuer (approved):
LGTM,
and without exporting them we keep them internal and might have in mind to do a proper isomorphism model later

PR #622, opened 2023-05-31
💡

kellertuer:
Since I learned in Manopt how this works, let's move our one tutorial to this as well.

Main features:
* We can now even allow cells to fail and do no longer need a truck to catch and display an error
* the results are cached, also the Julia and Conda dependencies, so the docs should probably even be a bit faster
* the tutorial is rendered from markdown into markdown (not from Pluto into HTML) which allows even for `@ref` links in the Quarto markdown file.


kellertuer:
Hm it would have been too easy if this just would work. I think `CondaPkg` and `PythonPlot` get angry with each other and then something avoids that `IJulia` is available in the docs environment or so? I have no clue why the CI precompilation fails.

kellertuer:
Took just 90 minutes to see that `IJulia` is probably not necessary in the docs/ environment, but I forgot to add it to the tutorials environment for sure (just not sure why it ran locally just fine). But now it seems to work.

Documenter timings:
* old run on Julia 1.8: 15:34 min
* new run with more packages and 3 instead of 2 envs on Julia 1.9:  15:52 min

I hope Caching will make that a bit nicer, since most time is spent in pre-compilation, we will see.

mateuszbaran:
Cool :rocket: . I will check the new docs tomorrow.

kellertuer:
🚀with cache 4:41 min.

tutorials/getstarted.qmd
```diff
@@ -0,0 +1,471 @@
...
+
+## Allocating and in-place computations
+
+Memory allocation is a [critical performace issue](https://docs.julialang.org/en/v1/manual/performance-tips/#Measure-performance-with-[@time](@ref)-and-pay-attention-to-memory-allocation) when programming in Julia. To take this into account, `Manifolds.jl` provides special functions to reduce the amount of allocations.
+
+We again look at the [📖 exponential map](https://en.wikip edia.org/wiki/Exponential_map_(Riemannian_geometry)). On a manifold `M` the exponential map needs a point `p` (to start from) and a tangent vector `X`, which can be seen as direction to “walk into” as well as the length to walk into this direction. In `Manifolds.jl` the function can then be called with `q = exp(M, p, X)` [🔗](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions.html#Base.exp-Tuple{AbstractManifold,%20Any,%20Any}). This function returns the resulting point `q`, which requires to allocate new memory.
```
mateuszbaran, line 345: Links where only the chain icon is clickable look inconsistent with the rest of this document.

mateuszbaran: And some links don't have the chain icon. Is that intentional? For example below the link to `BuresWassersteinMetric`.

tutorials/getstarted.qmd
```diff
@@ -0,0 +1,471 @@
...
+```
+
+
+This methodology is used for all functions that compute a new point or tangent vector. By default all allocating functions allocate memory and call the in-place function.
+This also means that if you implement a new manifold, you just have to implement the in-place version.
+
+## Decorating a manifold"
+
+As you saw until now, an
+[📎 `AbstractManifold`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/types.html#The-AbstractManifold)
```
mateuszbaran, line 388: Why is there a paperclip here?

tutorials/getstarted.qmd
```diff
@@ -0,0 +1,471 @@
...
+
+```{julia}
+p = [0 0; 0 1; 1 sqrt(2)]
+```
+
+```{julia}
+[is_point(M₂, p[:, 1]), is_point(M₂, p[:, 2])]
+```
+
+But of course the method we used previously also works for power manifolds:"
```
mateuszbaran, line 206:
```suggestion
But of course the method we used previously also works for power manifolds:
```

tutorials/getstarted.qmd
```diff
@@ -0,0 +1,471 @@
...
+```
+
+But the orthogonality of `p` and `X` is still valid, we can disable the point check,
+but even setting the error to true we get here
+
+```{julia}
+is_vector(M₃, [2, 0, 0], [0, 1, 1], true, false)
+```
+
+But of course it is better to use a valid point in the first place"
```
mateuszbaran, line 171:
```suggestion
But of course it is better to use a valid point in the first place
```

mateuszbaran (commented):
The new docs look quite nicely, I only have a few minor fixes. Are you planning to redo the other tutorials in this PR as well?

tutorials/getstarted.qmd
```diff
@@ -0,0 +1,471 @@
...
+
+Thinking of a generic implementation, we would like to have a way to specify one, that is available. This can be done by using [🔗 `default_retraction_method`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/retractions.html#ManifoldsBase.default_retraction_method-Tuple{AbstractManifold}) and [🔗 `default_inverse_retraction_method`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/retractions.html#ManifoldsBase.default_inverse_retraction_method-Tuple{AbstractManifold}), respectively. We implement
+"""
```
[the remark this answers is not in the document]
mateuszbaran: I mean the `"""` part.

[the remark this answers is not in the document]
kellertuer: s spurious end of a Pluto Markdown string I missed to remove.

tutorials/getstarted.qmd
kellertuer: the chain icon meant (in the Pluto still) to refer to links outside the Manifolds.jl docs – here one could link the whole `exp` but since then we have to put the chain behind, that would still be inconsistent.

tutorials/getstarted.qmd
kellertuer: they should all have because it is links to outside the Manifolds.jl docs :) they all should have that icon. But now I am back at my computer and can fix those.

kellertuer:
Now all links should be fixed

* Wikipedia links have a 📖 prefix
* Links to other docs have a 🔗 prefix (that is the UTF8-Symbol for link)
* new stuff: Since we render to Markdown, internal links now work* with `@ref`*

*nearly. In Pandoc runs everything with `[text](@Link)` will look for the reference with the cite key `@Link` and also convert all spaces in the link to `%20`s. But that can (both!) be avoided by using

```
``[`Euclidean`](@ref)``{=commonmark} or ``[Section link](@ref NameOfSectionInDocs)``{=commonmark}
```

so always clarifying these as raw common mark. This is not the nicest thing, but better than hardcoded http links to internal docs pages (where even 2 linked to old PR docs).

tutorials/getstarted.qmd
kellertuer: Ah missed that one and misread it maybe, its fixed now (I think in a very early version paperclip was meant to be  🔗)

kellertuer:
Can this be merged? We can surely register it together with the other PR in one version still :)

mateuszbaran:
I was hoping the other tutorials would also be converted before merging (hand gestures and working in charts).

kellertuer:
They do not have much text, are not proper Pluto Notebooks, so I am not sure what to convert a pure code file to?

kellertuer:
Ah, it is just hadn't gestures that is pure code file without any markdown text

https://github.com/JuliaManifolds/Manifolds.jl/blob/master/tutorials/hand-gestures.jl

the working in charts one seems to be Pluto notebook but it did not run when I tried running it, so not only would it need 1-2 hours to transfer that to Quarto it probably also needs an update to Manifolds 0.8 and some debugging?

mateuszbaran:
OK, then maybe just don't delete them here?

kellertuer:
Sure we can leave them in and when time permits we can turn them into proper Quarto tutorials.
Since my student is currently working with the shape of a rubber duck – that might be a better way to illustrate what Kendall's shape space is (28 points in R2) 🦆 (sadly no yellow rubber duck emoji available)

mateuszbaran:
Sure, we could definitely have a better tutorial with ducks.

mateuszbaran (approved):

kellertuer:
I disabled the Julia recompilation cache (which brings us back to 17 Minutes for the docs), but even then it seems that the CI and compilation/precompilation of Extensions is still a mess. Precompilation fails, but the docs still render fine. This feels more like a problem that the Documenter CI has 3 environments to handle where Extensions then struggle.

PR #623, opened 2023-06-03
Move ManifoldDiff extension here

mateuszbaran:
See here: https://github.com/JuliaManifolds/ManifoldDiff.jl/issues/20

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.8.68"
+version = "0.8.69"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
@@ -48,7 +48,7 @@ Einsum = "0.4"
 Graphs = "1.4"
 HybridArrays = "0.4"
 Kronecker = "0.4, 0.5"
-ManifoldDiff = "0.2.1, 0.3"
+ManifoldDiff = "0.3.3"
 ManifoldsBase = "0.14.1"
 MatrixEquations = "2.2"
 OrdinaryDiffEq = "6.31"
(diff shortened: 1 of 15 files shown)


kellertuer (approved):

mateuszbaran:
These new lines most likely won't be cover until [ManifoldDiff](https://github.com/JuliaManifolds/ManifoldDiff.jl/issues/20) is merged, and tests in that PR likely won't pass until this is merged. Which one do you prefer to merge and tag first?

kellertuer:
For me its ore reasonable to remove it from ManifoldDiff first and then add it here, since otherwise there is a version where the extension exists twice.

mateuszbaran:
I have no idea why Manopt gradient tests fail here. Locally they pass fine.

kellertuer:
The error message seems to be quite Diff-related though

mateuszbaran:
Yes, it doesn't see the `adjoint_Jacobi_field` method for `Circle`. That method was in the extension of ManifoldDiff. We have it in this PR now and it's covered by a test so this makes no sense to me.

kellertuer:
Ah, maybe Manopt still checks our some old version? Because currently Manopt/ManifoldDiff/Manifolds in their most recent version do not play well together, since the extension is nowhere (reverted to 0.3.2 to get it working again). So I am not sure how to resolve this. Register this version and hope for the best?

kellertuer:
The same happens for now on Manopt CI, I think the choice of having a version combination without this extension was not that good and a next release here should fix that.

On the other hand it means that there is now combinations of Manifolds and ManifoldDiff that do not work well together (one or two combinations where the extension is nowhere to be found), but I think this can not be avoided.

kellertuer:
I think I found a solution. I will restrict ManifoldDiff in Manopt to 0.3.2, then it should work fine; we register the stochastic PR (maybe also together with the new tutorial update?) as a new version (with the 0.3.2 bound). Then this PR defines a few functions double for Manopt, but that should be fine; afterwards bumping version to 0.3.3 and Manifolds <= this PR should also make Manopt work again.

mateuszbaran:
OK, that sounds like a good plan.

kellertuer:
It worked :) After this is merged and registered I can use the version number from here and Diff 0.3.3 as lower bounds in Manopt, for the next version.

mateuszbaran:
Great! So I think this PR can be merged now?

kellertuer:
Yes :)

PR #626, opened 2023-06-13
Add Exports, Improve GHA Caches.

kellertuer:
* introduce a few more exports
* update caching (that is what I want to check on this PR

It could maybe even be included in the next version that is nearly ready on master?


mateuszbaran (approved):
Yes, sure, I forgot to tag yesterday so we can merge this too before tagging.

kellertuer:
(Sorry had to retriever with the new tags). Sure will merge this once I see the new cache working correctly.

mateuszbaran:
Weird, IIRC docs for probability simplex worked for me locally, though I didn't `@ref` to the private functions.

Sure, we can add a new submenu for extensions, though I wouldn't like to wait with tagging a new version until that's ready.

kellertuer:
Sure I did not mean to do that here, and I fixed the references. It were those _within_ the new private functions that referred to other ones within Manifolds.
Also the `FisherRaoMetric` seems to be private? We could maybe export it. Also does not have to be here. I think this is now working and fine :)

mateuszbaran:
OK, thanks. I think we can export `FisherRaoMetric`.

mateuszbaran (approved):

PR #627, opened 2023-06-13
Fix change_metric, change_representer and project for ProbabilitySimplex

mateuszbaran:
This fixes the issue noticed in https://github.com/JuliaManifolds/Manifolds.jl/pull/604#issuecomment-1589290857 (and also projection).

### DIFF
--- a/src/manifolds/ProbabilitySimplex.jl
+++ b/src/manifolds/ProbabilitySimplex.jl
@@ -70,38 +70,29 @@ active_traits(f, ::ProbabilitySimplex, args...) = merge_traits(IsEmbeddedManifol
     change_representer(M::ProbabilitySimplex, ::EuclideanMetric, p, X)
 
 Given a tangent vector with respect to the metric from the embedding, the [`EuclideanMetric`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/manifolds.html#ManifoldsBase.EuclideanMetric),
-the representer of a linear functional on the tangent space is adapted as ``Z = p .* X``, since
-this “compensates” for the divsion by ``p`` in the Riemannian metric on the [`ProbabilitySimplex`](@ref).
+the representer of a linear functional on the tangent space is adapted as ``Z = p .* X .- p .* dot(p, X)``.
+The first part “compensates” for the divsion by ``p`` in the Riemannian metric on the [`ProbabilitySimplex`](@ref)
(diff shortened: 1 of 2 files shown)


src/manifolds/ProbabilitySimplex.jl
```diff
@@ -70,38 +70,29 @@ active_traits(f, ::ProbabilitySimplex, args...) = merge_traits(IsEmbeddedManifol
...
-
-and hence ``C(X)_i = X_i\sqrt{p_i}, i=1,…,n+1``.
+This can be achieved by rewriting representer change in matrix form as `(Diagonal(p) - p * p') * X`
+and taking square root of the matrix
 """
 change_metric(::ProbabilitySimplex, ::EuclideanMetric, ::Any, ::Any)
 
 function change_metric!(::ProbabilitySimplex, Y, ::EuclideanMetric, p, X)
-    return Y .= sqrt.(p) .* X
+    return Y .= sqrt(Diagonal(p) - p * p') * X
```
kellertuer, line 95: Here it would be nice to know whether this really holds. That the metric is preserved is clear to me (I think); since `X` (and `Y`) are tangents the last three summands of X'R'RY (R is the matrix we define here, and you can rotate ) should vanish. Whether the results here are tangent I do not see that fast.

mateuszbaran: A rough sketch: `w = [1, 1, ..., 1]` is an eigenvector of `R=Diagonal(p) - p * p'` with eigenvalue 0 because it's a projection onto the tangent space. So `w` is also an eigenvector of `sqrt(R)`, with the same eigenvalue. So `sqrt(R)` cannot add a component in the direction `w` to a vector it acts on.

kellertuer (approved):

src/manifolds/ProbabilitySimplex.jl
kellertuer: I think this should work. To me this looks like a projection matrix, so it is symmetric. I am not so sure its positive semidefinite – but then the square root is unique, see https://en.wikipedia.org/wiki/Square_root_of_a_matrix#Matrices_with_distinct_eigenvalues – further you can use the eigenvalue decomposition to compute this unique square root – and then the square roots of the eigenvalues of A are the eigenvalues of sqrt(A). This would keep your zero as an eigenvalue with the same eigenvector.

So I think (positive semidefiniteness left) this holds true.

PR #628, opened 2023-06-14
Towards integration on manifolds

mateuszbaran:
I've started some work on integration on manifolds, mostly with exponential-wrapped distributions and kernel density estimation in mind. It is primarily inspired by http://arxiv.org/abs/2009.01983 . @sethaxen I think the `volume_density` function may be interesting for you :slightly_smiling_face: . It connects integration on a manifold with integration in normal coordinates.


tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+    q = rand(M)
+    for i in 1:N
+        sum += f(M, q)
+        rand!(M, q)
+    end
+    return V * sum/N
+end
+```
+
+``[`manifold_volume`](@ref)``{=commonmark} and ``[`volume_density`](@ref)``{=commonmark} are closely related to each other, though very few sources explore this connection, and some even claiming a certain level of arbitrariness in defining `manifold_volume`.
```
kellertuer, line 47: This to mea reads like a very abrupt change of topic? Maybe introduce this a bit, maybe as, “We used the function ... which is related to”?

```suggestion
[`manifold_volume`](@ref) and [`volume_density`](@ref) are closely related to each other, though very few sources explore this connection, and some even claiming a certain level of arbitrariness in defining [`manifold_volume`](@ref).
```
the commonmark (and = to keep the code unchanged) is only necessary of there are spaces in the `(@ref)` part, since otherwise these are escaped in an HTML manner (i.e. to `%20`).

tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+Volume is sometimes considered arbitrary because Riemannian metrics on some spaces like the manifold of rotations are defined with arbitrary constants.
+However, once a constant is picked (and it must be picked before any useful computation can be performed), all geometric operations must follow in a consistent way: inner products, exponential and logarithmic maps, volume densities, etc.
+`Manifolds.jl` consistently picks such constants and provides a unified framework, though it sometimes results in picking a different constant than what is the most popular in some sub-communities.
+
+On Lie groups the situation is more complicated. Invariance under left or right group action is a desired property that leads one to consider Haar measures [^Tornier2020].
+It is, however, unclear what are the practical benefits of considering Haar measures over the Lebesgue measure of the underlying manifold, which often turns out to be invariant anyway.
+
+Last point before we turn to consider an example is the matter of integration through charts.
+This is an approach not currently supported by `Manifolds.jl`.
```
kellertuer, line 56:
```suggestion
This is an approach currently not supported by `Manifolds.jl`.
```

tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+
+## Distributions
+
+We will now try to verify that volume density correction correctly changes probability density of an exponential-wrapped normal distribution.
+`pdf_tangent_space` represents probability density of a normally distributed random variable $X_T$ in the tangent space $T_p \mathcal{M}$.
+`pdf_manifold` refers to the probability density of the distribution $X_M$ from the tangent space $T_p \mathcal{M}$ wrapped using exponential map on the manifold.
+``[`volume_density`](@ref)``{=commonmark} function calculates the necessary correction.
+
+Note that our simplified `pdf_manifold` assumes that the probability mass of `pdf_tangent_space` outside of (local) injectivity radius is negligible.
+If it was not, a summation over vectors in the tangent plane pointing at `q` would have to be performed.
```
kellertuer, line 69: This paragraph I do not understand, neither the “pointing at `q`” nor what you mean in general.

tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+    Xc = get_coordinates(M, p, X, DefaultOrthonormalBasis())
+    vd = abs(volume_density(M, p, X))
+    if vd > eps()
+        return pdf_tangent_space(M, Xc) / vd
+    else
+        return 0.0
+    end
+end
+
+println(simple_mc_integrate(Sphere(2), pdf_manifold; N=1000000))
```
kellertuer, line 88: Maybe mention where the `simple_mc_integrate` function is from?

tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+
+On Lie groups the situation is more complicated. Invariance under left or right group action is a desired property that leads one to consider Haar measures [^Tornier2020].
+It is, however, unclear what are the practical benefits of considering Haar measures over the Lebesgue measure of the underlying manifold, which often turns out to be invariant anyway.
+
+Last point before we turn to consider an example is the matter of integration through charts.
+This is an approach not currently supported by `Manifolds.jl`.
+One has to define a suitable set of disjoint charts covering the entire manifold and use a method for multivariate Euclidean integration.
+Note that ranges of parameters have to be adjusted for each manifold and scaling based on the metric needs to be applied.
+See [^BoyaSudarshanTilma2003] for some considerations on symmetric spaces.
+
```
kellertuer, line 60: For me the block from the last code block until here is a lot of technical detail what manifolds does and does not provide, that I do not completely unterstand. Could we maybe move this further down so that the tutorial first illustrates how Integration can be used (i.e. the distributions section) before discussion these details?

kellertuer (commented):
Here is a few comments.

tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+using Markdown
+```
+
+This part of documentation covers integration of scalar functions defined on manifolds.
+The basic concepts are derived from geometric measure theory.
+In principle, there are many ways in which a manifold can be equipped with a measure that can be later used to define an integral.
+One of the most popular ways is based on pushing the Lebesgue measure on a tangent space through the exponential map.
+Any other suitable atlas could be used, not just the one defined by normal coordinates, though each one requires different volume density corrections due to the Jacobian determinant of the pushforward.
+`Manifolds.jl` provides the function ``[`volume_density`](@ref)``{=commonmark} that calculates that quantity, denoted $\theta_p(X)$.
+See for example [^leBrigantPuechmorel2019], Definition 11, for a precise description using Jacobi fields.
```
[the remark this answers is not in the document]
mateuszbaran: I've tried the commonmark thing but IMO on its own it makes things worse: bibliography is still numbered instead of by key. `[1]` seems perfectly usable here and I'd prefer to leave to you updating bibliography here similarly to Manopt.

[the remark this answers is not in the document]
kellertuer: I fear I am too lazy to update your bibliography ;) 

My only sad point is that then this page is very different from the rest of the docs, but still in style close (with the footnotes).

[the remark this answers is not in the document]
mateuszbaran: Yes, it doesn't look great but I hope maybe at some point there will be an easier way to manage bibliography so I'm not particularly motivated to fight with it now.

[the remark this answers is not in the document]
kellertuer: Ys a proper way to handle bibliography in Documenter would be very welcome. Ok let's keep it like that for now.

tutorials/integration.qmd
```diff
@@ -0,0 +1,171 @@
...
+#| code-fold: true
+#| output: false
+using Pkg;
+cd(@__DIR__)
+Pkg.activate("."); # for reproducibility use the local tutorial environment.
+Pkg.develop(PackageSpec(; path=(@__DIR__) * "/../"))
+using Markdown
+```
+
+This part of documentation covers integration of scalar functions defined on manifolds.
```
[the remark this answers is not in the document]
mateuszbaran: OK, done.

tutorials/integration.qmd
mateuszbaran: Good point, I've moved that part to a separate section at the end. I've also removed the `commonmark` thing where it looked unnecessary.

tutorials/integration.qmd
mateuszbaran: I've added some details to it, what do you think?

tutorials/integration.qmd
mateuszbaran: Yes, that is a bit too technical. I've moved that paragraph to a new section for technical details.

tutorials/integration.qmd
```diff
@@ -0,0 +1,198 @@
...
+    q = rand(M)
+    for i in 1:N
+        sum += f(M, q)
+        rand!(M, q)
+    end
+    return V * sum/N
+end
+```
+
+We used the function [`manifold_volume`](@ref) to get the volume of the set over which integration is performed, as described in the linked Wikipedia article.
```
kellertuer, line 52:
```suggestion
We used the function [`manifold_volume`](@ref) to get the volume of the set over which the integration is performed, as described in the linked Wikipedia article.
```

tutorials/integration.qmd
```diff
@@ -0,0 +1,198 @@
...
+
+## Distributions
+
+We will now try to verify that volume density correction correctly changes probability density of an exponential-wrapped normal distribution.
+`pdf_tangent_space` represents probability density of a normally distributed random variable $X_T$ in the tangent space $T_p \mathcal{M}$.
```
kellertuer, line 57: Could you clarify here that `pdf_tangent_space` is a function that you plan to define (and do later)? same for `pdf_manifold`.

kellertuer (approved):
Here is two more remarks.

In general this is missing 4 lines of coverage, that would be nice to check.

I do not understand the theory here completely, but noticed that already quite a while ago I checked most files as “viewed” - so I think past-me found them ok.

So besides test coverage, this is fine I think.

tutorials/integration.qmd
mateuszbaran: Sure, done.

PR #629, opened 2023-06-15
SPD Fixed Determinant (Isochoric Matrices)

kellertuer:
This came as a small idea this morning and might be useful for someone, but it is indeed just a split of the SPDs into “constant determinant shells”.

Still a nice short thing to do.


mateuszbaran:
Nice! One small observation, in many many places you have "fixed rank" instead of "fixed determinant".

docs/src/manifolds/symmetricpositivedefinitefixedrank.md
```diff
@@ -0,0 +1,15 @@
+# [Symmetric positive definite matrices of fixed deretminant](@id SymmetricPositiveDefiniteFixedDeterminantSection)
```
mateuszbaran, line 1:
```suggestion
# [Symmetric positive definite matrices of fixed determinant](@id SymmetricPositiveDefiniteFixedDeterminantSection)
```

kellertuer:
> Nice! One small observation, in many many places you have "fixed rank" instead of "fixed determinant".

oops, yes I always saw the file on the side and was maybe confused by that every now and then

docs/src/manifolds/symmetricpositivedefinitefixeddeterminant.md
```diff
@@ -0,0 +1,15 @@
+# [Symmetric positive definite matrices of fixed deretminant](@id SymmetricPositiveDefiniteFixedDeterminantSection)
```
mateuszbaran, line 1:
```suggestion
# [Symmetric positive definite matrices of fixed determinant](@id SymmetricPositiveDefiniteFixedDeterminantSection)
```

mateuszbaran:
Sorry, I didn't mean to write "many" twice :sweat_smile: .

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,155 @@
...
+Additionally we store the tangent vectors as `X=p^{-1}Z`, i.e. symmetric matrices.
+
+# Constructor
+
+    SymmetricPositiveDefiniteFixedDeterminant(n, d=1.0)
+
+generates the manifold $\mathcal P_d(n) \subset \mathcal P(n)$ of determinant ``d``,
+which defaults to 1.
+"""
+struct SymmetricPositiveDefiniteFixedDeterminant{N,D} <: AbstractDecoratorManifold{ℝ} end
```
mateuszbaran, line 33:
```suggestion
struct SymmetricPositiveDefiniteFixedDeterminant{N,TD<:Real} <: AbstractDecoratorManifold{ℝ}
    d::TD
end
```
I think determinant should be just a field.

kellertuer:
> Sorry, I didn't mean to write "many" twice 😅 .

That‘s completely fine, I also did not mean to write rank 😅

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
kellertuer: Yes, you might be right, there is not much reason to ever dispatch on that.

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+with determinant ``\det(p) = d``.
+
+The tolerance for the determinant of `p` can be set using `kwargs...`.
+"""
+function check_point(
+    M::SymmetricPositiveDefiniteFixedDeterminant{n},
+    p;
+    kwargs...,
+) where {n}
+    if det(p) != M.d
```
mateuszbaran, line 61:
```suggestion
    if det(p) ≉ M.d
```
I think approximate equality should be sufficient.

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+        )
+    end
+    return nothing
+end
+
+"""
+    check_vector(M::SymmetricPositiveDefiniteFixedDeterminant{n}, p, X; kwargs... )
+
+Check whether `X` is a tangent vector to manifold point `p` on the
+[`SymmetricMatrices`](@ref) `M`, i.e. `X` has to be a symmetric matrix of size `(n,n)`
```
mateuszbaran, line 74:
```suggestion
[`SymmetricPositiveDefiniteFixedDeterminant`](@ref) `M`, i.e. `X` has to be a symmetric matrix of size `(n,n)`
```

mateuszbaran: The rest of the docstring doesn't seem to correspond to what is actually tested here.

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+and its values have to be from the correct [`AbstractNumbers`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/types.html#number-system).
+
+The tolerance for the symmetry of `X` can be set using `kwargs...`.
+"""
+function check_vector(M::SymmetricPositiveDefiniteFixedDeterminant, p, X; kwargs...)
+    if !isapprox(tr(X), 0.0; kwargs...)
```
mateuszbaran, line 80:
```suggestion
function check_vector(M::SymmetricPositiveDefiniteFixedDeterminant, p, X; atol::Real=sqrt(max_eps(X)), kwargs...)
    if !isapprox(tr(X), 0.0; atol=atol, kwargs...)
```
So that by default it's not an exact comparison with `0`.

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+````
+"""
+
+function manifold_dimension(M::SymmetricPositiveDefiniteFixedDeterminant)
+    return manifold_dimension(get_embedding(M)) - 1
+end
+
+@doc raw"""
+    q = project(M::SymmetricPositiveDefiniteFixedDeterminant{n,d}, p)
+    project!(M::SymmetricPositiveDefiniteFixedDeterminant{n,d}, q, p)
```
mateuszbaran, line 115:
```suggestion
    q = project(M::SymmetricPositiveDefiniteFixedDeterminant{n}, p)
    project!(M::SymmetricPositiveDefiniteFixedDeterminant{n}, q, p)
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+
+Project the symmetric positive definite (s.p.d.) matrix `p` from the embedding onto the
+(sub-)manifold of s.p.d. matrices of determinant d (in place of q).
```
mateuszbaran, line 118:
```suggestion
(sub-)manifold of s.p.d. matrices of determinant `M.d` (in place of `q`).
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+project(M::SymmetricPositiveDefiniteFixedDeterminant, p)
+
+function project!(M::SymmetricPositiveDefiniteFixedDeterminant{n}, q, p) where {n}
+    q .= (M.d / det(p))^(1 / n) .* p
+    return
+end
+
+@doc raw"""
+    Y = project(M::SymmetricPositiveDefiniteFixedDeterminant{n,d}, p, X)
+    project!(M::SymmetricPositiveDefiniteFixedDeterminant{n,d}, Y, p, X)
```
mateuszbaran, line 135:
```suggestion
    Y = project(M::SymmetricPositiveDefiniteFixedDeterminant{n}, p, X)
    project!(M::SymmetricPositiveDefiniteFixedDeterminant{n}, Y, p, X)
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+
+Project the symmetric matrix `X` onto the tangent space at `p` of the
+(sub-)manifold of s.p.d. matrices of determinant d (in place of `Y`),
```
mateuszbaran, line 138:
```suggestion
(sub-)manifold of s.p.d. matrices of determinant `M.d` (in place of `Y`),
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+by setting its diagonal (and hence its trace) to zero.
+
+"""
+project(M::SymmetricPositiveDefiniteFixedDeterminant, p, X)
+
+function project!(M::SymmetricPositiveDefiniteFixedDeterminant, Y, p, X)
+    copyto!(M, Y, p, X)
+    Y[diagind(Y)] .= 0.0
```
mateuszbaran, line 146:
```suggestion
    fill!(Y[diagind(Y)], 0)
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+
+The tangent space is modelled the same as for [`SymmetricPositiveDefinite`](@ref)`(n)`
+and consists of all symmetric matrices with zero trace
+```math
+    T_p\mathcal P_d(n) =
+    \bigl\{
+        X \in \mathbb R^{n×n} \big|\ X=X^\mathrm{T} \text{ and } \operatorname{tr}(p) = 0
+    \bigr\},
+```
+since for a constant determinant we require that `0 = D\det(p)[Z] = \det(p)\operatorname{tr}(p^{-1}Z)` for all tangent vectprs ``Z``.
```
mateuszbaran, line 23:
```suggestion
since for a constant determinant we require that `0 = D\det(p)[Z] = \det(p)\operatorname{tr}(p^{-1}Z)` for all tangent vectors ``Z``.
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
```diff
@@ -0,0 +1,152 @@
...
+
+function active_traits(f, ::SymmetricPositiveDefiniteFixedDeterminant, args...)
+    return merge_traits(IsEmbeddedSubmanifold())
+end
+
+@doc raw"""
+    check_point(M::SymmetricPositiveDefiniteFixedDeterminant{n}, p; kwargs...)
+
+Check whether `p` is a valid manifold point on the [`SymmetricPositiveDefiniteFixedDeterminant`](@ref)`(n,d)` `M`, i.e.
+whether `p` is a [`SymmetricPositiveDefinite`](@ref) matrix (checked by traits) ofsize `(n,n)`
```
mateuszbaran, line 50:
```suggestion
whether `p` is a [`SymmetricPositiveDefinite`](@ref) matrix (checked by traits) of size `(n, n)`
```

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
kellertuer: Oh because I forgot to update that. Sorry, I thought it was a quick fix but it seems I missed several things.

src/manifolds/SymmetricPositiveDefiniteFixedDeterminant.jl
kellertuer: hm, until now we never did that.

mateuszbaran: We have that in a few places, for example `KendallsShapeSpace`. I think we should move in this direction for other manifolds as well.

kellertuer: hm until now at least my philosophy was: The default from is approx should be fine enough

mateuszbaran: The default is enough when we don't compare to 0. When we compare to 0, we should IMO have a non-zero atol by default.

kellertuer: Hm, then is approx has a strange default for that case. I would prefer to maybe have one PR that changes the default behaviour of Manifolds.jl when we compare approximately vs zero overall? Otherwise this would just get a bit inconsistent throughout the package for now.

mateuszbaran: It's already inconsistent but if you prefer we can leave fixing that for later.

kellertuer:
I think its maybe best to stop for today, now project is broken and I do not see why, maybe your fill does not what you thought it should?

kellertuer:
Just learned these matrices are called Isochronic; maybe that is a nicer name than, but have to check that again.

mateuszbaran:
Ah, it doesn't because I forgot `view`. It should be `fill!(view(Y, diagind(Y)), 0)`

mateuszbaran:
We can of course finish this tomorrow.

kellertuer:
I switched that again; Well I want to think about the name, this is quite long, but isochoric would just be the volume preserving case `d=1`. So some thoughts about the name might be nice; and maybe IsochoricMatrices is still fine since the default is d=1.

mateuszbaran:
Do you have a source where such matrices are called isochoric?

kellertuer:
Until now I only found the systems, we had that in a note we worked with this morning, but I have to check for a source whether the matrices are then also called that.

And then we have to decide whether fixed determinant is fine as well.

kellertuer:
Yes it only refers to volume preservation, so determinant 1 https://en.m.wiktionary.org/wiki/isochoric#English; so then the interpretation question is whether it's fine to have a unified volume change as well (det != 1 but fixed).

mateuszbaran:
OK then let's call it IsochoricMatrices.

kellertuer:
I hope in renaming I no-where stumbled into isotropic or isochronic. ;)

sethaxen:
An alternative to avoid choosing a non-standard name that maybe 1 or 2 people will understand would be to add the alias `SPD` for `SymmetricPositiveDefinite` and then define `SPDFixedDeterminant`, which is not so long, and now the name is understandable by everyone.

mateuszbaran:
Hm, actually `SPDFixedDeterminant` seems more understandable to me than `IsochoricMatrices`.

kellertuer:
That's a bit sad, I was quite happy finding something short that is not an abbreviation – since for now we avoided to abbreviate manifolds (sure we have `SVDPoint` and `SPDPoint`), and I feel the property “volume preserving” (a bithidden in greek) is nicer that describing the equation that is added.

But sure, if I am in the minority I can change the name (back).

src/manifolds/SPDFixedDeterminant.jl
```diff
@@ -0,0 +1,153 @@
...
+\mathcal P_d(n) =
+\bigl\{
+p ∈ ℝ^{n × n} \ \big|\ a^\mathrm{T}pa > 0 \text{ for all } a ∈ ℝ^{n}\backslash\{0\}
+  \text{ and } \det(p) = d
+\bigr\}.
+````
+
+This manifold is modelled as a submanifold of [`SymmetricPositiveDefinite`](@ref)`(n)`.
+
+These matrices are sometimes also called [isochoric](https://en.wiktionary.org/wiki/isochoric) refers to the interpretation of
```
mateuszbaran, line 16:
```suggestion
These matrices are sometimes also called [isochoric](https://en.wiktionary.org/wiki/isochoric), which refers to the interpretation of
```

src/manifolds/SPDFixedDeterminant.jl
```diff
@@ -0,0 +1,153 @@
...
+the matrix representing an ellipsoid. All ellipsoids on this manifold have the same volume.
```
mateuszbaran, line 17:
```suggestion
the matrix representing an ellipsoid. All ellipsoids that represent points on this manifold have the same volume.
```

mateuszbaran (approved):

kellertuer:
I think this is not urgent so it can also wait for the others before we merge (or we merge it but wait before registering a new version).

PR #631, opened 2023-06-16
Stable hyperbolic functions

hajg-ijk:
Replaces distance, exponential, and logarithmic maps to more numerically stable versions, as per the Matlab Manopt version.

### DIFF
--- a/accuracy/Hyperboloid.qmd
+++ b/accuracy/Hyperboloid.qmd
@@ -1,6 +1,5 @@
 ---
 title: Accurcay tests on the Hyperboloid
-format: html
 ---
 
 
@@ -38,8 +37,9 @@ p = [0.0, 0.0, 0.1]
 pb = BigFloat.(p)
 X = [1 / sqrt(2), 1 / sqrt(2), 0.0]
 Xb = BigFloat.(X)
-lengths = [10.0^k for k in -16:1]
+lengths = [10.0^k for k in -16:4]
 Xbs = [s*Xb for s in lengths]
+Xbns = [norm(M, p, X) for X in Xbs]
 qbs = [exp(M, pb, Y) for Y in Xbs]
 Xs = [Float64.(X) for X in Xbs]
 qs = [Float64.(q) for q in qbs]
@@ -62,26 +62,25 @@ end
 and compare
 
 ```{julia}
-d_comp = [distance(M, pb, q) for q in qbs]
 d_1 = [distance_1(p, q) for q in qs]
 d_2 = [distance_2(p, q) for q in qs]
 fig = plot(;
-    legend=:topright,
+    legend=:topleft,
     xlabel=raw"lengths $s$ (log. scale)",
     ylabel=raw"Error to BigFloat (log. scale)",
     yaxis=:log,
     xaxis=:log,
     ylims=(1e-16, 1e2),
-    xlims=(1e-17, 10^1),
+    xlims=(1e-17, 10^3),
 )
-plot!(fig, lengths, abs.(d_comp .- d_1), color=indigo, label="1 - acosh-based")
-plot!(fig, lengths, abs.(d_comp .- d_2), color=teal, label="2 - sinh-based")
+plot!(fig, lengths, abs.(Xbns .- d_1), color=indigo, label="1 - acosh-based")
+plot!(fig, lengths, abs.(Xbns .- d_2), color=teal, label="2 - asinh-based")
 ```
 
 ```{julia}
-abs.(d_comp .- d_1)
+abs.(Xbns .- d_1)
 ```
 
 ```{julia}
-abs.(d_comp .- d_2)
+abs.(Xbns .- d_2)
 ```
--- a/accuracy/Hyperboloid.qmd
+++ b/accuracy/Hyperboloid.qmd
@@ -33,7 +33,7 @@ We choose a “base point”, a direction and create tangent vectors and correcp
 ```{julia}
 #| output: false
 M = Hyperbolic(2)
-p = [0.0, 0.0, 0.1]
+p = [0.0, 0.0, 1.0]
 pb = BigFloat.(p)
 X = [1 / sqrt(2), 1 / sqrt(2), 0.0]
 Xb = BigFloat.(X)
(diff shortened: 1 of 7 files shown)


mateuszbaran:
Improving accuracy for `Hyperbolic` would be a good idea. Could you make a comparison of accuracy of the old and new versions? For example on `Hyperbolic(2)` and `Hyperbolic(10)` using some random points. You can compute a reference using `BigFloat`.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -242,8 +246,8 @@ function exp!(M::Hyperbolic, q, p, X, t::Number)
 end
 function exp!(M::Hyperbolic, q, p, X)
     vn = sqrt(max(inner(M, p, X, X), 0.0))
-    vn < eps(eltype(p)) && return copyto!(q, p)
-    q .= cosh(vn) .* p .+ (sinh(vn) / vn) .* X
+    sn = vn == 0.0 ? 1.0 : sinh(vn) / vn
```
mateuszbaran, line 249:
```suggestion
    sn = vn == 0 ? one(vn) : sinh(vn) / vn
```
In case `vn` is of a different type.

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -226,7 +226,11 @@ d_{\mathcal H^n}(p,q) = \operatorname{acosh}( - ⟨p, q⟩_{\mathrm{M}}),
 where $⟨\cdot,\cdot⟩_{\mathrm{M}}$ denotes the [`MinkowskiMetric`](@ref) on the embedding,
 the [`Lorentz`](@ref)ian manifold.
 """
-distance(::Hyperbolic, p, q) = acosh(max(-minkowski_metric(p, q), 1.0))
+function distance(::Hyperbolic, p, q)
+    w = q - p
+    m = sqrt(max(0.0, minkowski_metric(w, w)))
+    return 2 * asinh(0.5 * m)
```
mateuszbaran, line 232:
```suggestion
    return 2 * asinh(m / 2)
```
Minor point: `m / 2` doesn't force promotion to `Float64`

src/manifolds/HyperbolicHyperboloid.jl
```diff
@@ -374,11 +378,10 @@ This employs the metric of the embedding, see [`Lorentz`](@ref) space.
...
 
 function log!(M::Hyperbolic, X, p, q)
-    scp = minkowski_metric(p, q)
-    w = q + scp * p
-    wn = sqrt(max(scp .^ 2 - 1, zero(scp)))
-    wn < eps(eltype(p)) && return zero_vector!(M, X, p)
-    X .= acosh(max(one(scp), -scp)) / wn .* w
+    d = distance(M, p, q)
+    s = sinh(d)
+    w = s == 0.0 ? 1.0 : d / s
```
mateuszbaran, line 383:
```suggestion
    w = s == 0 ? one(s) : d / s
```

kellertuer:
We also noticed that parallel transport might be more stable in a different implementation.

Actually the ideas come from Manopt/Matlab, where they are documented as more stable – but then where would we put such an accuracy comparison? A small quarto where we define and compare both variants?

kellertuer:
Thanks for your remarks Mateusz :)

I hacked a small start of a comparison; I can tell from that, that both implementations yield different results, but since I use the old one for the BigFloat, the old one has a smaller error than the new one. So I am not so sure what to compare to.

kellertuer:
I found the good comparison for dist - comparing to the vector norm (which does not need asinh/acosh). The image rendered in the cmd is in the end
Screenshot 2023-06-18 at 14 51 44

And I would assume that exp/log are similar when switching from asinh to acosh. I again would have to check how to best evaluate that, but maybe distance is enough to validate the switch here.

One can also see that beyond (Euclidean) $\lVert X \rVert > 1$ asinh is a bit better for short before running into a Nan-overflow. Acosh is for short worse but still works for (Euclidean) vector norms beyond 10.

accuracy/Hyperboloid.qmd
```diff
@@ -0,0 +1,86 @@
...
+teal = paul_tol["mutedteal"]
+```
+
+## Reference Data
+
+We choose a “base point”, a direction and create tangent vectors and correcponding points in several magnutudes of distances
+```{julia}
+#| output: false
+M = Hyperbolic(2)
+p = [0.0, 0.0, 0.1]
```
mateuszbaran, line 36: This is not a valid point on `M`.

mateuszbaran: I've re-run tests with ` p = [0.0, 0.0, 1.0]` and it looks like the larger error of `distance_2` for larger `Xs` coincides with `qs` being inaccurate themselves (`check_point` returns a huge error). So I'd say we can just use `distance_2`.

kellertuer: ups. Yes, That is what I concluded when I first compared to distance with BigFloat and not with norm. Note that the exp is also the old (asinh) one not the new acosh).

kellertuer: I'd also conclude that the new distance (2) is the one we should use, but it was still good to check that.

mateuszbaran: Yes, I think we should go with this change. Now we are sure this new approach is more accurate.

kellertuer: Hajg has already also adapted PT, maybe just not yet pushed here and I will already now accept your changes :)

mateuszbaran (approved):

kellertuer:
Hm, something broke on CI unrelated to this PR I think? See https://github.com/JuliaManifolds/Manifolds.jl/actions/runs/5304573985/jobs/9600993369?pr=631

mateuszbaran:
OrdinaryDiffEq precompilation fails, that looks unrelated to Manifolds.jl.

hajg-ijk:
Nightly checks are failing. Is it still okay to squash and merge or not?

mateuszbaran:
Yes, it's perfectly fine to squash and merge with nightly failures. I like to checked them once in a few months to see if there is something to report but that's all.

kellertuer (approved):

PR #634, opened 2023-06-21
Add injectivity radius on fixed rank matrices

mateuszbaran:
I've found a note that says it is 0. I'm going to make a PR to Manopt later that makes `max_stepsize` return 1 when injectivity radius is 0 to make algorithms usable with this addition.

### DIFF
--- a/src/manifolds/FixedRankMatrices.jl
+++ b/src/manifolds/FixedRankMatrices.jl
@@ -352,6 +352,21 @@ end
 
 get_embedding(::FixedRankMatrices{m,n,k,𝔽}) where {m,n,k,𝔽} = Euclidean(m, n; field=𝔽)
 
+"""
+    injectivity_radius(::FixedRankMatrices)
+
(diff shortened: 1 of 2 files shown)


kellertuer:
Whil I do not have read the paper, it should be locally for some points larger than that, right? We could use that to specify the infectivity radius for single points?

mateuszbaran:
That paper doesn't give the expression for local injectivity radius and I couldn't find a better reference. So I don't know.

kellertuer:
Hm, I remember a formula that was not Malays zero (depending on the point) but surely its infimum was. I will look for that .

kellertuer (approved):
I could not find a good reference for the local injectivity radii I had in mind, so for now let‘s just do what we have here.

mateuszbaran:
OK. With https://github.com/JuliaManifolds/Manopt.jl/pull/275 it should be a safe thing to do.

PR #638, opened 2023-06-30
Move tutorials to Quarto

mateuszbaran:
I've moved hand gesture tutorial to Quarto, I think it makes it easier to discover. And Quarto is nice.

I'm thinking about moving the chart/atlas tutorials but then we would lose all interactivity.


docs/Project.toml
```diff
@@ -10,6 +10,7 @@ HybridArrays = "1baab800-613f-4b0a-84e4-9cd3431bfbb9"
 IJulia = "7073ff75-c697-5162-941a-fcdaad2a7d2a"
 Manifolds = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 ManifoldsBase = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
+MultivariateStats = "6f286f6a-111f-5878-ab1e-185364afe411"
```
kellertuer, line 13: I think this should be added to the `tutorials` environment and not to the docs (yes yes we have 3 environments by now, but I think its better to split that).

kellertuer: Ah and CSV seems to be missing from the `tutorials/` project toml as well it seems.

mateuszbaran: I see, I was confused where to put that.

kellertuer: No problem. The idea is that the quarto part is a bit encapsulated besides the docs. I am not even 100% sure IJulia is needed in the docs, it should not.
Splitting docs and tutorials maybe makes the two environments a bit nicer themselves, less dependencies.

docs/make.jl
```diff
@@ -17,7 +17,7 @@ if Base.active_project() != joinpath(@__DIR__, "Project.toml")
 end
 
 # (b) Did someone say render? Then we render!
-if "--quarto" ∈ ARGS
+if true || "--quarto" ∈ ARGS
```
kellertuer, line 20: What does this approve? This means quarto is always run.

If you want to run this locally, the make.jl is executable and on terminal you can just use `docs/make.jl --quarto`.

mateuszbaran: Sorry, I just meant to use it locally. I run basically everything through VSCode's "execute file in REPL" command so it's easier to just add a hack like this.

kellertuer: ah ok, sure if I would know a nicer way we could add that as well, for now this allows to disable quarto un by default and only activate it if necessary. You can also just call `quarto render file.qmd` or even the quarto render shortcut in VS Code.

mateuszbaran:
Hm, GLMakie.jl doesn't work on CI. @kellertuer , what do you think about rendering images locally and uploading them? How should I do that?

kellertuer:
I think there are two possibilities, either pretender just the one image and load it in the file _or_ `quarto render` the whole example, commit the `md` file and the rendered images.

I would prefer the second I think, and do the same with two tutorials in Manopt.jl as well, you just have to exclude them then here in the `_quarto.yml`, see https://github.com/JuliaManifolds/Manopt.jl/blob/master/tutorials/_quarto.yml

You can also see that both markdown files and some images are actually committed https://github.com/JuliaManifolds/Manopt.jl/tree/master/docs/src/tutorials

For the regression example the reason is the same as yours, we do not have Asymptote on CI, the other one just runs too long on CI.

mateuszbaran:
I went with the first option, it looks a bit easier.

kellertuer:
I will try to find time to review this tomorrow. At first glance: You created two quarto notebooks but only deleted one old jl file. What about the other one?

mateuszbaran:
I haven't deleted the "working in charts" .jl file because it still has some interactivity parts that couldn't be ported to the Quarto version.

tutorials/hand-gestures.qmd
```diff
@@ -0,0 +1,114 @@
...
+    cur_hand_logs = red_coords[:, mask]
+    cur_t = MultivariateStats.transform(fp, cur_hand_logs)
+    scatter!(fig, cur_t[1, :], cur_t[2, :], label="gesture " * string(label_num))
+end
+xlabel!(fig, "principal component 1")
+ylabel!(fig, "principal component 2")
+fig
+```
+
+Now let's compute pairwise distances between gestures
```
kellertuer, line 106: You start a lot of sentences with also, now, ... – maybe leave them out?

tutorials/working-in-charts.qmd
```diff
@@ -0,0 +1,179 @@
...
+#| echo: false
+#| code-fold: true
+#| output: false
+using Pkg;
+cd(@__DIR__)
+Pkg.activate("."); # for reproducibility use the local tutorial environment.
+using Markdown
+```
+
+## Working in Charts: the embedded Torus
```
kellertuer, line 17:
```suggestion
```

I think this was the old title?

tutorials/hand-gestures.qmd
```diff
@@ -0,0 +1,114 @@
...
+using Markdown
+```
+
+
+```{julia}
+using Manifolds
+using CSV
+using DataFrames
+using Plots
+using MultivariateStats
```
kellertuer, line 23:
```suggestion
using Manifolds, CSV, DataFrames, Plots, MultivariateStats
```

just for unification with the other two tutorials

tutorials/working-in-charts.qmd
```diff
@@ -0,0 +1,179 @@
...
+bvp_a2 = [θ₂, φ₂]
+bvp_sol = Manifolds.solve_chart_log_bvp(M, bvp_a1, bvp_a2, A, bvp_i);
+# geo_r = [Point3f(get_point(M, A, bvp_i, p[1:2])) for p in bvp_sol(0.0:0.05:1.0)]
+
+# ax2, fig2 = torus_figure()
+# lines!(geo_r; linewidth=4.0, color=:green)
+# fig2
+```
+
+![fig-geodesic](working-in-charts/working-in-charts-geodesic.png)
```
kellertuer, line 179: You could add a final paragraph that mentions the (old) `.jl` file which is an interactive (Pluto notebook) version of this tutorial?

kellertuer (commented):
Here are a few comments, overall – nice to have them in the new format. I hope we get a few more somewhen (maybe one on metrics, one on Lie,...)

tutorials/hand-gestures.qmd
```diff
@@ -0,0 +1,114 @@
...
+
```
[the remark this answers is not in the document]
mateuszbaran: Sure, I can move that.

tutorials/hand-gestures.qmd
mateuszbaran: I can leave them out but then many sentences will just start with "let's". I'm not sure if that's better.

kellertuer: Then maybe rephrase them also a bit to avoid the repetition?

tutorials/Project.toml
```diff
@@ -1,8 +1,16 @@
...
 IJulia = "7073ff75-c697-5162-941a-fcdaad2a7d2a"
 Manifolds = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 ManifoldsBase = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
+MultivariateStats = "6f286f6a-111f-5878-ab1e-185364afe411"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 
 [compat]
```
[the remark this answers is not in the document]
mateuszbaran: Sure, I think the relevant dependencies are `CSV`, `DataFrames` and `MultivariateStats`.

mateuszbaran:
> Here are a few comments, overall – nice to have them in the new format. I hope we get a few more somewhen (maybe one on metrics, one on Lie,...)

Yes, Quarto looks nice. I think I will keep using it for new tutorials.

kellertuer:
In VS Code there is also a live html preview (Live as in re-render per save action, not as live as Pluto). Then writing these tutorials is very nice – and I feel a bit better fitting than Pluto even.

tutorials/Project.toml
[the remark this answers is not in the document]
kellertuer: If you could do it for all, that would make sure that the tutorials basically never break even with updates of the packages.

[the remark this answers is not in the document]
mateuszbaran: I think other dependencies are constrained transitively by Manifolds.jl anyway?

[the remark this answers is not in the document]
kellertuer: Hm, I am not sure about that.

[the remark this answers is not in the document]
mateuszbaran: I'd like to avoid putting too much here because I'm afraid we will just forget to update this file, and CompatHelper doesn't seem to work for such Project.toml files: https://github.com/JuliaRegistries/CompatHelper.jl/issues/408 .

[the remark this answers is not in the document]
kellertuer: We might have to think about those ourselves, yes. I 'd like to make at least sure that the tutorials – they are cached and will not run that often – suddenly fail to run, because we forgot a compat.
Also with the cache we use for the tutorials CompatHelper would not note that as well.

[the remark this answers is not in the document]
mateuszbaran: Tutorials can also fail to run because we they have obsolete compat entries. Unless we want to render tutorials with older versions of Manifolds.jl/ManifoldsBase.jl than the newest one? That doesn't seem ideal either. I think ideally we should have a loud and as early as possible error when tutorials stop working with the latest Manifolds.jl. I would prefer to avoid merging PRs that break tutorials.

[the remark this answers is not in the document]
kellertuer: That is manually done by having Manifolds in dev mode from the parent folder when setting everything up in the make.jl, so that we run on the most recent one, yes.
Hm, now I am unsure, because what I want to avoid is that someone wants to contribute and their PR fails because some tutorial fails (maybe they fixed a typo in there). But maybe there is no real solution to this.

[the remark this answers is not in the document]
mateuszbaran: > Hm, now I am unsure, because what I want to avoid is that someone wants to contribute and their PR fails because some tutorial fails (maybe they fixed a typo in there). But maybe there is no real solution to this.

The only solution to this problem is specifying exact versions of dependencies in Project.toml. Then nothing can break when someone contributes a typo fix. When we allow "non-breaking" updates of dependencies, CI can randomly break. And sometimes it does. I don't think that has ever been a big issue? We either add an upper bound to the dependency ourselves or wait a bit until it's fixed.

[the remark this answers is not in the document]
kellertuer: Well, ok, then let's leave it as is.

tutorials/working-in-charts.qmd
```diff
@@ -0,0 +1,180 @@
...
+
+An interactive Pluto version of this tutorial is available in file `tutorials/working-in-charts.jl`.
```
kellertuer, line 180: One could link this to a generic GitHub (`blob`)  link of the file?

mateuszbaran: Sure, done :slightly_smiling_face:

PR #639, opened 2023-07-01
Make `switch_direction` consistent

### LINKED ISSUE #637: Inconsistency with `switch_direction`?
Let me first clarify a point about left and right actions. From a *left* action on a manifold, (g,x) ↦ L(g,x), one can always construct a *right* action, namely: L'(g,x) := L(g⁻¹,x). (A right action can also be transformed into a left action in this manner.)

Now, on a group, there are two *distinct* actions, stemming from multiplying on the left, or on the right. Put differently, one can say that there are **two distinct left actions**, L₁, and L₂ defined by L₁(g,g') := g g' and L₂(g,g') := g' g⁻¹. Now, according to the process above, both each give rise to two distinct right actions, namely L₁', and L₂'.

Here comes the inconsistency.

For the `RotationAction`, the result of `switch_direction` on the action is to map L to L' (where `L` is the rotation action).

But for the `GroupOperationAction`, the result of `switch_direction` is to map L₁ to L₂', instead of mapping L₁ to L₁'.

It would perhaps be nice if it was better documented, or, better, fixed to make it consistent?

mateuszbaran:
This PR implements a solution that will resolve #637 by introducing four different action directions instead of just two.

TODO:
* [x] review rotation action directions.
* [x] implement forward-right and backward-left actions.


mateuszbaran:
I think this should be more or less ready. @olivierverdier , do you think this solves the issue? I was trying to make the default close to what we currently have but now you can express all four options and switch just the left/right or forward/backward part of the action. Rotation and translation actions are all forward now.

kellertuer (approved):

olivierverdier:
It does solve the issue, thank you!

Issue #625, opened 2023-06-12
Add non-size-specializing variants of manifolds


mateuszbaran:
Currently most manifolds have size embedded in their type. That's nice for runtime performance but adds additional time for compilation which is sometimes an undesirable trade-off.

Manifolds that should have non-size-specializing variants:
- [ ] AbstarctUnitaryMatrices
- [ ] CenteredMatrices
- [ ] Elliptope
- [ ] FixedRankMatrices
- [ ] Flag
- [ ] GeneralizedGrassmann
- [ ] GeneralizedStiefel
- [ ] Grassmann
- [ ] MultinomialMatrices
- [ ] MultinomialSymmetric
- [ ] Oblique
- [ ] OrthogonalMatrices
- [ ] ProbabilitySimplex
- [ ] ProjectiveSpace
- [ ] SphereSymmetricMatrices
- [ ] Stiefel
- [ ] SymmetricMatrices
- [ ] SymmetricPositiveDefinite
- [ ] SymmetricPositiveSemidefiniteFixedRank
- [ ] Tucker

kellertuer:
Can you give an example how we would provide both in a nice way?
For completeness (and me to understand); maybe with the sphere?

mateuszbaran:
For Sphere, we could do
```julia
struct DynSphere{𝔽} <: AbstractSphere{𝔽}
    N::Int
end
representation_size(M::DynSphere) = M.N
function get_embedding(M::DynSphere{𝔽}) where {𝔽}
    return DefaultManifold(representation_size(M)...; field=𝔽)
end
```
and more or less all other methods should work fine thanks to dispatching on `AbstractSphere`. We would then need something like `AbstractSphere` for other manifolds (`AbstractStiefel` etc.). For some manifolds it would be more important than for other, for example `Grassmann` and `Stiefel` look high-ish priority while `OrthogonalMatrices` would be less important IMO.

The big problem is coming up with reasonable naming. I think ideally the "plain" name should refer to the non-specializing variant, and something like `StaticSphere` should be the current specializing one but that would be slightly breaking.

kellertuer:
I still do not necessarily see the benefit of these new types.
They should have quite a benefit, since keeping both always means we have to carefully communicate that to the user.
If we can't decide which one is better and have two, how should the user decide?

kellertuer:
On the other side, if that is faster, we could also change all dimension-parameters from type parameters to fields for all manifolds (and do a breaking release)?
[...]

Issue #471, opened 2022-05-07
Inconsistency in invariant metrics


mateuszbaran:
Currently invariant metrics on groups are somewhat inconsistent: `inner` continues dispatch with traits without unwrapping the manifold while `exp` and `log` unwrap and run trait dispatch from scratch. This should be consistent.

mateuszbaran:
Related: bases for invariant metrics:
```julia

for (invtrait, invdir) in
    [(HasLeftInvariantMetric, LeftAction), (HasRightInvariantMetric, RightAction)],
    basis_type in [DefaultOrthonormalBasis, DefaultOrthogonalBasis]

    eval(
        quote
            function get_coordinates(
                ::TraitList{$invtrait},
                M::AbstractDecoratorManifold,
                p,
                X,
                B::$basis_type,
            )
                Xₑ = inverse_translate_diff(M, p, p, X, $invdir())
                return get_coordinates(M.manifold, p, Xₑ, B)
            end
            function get_coordinates!(
                ::TraitList{$invtrait},
                M::AbstractDecoratorManifold,
                c,
                p,
                X,
                B::$basis_type,
            )
                Xₑ = inverse_translate_diff(M, p, p, X, $invdir())
                return get_coordinates!(M.manifold, c, p, Xₑ, B)
            end

            function get_vector(
                ::TraitList{$invtrait},
                M::AbstractDecoratorManifold,
                p,
                c,
                B::$basis_type,
            )
                Xₑ = get_vector(M.manifold, p, c, B)
                return translate_diff(M, p, p, Xₑ, $invdir())
            end
            function get_vector!(
                ::TraitList{$invtrait},
                M::AbstractDecoratorManifold,
                X,
                p,
                c,
                B::$basis_type,
            )
                Xₑ = get_vector(M.manifold, p, c, B)
                return translate_diff!(M, X, p, p, Xₑ, $invdir())
            end
        end,
    )
end
```

mateuszbaran:
I think in general we just shouldn't dispatch on specific traits, and just use them for forwarding of specific functions. This part works reasonably well. Dispatching on traits allows for some nice fallback definitions but it is just too convoluted.

PR #642, opened 2023-07-10
Extend support for SpecialEuclidean, fiber bundles, optionally static sizes

mateuszbaran:
One of the things I'm currently working on is rigid body dynamics using SE(n). I intend to add a bunch of missing features and a tutorial in this PR. The tutorial should potentially supersede #366 .

EDIT:
TODO:
  - [x] Check docs.
  - [x] Check coverage.
  - [x] Re-check #637 new fix.
  - [x] Remove (?) `LinearAffineMetric`.


tutorials/rigid-body-dynamics.qmd
```diff
@@ -0,0 +1,246 @@
...
+
+Phase space in classical mechanics is typically defined as a contangent bundle.
+Cotangent bundles are, however, hard to use for computation.
+So we select Riesz representers of cotangent vectors according to the metric of our manifold.
+In this way we get an isomorphism between tangent and cotangent bundles.
+
+Note that a generic description of a phase space doesn't require a Riemannian metric.
+Generic tools come from symplectic geometry which imposes a less rich structure.
+
+### Forward Euler? Really? {#sec-forward-euler}
```
mateuszbaran, line 244: @kellertuer have you maybe experimented with section cross-links? This doesn't seem to work. I know Quarto docs suggest just using `@sec-foward-euler` and go with the default style but maybe we could have arbitrary text in a link?

kellertuer: I would just use Documenter.jl syntax?

mateuszbaran: Something like 
```
### `[Forward Euler? Really?](@ref sec-forward-euler)`{=commonmark}
```
?

kellertuer: nearly, the section definitions get `@id`s.

mateuszbaran:
The last commit partially resolves #625 (I haven't updated all manifolds yet).

kellertuer:
Since this one is breaking, could we base that on a breaking release of ManifoldsBase as well? Would that be a good reason to do that?

mateuszbaran:
Resolving the exp/retract asymmetry would require a breaking release of ManifoldsBase.jl but I want to do things that don't require breaking Base first.

kellertuer:
Ok, I was just thinking, since we are breaking something here, why not at the same time break ManifoldsBase as well, but that was also just an idea.

src/groups/heisenberg.jl
```diff
@@ -26,16 +33,17 @@ function active_traits(f, ::HeisenbergGroup, args...)
     )
 end
 
-function _heisenberg_a_view(::HeisenbergGroup{n}, p) where {n}
+function _heisenberg_a_view(M::HeisenbergGroup, p)
+    n = get_n(M)
```
kellertuer, line 37: here, the same for the name applies as before.

src/groups/rotation_action.jl
```diff
@@ -26,11 +26,11 @@ function Base.show(io::IO, A::RotationAction)
     return print(io, "RotationAction($(A.manifold), $(A.SOn), $(direction(A)))")
 end
 
-const RotationActionOnVector{N,F,TAD} = RotationAction{
-    <:Union{Euclidean{Tuple{N},F},TranslationGroup{Tuple{N},F}},
-    SpecialOrthogonal{N},
+const RotationActionOnVector{TAD,𝔽,TE,TSO} = RotationAction{
```
kellertuer, line 29: You turned these around since then the dispatch below is easier?

src/groups/rotation_translation_action.jl
```diff
@@ -0,0 +1,310 @@
...
+"""
+struct RotationTranslationAction{
+    TAD<:ActionDirection,
+    TM<:AbstractManifold,
+    TSE<:SpecialEuclidean,
+} <: AbstractGroupAction{TAD}
+    manifold::TM
+    SEn::TSE
+end
+
```
kellertuer, line 22: Could we document a few of the functions below? Mainly for a new user to more understand how the functions are related to actions happening in theory.

src/groups/rotation_translation_action.jl
```diff
@@ -0,0 +1,310 @@
...
+    a::ArrayPartition,
+    p,
+    X,
+)
+    return a.x[2] * X
+end
+
+###
+
+@doc raw"""
```
kellertuer, line 226: Could you motivate how. this is used later? I would have thought this is more like a product of actions, so that I am not sure how much this type is needed.

src/groups/special_euclidean.jl
```diff
@@ -25,34 +25,45 @@ $\mathrm{T}(n) × \mathrm{SO}(n)$. For group-specific functions, they may also b
...
+function Base.show(io::IO, G::SpecialEuclidean{Tuple{Int}})
+    n = get_n(G)
+    return print(io, "SpecialEuclidean($(n))")
+end
 
 @inline function active_traits(f, M::SpecialEuclidean, args...)
     return merge_traits(IsGroupManifold(M.op), IsExplicitDecorator())
 end
 
+get_n(::SpecialEuclidean{TypeParameter{Tuple{N}}}) where {N} = N
```
kellertuer, line 62: here again the same – I think the generic get_parameter should be fine as well?

src/groups/special_euclidean.jl
```diff
@@ -72,34 +83,38 @@ Base.@propagate_inbounds function Base.setindex!(
 end
 
 Base.@propagate_inbounds function submanifold_component(
-    ::Union{SpecialEuclidean{n},SpecialEuclideanManifold{n}},
```
kellertuer, line 75: So in the old form we implemented it for every union of same size but actually that was ten then not necessary? Ah, we still need the `n` now, I see. But here, we would loose the “dispatch-advantage” of the type-manifold, right?

src/groups/special_euclidean.jl
```diff
@@ -128,7 +144,11 @@ The formula for the coefficients reads ``t×(R⋅ω) + R⋅r`` for the translati
 matrix part of `p`, `r` is the translation part of `fX` and `ω` is the rotation part of `fX`,
 ``×`` is the cross product and ``⋅`` is the matrix product.
 """
-function adjoint_action(::SpecialEuclidean{3}, p, fX::TFVector{<:Any,VeeOrthogonalBasis{ℝ}})
```
kellertuer, line 131: Would this still work for the field-manifold?

kellertuer (commented):
I had a bit of time during breaks of my seminar, so here are a few remarks.

In general maybe

* `get_n` is not a nice name and why can't we use `get_parameter` (or another function that “extracts” the old `{N}` type parameter)?
* is it correct that several functions for now only work for the type parameter variant, while the field one is the default?

NEWS.md
```diff
@@ -0,0 +1,95 @@
...
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.9.0] - 2023-mm-dd
+
+### Added
+
+- Vector bundles are generalized to fiber bundles.
+- `RotationTranslationAction`.
```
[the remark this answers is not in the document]
mateuszbaran: `RotationTranslationAction` is still WIP

benchmark/benchmarks.jl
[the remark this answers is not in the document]
mateuszbaran: I haven't run that benchmark in a very long time but I don't want to work on it more than necessary in this PR.

ext/ManifoldsRecipesBaseExt.jl
```diff
@@ -24,7 +26,7 @@ SURFACE_RESOLUTION_DEFAULT = 32
 # Plotting Recipe – Poincaré Ball
 #
 @recipe function f(
-    M::Hyperbolic{2},
```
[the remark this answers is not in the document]
mateuszbaran: This can be discussed as part of the broader issue of "things working for type parameter but not for field parameter".

ext/ManifoldsTestExt/tests_group.jl
```diff
@@ -741,10 +741,10 @@ function test_action(
 
                 eX = allocate(X)
                 Test.@test apply_diff!(A, eX, e, m, X) === eX
-                Test.@test isapprox(G, m, eX, X; atol=atol)
```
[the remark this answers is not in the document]
mateuszbaran: I don't remember exactly, something was lacking a pass-through and extending them for groups is very convoluted due to `Identity`.

src/groups/general_linear.jl
```diff
@@ -88,12 +92,12 @@ end
 function exp!(G::GeneralLinear, q, p, X, t::Number)
     return exp!(G, q, p, t * X)
 end
-function exp!(::GeneralLinear{1}, q, p, X)
```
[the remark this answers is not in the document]
mateuszbaran: This is again the problem of forwarding from type to field parameter implementation. That is possible but requires adding forwarding methods. In general things should work for field parameters but performance-wise, it's easier to require users to use type-parameters for parameter-specific optimizations rather than checking values of parameters in the generic method for possible faster implementations. But doing such checks and forwards is definitely possible.

src/groups/general_linear.jl
```diff
@@ -105,50 +109,62 @@ function exp!(G::GeneralLinear{2}, q, p, X)
...
-) where {n}
+)
     return copyto!(Xⁱ, X)
 end
 
-get_embedding(::GeneralLinear{n,𝔽}) where {n,𝔽} = Euclidean(n, n; field=𝔽)
+function get_embedding(::GeneralLinear{TypeParameter{Tuple{n}},𝔽}) where {n,𝔽}
+    return Euclidean(n, n; field=𝔽, parameter=:type)
+end
+function get_embedding(M::GeneralLinear{Tuple{Int},𝔽}) where {𝔽}
```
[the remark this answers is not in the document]
mateuszbaran: We could have only one method of `get_embedding` per manifold but then we would have to delegate dispatch on parameter to another function with two methods. It would make sense if that function had other uses but I haven't found them yet.

src/groups/rotation_action.jl
mateuszbaran: Yes, exactly.

src/groups/rotation_translation_action.jl
mateuszbaran: This is still WPI, I will do that later.

kellertuer:
> > * `get_n` is not a nice name and why can't we use `get_parameter` (or another function that “extracts” the old `{N}` type parameter)?
> 
> We can have something different, I just wanted names somewhat relevant to the manifold being implemented. For example "get n for SE(n)" sounds to me more straightforward than "get the first parameter of SE(n)". Do you have a better name for that?

But `get_n` is too generic and requires too much knowledge on the user side what `n` is.
Sure `get_parameter` is a technical name, but maybe nicer to have a generic name and having less code per manifold? I would prefer just one times this implementation.

> > * is it correct that several functions for now only work for the type parameter variant, while the field one is the default?
> 
> I'm not sure. I've done the core part of this change and left such problems for later 🙂 . I'm generally fine with both only working for the type parameter variant and having an implementation for field parameter that forwards to the type parameter implementation.

I am not so much a fan of having them only for the type parameter when the field parameter is the default, since a new user might miss functionality that way that actually exists, just not for their (default, and quite technical) manifold type. How would one to the forwards?

src/groups/special_euclidean.jl
mateuszbaran: What "dispatch-advantage” of the type-manifold" do you mean?

mateuszbaran:
> But `get_n` is too generic and requires too much knowledge on the user side what `n` is.
> Sure `get_parameter` is a technical name, but maybe nicer to have a generic name and having less code per manifold? I would prefer just one times this implementation.

In the context of a specific manifold `get_n` seems quite clear to me but I can change it to use `get_parameter`.

> I am not so much a fan of having them only for the type parameter when the field parameter is the default, since a new user might miss functionality that way that actually exists, just not for their (default, and quite technical) manifold type. How would one to the forwards?

The simplest forward would be just
```julia
project(M::UnitaryMatrices{Tuple{Int},ℍ}, p) = project(UnitaryMatrices(M.size[1]; parameter=:type), p)
```
and so on for other methods.

kellertuer:
I would prefer the slightly technical `get_parameter` variant, yes.

Concerning the forward – doing that for every manifold is _a lot_ of boiler plate code.

mateuszbaran:
> I would prefer the slightly technical `get_parameter` variant, yes.

OK.

> Concerning the forward – doing that for every manifold is _a lot_ of boiler plate code.

It is, that's why I haven't done it yet. I don't see a better way though.

kellertuer:
The amount of boiler plate needed is – for me – and argument against this doubling / Parameter/field thing.

For now I do not yet see the actual benefit.
It is quite a technical detail – so _if_ we do this, it should not be to the disadvantage of the usability, especially for new users.
Missing functions would be a disadvantage I think.
But I also do not like boiler plate code.

Hm. What can we best do here?

mateuszbaran:
> For now I do not yet see the actual benefit.

The benefit is in the amount of compilation required to get some results. It's particularly important when someone needs for their computation some manifolds with many different sizes. With precompilation it can also significantly reduce TTFX.

> It is quite a technical detail – so _if_ we do this, it should not be to the disadvantage of the usability, especially for new users. Missing functions would be a disadvantage I think. But I also do not like boiler plate code.
> 
> Hm. What can we best do here?

Maybe the easiest option would be to keep type-parameter as the default. New users would have full functionality by default with no boilerplate, at the cost of larger TTFX when we start doing precompilation.

kellertuer:
I see the idea you want to get to, but with parameter as type, your times are much smaller ( factor 3?) than for the field one - or is that some randomness?

Also this will only affect users _reusing_ manifolds but also changing the dimension/size of the manifold. To me this seems to also hint to using `parameter=:type` as a default.
We can maybe switch the default when we notice that `parameter=:field` is well-supported in most manifolds and indeed in some sense beneficial.

But you convinced me, we can surely have both in parallel for now :)
Just, then, please keep `:type` as the default for now, since it is the one where we are more sure things work.

mateuszbaran:
> I see the idea you want to get to, but with parameter as type, your times are much smaller ( factor 3?) than for the field one - or is that some randomness?

No, with field parameter the TTFX is 20x smaller ( 0.278328 seconds vs 0.013193 seconds)

> Also this will only affect users _reusing_ manifolds but also changing the dimension/size of the manifold. To me this seems to also hint to using `parameter=:type` as a default. We can maybe switch the default when we notice that `parameter=:field` is well-supported in most manifolds and indeed in some sense beneficial.

It will affect anyone whose size isn't precompiled, not just people reusing the same manifold with a different size (see first comment).

> But you convinced me, we can surely have both in parallel for now :)
> Just, then, please keep `:type` as the default for now, since it is the one where we are more sure things work.

OK, cool, I will make `:type` default in this version -- it can be easily changed in the next breaking release if we decide so.

kellertuer:
> > I see the idea you want to get to, but with parameter as type, your times are much smaller ( factor 3?) than for the field one - or is that some randomness?
> 
> No, with field parameter the TTFX is 20x smaller ( 0.278328 seconds vs 0.013193 seconds)
> 

Now I am confused, maybe since you did not provide all details. So the _very first_ call is a field-type manifold and takes 0.27 seconds? It does not have keyword and currently the default is field.

Then you have the precompiled stuff (that we could introduce) which reduces your recompilation by a factor of 20. But that is not the switch from parameter to field _but_ the precompilation?!

The lat 2 ones on the other hand (our good old parameter type friends) take 0.08 seconds, which is 3x faster than the initial one without precompilcation. That is what I referred to, since for now we do not have precompilation.
But you are right in the sense that – for field manifolds we _could_ do precompilations.

mateuszbaran:
> Now I am confused, maybe since you did not provide all details. So the _very first_ call is a field-type manifold and takes 0.27 seconds? It does not have keyword and currently the default is field.
> 
> Then you have the precompiled stuff (that we could introduce) which reduces your recompilation by a factor of 20. But that is not the switch from parameter to field _but_ the precompilation?!
> 
> The lat 2 ones on the other hand (our good old parameter type friends) take 0.08 seconds, which is 3x faster than the initial one without precompilcation. That is what I referred to, since for now we do not have precompilation. But you are right in the sense that – for field manifolds we _could_ do precompilations.

To make it more straightforward: without precompilation the very first call takes 0.27s either way (both field-parameter and type-parameter). Precompilation (for a different size) + type parameter took 0.08s, while precompilation + field parameter took 0.013s.

mateuszbaran:
> Then you have the precompiled stuff (that we could introduce) which reduces your recompilation by a factor of 20. But that is not the switch from parameter to field _but_ the precompilation?!

So the factor of 20 requires both actually: precompilation and switching to field (or precompiling for the exact size user wants).

kellertuer:
Ok, I now see that the new type has areas where it might be beneficial.
It might still be a bit difficult to convey to the user, but we can maybe to a small tutorial somewhen (maybe even in ManifoldsBase).

src/groups/special_euclidean.jl
kellertuer: Ah maybe not so relevant – the new form looks nice and seems to work for all ur diverse manifold types, that nice.

src/manifolds/DirectSumBundle.jl
```diff
@@ -0,0 +1,232 @@
+
```
kellertuer, line 1: This is probably a bit WIP still, but again the functions should be documented.

kellertuer (commented):
Here are a few more comments, but I think most of the remaining part is already mention, so if that is reworked, I will take a look at the files after Euclidean :)

src/groups/special_euclidean.jl
mateuszbaran: This is one place where `get_parameter` is less clear, as `SpecialEuclidean` requires a lot of unwrapping: `get_parameter(M.manifold.manifolds[1].manifold.size)[1]`. Can we have a helper function that does that?

kellertuer: We could indeed just define `get_parameter(::SpecialEuclidean{TypeParameter{Tuple{N}}}) where {N} = N` or would that lead to ambiguities since `SpecialEuclidean` is some strange type?

kellertuer: For me the problem is that similar to “magic numbers” the `n` in the function name is a “magic name”.
We have a few ones we document and use a lot (M, p, q, X, Y, and such) but not `n` in that general sense yet, that is my argument against the function name.

mateuszbaran: `get_parameter` was specifically written to unwrap `TypeParameter`. Adding special behavior for manifolds may be problematic in case a manifold is used as a (type) parameter. I'm open to calling this helper function something different than `get_n` but don't have an idea for a good name.

src/groups/special_euclidean.jl
mateuszbaran: No, it wouldn't, but I think we concluded that it's fine if type-parameter is the default (which it is now).

kellertuer: Ok, cool.
For me this is a good tradeoff, now that I thought about that for a while.

It is like, for the day-to-day work you have a, maybe not that efficient, but reliable and tested car (parameter type), that does the job, but is maybe not fast in first runs.
And if you have to race, then you switch to the race car (field type) that is fast also in first runs, but (not yet) so reliable and maybe without airbags.

mateuszbaran: Cool, I also think it makes sense.

src/groups/special_euclidean.jl
kellertuer: it is, after all, still getting the parameter, what the function does, though :D

Would `_get_parameter` be ok? That indicates that this is a local function, here even meant in the sense only local to special euclidean even?

mateuszbaran: `_get_parameter` is definitely fine. I just don't want `get_parameter` to ever return a parameter of a parameter instead of a parameter (a bit like `1\in \{1\}` but `1 \notin \{\{1\}\}`)

kellertuer: I see your point and agree. But then let's use `_get_parameter` here.

ext/ManifoldsTestExt/tests_group.jl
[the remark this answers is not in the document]
mateuszbaran: I've checked it and having `G` is actually a bug: `M` may be a topologically different manifold than the one corresponding to `G`, so it was actually a bit of luck that `G` worked here so far. And `m`, `eX`, `X` are definitely from `M` (or its tangent space), not `G`.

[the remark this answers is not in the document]
kellertuer: Ah, thanks for checking; that makes sense :)

kellertuer:
Since I am currently asking so many questions – could you append to the comment here https://github.com/JuliaManifolds/Manifolds.jl/blob/mbaran/special-euclidean-revisited/src/trait_recursion_breaking.jl#L12 a bit? I am unsure what that file does and the comment is more like: “This is here because it is here.”

mateuszbaran:
Sure, done. If you need a simple explanation of "what do I need to know to modify this file", then unfortunately it requires delving into profiling and some trial and error, and much more text.

kellertuer:
Hm, I was expecting a bit more detail, but I at least understand, that the code is not there, because otherwise we get errors, but again to enter racing-areas ;) 

In the long run a bit more explanation might be nice though. But I was just stumbling into this while working on the manifold features and was curious. I think long-term it is good to have a bit more comments or explanations in such situations.

NEWS.md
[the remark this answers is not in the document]
mateuszbaran: Regarding BTW, ManifoldsBase.jl has `NEWS.md`. And https://keepachangelog.com/en/1.0.0/ suggests capitalized `CHANGELOG`. So it's somewhat inconsistent.

ext/ManifoldsRecipesBaseExt.jl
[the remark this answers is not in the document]
mateuszbaran: I think this has been resolved.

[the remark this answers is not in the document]
kellertuer: It still would not work for the field parameter, but with the default being parameter, this is fine :)

src/groups/rotation_translation_action.jl
mateuszbaran: We don't currently have a generic "product" of actions, and I'm not sure how much it would be needed. We can always refactor it later when the appropriate product (I can think of many different constructions that can be thought of as a product of actions) is introduced.

kellertuer: Ok, this sounds fair.

src/manifolds/DirectSumBundle.jl
mateuszbaran: I've decided to leave it for a future PR.

mateuszbaran:
@kellertuer I've tried addressing https://github.com/JuliaManifolds/Manifolds.jl/issues/637#issuecomment-1727115435 and it sort of works but `ActionDirectionAndSide` looks a bit ugly now IMO. Do you have a better idea? It's mainly needed for all those `translate` and `translate_diff` methods.

I've also started making a bit of a cleanup of invariant metrics but I think I will leave a proper rework addressing #471  for later. The change here is "don't use invariance traits if you don't want to use the fallbacks", which limits the impact of a future rework. The problem is that these defaults are introduced in a very non-intuitive way and if we ever wanted patterns that work with group manifolds with partial use of these fallbacks and seamless wrapping in additional decorators, that is simply too much complexity for me.

kellertuer:
In said comments you and Olivier switched opinions more often than Action Directions, so I have no opinion nor overview on that any longer.

Concerning the metric, I would prefer anyways to have the metric interface in `ManifoldsBase` at some point. Today I noticed also TangentSpaceAtPoint would be neat in there, but there was something that avoided that we moved that to base? However, for metrics that would be another step unrelated to this PR I think.

mateuszbaran:
> In said comments you and Olivier switched opinions more often than Action Directions, so I have no opinion nor overview on that any longer.

Yes, this is confusing topic but I hope this time I've got it right :slightly_smiling_face: . The main remaining questions are:
1. Do we want `translate(M::AbstractManifold, p, q, conv::ActionDirectionAndSide)` or `translate(M::AbstractManifold, p, q, conv::ActionDirection, ::GroupActionSide)`?
2. Maybe `ForwardSide` and `BackwardSide` would be better than `LeftSide` and `RightSide`? I'd like it to not be too confusing for new users, if possible. It's quite unfortunate that many sources don't pay too much attention to this distinction.

> Concerning the metric, I would prefer anyways to have the metric interface in `ManifoldsBase` at some point. Today I noticed also TangentSpaceAtPoint would be neat in there, but there was something that avoided that we moved that to base? However, for metrics that would be another step unrelated to this PR I think.

That would be a good idea but it would require iterating upon it IMO. I wanted to make a small change in this PR that I think is necessary because we make a breaking release anyway but if you prefer a single, comprehensive metric rework later I can leave it out of this PR. I can also put `TangentSpaceAtPoint` in ManifoldsBase.jl (would be a part of 0.15 together with https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/167).

kellertuer:
Concerning 1) I do not have a preference. Maybe the second is easier to provide a default for the last parameter and that makes it easier to use this way?
Concerning 2) here I really have no preference, left/right forward/backward or even hubward/rimward or turnwise/widdershins (actually the four directions to walk on https://en.wikipedia.org/wiki/Discworld),... I have no preference.

Concerning the metric – yes let‘s do that in another PR to keep this one of finite length.

Concerning TangentSpaceAtPoint – yes for example as part of 0.15, but if I remember correctly there was some reason (dependency?) why you did not want to move all FibreGames to ManifoldsBase?

mateuszbaran:
> Concerning 1) I do not have a preference. Maybe the second is easier to provide a default for the last parameter and that makes it easier to use this way?
> Concerning 2) here I really have no preference, left/right forward/backward or even hubward/rimward or turnwise/widdershins (actually the four directions to walk on https://en.wikipedia.org/wiki/Discworld),... I have no preference.

OK, I will think about it a bit more or maybe ask on Slack or zulip.

> Concerning the metric – yes let‘s do that in another PR to keep this one of finite length.

OK.

> Concerning TangentSpaceAtPoint – yes for example as part of 0.15, but if I remember correctly there was some reason (dependency?) why you did not want to move all FibreGames to ManifoldsBase?

No, there was no issue with fibres or tangent spaces. `ProductManifold` and `VectorBundle` were problematic because it can't work without `RecursiveArrayTools.jl`. But now we can make `RecursiveArrayTools.jl` a package extension to ManifoldsBase.jl so they would be doable in `ManifoldsBase.jl` too.

kellertuer:
For Manopt I would _love_ to have TangentSpaceAtPoint in ManifoldsBase (maybe even before finishing the rework on Trust Regions), since it allows to have a nice way to define the lifted (=defined on a tangent space) problems for the sub solvers. I will take a look at tangent space then, probably.
But for best of cases, all these meta-manifold should be defined in ManifoldsBase, because then everyone using ManifoldsBase defining their own manifold can have the meta ones already for free.

kellertuer:
One comment on the inial post already – yes please remove the `LinearAffineMetric` which is deprecated already and should be `AffineInvariantMetroc` by now :)

kellertuer:
Oops sorry for removing the `===` was probably during debugging something.

mateuszbaran:
I think this is more or less ready now. We can tag ManifoldsBase.jl 0.15 and check coverage here.

kellertuer:
Nice! Let‘s do that.

kellertuer:
In the long run this dependency on each other should be removed, maybe even for both, but especially this here should not (weak/extra) depend on Diff.

mateuszbaran:
Hm... why does quarto activate a separate project instead of using the same one as docs? I think it just makes it impossible to use this unreleased branch for tutorials?

kellertuer:
That was initially the idea to make the tutorials more safe / reproducible – and independent of the docs environment.
What you can do, is to put Manifolds.jl in development mode in the tutorials environment (and remove it on the next PR after this)?

mateuszbaran:
This is how it resolves:
```
  Activating project at `~/work/Manifolds.jl/Manifolds.jl/tutorials`
   Installed InlineStrings ─────── v1.4.0
   Installed BoundaryValueDiffEq ─ v5.1.0
   Installed DiffEqBase ────────── v6.133.1
   Installed NonlinearSolve ────── v2.4.0
   Installed LinearSolve ───────── v2.11.1
   Installed SciMLBase ─────────── v2.4.3
   Installed SentinelArrays ────── v1.4.0
   Installed WorkerUtilities ───── v1.6.1
   Installed WeakRefStrings ────── v1.4.2
   Installed DataFrames ────────── v1.6.1
   Installed PooledArrays ──────── v1.4.3
   Installed Arpack_jll ────────── v3.5.1+1
   Installed MultivariateStats ─── v0.10.2
   Installed ManifoldDiff ──────── v0.3.6
   Installed Arpack ────────────── v0.5.4
   Installed FilePathsBase ─────── v0.9.21
   Installed StringManipulation ── v0.3.4
   Installed PrettyTables ──────── v2.2.8
   Installed ManifoldsBase ─────── v0.14.12
   Installed Manifolds ─────────── v0.8.74
   Installed CSV ───────────────── v0.10.11
   Installed InvertedIndices ───── v1.3.0
    Updating `~/work/Manifolds.jl/Manifolds.jl/tutorials/Project.toml`
  [764a87c0] + BoundaryValueDiffEq v5.1.0
  [336ed68f] + CSV v0.10.11
  [a93c6f00] + DataFrames v1.6.1
  [459566f4] + DiffEqCallbacks v2.33.1
  [31c24e10] + Distributions v0.25.102
  [7073ff75] + IJulia v1.24.2
⌃ [1cead3c2] + Manifolds v0.8.74
⌅ [3362f125] + ManifoldsBase v0.14.12
  [6f286f6a] + MultivariateStats v0.10.2
  [1dea7af3] + OrdinaryDiffEq v6.58.0
  [91a5bcdd] + Plots v1.39.0
  [731186ca] + RecursiveArrayTools v2.38.10
  [276daf66] + SpecialFunctions v2.3.1
  [90137ffa] + StaticArrays v1.6.5
```
And we get an error that looks like tutorial rendering using these old versions:
```
An error occurred while executing the following cell:
------------------
is_point(M₆, p₃; error=:error)
------------------


MethodError: no method matching isapprox(::Int64, ::Float64; error::Symbol)

Closest candidates are:
  isapprox(::Number, ::Number; atol, rtol, nans, norm) got unsupported keyword argument "error"
   @ Base floatfuncs.jl:304
  isapprox(::Union{StatsBase.PValue, StatsBase.TestStat}, ::Real; kwargs...)
   @ StatsBase ~/.julia/packages/StatsBase/WLz8A/src/statmodels.jl:104
  isapprox(::SemidirectProductGroup, ::Any, ::Any; kwargs...)
   @ Manifolds ~/.julia/packages/Manifolds/TizHt/src/groups/semidirect_product_group.jl:225
  ...


Stacktrace:
  [1] kwerr(::NamedTuple{(:error,), Tuple{Symbol}}, ::Function, ::Int64, ::Float64)
    @ Base ./error.jl:165
  [2] check_point(M::Hyperbolic{2}, p::Vector{Int64}; kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:error,), Tuple{Symbol}}})
    @ Manifolds ~/.julia/packages/Manifolds/TizHt/src/manifolds/HyperbolicHyperboloid.jl:29
  [3] (::Manifolds.var"#177#180"{Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:error,), Tuple{Symbol}}}})(t::Tuple{Int64, Hyperbolic{2}, Vector{Int64}})
    @ Manifolds ./none:0
  [4] iterate
    @ ./generator.jl:47 [inlined]
  [5] collect(itr::Base.Generator{Tuple{Tuple{Int64, Hyperbolic{2}, Vector{Int64}}, Tuple{Int64, Sphere{2, ℝ}, Vector{Int64}}}, Manifolds.var"#177#180"{Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:error,), Tuple{Symbol}}}}})
    @ Base ./array.jl:782
  [6] check_point(M::ProductManifold{ℝ, Tuple{Hyperbolic{2}, Sphere{2, ℝ}}}, p::ArrayPartition{Int64, Tuple{Vector{Int64}, Vector{Int64}}}; kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:error,), Tuple{Symbol}}})
    @ Manifolds ~/.julia/packages/Manifolds/TizHt/src/manifolds/ProductManifold.jl:231
  [7] check_point
    @ ~/.julia/packages/Manifolds/TizHt/src/manifolds/ProductManifold.jl:229 [inlined]
  [8] #is_point#107
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/ManifoldsBase.jl:669 [inlined]
  [9] is_point
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/ManifoldsBase.jl:663 [inlined]
 [10] invoke
    @ ./Base.jl:112 [inlined]
 [11] is_point(::ManifoldsBase.EmptyTrait, M::ProductManifold{ℝ, Tuple{Hyperbolic{2}, Sphere{2, ℝ}}}, p::ArrayPartition{Int64, Tuple{Vector{Int64}, Vector{Int64}}}, te::Bool; kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:error,), Tuple{Symbol}}})
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:346
 [12] #is_point#192
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:313 [inlined]
 [13] is_point
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:308 [inlined]
 [14] is_point(_t::ManifoldsBase.TraitList{IsDefaultMetric{ProductMetric}, ManifoldsBase.TraitList{IsMetricManifold, ManifoldsBase.EmptyTrait}}, M::ProductManifold{ℝ, Tuple{Hyperbolic{2}, Sphere{2, ℝ}}}, p::ArrayPartition{Int64, Tuple{Vector{Int64}, Vector{Int64}}}, te::Bool; kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:error,), Tuple{Symbol}}})
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:313
 [15] is_point
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:308 [inlined]
 [16] #is_point#191
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:306 [inlined]
 [17] is_point
    @ ~/.julia/packages/ManifoldsBase/vKGuo/src/nested_trait.jl:305 [inlined]
 [18] top-level scope
    @ In[31]:1
```

kellertuer:
...otherwise, feel free to propose a complete rewrite of this approach, I was happy to find a stable reproducing way after a few weeks of working on this. If you see a better way – sure. Go for that. I would still prefer to have 2 environments for tutorials and docs.

mateuszbaran:
At the moment I'm not even sure what exactly the interactions of CI, Pkg, Conda and quarto are and all I want is to make the docs working. We can revert whatever makes them work in the next PR or discuss an alternative later.

docs/make.jl
```diff
@@ -21,11 +21,11 @@ if "--quarto" ∈ ARGS
...
         @info "Rendering Quarto"
-        tutorials_folder = (@__DIR__) * "/../tutorials"
-        # instantiate the tutorials environment if necessary
-        Pkg.activate(tutorials_folder)
-        Pkg.resolve()
-        Pkg.instantiate()
+        # tutorials_folder = (@__DIR__) * "/../tutorials"
+        # # instantiate the tutorials environment if necessary
+        # Pkg.activate(tutorials_folder)
+        # Pkg.resolve()
```
kellertuer, line 27: This will not help much and just lead to tutorials failing to start every now and then; every tutorial runs in its own Julia call and every tutorial starts with activating this environment, so not updating/resolving it – led to every now and then Python not finding Julia.

kellertuer: But I have to be honest, I do not fully understand all the resolving that is done here, so if you see a better way feel free to improve this. I noticed, these lines are necessary.

mateuszbaran: I was hoping it would just go back to the environment used for docs but as I said, I don't really understand all this stuff.

kellertuer:
> At the moment I'm not even sure what exactly the interactions of CI, Pkg, Conda and quarto are and all I want is to make the docs working. We can revert whatever makes them work in the next PR or discuss an alternative later.

Here is a trick to get a single tutorial use the current dev one https://github.com/JuliaManifolds/Manopt.jl/blob/d6acf643e41a399f4a4e678eeceda2a8a54156cd/tutorials/CountAndCache.qmd#L52-L59 – but I am not sure that works for breaking ones.

As I already wrote – I am not saying my tutorial rendering is the very best solution when it comes to breaking version changes – but after quite a few weeks of work that was the version that did work in a reproducible way.

docs/make.jl
kellertuer: Every quarto is run in a separate Julia call and every notebook in the very first cell activates the tutorials environment.

mateuszbaran:
Thanks for the tip, I will try it.

docs/make.jl
mateuszbaran: OK, so I won't touch the tutorials that still work.

kellertuer:
If it does not I can try to fix it locally here, since I have done that once at least (as I said I first did not have that as a case we do not want to do). The trick should be to locally (in the notebook) dev the right packages ;)

mateuszbaran:
Documentation works now but normal tests seem to run out of memory on CI. I think we can cut some power or product manifold tests to help with that.

kellertuer:
Sure, I mean some time after the TR rework I want to get back to minimising the tests anyways. So feel free to minimise the tests (as long as code coverage remains).

mateuszbaran:
Tests are passing again and coverage is better than before :tada:

src/groups/circle_group.jl
```diff
@@ -15,7 +15,6 @@ CircleGroup() = GroupManifold(Circle{ℂ}(), MultiplicationOperation())
     else
         return merge_traits(
             IsGroupManifold(M.op),
-            HasBiinvariantMetric(),
```
kellertuer, line 18: So it‘s invariant but not bi-invariant? Should we maybe have a trait for the single-invariant metric?

mateuszbaran: Eh, I just haven't fully undone the removal of this trait. I think we need to redo invariant metrics anyway. Note that it only appears (on master branch) in the *non*-metric path of `active_traits`, so it was useless anyway.

kellertuer: Ok, we can surely redo that later (maybe when moving MatricManifold to ManifoldsBase, as I once tried to start).

src/groups/rotation_translation_action.jl
mateuszbaran: I've added some documentation here.

kellertuer:
Yes we should probably split that – or rethink even the idea of doing a separate Lie groups package (i.e. define the manifolds parts here and the group “extensions” to said manifolds over there).

kellertuer (commented):
Overall this looks fine, just a few small remarks and three further points to discussion

* I feel the `FiberBundleProductRetraction` could have two retractions contained, one for the manifold part, one for the fiber part?
* the name `direction_and_side` sounds a bit strange
* For the new documentation like
  ```julia
  M::MetricManifold{ℝ,<:Stiefel{<:Any,ℝ},<:StiefelSubmersionMetric}
  ```
  it might be a bit hard to read in the docs compared to the old `n,k` in place of the `<:Any` I feel, since the `n,k` transported more information as well.

NEWS.md
```diff
@@ -0,0 +1,107 @@
+# Changelog
```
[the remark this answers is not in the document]
mateuszbaran: Sure, I will try that.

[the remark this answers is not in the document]
mateuszbaran: I've added it, let me know if you prefer a different placement.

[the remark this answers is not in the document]
kellertuer: While I initially said maybe as a last entry, I see the Misc section is ordered alphabetically, so that is a good idea to stay that way and the placement is nice :)

mateuszbaran:
> If it has a reasonable behaviour by now (does it map to both defaults by now?) then this is fine here but the more general case where one can even set them would be a reasonable extension for a next PR then.

Currently it's only implemented for vector bundles and uses `exp`/`+`. That was sufficient so far. Anyway, I will work on extending it in a reasonably near future (probably after Manopt.jl/Optimization.jl integration).

> I do see that this is useful, I just feel the name is a bit strange, since it sounds like it would do two things at the same time.

It kind of does two things :sweat_smile:: returns both direction and side of group operation action, as a two-element tuple.

> One could – if `n` and `k` are required in the doc string (I did not check all docs for this) add them either as a reminder or the math formula to specify n and k again.

Yes, something like that. If you had an idea for a copy-paste'able phrase I could quickly add, I could add it. Otherwise I'd prefer to leave that for later.

mateuszbaran:
Great! I will wait for CI and merge it then.

PR #648, opened 2023-09-08
Add Julia 1.10-beta to CI, remove nightly

mateuszbaran:
I think currently Julia nightly tests are more confusing than helpful and we should limit testing to alpha/beta builds of Julia.


mateuszbaran:
Regarding failure, see https://github.com/JuliaLinearAlgebra/LinearMaps.jl/issues/214

kellertuer:
I agree on the nightly issue, they were broken for some time now and not helpful; so let's not run them; we could still keep beta versions in a nightly-like fashion to not break tests?

mateuszbaran:
I'm not sure about running alpha/beta builds in a "nightly" CI. Once we merge the PR enabling support for it, ideally things should not break. If things don't work, we would just delay merging that PR and regressions should be very rare.

kellertuer:
Ok, that is also fine with me, since you might be right, the beta versions should be more stable than Nightly.

kellertuer (approved):

mateuszbaran:
LinearMaps strikes again: https://github.com/JuliaLinearAlgebra/LinearMaps.jl/issues/219 .

PR #651, opened 2023-09-19
Hyperbolic Riemann tensor

hajg-ijk:
This small PR implements the Riemann tensor for the `Hyperbolic` manifold, analogous to the one on the `Sphere`.

I know I should probably open an issue about this, but since it's related: I was wondering whether it would make sense to implement a `sectional_curvature` method for the manifolds where a `riemann_tensor` is available.

I also added a _new_ textbook reference for the formula and myself to the list of contributors.


src/manifolds/Hyperbolic.jl
```diff
@@ -380,3 +380,22 @@ for (P, T) in zip(_ExtraHyperbolicPointTypes, _ExtraHyperbolicTangentTypes)
...
+````math
+R(X,Y)Z = - (\langle Z, Y \rangle X - \langle Z, X \rangle Y)
+````
+"""
+riemann_tensor(::Hyperbolic, p, X, Y, Z)
+
+function riemann_tensor!(M::Hyperbolic, Xresult, p, X, Y, Z)
+    innerZX = inner(M, p, Z, X)
+    innerZY = inner(M, p, Z, Y)
+    Xresult .= -(innerZY .* X .- innerZX .* Y)
```
mateuszbaran, line 399:
```suggestion
    Xresult .= innerZX .* Y .- innerZY .* X
```
This should reduce the number of operations required here.

hajg-ijk:
> What is computing sectional curvatures useful for? I'm generally trying to avoid adding functions that don't have a practical use. Note that most of MetricManifold stuff needs to be revisited so don't pay too much attention to it.

Honestly, I only thought about it because I am using it in what I am doing at the moment, which is also the reason why I implemented the Riemann tensor for the hyperbolic space. I thought about it as just a utility function, much like some cost functions are exported by Manopt. However, I am not sure about all the possible use cases, and one might of course argue that it only takes a few lines to implement if needed.

mateuszbaran:
OK, sure, if you find it useful then we can add it to Manifolds.jl :slightly_smiling_face:

kellertuer (approved):
looks good to me (lgtm).

hajg-ijk:
I still don't understand why codecov/project fails. It laments an indirect change in the circle group, but when you look at it, it has no relation whatsoever to my changes.

kellertuer:
Nighlty is ok to fail and the rest looks good. all is fine.

PR #653, opened 2023-09-22
Fix random point / vector allocations 

### LINKED ISSUE #652: Allocation for TagentSpaceAtPoint fails
With
```
k = 5
N = 20
M = 30
M = FixedRankMatrices(M,N,k)
p = rand(M)
TpM = TangentSpaceAtPoint(M,p)
```
We can easily generate a random tangent vector with
```
rand(M; vector_at=p)
```
which is a UMVTVector as expected, but it seems for `rand` on the tangent space the allocation of the result is wrong, calling
```
rand(TpM)
```
fails because
```
 allocate_result(TpM,rand)
```
creates a `30x20` matrix (from the embedding?) instead of an `UMVTVector`. I could patch this for exactly this one case, but is there maybe a more general way to do this?

kellertuer:
This is the companion PR to https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/168 to fix using TangentSpaceAtPoint on non-matrix-type manifolds.

Resolves #652


kellertuer:
For positive numbers the problem was, that the allocating version has to return a number and can not allocate anything,
for SPDPoint we needed an allocation of a tangent vector – but then now rand should be overall more stable when it comes especially to random tangents.

mateuszbaran (approved):
LGTM. Maybe up to the point that we still might want to have in-place `rand!` for `PositiveNumbers`?

kellertuer:
Sure, then we just have to check that it is also covered by tests, the new (allocating) one basically covers all cases we currently have in the tests.

kellertuer:
The new four uncovered lines are false positives (function headers but the functions are green

PR #654, opened 2023-09-25
Update Documenter.jl to 1.0

mateuszbaran:


docs/src/assets/citations.css
```diff
@@ -0,0 +1,17 @@
+.citation dl {
```
kellertuer, line 1: Overall this looks good to me, but where do these CSS tweaks come from and why do we need them?

kellertuer:
I am surprised about the large reduction of test coverage, any idea where that might come from?

docs/src/assets/citations.css
mateuszbaran: It's from DocumenterCitations.jl's docs: https://juliadocs.org/DocumenterCitations.jl/dev/styling/ . I think it makes bibliography look nicer :slightly_smiling_face: .

mateuszbaran:
It's again that weird issue with first line of method not being covered :confused:

kellertuer:
Seems to just be a hiccup every now and then.

Though I usually just retriever the action (under actions – rerun all) that looks a bit nicer in history compared to the close-reopen-trick.

docs/src/assets/citations.css
kellertuer: Ah ok, maybe we could add that as a comment in the CSS?

mateuszbaran: Sure, I will add a comment.

mateuszbaran: Done.

kellertuer (approved):

PR #655, opened 2023-10-06
Improve performance of selected operations on SE(n)

mateuszbaran:
See https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1546#issuecomment-1749587523 for explanation.

### DIFF
--- a/src/groups/special_euclidean.jl
+++ b/src/groups/special_euclidean.jl
@@ -692,3 +692,27 @@ end
 function project!(M::SpecialEuclideanInGeneralLinear, Y, p, X)
     return copyto!(Y, project(M, p, X))
 end
+
+### Special methods for better performance of selected operations
+
+function exp(M::SpecialEuclidean, p::ArrayPartition, X::ArrayPartition)
+    M1, M2 = M.manifold.manifolds
+    return ArrayPartition(
+        exp(M1.manifold, p.x[1], X.x[1]),
+        exp(M2.manifold, p.x[2], X.x[2]),
+    )
+end
+function log(M::SpecialEuclidean, p::ArrayPartition, q::ArrayPartition)
+    M1, M2 = M.manifold.manifolds
+    return ArrayPartition(
+        log(M1.manifold, p.x[1], q.x[1]),
+        log(M2.manifold, p.x[2], q.x[2]),
+    )
+end
+function vee(M::SpecialEuclidean, p::ArrayPartition, X::ArrayPartition)
+    M1, M2 = M.manifold.manifolds
+    return vcat(vee(M1.manifold, p.x[1], X.x[1]), vee(M2.manifold, p.x[2], X.x[2]))
+end
+function compose(::SpecialEuclidean, p::ArrayPartition, q::ArrayPartition)
+    return ArrayPartition(p.x[2] * q.x[1] + p.x[1], p.x[2] * q.x[2])
+end
--- a/src/groups/special_euclidean.jl
+++ b/src/groups/special_euclidean.jl
@@ -713,6 +713,20 @@ function vee(M::SpecialEuclidean, p::ArrayPartition, X::ArrayPartition)
     M1, M2 = M.manifold.manifolds
     return vcat(vee(M1.manifold, p.x[1], X.x[1]), vee(M2.manifold, p.x[2], X.x[2]))
 end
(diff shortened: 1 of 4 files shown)


Affie:
I tried something similar but missed the SE(N) dispatches.
I did get_vector_orthogonal, exp, log,  and get_coordinates. 
I see you already have get_vector_orthogonal and exp. 
Is it worth it to include log and get_coordinates as well?
this is what I tried:
```julia
function Manifolds.log(M::Manifolds.GeneralUnitaryMatrices{3,ℝ}, p::SMatrix, q::SMatrix)
  U = transpose(p) * q
  cosθ = (tr(U) - 1) / 2
  if cosθ ≈ -1
      eig = Manifolds.eigen_safe(U)
      ival = findfirst(λ -> isapprox(λ, 1), eig.values)
      inds = SVector{3}(1:3)
      #TODO this is to stop convert error of ax as a complex number
      ax::Vector{Float64} = eig.vectors[inds, ival]
      return get_vector(M, p, π * ax, DefaultOrthogonalBasis())
  end
  X = U ./ Manifolds.usinc_from_cos(cosθ)
  return (X .- X') ./ 2
end

function Manifolds.get_coordinates(
  ::Manifolds.GeneralUnitaryMatrices{3,ℝ},
  p::SMatrix,
  X::SMatrix,
  ::DefaultOrthogonalBasis{ℝ,Manifolds.TangentSpaceType},
)
  return SA[X[3, 2], X[1, 3], X[2, 1]]
end
```

src/groups/special_euclidean.jl
```diff
@@ -692,3 +692,27 @@ end
 function project!(M::SpecialEuclideanInGeneralLinear, Y, p, X)
     return copyto!(Y, project(M, p, X))
 end
+
+### Special methods for better performance of selected operations
+
+function exp(M::SpecialEuclidean, p::ArrayPartition, X::ArrayPartition)
```
kellertuer, line 698: would it maybe make sense to change this to the right allocation methods (for the result) and implementing `exp!` instead?

mateuszbaran:
> Is it worth it to include log and get_coordinates as well?
> this is what I tried:

Sure, I will add those too.

src/groups/special_euclidean.jl
mateuszbaran: That's not really an option until Julia gets some form of inter-procedural allocation elision. Right now we can't mutate anything allocated on stack, and stack allocation is what we want here.

kellertuer: Hm, this leave my German-ness for unified code a bit unsatisfied ;)

kellertuer: Ah, but my other-german-ness has to add: Please document these 🤓

mateuszbaran: Yes, it's not particularly satisfying and I hope Julia will get these optimizations.

How much documentation do you expect here? These methods don't introduce any new functionality.

kellertuer: It would increase visibility? For now the user has to know that these exist to use them, right? Or are points on SE(n) already always ArrayPartitions anyways?

mateuszbaran: I don't really intend to write a long explanation on performance for SE(n) here. These methods will be automatically used if user uses `ArrayPartition` with `SArray`, which is standard generic performance advice. Points on SE(n) aren't necessarily `ArrayPartition` but we can add similar optimizations for affine matrices too if needed.

kellertuer: Ok.

kellertuer: Should we maybe – for code clarity and structure – at some point aim for RecursiveArrayTools and StaticArrays becoming weak dependencies? That would collect these performance dispatches in the extension – so in one place?

mateuszbaran: Yes, we could convert both libraries to weak dependencies. `RecursiveArrayTools` more easily than `StaticArrays` because the latter is used in some methods not restricted to StaticArrays types.

kellertuer: Nice. I will open an issue later today to keep track of this (but we also have enough open PRs already by now ;))

Affie:
@dehann, it looks like with these kinds of optimizations we can get down to zero allocations in the factor residual functions.

kellertuer (approved):

PR #657, opened 2023-10-11
up compat BoundaryValDiffEq

dehann:
Hi, I found the following compat issue.  Will let PR CI tests run and see if there are any issues.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -42,7 +42,7 @@ ManifoldsRecipesBaseExt = ["Colors", "RecipesBase"]
 ManifoldsTestExt = "Test"
 
 [compat]
-BoundaryValueDiffEq = "4"
+BoundaryValueDiffEq = "4, 5"
 Colors = "0.12"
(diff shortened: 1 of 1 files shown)


dehann:
Not sure why CompatHelper did not catch this.  I looked around for related issues or PR but didn't immediately see one.

kellertuer:
Interesting; it might be that CompatHelper does not check weak dependencies? At least the CI runs do not look like it would consider those; maybe we should check that those are included as well then.

kellertuer:
Yes, my impression was correct, see https://github.com/JuliaRegistries/CompatHelper.jl/issues/452

mateuszbaran:
Thanks! Before merging it would be good to check if BoundaryValueDiffEq v5 is actually used on CI.

mateuszbaran:
Hm, there seems to be a *huge* problem with BoundaryValueDiffEq v5

mateuszbaran:
I will close this in favor of #672 .

PR #660, opened 2023-10-12
Add more SE2 and SE3 performance dispatches

Affie:
- Added get_coordinates and get_vector specialized dispatches: vee and hat do not always seem to dispatch without allocations and there are places where we use get_vector/coordinates.
- Not only SVector for coordinates: we take the coordinates from a larger vector in some places.

### DIFF
--- a/src/groups/semidirect_product_group.jl
+++ b/src/groups/semidirect_product_group.jl
@@ -173,7 +173,7 @@ end
 # We need to prevent decorator unwrapping so that the correct `get_vector!` gets called
 # and applies proper padding to the result if `X` happens to be a matrix.
 # Otherwise rare random bugs happen where the padding is not applied.
-function get_vector(G::SemidirectProductGroup, p, X, B::AbstractOrthogonalBasis)
+function get_vector(G::SemidirectProductGroup, p, X, B::VeeOrthogonalBasis)
     Y = allocate_result(G, get_vector, p, X)
     return get_vector!(G, Y, p, X, B)
 end
--- a/src/groups/semidirect_product_group.jl
+++ b/src/groups/semidirect_product_group.jl
@@ -173,7 +173,7 @@ end
 # We need to prevent decorator unwrapping so that the correct `get_vector!` gets called
 # and applies proper padding to the result if `X` happens to be a matrix.
 # Otherwise rare random bugs happen where the padding is not applied.
-function get_vector(G::SemidirectProductGroup, p, X, B::VeeOrthogonalBasis)
+function get_vector(G::SemidirectProductGroup, p, X, B::AbstractOrthogonalBasis)
     Y = allocate_result(G, get_vector, p, X)
     return get_vector!(G, Y, p, X, B)
 end
--- a/src/groups/special_euclidean.jl
+++ b/src/groups/special_euclidean.jl
@@ -748,7 +748,7 @@ function hat(M::SpecialEuclidean{3}, p::ArrayPartition, c)
     M1, M2 = M.manifold.manifolds
     return ArrayPartition(
         get_vector_orthogonal(M1.manifold, p.x[1], c[SOneTo(3)], ℝ),
-        get_vector_orthogonal(M2.manifold, p.x[2], c[SA[4,5,6]], ℝ),
+        get_vector_orthogonal(M2.manifold, p.x[2], c[SA[4, 5, 6]], ℝ),
     )
 end
 function get_vector(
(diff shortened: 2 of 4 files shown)


src/groups/semidirect_product_group.jl
```diff
@@ -173,7 +173,7 @@ end
 # We need to prevent decorator unwrapping so that the correct `get_vector!` gets called
 # and applies proper padding to the result if `X` happens to be a matrix.
 # Otherwise rare random bugs happen where the padding is not applied.
-function get_vector(G::SemidirectProductGroup, p, X, B::VeeOrthogonalBasis)
+function get_vector(G::SemidirectProductGroup, p, X, B::AbstractOrthogonalBasis)
```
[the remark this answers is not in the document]
Affie: but it looks like it broke something else

Affie:
@mateuszbaran would you mind helping to implement the performance dispatches I tried to do in this PR? I thought I could help, but I'm still figuring out Manifolds.jl

mateuszbaran:
Sure, I will check it out.

src/groups/special_euclidean.jl
```diff
@@ -713,18 +713,52 @@ function vee(M::SpecialEuclidean, p::ArrayPartition, X::ArrayPartition)
     M1, M2 = M.manifold.manifolds
     return vcat(vee(M1.manifold, p.x[1], X.x[1]), vee(M2.manifold, p.x[2], X.x[2]))
 end
-function hat(M::SpecialEuclidean{2}, p::ArrayPartition, c::SVector)
+function get_coordinates(
+    M::SpecialEuclidean,
+    p::ArrayPartition,
+    X::ArrayPartition,
+    basis::AbstractBasis,
```
mateuszbaran, line 720: I'm trying to fix this but I'm not sure what call this is supposed to handle. In general, you should try to use as narrow basis types as possible here. In particular. not all bases can be propagated like in this method (for example `CachedBasis`). Please try using either of these here:
- `DefaultOrthonormalBasis`
- `DefaultOrthogonalBasis`
- `VeeOrthogonalBasis`,
and similarly in `get_vector`.

Affie: Thanks for the tip, I went with `DefaultOrthogonalBasis`. but I'm not sure what the difference between `VeeOrthogonalBasis `and `DefaultOrthogonalBasis` is.

mateuszbaran: There isn't much of a difference currently, and I'm not sure we even need `DefaultOrthogonalBasis` for anything. If you like using `vee`/`hat` then it's probably better to just use `VeeOrthogonalBasis`.

Affie: To better follow the Manifolds.jl standard, we were moving away from vee/hat to using `get_coordinates/vector` with `DefaultOrthogonalBasis` (I didn't know about `VeeOrthoganolBasis`) so it's used in a few places. We currently only support Lie Groups, but the long-term plan is to support all manifolds in Manifolds.jl.

Affie:
> Not only SVector for coordinates: we take the coordinates from a larger vector in some places.

It looks like the view causes allocations in some instances, so perhaps your original SA is better. Will it be a problem with the coordinates `c::AbstractVector` and not `s::SArray`. What I tested so far looks like it doesn't have allocations.

mateuszbaran:
> It looks like the view causes allocations in some instances, so perhaps your original SA is better. Will it be a problem with the coordinates `c::AbstractVector` and not `s::SArray`. What I tested so far looks like it doesn't have allocations.

`SA` should not be a problem.

src/groups/special_euclidean.jl
```diff
@@ -713,20 +713,55 @@ function vee(M::SpecialEuclidean, p::ArrayPartition, X::ArrayPartition)
...
     return ArrayPartition(
         get_vector_orthogonal(M1.manifold, p.x[1], c[SOneTo(2)], ℝ),
         get_vector_orthogonal(M2.manifold, p.x[2], c[SA[3]], ℝ),
     )
 end
-function hat(M::SpecialEuclidean{3}, p::ArrayPartition, c::SVector)
+function get_vector(
+    M::SpecialEuclidean{2},
+    p::ArrayPartition,
+    c::AbstractVector,
```
mateuszbaran, line 738: I've added `::AbstractVector`. We aim to only represent coordinates of vectors using subtypes of `AbstractVector`.

mateuszbaran (approved):

PR #662, opened 2023-10-20
Introduction of the orthographic retraction for the fixed rank manifold

Klingbiel:


src/manifolds/FixedRankMatrices.jl
```diff
@@ -108,6 +108,111 @@ Base.:-(v::UMVTVector) = UMVTVector(-v.U, -v.M, -v.Vt)
...
+    ::OrthographicInverseRetraction;
+    kwargs...,
+)
+    return inverse_retract_orthographic(M, p, q; kwargs...)
+end
+
+"""
+    inverse_retract_orthographic(M::AbstractManifold, p, q)
+
+computes the allocating variant of the [`OrthographicInverseRetraction`](@ref),
```
mateuszbaran, line 164:
```suggestion
Compute the allocating variant of the [`OrthographicInverseRetraction`](@ref),
```

src/manifolds/FixedRankMatrices.jl
```diff
@@ -496,6 +635,49 @@ the size of matrices on this manifold ``(m,n)``.
...
+```math
+    p + X + N_{p}\mathcal M \cap \mathcal M
+```
+
+where ``N_{p}\mathcal M `` is the Normal Space of ``T_{p}\mathcal M ``.
+
+If ``X`` is sufficiently small, then the nearest such point is unique and can be expressed by
+
+```math
+    q = (U(S + M) + U_{p})(S + M)^{-1}((S + M)V^{\mathrm{T}} + V^{\mathrm{T}}_{p}),
```
mateuszbaran, line 652: Is Eq. (4.9) in Absil and Malik supposed to give the same thing? It looks more computationally efficient in this case.

mateuszbaran (commented):
Thank you for contribution! I've left a few small comments.

kellertuer:
Since we had some (breaking) changes in ManifoldsBase (concerning especially the data flow / allocation flow for retractions) and Manifolds (concerning statically parametrised manifolds) I just adapted this PR to those changes.

@Klingbiel could you check the comments above?

src/manifolds/FixedRankMatrices.jl
Klingbiel: Yes it gives the same thing, but in Absil and Malick they consider the SVD with orthogonal matrices U and V (i.e U,V are square matrices). But in Manifolds.jl we consider U and V as Stiefel matrices, so our SVD and corresponding tangent space look different than in Absil and Malick. Therefore, we cannot use this form of the orthographic retraction directly.

mateuszbaran: I see, thank you for this explanation.

mateuszbaran (approved):

kellertuer (approved):
Everything seems to be resolved – I just did a final commit with

a) Updating (bump) the version number
b) adding an entry to our changelog (NEWS.md)

Great to see further people contributing to Manifolds.jl, danke Lukas!

As soon as tests pass here, I will merge this and register a new version.

PR #666, opened 2023-10-25
Fix #665

### LINKED ISSUE #665: Minor bug with `rand` at `Identity`
I'm trying the very last version of the `Manifolds` ecosystem (love it so far!), and the following code that worked before doesn't anymore:
```julia
G = SpecialOrthogonal(3)
rand(G; vector_at=Identity(G))
```
It's not a big deal, as `rand(G; vector_at=identity_element(G))` *does* work, but perhaps it could hide a potentially more harmful bug? If not, feel free to close the issue right away.

mateuszbaran:
Fixes #665 , and one other minor bug with `is_vector`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `rand(G; vector_at=Identity(G))` now works for translation, special orthogonal and special Euclidean groups `G`.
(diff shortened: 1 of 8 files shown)


kellertuer (approved):
You are right, in general this is not possible to do – I wonder why that did work before then, though. Did we “materialize” (allocate) the identity as an element on the manifold?

kellertuer:
Ah, I saw in other Changelogs, that they would link to errors/issues that are resolved by changes, do we want to do the same? Might be nice in the rendered docs.

mateuszbaran:
I'm not sure, I haven't checked what we did before.

Sure, I can add a link to the issue that was fixed.

PR #670, opened 2023-10-26
Add `get_embedding` for `GeneralUnitaryMultiplicationGroup`

mateuszbaran:
This is needed for example by extrinsic mean calculation.

I've also added compat entries for stdlibs as written here: https://github.com/JuliaRegistries/General/pull/94151.


src/groups/general_unitary_groups.jl
```diff
@@ -136,6 +136,10 @@ function exp_lie!(::GeneralUnitaryMultiplicationGroup{TypeParameter{Tuple{4}},
     return q
 end
 
+function get_embedding(G::GeneralUnitaryMultiplicationGroup)
```
kellertuer, line 139: Would it maybe make sense to define the embedding of an abstract group to be the embedding of the (stored within) manifold?

mateuszbaran: I don't want to go that route. Basic dispatches are not sufficient to prevent incorrect dispatch, and I don't want to add more trait-based dispatch than necessary.

kellertuer: Ok, that's why I asked :)

kellertuer:
You could still copy over the new action for the News.md change check and if you like even the new make.jl args if you like

kellertuer (approved):

mateuszbaran:
I've added the action but I will leave make.jl changes for later. I will try to fix the new bug first but today I'm too tired to figure it out.

kellertuer:
No problem, I can also do the make.jl in a later PR.

PR #672, opened 2023-10-27
Update for BoundaryValueDiffEq v5

### LINKED ISSUE #664: Please update dep of BoundaryValueDiffEq.jl to v5.
This prevents Manifolds.jl to be compatible with latest DifferentialEquations.jl

mateuszbaran:
This may resolve the never-ending solving problem on BoundaryValueDiffEq v5. Resolves #664 . Replaces #657 .

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.3] - 2023-10-28
+
+### Added
+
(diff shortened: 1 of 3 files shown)


kellertuer (approved):
This even looks like a relatively easy fix in the end – are we sure the new tests run on v5? And then maybe check why we currently loose a bit of coverage.

But for the rest (besides of course a short note in the news) this looks nice!

mateuszbaran:
> This even looks like a relatively easy fix in the end – are we sure the new tests run on v5? And then maybe check why we currently loose a bit of coverage.

Yes, Julia v1.9 tests run on BoundaryValueDiffEq v5 while Julia v1.6 tests run on BoundaryValueDiffEq v4. I just don't feel particularly good about tests running more than an hour :disappointed: .

Coverage loss looks like false positives.

PR #673, opened 2023-10-27
Fix `apply_diff_group` for some actions

mateuszbaran:
I've also added differential of matrix inversion.

TODO:

- [ ] Check if this also affects `translate_diff`, `inverse_translate_diff` and adjoint variants.
- [ ] Add tests & check coverage.


src/Manifolds.jl
```diff
@@ -989,6 +989,8 @@ export adjoint_action,
     identity_element!,
     inv,
     inv!,
+    inv_diff,
```
kellertuer, line 992: would `group_inv_diff`  or even `group_inverse_diff` maybe be a better name? Since we use inv already the first might be better...

mateuszbaran: OK, `group_inv_diff` sounds good.

kellertuer: Hm, maybe the current is also fine since the original function is called `inv`,

mateuszbaran:
Sorry that it takes a while but it's fairly convoluted. @olivierverdier if I'm on the right track, it looks like are incorrect exactly up to sign when differentiating wrt. something taken with an inverse.

olivierverdier:
> Sorry that it takes a while but it's fairly convoluted. @olivierverdier if I'm on the right track, it looks like our differentials are incorrect exactly up to sign when differentiating wrt. something taken with an inverse.

Do you think it is a math mistake or an implementation mistake? 🤔

mateuszbaran:
There isn't enough math comments left from the time I've added left-backward and right-forward actions to be sure. I'm just re-doing the math and checking it against implementation.

mateuszbaran:
Could you check if `apply_diff_group` gives the correct result in this PR for group operation actions on SO(3) when `a` and `p` are not identity? I've made some mistakes in my math. The current state of implementation was derived from the Giles' paper I've cited but all these inverses and representation adjustments are easy to get wrong.

mateuszbaran:
This fix was a little harder then I expected but I think this is finished now.

Issue #679, opened 2023-11-08
Implement `translate_diff` and `inv_diff` for all groups at once?


olivierverdier:
Here is a suggestion to decrease the amount of code, and improve the user interface of `Manifolds.jl`.

---

I suggest the **basic method to be implemented to be the left and right adjoint action**. Something like
```julia
adjoint_action!(G::MyGroup, Y, p, X, ::LeftAction) = ... # essentialy, pXp^{-1}
adjoint_action!(G::MyGroup, Y, p, X, ::RightAction) = ... # essentially p^{-1}Xp
```
This has to be implemented for every group (if only the left adjoint action is implemented, there is an easy fallback (can also be made as a test) such as `adjoint_action!(G::MyGroup, Y, p, X, ::RightAction) = adjoint_action!(G, Y, inv(G,p), X, LeftAction())`).

---

Now, **the rest is automatic**.

1) To state the obvious, implement a general `adjoint_action(G, p, X)` by allocation and using `adjoin_action!`, something like
```julia
function adjoint_action(G, p, X, conv)
    Y = allocate_result(G, adjoint_action, p, X, conv)
    return adjoint_action!(G, Y, p, X, conv)
end
```
2) Then also, for backward compatibility:
```julia
adjoint_action(G, p, X) = adjoint_action(G, p, X, LeftAction()) # backward compatibility
```

3) Now with the *left invariance convention* for storing tangent vectors, one has:
```julia
translate_diff!(G, Y, ::Any, ::Any, X, ::LeftForwardAction,) = copyto!(G, Y, X)
translate_diff!(G, Y, ::Any, ::Any, X, ::RightForwardAction,) = copyto!(G, Y, X)
translate_diff!(G, Y, p, ::Any, X, ::LeftBackwardAction,) = adjoint_action!(G, Y, p, X, LeftAction())
translate_diff!(G, Y, p, ::Any, X, ::RightBackwardAction,) = adjoint_action!(G, Y, p, X, RightAction())
```
 Then, *remove* all the specific instances of `translate_diff` that are implemented (except those with a different convention for storing tangent vectors). That is less code to manage and test.
4) and also
```
inv_diff(G, p, X) = -adjoint_action(G, p, X, LeftAction())
```
(and then all the `apply_diff_group` and `apply_diff` for group operation actions come for free, since `inv_diff` is always defined; it is already implemented this way in `group.jl`).

---

**Benefits**: 
[...]

PR #683, opened 2023-11-13
Implement `translate_diff` and `inv_diff` for all groups (#679)

olivierverdier:
The main work of implementing `translate_diff` for all groups is done.

The main method that has to be implemented on any given group is `adjoint_action!`, the rest is done automatically.

- [x] Replace `translate_diff` by `adjoint_action`
- [x] Implement `inv_diff` as well
- [x] Fixes for `RealCircleGroup`
- [x] Fixes for `SpecialEuclidean`
- [x] ~Add the extra erroring methods for `SemiDirectProductGroup`~ (no longer necessary)
- [x] Fix the "wrapped" `SpecialEuclidean` cases (for instance, `ConnectionManifold(SpecialEuclidean(2), CartanSchoutenMinus())`)
- [x] ~Implement `inv_diff` for Product/Power ~manifolds~ groups~ (no longer necessary)
- [x] ~Add tests with a product/power ~manifold~ groups containing a `SpecialEuclidean` or `CircleGroup`, two groups which do not use the left-invariant storage of tangent vectors~ (no longer necessary)
- [x] Implement `adjoint_action` for `SpecialEuclidean`
- [x] Use left invariant storage for `SpecialEuclidean`
- [x] Implement `exp` and `log` for Lie group based on `exp_lie` and `log_lie`


olivierverdier:
The PR is mostly implemented although it doesn't pass all the tests. Two problems essentially remain, which I'm not sure how I can fix:
1. The `RealCircleGroup` stores its elements as float, so the `translate_diff!` which is called after allocation doesn't work. Do you know how to fix it?
2. For a group such as `SpecialEuclidean`, everything works, but stops working when it is wrapped, for instance as a `MetricSpace`, because then the general `translate_diff` methods are called (instead of the `SpecialEuclidean` ones). Do you know how to fix that?

kellertuer:
Hi, I just saw that the format check failed. If you install `JuliaFormatter.jl` (in your favourite environment) and you switch with your current folder into the `Manifolds.jl` base folder (we have an individual setup for formatter specified in that folder) you can run

```
using JuliaFormatter; format(".")
```

to format the code according to our rules. Actually the Formatter check does nothing else than that and checks whether the run changed the code (and if it changed on CI, it complains).

kellertuer:
The other test that is currently not passed is the one that checked, that every PR changes the `NEWS.md` file to report what was changed with that PR (so that it is in the list of changes for the next released version).

kellertuer:
That is one reason why I asked you to start working on this, only when working on this, one notices the remaining gaps in the ideas. I personally usually try to find some time nearly every day to work on these packages here, but for the next few weeks, also called semester end time, this will not be possible from my side, due to too many other duties.

olivierverdier:
I’d be happy to finish this myself if you give me pointers as to how to proceed, that’s all I’m asking.

mateuszbaran:
Thank you for this work, I will try to help finish this. I need to clarify a couple of things first.

If I understand correctly, adjoint_action for the left action is differential of $Ψ_p(q) = pqp^{-1}$, while for the right action it is the differential of $Ψ_p(q) = p^{-1}qp$? Also, why did you remove `inv_diff` and `translate_diff` for product and power manifolds? That would only work if `inv_diff` and `translate_diff` are never going to be specialized for particular manifolds.

> 1. The `RealCircleGroup` stores its elements as float, so the `translate_diff!` which is called after allocation doesn't work. Do you know how to fix it?

I will fix that myself.

olivierverdier:
> Thank you for this work, I will try to help finish this. I need to clarify a couple of things first.

Thank you, that's much appreciated! 

> If I understand correctly, adjoint_action for the left action is differential of Ψp(q)=pqp−1, while for the right action it is the differential of Ψp(q)=p−1qp? 

Yes, exactly. Let's fix $ψ_p(q) := p q p^{-1}$. Then the adjoint action is the derivative of $ψ_p$ at the identity. The right action is the left adjoint action, but at $p^{-1}$ so the derivative of $ψ_{p^{-1}}$ at the identity, as you say (is is the same relation as between any other left/right actions).

> Also, why did you remove `inv_diff` and `translate_diff` for product and power manifolds? That would only work if `inv_diff` and `translate_diff` are never going to be specialized for particular manifolds.

Excellent point, I made a mistake there. It would only work if all the group use the left-invariant storage, which shouldn't be assume indeed.

> 
> > 1. The `RealCircleGroup` stores its elements as float, so the `translate_diff!` which is called after allocation doesn't work. Do you know how to fix it?
> 
> I will fix that myself.

Thanks a lot!

olivierverdier:
> Also, why did you remove `inv_diff` and `translate_diff` for product and power manifolds? That would only work if `inv_diff` and `translate_diff` are never going to be specialized for particular manifolds.

This should be fixed.

However, the fact that the tests didn't catch that indicates that all the tests for product/power manifolds never involve groups with no left-invariant storage, so it could be an idea to add such tests?

olivierverdier:
Unless I am mistaken, `inv_diff` was not implemented for product/power manifolds, but it should be (similarly to `translate_diff`) if one of the components does not use the left-invariant storage. I add that as a task.

kellertuer:
just to aim for the same precision as you do – software wise we try to split for some time, to have first manifolds (e.g. power manifold and product manifold) and have the Lie part as an add-on (not yet 100% consistent, since some time all new stuff is, cf. rotations and their split from SO(n)).

So with that in mind, please implement `inv_diff` for the product and power _group_ and not the manifold :)

olivierverdier:
> So with that in mind, please implement `inv_diff` for the product and power _group_ and not the manifold :)

Yes, I meant "product/power group", as `inv_diff` doesn't make sense for manifolds.

For the record, before this PR, `inv_diff` was not available at all for product/power groups, as far as I know.

kellertuer:
Sure, sometimes we (a) might not have implemented existing methods for a new manifold directly or (b) when implementing a new method we might not have taken the whole ecosystem under revision; both cases might espcially happen, when a method is still new and needs more experience or an “external contributor“ started them.

mateuszbaran:
Thanks for clarification, I should have some time tomorrow to work on this PR.

mateuszbaran:
I've fixed `CircleGroup` and `RealCircleGroup`. They can be sometimes a little annoying to work with because numbers can't be allocated and we need specialized non-allocating methods.

kellertuer:
HI @olivierverdier,
do you have plans and time to continue this? That would be great. We could help merging the current master here for example.

I think in the long run it would be really great to move the Lie group parts to a dedicated `LieGroups.jl` package (depedingin on ManifoldsBase and Manifolds – we could also do a `LieGroupsBase.jl` just depending on ManifoldsBase, but that can also be done in a later step). Such a “restart” is maybe also a good point to rethink a few of the design choices.

olivierverdier:
No, I don't have plans to work on this. The main issue is that some of the groups store tangent vectors differently than others (I think it is `SpecialEuclidean` and `RealCircleGroup`), and that creates a lot of unnecessary complications. 

My personal solution: I'm running my own version of `Manifolds.jl` which is this PR along with these complications removed and I'm happy with that solution so far.

kellertuer:
That  is interesting to know.

I think the reason here might be that until now we often inherited the representation from the manifold – which for some manifolds might not make sense, you are right.

I think switching to your variant might also be a good idea, with the small caveat, that this would be breaking.
on the other hand, if your new representation is worth that, maybe it is worth releasing a breaking change.

kellertuer:
If you do not plan to propose a breaking change here, we should maybe see how we can adapt this PR (maybe @mateuszbaran you can take a look)? And we should close this PR.

olivierverdier:
The essential difference in my version is that I just removed the `translate_diff!` method in `semidirect_product_group.jl`. This allows me to use the `SemiDirectProductGroup` the way I wanted, with left translated vectors as implemented in this PR.

(I also had to remove the `get_vector`, `get_vector!` and `get_coordinates!` method for the `VeeOrthogonalBasis`, in that same file, for some reason, I can't remember why)

mateuszbaran:
One possible approach would be to allow both representations and have a flag in Lie groups that tells which one is used. That's something I could do but there are a few other things I'd like to do first (some extensions and improvements to Manopt, maybe pushing a bit forward my work on fiber bundles and state estimation).

kellertuer:
If we allow both, we could do that maybe with different tangent vector types? Until now that was the way we distinguished different tangent vector types.

kellertuer:
That one could counter a bit by making our current modelling of “what does an array represent” a bit more public (I am no sure how public that is now directly), so a user can easily (or easier) specify his/her default behaviour of what tangent vectors are usually represented in.

mateuszbaran:
> That one could counter a bit by making our current modelling of “what does an array represent” a bit more public (I am no sure how public that is now directly),

Currently it's up to each manifold what an array means but that is a part of the public interface (of that particular manifold).

> so a user can easily (or easier) specify his/her default behaviour of what tangent vectors are usually represented in.

And that would have to be done through a flag in the manifold/Lie group as far as I can see.

kellertuer:
I see. What I meant was is that currently we have lines like

```
ManifoldsBase.@default_manifold_fallbacks Hyperbolic HyperboloidPoint HyperboloidTVector value value
```

and sure these are macros that define the default (fallback) that specify what arrays are the same as. We could also have a function dispatch for that instead of a parameter maybe?

Sure in the end we have to store the information somewhere, just that maybe a dispatch is nicer than a parameter in a struct (since we would have to add that to a lot of structs).

But this is maybe also a discussion beyond this PR.
The question here is, how do we resolve the bugs mentioned / noticed here? Or is that just a matter of how to represent tangent vectors?

mateuszbaran:
> and sure these are macros that define the default (fallback) that specify what arrays are the same as. We could also have a function dispatch for that instead of a parameter maybe?
> 
> Sure in the end we have to store the information somewhere, just that maybe a dispatch is nicer than a parameter in a struct (since we would have to add that to a lot of structs).

That dispatch is very different though, I don't see how we could re-use a trick like that for having two different representations of tangent vectors.

> But this is maybe also a discussion beyond this PR.
> The question here is, how do we resolve the bugs mentioned / noticed here? Or is that just a matter of how to represent tangent vectors?

IIRC there were no bugs addressed here, only change of representation and implementation of a few functions that are currently not implemented in Manifolds.jl for some groups. New implementations can be added without change of representation but only in a less nice way.

kellertuer:
You are right, it might not be super easy.

Ah, I did not see that in the changes so clearly. But then we can close this PR?

mateuszbaran:
Closing this PR would make it easier to forget about it, and I don't want to forget. It has valuable contributions to Lie groups which I would like to incorporate into Manifolds.jl next time I have time to work on this area.

olivierverdier:
[Motivation for this PR](#679)

kellertuer:
Hm, I think you meant to refer to #679 ? I am aware of that issue, that is why I was wondering whether this PR was still an active approach to fixing this. Thank for the info, that that is not the case any longer. But then one of the developers will take care of this, when our own schedules permit – I for example am a bit busy until end of May.

olivierverdier:
This PR is finished. It assumes left-invariant storage for all the groups.
Now, when creating a new group:
- implementing `adjoint_action` allows to use `translate_diff` in all its flavours, as well as `inv_diff`.
- implementing `exp_lie` and `log_lie` allows to use `exp` and `log`

And more specifically, now the adjoint action is finally defined for the special euclidean group.

kellertuer:
We would have to check with the conflicts to current master (hopefully that's mainly the changelog),

But great to read that you continued this! Mateusz is probably the better person to review the last changes.
Then we just also check for tests and documentation.

But if we can push to this branch, I can surely help with tests and docs as well.

kellertuer:
I checked the conflicts to master (but I think I can not even push to your branch here in the GitHub web interface, where I tried that on purpose)

in groups and power groups it seems both on master and here similar lines where changed (just a few). Would be great if you could check that

In the news just add your 3 entries to the current unreleased 0.9.19 version and keep all existing content from master.

When that is resolved, the tests will run.

mateuszbaran:
Thank you for continuing your work here. I will try to allocate enough time during this summer to review and hopefully merge this PR.

kellertuer:
Cool!
Version 8.19 was in preparation but not yet registered (nothing urgent basically) so yo can add your changes to that version as well :) the rest looks good.

Can you check the initial post? It has a list. Have you done the points that are still unchecked? Maybe if you do not plan to do them, strike them out?

mateuszbaran:
I've taken another look at this PR. Here are some notes:
1. Adding direction to `adjoint_action` is completely fine and can be made non-breaking.
2. I don't like the idea of abandoning support for the current tangent vector representation. For all groups without left-invariant storage, I would suggest adding such option (but the default is still the current behavior to avoid breaking changes). In the future we could change the default if it appears to be the better default.
3. I don't think the defaults are going to work with disconnected groups outside of their identity component. It could then cause wrong results for other components which would be difficult to track. As a solution I propose adding an opt-in flag to `IsGroupManifold` and `GroupManifold` that enables these defaults, for example an additional type parameter `TangentStorage` which can be `LeftInvariantStorage` and if so, your defaults are activated.

olivierverdier:
Thank you for reviewing this PR. It took me a lot of time to write and I appreciate that you take time to look into it thoroughly.

I also appreciate that you had this great idea of using left invariant storage on groups. I just pushed the idea a little further.

---

>  1. Adding direction to `adjoint_action` is completely fine and can be made non-breaking.

Great!

>  2. I don't like the idea of abandoning support for the current tangent vector representation. 

This PR does not *force* the user to any representation at all. 

It's just that if you **optionally** use left-invariant storage and implement adjoint action, **the other methods are implemented for you automatically**. 

**You are absolutely free to use another storage method**. This means implementing all the methods accordingly (but that is a *lot* of work, of course). An example was the previous implementation of SE(n), which was largely incomplete. The current implementation of SE(n) is **shorter** and **complete**. What's not to like about that?

> For all groups without left-invariant storage,

There aren't any, currently. All groups use the left invariant storage. Again, it means that all the standard methods are automatically defined for them, all this with **less code** and **more functionality**.

(For the record, before this PR, all the groups except for SE(n) were already using left-invariant storage)

>  3. I don't think the defaults are going to work with disconnected groups outside of their identity component.

I understand your concern, however (a) left invariant storage was already implemented for disconnected groups before this PR (b) *all the tests pass*, even with disconnected groups. (c) the approach is mathematically sound, and works on **all Lie groups** no matter what (connected or not).

But if you figure out a concrete example where this (*optional*) approach does not work, and that the tests would then have missed, please let me know.

---
Let me know if you have more questions about this PR.

mateuszbaran:
> Thank you for reviewing this PR. It took me a lot of time to write and I appreciate that you take time to look into it thoroughly.

I also appreciate your contribution :+1: . My main remaining concern is that I would like to keep the current storage of `SpecialEuclidean` (and `CircleGroup`, which is the other exception to left-invariant storage) as an option, even in its incomplete state. I can work on it myself though.

> There aren't any, currently. All groups use the left invariant storage. Again, it means that all the standard methods are automatically defined for them, all this with **less code** and **more functionality**.

> I understand your concern, however (a) left invariant storage was already implemented for disconnected groups before this PR (b) _all the tests pass_, even with disconnected groups. (c) the approach is mathematically sound, and works on **all Lie groups** no matter what (connected or not).

Yes, thanks for pointing it out, I've reconsidered it and it indeed looks sound.

olivierverdier:
I stand corrected, `CircleGroup` does indeed not use left-invariant storage. (As of this PR, it is the only group which does not.) This proves the point that it is possible to opt out of the left invariant storage.

In the future, one could be given the choice of the storage for a given group. There are only a handful of methods which depend on the storage. I believe it's only `translate_diff` and `inv_diff`. (But such a change would probably warrant another separate PR.)

src/groups/group.jl
```diff
@@ -426,7 +426,7 @@ function is_vector(
 end
 
 @doc raw"""
-    adjoint_action(G::AbstractDecoratorManifold, p, X)
+    adjoint_action(G::AbstractDecoratorManifold, p, X, dir)
 
 Adjoint action of the element `p` of the Lie group `G` on the element `X`
```
kellertuer, line 431: Could we add a small comment about the new `dir` here as well?
And, is this non-breaking? Sure internally with dispatches the new argument is fine, but this is exported, so how are those potentially affected?

src/groups/group.jl
```diff
@@ -562,12 +581,18 @@ inv_diff(G::AbstractDecoratorManifold, p)
...
 @trait_function inv_diff(G::AbstractDecoratorManifold, p, X)
 function inv_diff(::TraitList{<:IsGroupManifold}, G::AbstractDecoratorManifold, p, X)
-    Y = allocate_result(G, inv_diff, X, p)
-    return inv_diff!(G, Y, p, X)
+    return -adjoint_action(base_group(G), p, X)
 end
 
 @trait_function inv_diff!(G::AbstractDecoratorManifold, Y, p, X)
 
+# true only if tangent vectors are stored with the left-invariant convention
```
kellertuer, line 589: Do we ensure that somehow or is this implicitly assumed to be the default?

src/groups/group.jl
```diff
@@ -913,6 +938,46 @@ end
     X,
     conv::ActionDirectionAndSide=LeftForwardAction(),
 )
+# the following are true if the tangent vectors are stored with the left invariant convention
```
kellertuer, line 941: Same question: Do we assume this implicitly or make that sure somehow?

src/groups/group.jl
```diff
@@ -1121,6 +1186,43 @@ end
 direction_and_side(::GroupExponentialRetraction{D}) where {D} = D()
 direction_and_side(::GroupLogarithmicInverseRetraction{D}) where {D} = D()
 
+function log(::TraitList{<:IsGroupManifold}, G::AbstractDecoratorManifold, p, q)
+    # TODO: use the following version when `allocate_result` is fixed
```
kellertuer, line 1190: Maybe more something for Mateusz – what needs to be fixed here?

src/groups/semidirect_product_group.jl
```diff
@@ -135,41 +135,6 @@ function _compose!(G::SemidirectProductGroup, x, p, q)
     return x
 end
 
-@doc raw"""
```
kellertuer, line 138: Are these docs kept somewhere?
Or phrased differently: It is maybe useful to keep these attached to a new function or at least to the signature here?

src/groups/special_euclidean.jl
```diff
@@ -619,28 +582,15 @@ function lie_bracket!(G::SpecialEuclidean, Z, X, Y)
     return Z
 end
 
-"""
```
kellertuer, line 622: Same question here, we could keep the docs attached to the signature.

src/groups/special_euclidean.jl
```diff
@@ -721,7 +670,7 @@ function project(M::SpecialEuclideanInGeneralLinear, p, X)
     G = M.manifold
     np, hp = submanifold_components(G, p)
     nX, hX = submanifold_components(G, X)
-    return ArrayPartition(hp * nX, hX)
+    return ArrayPartition(nX, hX)
```
kellertuer, line 673: Does that mean we changed representation of tangent vectors? Just curious, whether this is breaking.

src/groups/special_linear.jl
```diff
@@ -78,8 +78,9 @@ function get_embedding(M::SpecialLinear{Tuple{Int},𝔽}) where {𝔽}
     return GeneralLinear(n, 𝔽; parameter=:field)
 end
 
-inverse_translate_diff(::SpecialLinear, p, q, X, ::LeftForwardAction) = X
-inverse_translate_diff(::SpecialLinear, p, q, X, ::RightBackwardAction) = p * X / p
+# note: this implementation is not optimal
```
kellertuer, line 81: Could we write that as `(p * X)  / p` to avoid the inv?

src/groups/circle_group.jl
```diff
@@ -24,9 +24,13 @@ end
 
 Base.show(io::IO, ::CircleGroup) = print(io, "CircleGroup()")
 
-adjoint_action(::CircleGroup, p, X) = X
+adjoint_action(::CircleGroup, p, X, ::LeftAction) = X
```
kellertuer, line 27: Is it breaking that we no longer have the original signature `adjoint_action(::CircleGroup, p, X)`? Or is there a global default that fills in a `LectAction()`for that case (maybe I missed that then)?

kellertuer (commented):
Thanks for your thorough contribution – much appreciated.

I looked through the code, and I added a few small questions.

besides that it would maybe be great to document the left-right-action thing on `group.md` a bit more so an only midly-Lie-group-experienced person like me has a chance to read up on the detail fixed here?
If that would be possible that'd be great.

src/groups/group.jl
olivierverdier: It is implicitly assumed to be the default.

src/groups/group.jl
olivierverdier: Same here. The purpose of those comments is to emphasise precisely which functions depend on that storage convention.

src/groups/group.jl
olivierverdier: One problem is that `allocate_result` does not respect the underlying numerical type of the group. So if `G` is a complex group, the allocated variable is real. There are other problems: some groups do not use vectors (e.g. `CircleGroup`, and allocation does not work.

src/groups/special_euclidean.jl
olivierverdier: Yes, the representation is now left invariant instead.

src/groups/special_linear.jl
olivierverdier: I don't think you can avoid the `inv` because `X` is a matrix. What I meant is that one can probably avoid some unnecessary allocations (the allocated `Y` is not used except in the end with a `copyto`).

src/groups/circle_group.jl
olivierverdier: Yes, it is a global default. You can always use `adjoint_action(G, p, X)`, and this is redirected to `adjoint_action(G, p, X, LeftAction())`.

src/groups/semidirect_product_group.jl
olivierverdier: I think that this docstring is now obsolete.

src/groups/special_euclidean.jl
olivierverdier: Same answer.

src/groups/group.jl
olivierverdier: Yes, the doc was missing here, thank you. I just pushed a new commit that should address this.

olivierverdier:
> besides that it would maybe be great to document the left-right-action thing on group.md a bit more so an only midly-Lie-group-experienced person like me has a chance to read up on the detail fixed here?

I agree. I'll try to do that.

src/groups/special_linear.jl
kellertuer: I did not check that yet but I can, it just reminded me a lot of the $p^{1/2}Xp^{-1/2}$ and such we do on SPDs, where `/`is exactly what we do; and sure both `p` and `X` are matrices.

src/groups/special_euclidean.jl
kellertuer: Ok, but this is a breaking change isn't it?

src/groups/semidirect_product_group.jl
kellertuer: Ok.

src/groups/group.jl
kellertuer: I think, then it would be better to document that somewhere in a doctoring and remove this comment.
The comment reads a bit like there is something left to do in the code for the case it is not stored like that.

src/groups/group.jl
kellertuer: Mateus is the allocation expert (it works for the circle and complex manifolds for example), so then we should adapt that for sure.

src/groups/special_euclidean.jl
kellertuer: To be clear, I do not mind doing a breaking release, just that _if_ we only do them very carefully in order to not do them too often; until now we had 8 breaking releases in a bit more than 5 years, recently more like once a year or less. 

If this is breaking we should definetly remark that in the News.md what we broke and how to change existing code to work the same as before. That's the reason for the question.

mateuszbaran: I'm going to make some additional work myself so that the old representation of SE(n) can still be used but there still likely will be something breaking.

src/groups/group.jl
mateuszbaran: I will take a look what is the issue with the `allocate_result` variant. We generally assume in generic code that points and tangent vectors are mutable and make manifold-specific adjustments when they are not (like for example for `CircleGroup`).

src/groups/special_euclidean.jl
olivierverdier: If you don't want breaking change, just rename the current `SpecialEuclidean` to something else, like `Displacement`, and keep the old, incomplete implementation of `SpecialEuclidean`.

src/groups/group.jl
olivierverdier: Here is an example:
```julia
Q2 = Unitary(2, ℍ)
allocate_result(Q2, log) # matrix of floats instead of quaternions
```
I suggest to remove this comment for now.

src/groups/group.jl
olivierverdier: Done: put in the docstring.

src/groups/group.jl
olivierverdier: Done: put in the docstring.

src/groups/group.jl
olivierverdier: I've removed the TODO. This can be fixed later.

src/groups/special_linear.jl
olivierverdier: I removed `inv`.

src/groups/special_euclidean.jl
kellertuer: No, we have a few things we could do for a breaking change, just that we consider that carefully. Here, that the tangent vectors are then unified for the Lie groups is a good reason – my main point is then to carefully document (in the News.md) what a user has to change in their code.

olivierverdier: If you have a **stored** tangent vector `X` at point `p` of `G = SpecialEuclidean(n)`, you would have to replace it with `translate_diff(G, inv(G,p), p, X)`. **Unfortunately** this method (`translate_diff`) is not available in `SpecialEuclidean` (because the current implementation of `SpecialEuclidean` is incomplete).

Other than that, (if you do not *store*, or manually manipulate, any tangent vectors), the transition should be transparent.

mateuszbaran: I would just add a new type parameter for `SpecialEuclidean` that determines the storage. AFAIK JuliaRobotics currently relies on our storage and I wouldn't like to force them to adapt everything.

olivierverdier:
> > besides that it would maybe be great to document the left-right-action thing on group.md a bit more so an only midly-Lie-group-experienced person like me has a chance to read up on the detail fixed here?
> 
> I agree. I'll try to do that.

Done. Let me know if it is understandable.

docs/src/manifolds/group.md
```diff
@@ -43,6 +43,56 @@ Pages = ["groups/GroupManifold.jl"]
 Order = [:constant, :type, :function]
 ```
 
+### Default Storage of Tangent Vectors
+
+In most groups, the storage of a tangent vector
```
kellertuer, line 48: I would maybe prefer to speak about representation instead of storage.

Storage for me sounds like an _extra_ memory allocated for something (maybe some `X.p` such that `X` knows its base point); you do not do that, the text just reads a bit like that.

kellertuer: And I think representation is not much my problem, we do a very similar “triick” on SPDs – the point why we need the action is maybe something I would like to understand a bit more (and since I want to understand it, documenting it somewhere would be great).

olivierverdier: Which action are you talking about here? The adjoint action? 

The piece of documentation I wrote explains exactly why the adjoint action is needed to compute the derivatives of the group operations when using left invariant ~storage~ representation of tangent vectors.

What else would you need in the docs?

kellertuer: Ah, I did not read that careful enough, yes that should suffice for now. Also thanks for using representation, that is what we usually use for that as well when talking about different representations of points or tangent vectors.

src/groups/group.jl
kellertuer: @Mateusz should we open an issue for this to keep that in mind? Besides that this comment is finished :)

src/groups/special_euclidean.jl
kellertuer: One idea could be to use some `TangentvectorRepresentation` and `LIeAlgebraRepresentation` for that distinction? Do we have a `TVectorRepresentation` supertype? If not we could either introduce that or use subtypes of `TVector` (though that might be misleading).

mateuszbaran: So far we've only differentiated representations by custom types, so it will be a new thing. I think we could have `TVectorRepresentation` abstract type for that purpose.

src/groups/group.jl
mateuszbaran: I will fix that quickly in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/192 since that is also related to allocation and not merged yet.

kellertuer: Then this point here is resolved.

mateuszbaran: Yes, I've just pushed the fix.

mateuszbaran:
I'm going to merge this PR into #732 next week and then make necessary adaptations there. Thank you again for this contribution :+1:

src/groups/special_linear.jl
```diff
@@ -153,9 +153,4 @@ function Base.show(io::IO, M::SpecialLinear{Tuple{Int},𝔽}) where {𝔽}
     return print(io, "SpecialLinear($n, $(𝔽); parameter=:field)")
 end
 
-translate_diff(::SpecialLinear, p, q, X, ::LeftForwardAction) = X
-translate_diff(::SpecialLinear, p, q, X, ::RightBackwardAction) = p \ X * p
-
-function translate_diff!(G::SpecialLinear, Y, p, q, X, conv::ActionDirectionAndSide)
-    return copyto!(Y, translate_diff(G, p, q, X, conv))
-end
+adjoint_action!(G::SpecialLinear, Y, p, q, X, conv::LeftAction) = p \ X * p
```
mateuszbaran, line 156: I don't think this method looks correct? There are two `adjoint_action!` methods for `SpecialLinear` above so it's probably not needed.

olivierverdier: This method doesn't make any sense: wrong number of arguments, not using the pre-allocated `Y`, it's all wrong.

mateuszbaran:
By the way, I've already merged this PR into #732 and I will polish it there.

src/groups/group.jl
```diff
@@ -1121,6 +1200,35 @@ end
 direction_and_side(::GroupExponentialRetraction{D}) where {D} = D()
 direction_and_side(::GroupLogarithmicInverseRetraction{D}) where {D} = D()
 
+function log(::TraitList{<:IsGroupManifold}, G::AbstractDecoratorManifold, p, q)
```
mateuszbaran, line 1203: In my PR I've restricted these `log` and `exp` methods to semidirect product groups with with left-invariant tangent vector representation. On many groups we have optimized implementation on the manifold so I'd prefer to have that to be picked up first before this relatively generic fallback.

olivierverdier: The solution here is to call the exponential/logarithms defined in `groups/connections.jl`. It's the only acceptable `exp` and `log` on a group manifold. This way, it will work for *all* groups, regardless of the tangent vector representation. Should I fix this in this PR?

mateuszbaran: I know your solution is correct, this is mostly about performance (and a bit about consistency, see the next paragraph). Substantial amount of work has been put to make practically relevant calls fast but unfortunately it's not gathered in an easy to run benchmark suite, so verifying what could have been slowed down and fixing issues would take me hours. So, unless you know of a case where my solution gives a wrong result for an existing group, I'd prefer to keep it like that for now.

Additionally, considering your implementation here makes me rethink our entire group -> manifold pass-through and it's not really something I'd like to do in the near future. If we don't pass exp and log to the manifold, why do we pass vector transports, inner product, retractions, inverse retractions, Riemann tensor, affine connection and so on? Maybe the current solution is not ideal but reworking it would take a very long time and I just don't see enough benefits.

> Should I fix this in this PR?

Since I've already merged your branch, it would be easier for me if you made any further fixes on top of the branch from #732 .

kellertuer: vector transport and retractions (and their inverses) are still valid, at least after turning back to tangent vectors in tangent spaces (not 100% sure about Lie algebra things); but it might be beneficial to maybe not pass through exp/log. But I agree that that is something that has to be considered carefully and in the near future I do not have the capacity for that.

olivierverdier: Thank you for your feedback and explanations! I hope I did not completely misunderstand.

1. Before my PR, the `exp` that was used was not necessary the group exponential (the one in `groups/connections.jl`). For instance, for semidirect products, the exponential of the underlying *product* manifold was used, which was wrong, because it was not invariant. My implementation fixed that, but was also flawed since it assumed left invariant representation. This is easily fixed as I mentioned above. I'd be happy to do that.

2. I'm not sure I understand the implications of what you call the "pass through" here... I'd be happy to be enlightened.
As far as speed is concerned, `exp` is always an expensive operation, and I doubt that the overhead of function calls makes any difference. Am I wrong to assume that?

3. The only optimized `exp` that I can find in the code is that of `GeneralLinear`. It comes from a choice of left invariant metric, and I'm pretty sure it is not the same as the invariant exponential in `groups/connections.jl`. (So that optimized exponential is *not* invariant, unless I am mistaken) Have I missed any other optimized `exp` implementation?

mateuszbaran: > 1. Before my PR, the `exp` that was used was not necessary the group exponential (the one in `groups/connections.jl`). For instance, for semidirect products, the exponential of the underlying _product_ manifold was used, which was wrong, because it was not invariant. My implementation fixed that, but was also flawed since it assumed left invariant representation. This is easily fixed as I mentioned above. I'd be happy to do that.

Yes, `exp` and `log` were not invariant because we chose to separate Riemannian and group operations into two categories and on group manifolds the Riemannian operations were operating on the manifold, with possibly non-invariant metric. `exp` and `log` can indeed be taken from the group structure but back when the decision to pass through was made, we didn't have Cartan-Schouten connections figured out yet. I'm open to changing this but as I said, it would be a bigger project. Right now to have an invariant `exp` and `log` you are supposed to use the `ConnectionManifold` type, which also selects the desired parallel transport (which is in general not unique for a group).

I'd say the current state is consistent enough to keep for a while until we have time to rethink the Riemannian/connection manifold operations we want on groups.

> 2\. As far as speed is concerned, `exp` is always an expensive operation, and I doubt that the overhead of function calls makes any difference. Am I wrong to assume that?

Actually it's not always expensive. On SE(3) with the old tangent vector representation we got down to about 50 ns using StaticArrays. Trait dispatch cost can easily reach an order of magnitude more. It's low enough that we can have it as a default but not low enough to never care about it.

> 3\. The only optimized `exp` that I can find in the code is that of `GeneralLinear`. It comes from a choice of left invariant metric, and I'm pretty sure it is not the same as the invariant exponential in `groups/connections.jl`. (So that optimized exponential is _not_ invariant, unless I am mistaken) Have I missed any other optimized `exp` implementation?

GL(N) in general does not admit a bi-invariant metric so the choice described in its docstring sounds fair to me. If the implementation doesn't match the docstring please make an issue.

Additionally we have some optimized implementations for `GeneralUnitaryMatrices`, SE(2) and SE(3).

olivierverdier: The invariant exponential for all Lie groups (regardless of tangent vector representation) is that one:
```julia
exp(G, p, X) = compose(G, p, exp_lie(G, inverse_translate_diff(G, p, p, X, LeftForwardAction())))
```

This implementation is **canonical** (i.e., the only invariant one) and works on any Lie group. I think that it should be the "reasonable default". 
(It does *not* depend on the choice of Cartan–Schouten parallel transport, which cannot be canonical). 

---

On SE(N) the previous implementations of `exp` were *not invariant*. This is not acceptable in robotics (the orientation of the robot at identity is arbitrary, like a choice of origin). (For SE(3), I do not know if it was invariant or not)

On GL(N), the canonical implementation above is invariant and canonical even though there are no bi-invariant metrics (and no invariant connections either).

As to `GeneralUnitaryMatrices`, I believe it is in fact an optimized implementation of `exp_lie`, but I may be wrong? Anyhow, it is a *manifold* exponential, not a *group* exponential. It should be the difference between `GeneralUnitaryMatrics` (the manifold) and `GeneralUnitaryMultiplicationGroup` (the group).

---

So I guess we fundamentally disagree for now: I would like to see the invariant exponential implementation by default for all Lie groups (even if it can be specialized), and, for now, you do not. Is that a correct assessment?

mateuszbaran: > On SE(N) the previous implementations of exp were not invariant. This is not acceptable in robotics (the orientation of the robot at identity is arbitrary, like a choice of origin). (For SE(3), I do not know if it was invariant or not)

> So I guess we fundamentally disagree for now: I would like to see the invariant exponential implementation by default for all Lie groups (even if it can be specialized), and, for now, you do not. Is that a correct assessment?

As I said, in the current Manifolds.jl you are supposed to use `ConnectionManifold(SpecialEuclidean(3), CartanSchoutenMinus())` (or some other connection) if you want `exp` to be the invariant one. You don't have to use `exp` on `SpecialEuclidean`. I'm open to changing the default of `exp` and `log` on groups but not by just patching these two functions but by not passing any metric-related functions to the underlying manifold. So from traits of groups with non-bi-invariant metrics you'd have to remove `IsEmbeddedManifold`, `IsDefaultMetric`, `active_traits(f, M.manifold, args...)` and most likely `IsExplicitDecorator` and see if everything that's supposed to work still works. Then check performance regressions and finally update documentation. It would probably take me a few days to sort it out and there is an easy workaround so I can't justify to myself that it's a productive way to spend time.

In the meantime we can maybe have an alias for `ConnectionManifold(SpecialEuclidean(N), CartanSchoutenMinus())`?

> As to GeneralUnitaryMatrices, I believe it is in fact an optimized implementation of exp_lie, but I may be wrong? Anyhow, it is a manifold exponential, not a group exponential. It should be the difference between GeneralUnitaryMatrics (the manifold) and GeneralUnitaryMultiplicationGroup (the group).

I thought these have a bi-invariant metric?

olivierverdier: Just to be perfectly clear: in general, on any manifold, `exp` only depends on a **connection modulo its torsion**. (Example: all the Cartan–Schouten connections are the same except for their torsion; any of them defines the same exponential). To be clearer: a metric is not needed at all for geodesics (and thus exp) to be defined.

So, what I am suggesting is that all groups are by default endowed with one of the Cartan–Schouten connections.

If I had to choose one, it would be `CartanSchoutenZero`, since it has zero torsion.

So, any of the following would be acceptable solutions for me (it's ok if you don't want any of them of course!):
1. *No exponential at all on groups.* You'd have to explicitly choose a connection. This way, mistakes are not possible.
2. *All groups are endowed with one of the Cartan–Schouten connection by default* (preferably `CartanSchoutenZero`, although it doesn't matter for `exp`). Is it easy to do?
3. *All groups have `exp` and `log` defined invariantly*. That was my choice, it's easy to do, and all the tests were passing.

For performance reason, you do not want 3., but it's the easiest to implement, I guess?

(The existence of a bi-invariant metric is irrelevant, in my opinion. Some groups have it some don't, but they all have an invariant connection, any of which gives the same exponential.)

> In the meantime we can maybe have an alias for ConnectionManifold(SpecialEuclidean(N), CartanSchoutenMinus())?

Yes, but can this be done for all groups, as suggested above in 2.? Something like that (alias or default) would be an acceptable solution, I think.

mateuszbaran: I'm OK with either (2) or (3) but neither of them is easy to do right I fear. The problem with (3) is performance and that nothing related the the Riemannian structure of the underlying manifold should work. We don't have test for either of those things currently and just because we don't doesn't mean we don't care about them. If tests are passing with your change, it means there is not enough tests.

Keep in mind that the current design of Manifolds.jl was centered around Riemannian manifolds and groups were seen as a useful addition. Most of the things regarding Lie group theory you mention here are things I wasn't aware of at the time the design was made but I've learned about them around the time I've added Cartan-Schouten connections, which I did in a way consistent with existing design choices. I can see the drawbacks but at least it has some consistent logic and just replacing `exp` and `log` would violate that logic.

> > In the meantime we can maybe have an alias for ConnectionManifold(SpecialEuclidean(N), CartanSchoutenMinus())?
> 
> Yes, but can this be done for all groups, as suggested above in 2.? Something like that (alias or default) would be an acceptable solution, I think.

Sure, we can have for example `SpecialEuclideanZero` etc. aliases for all groups.

olivierverdier: >  If tests are passing with your change, it means there is not enough tests.

Any concrete idea of a test missing?

I mean, `exp` was defined for some groups before, and everything was working? My change does not change very much what was already in place (see below).

> just replacing exp and log would violate that logic.

But `exp` and `log` were *already replaced* in many cases. For instance, the `exp` for `SpecialEuclidean` or `GeneralLinear` are not specifically related to a metric, and nothing wrong happens.

---
I don't really understand the concern here, but I think that at least, to make it very clear for the user, there should be a table explaining from which zero-torsion connection each `exp` is based on. Here are the ones that I could figure out so far:

| Group | Zero torsion connection | Invariant |
| --- | --- | --- |
| `SemidirectProductGroup{<:Any, <:LeftInvariantRepresentation}`  | `CartanSchoutenZero` (explicitly in `semidirect_product_group.jl:296` | ✅ |
| `SemidirectProductGroup{<:Any, <:HybridTangentRepresentation}` | Metric connection from the product of implicit metrics on `N` and `H` | ❌ |
| `GeneralLinearGroup` | Metric connection from the left invariant metric induced from the standard basis on the Lie algebra | ❌ |
| `GeneralUnitaryMultiplicationGroup` | `CartanSchoutenZero` (explicitly as such in `GeneralUnitaryMatrices.jl:225`)  | ✅ |
| `HeisenbergGroup` | `CartanSchoutenZero`? (probably defined explicitly as such in `heisenberg.jl:220`) | ✅? |
| `SpecialLinearGroup` | Same as `GeneralLinear` | ❌ |

Let me know if I have forgotten a group.

As you see, in these cases, `exp` is redefined, without reference to a connection or a metric, and everything is just fine.

mateuszbaran: > Any concrete idea of a test missing?
> 
> I mean, `exp` was defined for some groups before, and everything was working? My change does not change very much what was already in place (see below).
> 
> >

Things like "call X executes below a certain time threshold" or "all metric/connection functions on manifold X are either consistent or throw an error".

> I mean, `exp` was defined for some groups before, and everything was working? My change does not change very much what was already in place (see below).

`exp` was meant to be the Riemannian or connection exponential, universally in all Manifolds.jl, not the group one. The group one is `exp_lie`.

> But `exp` and `log` were _already replaced_ in many cases. For instance, the `exp` for `SpecialEuclidean` or `GeneralLinear` are not specifically related to a metric, and nothing wrong happens.

As far as I am aware `SpecialEuclidean` was replaced with a faster implementation of the Riemannian logic and `exp` of `GeneralLinear` should implement the Riemannian exponential. If either of these are not true, it's a bug, not a feature.

> I don't really understand the concern here, but I think that at least, to make it very clear for the user, there should be a table explaining from which zero-torsion connection each `exp` is based on. Here are the ones that I could figure out so far:

I see, we just didn't care until this point whether `exp` is group-invariant or not. So we have no such table and the interface wasn't designed with consistency in this part in mind. And I thought semidirect products are the only case where we don't have left-invariance?

I'm not sure why you marked the GL and SL `exp` as not invariant? I thought it should be left-invariant.

Remember that existence of a specialized implementation doesn't always mean that it's supposed to change logic, in some cases it's just for performance.

olivierverdier: > I'm not sure why you marked the GL and SL exp as not invariant? I thought it should be left-invariant.

What I mean by invariance is that it is completely canonical, uniquely defined from `exp_lie` by invariance. The math definition is
```math
\exp_p(pX) = p \exp(X)
```
where the second `exp` is the Lie exponential `exp_lie`.

The `GL` exponential as it is implemented comes from a *non-invariant* metric connection coming from a *left-invariant* metric (which is not canonical, it depends on the basis). It is confusing.

> existence of a specialized implementation doesn't always mean that it's supposed to change logic, in some cases it's just for performance.

My point exactly: these specialized implementations *actually change the logic*. Shouldn't that be clearly documented for the users?

mateuszbaran: > The `GL` exponential as it is implemented comes from a _non-invariant_ metric connection coming from a _left-invariant_ metric (which is not canonical, it depends on the basis). It is confusing.

Yes, currently the Riemannian metric or connection structure is fundamental for Manifolds.jl, not the group one. We can change that but it's not a small project. Maybe seeing the library from this point of view would make it easier to understand.

We wanted to have a Riemannian metric on GL(n) and IIRC we didn't find any that gives invariant geodesics. We preferred to have a Riemannian metric to having invariant exponential map with a non-metric connection.

> My point exactly: these specialized implementations _actually change the logic_. Shouldn't that be clearly documented for the users?

Could you be more specific which methods change the logic? Anyway, if some implementation changes the logic, we should change those methods so that the no longer change it. This is the fundamental part. They should not change the logic. `exp` and `log` should always come from some connection, their invariance to group action must follow from the choice of such connection.

Again, in Manifolds.jl `exp` is **not** the Lie exponential and where it is, it is a coincidence. Maybe we should make it more clear in documentation? I could surely write about it if you can see a good place for it, or maybe there should be even a new tutorial about groups.

kellertuer: The main point we did introduce `exp_lie` was _because_ `exp` was always meant to be one for a Riemannian metric on the manifold.

> we should change those methods so that the no longer change it. 

I agree with that.

The more I read these discussions the more I think it would be clearer to really make the distinction more clear between Riemannian manifolds (from which we sometimes only keep the set) and Lie groups build upon that.

>  there should be even a new tutorial about groups.

I‘d love to have that, but miss the capacities currently myself.

mateuszbaran: > The main point we did introduce `exp_lie` was _because_ `exp` was always meant to be one for a Riemannian metric on the manifold.

Yes, `exp_lie` and `log_lie` are the functions that guarantee invariance. Not `exp` and `log`.

olivierverdier: Thank you for all your answers!

I think a summary of our positions is this:
*My position*: `exp` should be defined from the natural connection on Lie groups (because many users will assume this)
*Your position*: `exp` should be defined from *some arbitrary connection* (sometimes the natural one by coincidence, sometimes not; aka sometimes `exp` is invariant, sometimes not)

Is it an accurate statement?

---

Here are the two definitions that guarantee that `log` and `exp` are invariant (i.e., defined from the natural connection). They are **fast** with left invariant representation, and work on all Lie groups:
```julia
log(G, p, q) = translate_diff(G, p, Identity(G), log_lie(G, compose(G, inv(G, p), q)), LeftForwardAction())
exp(G, p, X) = compose(G, p, exp_lie(G, inverse_translate_diff(G, p, p, X, LeftForwardAction())))
```
---
> We can change that but it's not a small project. 

I don't think so: I did it and it worked since all the tests were passing? Would you have a *very specific* test function that wouldn't pass?

> Could you be more specific which methods change the logic? Anyway, if some implementation changes the logic, we should change those methods so that the no longer change it. 

"Change the logic" means for me: not being invariant. **Specifically**, on `GL`, on `SL` and on semi-direct product with hybrid representation (see table above).

> This is the fundamental part. They should not change the logic. `exp` and `log` should always come from some connection, 

They all do! All the `exp` implemented come from *some* connection (see table above). But on groups, I think that they should depend on *the natural* connection, instead of some arbitrary other one.

If "Change the logic" means having `exp` not being defined from a connection, this is *not* what I'm suggesting.

> their invariance to group action must follow from the choice of such connection.

**There is no choice of a connection**: if `exp` is to be invariant, it **must** come from the natural connection.

> Again, in Manifolds.jl `exp` is **not** the Lie exponential and where it is, it is a coincidence. 

I'm not sure I'm following: these are two different concepts since `exp_lie` maps the Lie algebra to the group, `exp` maps any tangent space to the group. But the implementation above defines `exp` from `exp_lie` in a canonical way on any Lie group.

kellertuer: Currently the central object is a manifold. That will stay that way, see how the package is called.
We interpret a Lie group here as “wrapping” the Manifold in some thing that adds a group action.
`exp` is a function (in  the package) that belongs to the manifold. So we respect that and pass that down.
But “adding a group action” currently does not change the metric of the Manifold.
Or in other words: `exp` on a manifold changes only if you explicitly change the metric (see `MetricManifold` but not implicitly just because you add a group action.

That is a purely design decision.

You arguments are completely valid and the default exp on a Lie group sounds nice, but that would be something I would consider a good design decision in a package called `LieGroups.jl`. I am not even against doing such a package, I think that would be nice, taking all designs we have here, taking `Manifolds.jl` as a dependency and doing such a package. I even proposed such an Idea in an earlier longer discussion.

That does not mean your arguments are invalid, they are concise and nice, they would just change one of the main design decisions of this package with probably huge implications on the total code base.

I just do not have the capacity for starting another package. The ones I have (questions, inquiries, improvements, bug fixes) keep me busy enough.
Maybe that changes some time, but probably not soon.

kellertuer: Overall I am not yet so sure what you want to achieve here through this longish discussion:

1) Do you want to convince us that we are stupid? I feel your tone if often quite agressive and that at least for me sounds like you would like to convince us, that we are stupid – and maybe should stop developing this package?

2) Since we alreaady agreed to all your metric agruments – do you want us to rewrite the whole package? Before we do that I would really recommend to start Liegroups.jl and “do Lie Groups right” following your arguments and use Manifolds.jl as a kind of backend (for example for point checks or such).
I am even on board helping with that a bit, setting everything up here in JuliaManifolds, CI / docs / ... / review PRs.

kellertuer: Ah just when I wanted to call it a day, I had an idea: We could add a keyword argument to `LIeGroup` that says: I want to have the Olivier-approach (name to be found still)
Then the manifold that is passed would be wrapped in a MetricManifold with the corresponding metric (the one that is the best one in Oliviers sense) and on that dispatch the default from Olivier would work.

If we later notice that the approach of changing the metric is for some reason better, it would just mean to change the default in the constructor (from not wrapping to wrapping in the nicest-and-best-of-all-for-Lie metric). That would by then be breaking but until then we have both solution relatively easily in code.

mateuszbaran: > I think a summary of our positions is this: _My position_: `exp` should be defined from the natural connection on Lie groups (because many users will assume this) _Your position_: `exp` should be defined from _some arbitrary connection_ (sometimes the natural one by coincidence, sometimes not; aka sometimes `exp` is invariant, sometimes not)
> 
> Is it an accurate statement?

It isn't entirely accurate; the connection is currently arbitrary but I'm open to changing that, just in a consistent manner.

> Here are the two definitions that guarantee that `log` and `exp` are invariant (i.e., defined from the natural connection). They are **fast** with left invariant representation, and work on all Lie groups:

I got that point but it appears you don't understand my concerns at all.

> I don't think so: I did it and it worked since all the tests were passing? Would you have a _very specific_ test function that wouldn't pass?

Sure. Start with making `inner`, `norm`, `get_vector` and `get_coordinates` (for orthonormal and orthogonal bases) throw `MethodError` on every Lie group where `exp` and `log` does not come from a Levi-Civita connection of some metric. That's not all but it would be a good start. Also make sure the current non-invariant `exp` and `log` on GL and SL is usable in some (consistent) way.

> "Change the logic" means for me: not being invariant. **Specifically**, on `GL`, on `SL` and on semi-direct product with hybrid representation (see table above).

This is not at all what I refer to "change the logic". You made an incorrect assumption about how Manifolds.jl operates and refuse to adjust your expectations despite my detailed explanations.

> They all do! All the `exp` implemented come from _some_ connection (see table above). But on groups, I think that they should depend on _the natural_ connection, instead of some arbitrary other one.
> 
> If "Change the logic" means having `exp` not being defined from a connection, this is _not_ what I'm suggesting.

Whenever I was talking about changing the logic, I meant `exp` not corresponding to the connection chosen for a group. This is the fundamental logic of `exp`. If you don't like it, don't use `exp`. This is one part where I see no point in changing anything about Manifolds.jl: `exp` must come from a consistent choice of connection. We can change which connection it is but invariance is not a requirement for `exp`.

> **There is no choice of a connection**: if `exp` is to be invariant, it **must** come from the natural connection.

There is some literature on non-invariant connections. Some people find them useful. I don't see why we can't support such use cases, specially given that some of it is already implemented.

> I'm not sure I'm following: these are two different concepts since `exp_lie` maps the Lie algebra to the group, `exp` maps any tangent space to the group. But the implementation above defines `exp` from `exp_lie` in a canonical way on any Lie group.

Maybe a good way to resolve this disagreement would be to introduce `exp_inv` and `log_inv` (invariant exp and invariant log) which would be implemented in the exact way you described?

kellertuer: I think `exp_inv` and `log_inv` would also be good ideas – they would be a bit more specific than wrapping the manifold in a metric (as I proposed in my last comment) and dispatch with `exp` on that. I would be fine with both. Maybe the wrapping idea avoids introducing (yet) another function name.

olivierverdier: I get it now (I think): you want `exp` to be defined from a **metric** connection.

If `exp` is defined from a nonmetric connection, you are quite sure that something is going to break (because Lie groups are built on top of metric manifolds).

> We could add a keyword argument to `LIeGroup` that says: I want to have the Olivier-approach (name to be found still) Then the manifold that is passed would be wrapped in a MetricManifold with the corresponding metric (the one that is the best one in Oliviers sense) and on that dispatch the default from Olivier would work.

Yes, that would be nice, but it may not be possible, since the invariant connection is not necessarily metric (example: GL, semidirect products).

> It isn't entirely accurate; the connection is currently arbitrary but I'm open to changing that, just in a consistent manner.

For a metric connection, there is no consistent solutions, your solution is as good as any other. If you agree with (possibly) nonmetric connections, the invariant one is the only consistent choice across all Lie groups.

> You made an incorrect assumption about how Manifolds.jl operates and refuse to adjust your expectations despite my detailed explanations.

I guess I did. Sorry for that.
 
> Whenever I was talking about changing the logic, I meant `exp` not corresponding to the connection chosen for a group. This is the fundamental logic of `exp`. If you don't like it, don't use `exp`. This is one part where I see no point in changing anything about Manifolds.jl: `exp` must come from a consistent choice of connection. We can change which connection it is but invariance is not a requirement for `exp`.

You mean a metric connection. Tu sum it up, you prefer a *metric noninvariant* connection over an *invariant nonmetric* connection.

This is our disagreement in a nutshell.

> Maybe a good way to resolve this disagreement would be to introduce `exp_inv` and `log_inv` (invariant exp and invariant log) which would be implemented in the exact way you described?

Yes, that would be very nice. I think those two functions deserve to be defined somewhere.

---

>  1. Do you want to convince us that we are stupid? I feel your tone if often quite agressive 

It's just a misunderstanding, a disagreement. I like to understand things. I didn't and it's my fault, I'm slow, it makes me sound aggressive. But I think I get it now. Nobody is stupid here, or maybe we're all stupid in the face of maths.

>   I would really recommend to start Liegroups.jl and “do Lie Groups right” following your arguments and use Manifolds.jl as a kind of backend (for example for point checks or such).
>        I am even on board helping with that a bit, setting everything up here in JuliaManifolds, CI / docs / ... / review PRs.

I really appreciate that. I kind of already started actually. I'll let you know how it goes.

Tell me if we finally understand each other (i.e.: you only want `exp` defined from a metric connections, otherwise you know something will break). If we do, we can resolve this conversation.

kellertuer: I am fine with us all being stupid ;)
Sorry your tone (and repetition as it came across to me) was getting to me as agressive – thanks for clarifying that this was not meant like that.

My main problem until now was, that to me this thread slowly read like we are repeating arguments again and again.
But if you are fine with the `exp_inv`/`lie_inv` approach, that is fine with me as well. And yes, I think we by now understand each other, what we want from exp and why that is different fro yours.

For the `LieGroups.jl` idea; I think a nice thing would also be that we do not “overcrowd” `Manifolds.jl` but modularise features a bit. If you want and have done some work on that, as I said, I am very happy to help with all technical details in setting that up; I would prefer that in JuliaManifolds/ – just to not have a package in a persons namespace – but you would be admin of that package as well of course.

mateuszbaran: > I get it now (I think): you want `exp` to be defined from a **metric** connection.

Yes, we chose the connections for GL and SL because we prefer the defaults to be metric non-invariant connections over non-metric invariant connections.  Not all `exp` methods come from a metric connection but we generally prefer metric connections because they give more rich structure required by certain optimization or statistical algorithms; moreover, not all manifolds are Lie groups.

> You mean a metric connection. Tu sum it up, you prefer a _metric noninvariant_ connection over an _invariant nonmetric_ connection.
> 
> This is our disagreement in a nutshell.
> 

Yes, that's correct.

> Yes, that would be very nice. I think those two functions deserve to be defined somewhere.

OK, great, then I will add them in my PR and mention in the new tutorial.

> > 1. Do you want to convince us that we are stupid? I feel your tone if often quite agressive
> 
> It's just a misunderstanding, a disagreement. I like to understand things. I didn't and it's my fault, I'm slow, it makes me sound aggressive. But I think I get it now. Nobody is stupid here, or maybe we're all stupid in the face of maths.

For me developing this library is a great learning opportunity. I'm not a mathematician, and primarily see various branches of differential geometry and Lie groups as a nice language for generalizing classical numerical algorithms. I'm sometimes wrong, but when someone shows me why I'm wrong I can learn something new. There are definitely dozens of people who would do much better job developing JuliaManifolds but they are too busy with more important stuff.

I'm happy when people find JuliaManifolds useful and I generally consider any suggestions regarding interface and functionality of libraries. However, it's impossible to satisfy everyone and we have to set some priorities.

I generally don't mind the tone of messages I get about JuliaManifolds, either yours and from other people. I know it's sometimes a real hassle to express in a positive way, even more so if someone is not a native speaker.

> Tell me if we finally understand each other (i.e.: you only want `exp` defined from a metric connections, otherwise you know something will break). If we do, we can resolve this conversation.

Yes, we prefer metric connections if they are available; I think we've managed to resolve the misunderstanding here finally :slightly_smiling_face: .

> For the `LieGroups.jl` idea; I think a nice thing would also be that we do not “overcrowd” `Manifolds.jl` but modularise features a bit. If you want and have done some work on that, as I said, I am very happy to help with all technical details in setting that up; I would prefer that in JuliaManifolds/ – just to not have a package in a persons namespace – but you would be admin of that package as well of course.

Not-so-fun fact: https://github.com/yuehhua/LieGroups.jl already exists, is registered and looks abandoned. I think it would be fine to develop Lie groups in a separate package but currently we have a bunch of functionality here and maybe some part of it should be moved there?

kellertuer: Oh, I did not see that package, well, if ours gets a bit more serious we could reach out to hopefully take over the name, if that is abandoned. It also seems to not cover many (only one?) Lie group.

olivierverdier:
I realised I forgot to redefine `inverse_translate_diff`. The default (for left invariant representation) should be:
```julia
inverse_translate_diff!(G, Y, ::Any, ::Any, X, ::LeftForwardAction,) = copyto!(G, Y, X)
inverse_translate_diff!(G, Y, ::Any, ::Any, X, ::RightForwardAction,) = copyto!(G, Y, X)
inverse_translate_diff!(G, Y, p, ::Any, X, ::LeftBackwardAction,) = adjoint_action!(G, Y, p, X, RIghtAction())
inverse_translate_diff!(G, Y, p, ::Any, X, ::RightBackwardAction,) = adjoint_action!(G, Y, p, X, LeftAction())
```
The current definition is not wrong, and it works for all representations, but it is very inefficient, as it potentially inverts a matrix three times instead of once, or once instead of zero.

This change does not have to occur within the scope of this PR, it can be done later, maybe in a separate issue.

kellertuer:
We coudls surely still ad it to the 0.10 PR that is still open?

mateuszbaran:
Thanks, I will add that to the 0.10 PR.

mateuszbaran:
Uh, it seems to cause lots of ambiguities, so maybe let's resolve that in a new PR since it not breaking. Anyway, here is the adapted code:
```julia
function inverse_translate_diff!(
    ::TraitList{<:IsGroupManifold{<:AbstractGroupOperation,LeftInvariantRepresentation}},
    G::AbstractDecoratorManifold,
    Y,
    ::Any,
    ::Any,
    X,
    ::LeftForwardAction,
)
    return copyto!(G, Y, X)
end
function inverse_translate_diff!(
    ::TraitList{<:IsGroupManifold{<:AbstractGroupOperation,LeftInvariantRepresentation}},
    G::AbstractDecoratorManifold,
    Y,
    ::Any,
    ::Any,
    X,
    ::RightForwardAction,
)
    return copyto!(G, Y, X)
end
function inverse_translate_diff!(
    ::TraitList{<:IsGroupManifold{<:AbstractGroupOperation,LeftInvariantRepresentation}},
    G::AbstractDecoratorManifold,
    Y,
    p,
    ::Any,
    X,
    ::LeftBackwardAction,
)
    return adjoint_action!(G, Y, p, X, RightAction())
end
function inverse_translate_diff!(
    ::TraitList{<:IsGroupManifold{<:AbstractGroupOperation,LeftInvariantRepresentation}},
    G::AbstractDecoratorManifold,
    Y,
    p,
    ::Any,
    X,
    ::RightBackwardAction,
)
    return adjoint_action!(G, Y, p, X, LeftAction())
end
```

mateuszbaran:
I've merged #732 so this PR is already merged indirectly into master.

Issue #684, opened 2023-11-13
is_flat(::CholeskySpace)  and is_flat(::MetricManifold{ℝ,<:SymmetricPositiveDefinite,LogCholeskyMetric}) return false


SimonKvantdator:
`is_flat()` returns false for both CholeskySpace and SPD with the LogCholeskyMetric, but they are flat manifolds. They are flat because `distance(M::CholeskySpace, p, q)` is of the form $ds^2 = f_1(x_1) dx_1^2 + \dots + f_k(x_k) dx_k^2$, and so it is a reparametrization of $\mathbb{R}^k$. See Lin 2019 Proposition 8 https://arxiv.org/pdf/1908.09326.pdf.

olivierverdier:
Proposition 8 seems to say that the *sectional curvature* is zero, not the full curvature tensor, and `is_flat` returns true only if the full tensor is zero.

kellertuer:
Thanks for stepping in Olivier, but I think with Prop 8.36 Lee, we should return true here for the two cases in the title – I am not directly sure for the positivesemidefinite third case mentioned in the last comment above.

@SimonKvantdator can you maybe do a PR? e can surely help with the details (test coverage, formatting, changelog) :)

SimonKvantdator:
@kellertuer, I am not very proficient in git/github but I will try to do a pull request!

As mentioned, I am unsure about SymmetricPositiveSemidefiniteFixedRank. But here is a script I sometimes use to numerically test the curvature of my manifolds:
```
p = rand(M)

u = rand(M, vector_at=p)
u = u / norm(M, p, u)
v = rand(M, vector_at=p)
v = v - inner(M, p, u, v) * u
v = v / norm(M, p, v)

r = 1e-3
ps = [exp(M, p, r * (cos(theta) * u + sin(theta) * v)) for theta in 0.0:1e-5:(2 * pi)]
ds = [norm(embed(M, p1) - embed(M, p2)) for (p1, p2) in zip(ps, [ps[2:end]..., ps[1]])]
C = sum(ds)
K = 3 * (2 * pi * r - C) / (pi * r^3)
println(K)
```
It uses this formula https://en.wikipedia.org/wiki/Bertrand%E2%80%93Diguet%E2%80%93Puiseux_theorem

olivierverdier:
[...]

PR #685, opened 2023-11-13
is_flat returns true for Cholesky space and SPD with log-Cholesky met…

SimonKvantdator:
…ric.

See issue #684

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.7] - 2023-11-14
+
+### Fixed
+
+- Fixed `is_flat` for `CholeskySpace` and `SymmetricPositiveDefinite` with `LogCholeskyMetric` [https://github.com/JuliaManifolds/Manifolds.jl/issues/684](https://github.com/JuliaManifolds/Manifolds.jl/issues/684).
+
 ## [0.9.6] - 2023-11-09
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


kellertuer:
Good morning and well done – the git and GitHub parts worked super well :)

For the CI (the sometimes a bit annoying part, but nonetheless quite important to keep our code well-maintained):

## Tests

Currently the tests fail – we write tests for nearly all lines of code (>99% of our lines are covered) – so were the two lines you changed, as you can see

* here https://github.com/JuliaManifolds/Manifolds.jl/actions/runs/6855569192/job/18640887683?pr=685#step:6:1452
* and here https://github.com/JuliaManifolds/Manifolds.jl/actions/runs/6855569192/job/18640887683?pr=685#step:6:1302

both even tell you the file and the line that fails and has to be adapted. You can even run just these tests locally since all our test files are standalone runnable by just including them.

## Documentation

We should document the answers more properly with references why values are returned as they are. We have documenter citations for that (you should see them for example in this line https://github.com/JuliaManifolds/Manifolds.jl/blob/5d267d532036399ed9d0e31d19345dee10b64bdb/src/manifolds/CholeskySpace.jl#L6 how they work. Could you add the reason for the new return value similarly in the doc strings of the functions you changed? All references are in this file https://github.com/JuliaManifolds/Manifolds.jl/blob/master/docs/src/references.bib so feel freee to add papers (or look them up) for referencing

## Changelog

Finally we keep track of things we changed in the NEWS.md file in the main directory (that is the other test failing), please add an entry there – and also update the version number in the Project toml accordingly.

Thanks for helping out :)

mateuszbaran (approved):
Thank you for the fix! I can help with failing tests if you have problems fixing them.

SimonKvantdator:
> Thank you for the fix! I can help with failing tests if you have problems fixing them.

That would be much appreciated I think. I don't know where to fix that.

> Could you add the reason for the new return value similarly in the doc strings of the functions you changed?

> we keep track of things we changed in the NEWS.md file in the main directory (that is the other test failing), please add an entry there – and also update the version number in the Project toml accordingly.

Okay, does this require a new pull request then?

mateuszbaran:
I can push changes to your pull request :slightly_smiling_face:

SimonKvantdator:
Sure :)

kellertuer:
to answer the new PR part: No you just need to push to the same branch you were working on, then PRs are automatically updated :)

kellertuer (approved):

SimonKvantdator:
> Before this was merged I forgot to write: Thanks for both reporting and fixing this!
> 
> All contributions are always welcome and important to the package :)

Thanks Ronny! I am aiming to contribute more things in the futute.

kellertuer:
That is great to hear! Looking forward to your next contributions then - and of course we help with details similar to we did here :)

Issue #686, opened 2023-11-15
Possible issues with random elements of `CircleGroup`?


olivierverdier:
Consider the circle group
```julia
G = CircleGroup()
```
This works:
```julia
g = rand(G)
```

This also works, although it gives a tangent vector which is not stored as left invariant (this is fine, but worth being aware of it):
```julia
rand(G; vector_at=g)
```
---

Now:

This throws a `MethodError`:
```julia
rand(TangentSpace(G,g)) # same error if g == identity_element(G)
```

This only ever returns 0:
```julia
rand(G; vector_at=identity_element(G)) # 0.0
```

olivierverdier:
Closing, since this is probably working as expected (see https://github.com/JuliaManifolds/Manifolds.jl/issues/688#issuecomment-1816051476).

PR #689, opened 2023-11-17
Improve random tangent generation on complex circle and rotation matr…

### LINKED ISSUE #688: Improve `rand` for tangent vectors on `SpecialOrthogonal(2)`
If M is defined as `[0 1; -1 0]`, then
```julia
SO2 = SpecialOrthogonal(2)
rand(SO2; vector_at=identity_element(SO2)) # always return ±M/sqrt(2)
```
This contaminates products or semidirect products with `SO(2)`, for instance, tangent vector of `SpecialEuclidean(2)` are not random either.

mateuszbaran:
…ices

Fixes #688 and #686.

### DIFF
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -330,7 +330,6 @@ function Random.rand!(
         (manifold_dimension(M) == 0) && return fill!(pX, 0)
         A = σ .* randn(rng, representation_size(M))
         pX .= triu(A, 1) .- transpose(triu(A, 1))
-        normalize!(pX)
     end
     return pX
 end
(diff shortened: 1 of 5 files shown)


src/manifolds/Rotations.jl
```diff
@@ -330,7 +330,6 @@ function Random.rand!(
         (manifold_dimension(M) == 0) && return fill!(pX, 0)
         A = σ .* randn(rng, representation_size(M))
         pX .= triu(A, 1) .- transpose(triu(A, 1))
-        normalize!(pX)
```
kellertuer, line 333: I would have red this line as we get back to a determinant 1 and for points this might be necessary? Yes I know you do tangents here, but maybe we should still normalise points ;) ?

mateuszbaran: Normalization doesn't make determinant equal to 1 though. Random point generation seems to work correctly, and tangent vector generation doesn't need normalization.

kellertuer: I do agree that tangent vectors do not need this :)
Then I maybe had the wrong thing in mind what is normalised here. If tests pass, this should be fine with me then.

kellertuer (approved):

PR #692, opened 2023-12-14
Unify use of an `atol` when comparing to zero

### LINKED ISSUE #630: Improve default `atol` when comparing to 0
Currently most places when we compare to 0 with `isapprox` have defaults that cause the comparison to be exact. `KendalShapeSpace` is one exception. I think we should change other defaults (when `isapprox` with 0 happens) to match approximate comparison in `KendalShapeSpace`.

kellertuer:
As discussed therein, this fixes #630.

I carefully went through all `isapprox` and as soon as there was a comparison to zero, I adapted from existing  cases a nonzero `atol` in the signature and passed that down.

this still needs a bit of checks for failing tests, but should work otherwise I hope.


src/groups/addition_operation.jl
```diff
@@ -83,8 +83,14 @@ function inv_diff!(::AdditionGroupTrait, G::AbstractDecoratorManifold, Y, p, X)
     return Y
 end
 
-function is_identity(::AdditionGroupTrait, G::AbstractDecoratorManifold, q; kwargs...)
-    return isapprox(G, q, zero(q); kwargs...)
+function is_identity(
+    ::AdditionGroupTrait,
+    G::AbstractDecoratorManifold,
+    q::T;
+    atol=sqrt(prod(representation_size(G)) * eps(T)),
```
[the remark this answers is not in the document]
kellertuer: ah, nice, and I also was not aware that number_eltype is useful here. Will replace corresponding Ts :)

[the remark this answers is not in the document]
kellertuer: I also needed a `real` outside the `float` since for `ComplexF64` there is no eps – but now it seems to work like a charm.

kellertuer:
Locally the tests run fine by now – here it seems on Julia 1.10 there is a problem with DiffEq (again?).

mateuszbaran:
BoundaryValueDiffEq.jl again :disappointed:

mateuszbaran:
I've opened an issue: https://github.com/SciML/BoundaryValueDiffEq.jl/issues/148

kellertuer:
With the last fix it really runs locally just fine (probably with an older version of BoundaryValueEq still), that is indeed a bit annoying.

src/manifolds/SymplecticStiefel.jl
```diff
@@ -142,14 +147,20 @@ The tolerance can be set with `kwargs...` (e.g. `atol = 1.0e-14`).
 """
 check_vector(::SymplecticStiefel, ::Any...)
 
-function check_vector(M::SymplecticStiefel{<:Any,field}, p, X; kwargs...) where {field}
+function check_vector(
+    M::SymplecticStiefel{S,𝔽},
+    p,
+    X::T;
+    atol=sqrt(prod(representation_size(M))) * eps(real(float(number_eltype(T)))),
```
mateuszbaran, line 154:
```suggestion
    atol::Real=sqrt(prod(representation_size(M))) * eps(real(float(number_eltype(T)))),
```

mateuszbaran (commented):
Everything else looks fine :+1:

src/manifolds/EssentialManifold.jl
```diff
@@ -125,8 +125,14 @@ exp(::EssentialManifold, ::Any...)
 
 get_iterator(::EssentialManifold) = Base.OneTo(2)
 
-function _isapprox(M::EssentialManifold, p, q; kwargs...)
-    return isapprox(distance(M, p, q), 0.0; kwargs...)
+function _isapprox(
+    M::EssentialManifold,
+    p,
+    q::T;
+    atol=eps(real(float(number_eltype(number_eltype(T))))),
```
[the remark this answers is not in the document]
kellertuer: But we have to unwrap twice – then we have to reimplement `number_eltype` to handle this. With just one `number_eltype` this returns `Matrix{Float64}`which then errors on the eps (or the float?)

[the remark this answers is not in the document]
mateuszbaran: I'd say that this is a bug in `number_eltype` which should always return a scalar type.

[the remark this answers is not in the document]
mateuszbaran: I will fix that bug.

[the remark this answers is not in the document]
kellertuer: Ok, the case here is that the essential manifold has points as a vector (of length 2, one could also take a tuple I think) of matrices, so the element is a matrix, but I agree with the idea that number_eltype could also work completely recursive. Can you fix that or should I take a look?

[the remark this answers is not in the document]
mateuszbaran: done: https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/179 :slightly_smiling_face:

mateuszbaran:
@kellertuer could you check if there is anything else to adapt due to https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/177 ?

kellertuer:
At first glance there might be two further things (but I am on mobile)

* `default_approximation_method` is the new function and we have an old (deprecated) `default_estimation_method` now here
* the `ExtrinsicEstimationMethod` got a little new feature to store the method using in the extrinsic space (since usually that was only implicitly mentioned in the docs that one uses gradient in the embedding or such) – so the constructor without parameter (defaulting to `GradientDescentEstimation` I would think)

mateuszbaran:
> * 
> * the `ExtrinsicEstimationMethod` got a little new feature to store the method using in the extrinsic space (since usually that was only implicitly mentioned in the docs that one uses gradient in the embedding or such) – so the constructor without parameter (defaulting to `GradientDescentEstimation` I would think)

The most common use case for extrinsic estimation is when the embedding in Euclidean and we use EfficientEstimator there.

kellertuer:
Sure that also sounds like a very good default. I think I just saw the gradient one somewhere, but yours is indeed maybe the more common case.

src/statistics.jl
```diff
@@ -515,15 +413,19 @@ function Statistics.mean!(
...
-        get_embedding(M),
-        mean,
-    ),
+    e::ExtrinsicEstimation;
+    extrinsic_method=nothing,
     kwargs...,
 )
+    if !isnothing(extrinsic_method)
+        Base.depwarn(
+            "The Keyword Argument `extrinsic_method` is deprecated use `ExtrinsicEstimators field instead",
```
mateuszbaran, line 422:
```suggestion
            "The Keyword Argument `extrinsic_method` is deprecated use `ExtrinsicEstimators` field instead",
```

src/statistics.jl
```diff
@@ -771,15 +670,19 @@ function Statistics.median!(
...
-        get_embedding(M),
-        median,
-    ),
+    e::ExtrinsicEstimation;
+    extrinsic_method=nothing,
     kwargs...,
 )
+    if !isnothing(extrinsic_method)
+        Base.depwarn(
+            "The Keyword Argument `extrinsic_method` is deprecated use `ExtrinsicEstimators field instead",
```
mateuszbaran, line 679:
```suggestion
            "The Keyword Argument `extrinsic_method` is deprecated use `ExtrinsicEstimators` field instead",
```

test/statistics.jl
```diff
@@ -796,12 +797,12 @@ end
...
         w = pweights([rand(rng) for _ in 1:length(x)])
         m = normalize(median(Euclidean(3), x, w))
-        mg = median(S, x, w, ExtrinsicEstimation())
+        mg = median(S, x, w, ExtrinsicEstimation(EfficientEstimator()))
         @test isapprox(S, m, mg)
     end
 
     @testset "Covariance Default" begin
-        @test default_estimation_method(TestStatsSphere{2}(), cov) ==
+        @test default_approximation_mthod(TestStatsSphere{2}(), cov) ==
```
mateuszbaran, line 805:
```suggestion
        @test default_approximation_method(TestStatsSphere{2}(), cov) ==
```

kellertuer:
The new approximation methods and their dispatch should now be fixed. I also circumvent the inconsistency from StatsBase for now, which mainly means we have a bit of code that can be considered duplication (if StatsBase would not be inconsistent).

NEWS.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.9] – 2023-12-14
+
+### Fixed
+
+* introduced a nonzero `atol` for all point and vector checks that compre to zero.
+  This makes those checks a bit more relaxed by default and resolves [#630](https://github.com/JuliaManifolds/Manifolds.jl/issues/630).
+* `get_estimation_method(M)` is deprecated, use `get_approximation_method(M, f)` for your specific mathod `f` on the manifold `M`.
```
mateuszbaran, line 14:
```suggestion
* `default_estimation_method(M, f)` is deprecated, use `default_approximation_method(M, f)` for your specific method `f` on the manifold `M`.
* `AbstractEstimationMethod` is deprecated, use `AbstractApproximationMethod` instead.
```
There was no `get_estimation_method`. `AbstractEstimationMethod` is deprecated though.

src/manifolds/EssentialManifold.jl
mateuszbaran, line 132:
```suggestion
    atol::Real=eps(real(float(number_eltype(number_eltype(T))))),
```

src/statistics.jl
```diff
@@ -515,26 +406,37 @@ function Statistics.mean!(
     y,
     x::AbstractVector,
     w::AbstractVector,
-    ::ExtrinsicEstimation;
-    extrinsic_method::AbstractEstimationMethod=default_estimation_method(
-        get_embedding(M),
-        mean,
-    ),
+    e::ExtrinsicEstimation;
+    extrinsic_method=nothing,
```
mateuszbaran, line 410:
```suggestion
    extrinsic_method::Union{AbstractEstimationMethod,Nothing}=nothing,
```

src/manifolds/Rotations.jl
```diff
@@ -375,12 +375,6 @@ end
 function parallel_transport_direction!(::Rotations{TypeParameter{Tuple{2}}}, Y, p, X, d)
     return copyto!(Y, X)
 end
-function parallel_transport_direction(M::Rotations, p, X, d)
-    expdhalf = exp(d / 2)
-    q = exp(M, p, d)
-    return transpose(q) * p * expdhalf * X * expdhalf
-end
-parallel_transport_direction(::Rotations{TypeParameter{Tuple{2}}}, p, X, d) = X
```
mateuszbaran, line 383: Why exactly isn't this needed? IIRC it was a small optimization for static arrays. Is it done on unitary matrix level?

mateuszbaran (commented):
Cool, it looks almost ready :+1:

src/manifolds/Rotations.jl
kellertuer: These are no longer hit, since we now also allocate first for the vector transports.

NEWS.md
kellertuer: There was in just a few places.

src/manifolds/Rotations.jl
mateuszbaran: But we don't if this function gets called directly?

kellertuer: If we indeed need them, we would have to add that on layer 2 (the `_vector_transport_` level), but I think the (a) were only fallbacks since these were not unified to allocate first (yet) and the last line is even cheating a bit, since it would be mutating (or returning its input to be precise).

NEWS.md
mateuszbaran: Hm, I couldn't find it then :sweat_smile:

src/manifolds/Rotations.jl
kellertuer: > But we don't if this function gets called directly?

They are no longer covered by tests, that is mainly why I removed them. The last update in base moved vector transports also to allocate-first (like we did with retractions).

mateuszbaran: Returning input is not cheating if no change is necessary. And I'd prefer to keep parallel transport non-allocating for `SMatrix` is possible.

NEWS.md
kellertuer: See https://github.com/JuliaManifolds/Manifolds.jl/blob/b2d83b337e214158f64d22a39776b9a80ffd2981/src/manifolds/Grassmann.jl#L174-L176 (that's on current master) – but I also only saw them when reworking Manifolds.jl

src/manifolds/Rotations.jl
kellertuer: hm oh that is even PT, that is a bit surprising, I did not rework that, but maybe some fallback is then no longer hitting. hen we maybe should get that back (I was probably not careful enough and thought they were vector transports, those now allocate earlier).

But wer have to come up with tests then, and I am super uncreative with that currently.

NEWS.md
mateuszbaran: Yes, that's `default_estimation_method`, not `get_estimation_method`. I'm not sure where I was wrong?

src/manifolds/Rotations.jl
mateuszbaran: I will try to add a test then.

NEWS.md
kellertuer: ah I just (as for the other comment) did not read your comment carefully enough, sorry. This PR was three days (not straight in a row) of fighting wrong dispatches. You are right, the fix is ok.

mateuszbaran: OK, it's fine :slightly_smiling_face: . That was quite a lot of fighting then.

src/manifolds/Rotations.jl
kellertuer: Most time I spent on finding out about the `mean!` inconsistency and then fixing it. Well – I think this got better now and it is nice to finish this to continue with the other ones that depend on this one.

Issue #656, opened 2023-10-07
Consider making RecursiveArrayTools and StaticArrays weak dependencies


kellertuer:
It might be a bit counter-intuitive to the current structure, where we have these methods with specific manifolds,
but I think for the package it would be neat if these packages that introduce speedups for certain data types would be weak dependencies.

kellertuer:
I will maybe try this next, though for `RecursiveArrayPartition` this might be breaking, since previously we exported `ArrayPartition`.
The rest seems mainly to be reorganising code.

mateuszbaran:
This may indeed be a bit breaking so I'd prefer to have it scheduled for the next breaking release instead of rushing it now.

kellertuer:
I started and noticed it is mainly code-reorganisation, but there might be 2-3 small breaking things

* `ArrayPartition` is reexported and that is no longer the case after this change
*  `we might loose a few default implementations that are no longer available when you “just” load Manifolds (since a random function creates an ArrayPartition for example)

So – I will slowly do this, open a PR, but that should really only be merged when we have more reasons for a breaking change – since this decoupling will also not have such a large effect on loading and such.

So I fully agree with your opinion :)

mateuszbaran:
We can probably close this because RecursiveArrayTools is now a weak dependency and StaticArrays is not worth extracting at the moment?

kellertuer:
Yes.

PR #693, opened 2023-12-15
Move more packages to weak dependencies

kellertuer:
As discussed in #656, this is an idea to move two further packages to weak dependencies.

I am not saying we should necessarily do that, but it is something to discuss, and I for now just spent about an hour to reorganise the code. I am not yet sure the advantages outweigh the disadvantages here

This would be a breaking change

* `ArrayPartition` will no longer be exported in Manifolds.jl
* we currently loose some default implementations on product manifolds and groups (for RecursiveArrays) and a few random functions (for staticArrays) which could also be provided again as default without the package dependency for sure.

Before checking and adopting the tests I would like to discuss whether we actually want to do this, with the disadvantages above the question is, does the reduced loading time make this worth the change?


kellertuer:
Yes, now that I noticed that Oliviers PR is breaking, we could combine that with this PR and check a bit what else to break since we are at it ;)

Thanks for looking into it.

kellertuer:
We could for example also do #701 then. That should even be a quite short PR.

mateuszbaran:
Yes, #701 could also be done, or any other quick and breaking thing we intended to do.

mateuszbaran:
#732 is now merged and supersedes this PR so we can close this.

PR #694, opened 2023-12-25
Fix documentation on master

kellertuer:
This fixes the error but also most warnings for the docs on master, since we missed to preview on last PR.

* [x] There seems to be one `$` somewhere in the docs now
* [x] there seem to be a few warnings for missing docs from extensions?
* [x] We could maybe use an emoji upfront links that point to ManifoldsBase? I just did not yet have a good idea which one – or we could use the base symbol if that is not too hard to include in text.


mateuszbaran:
> [ ] There seems to be one `$` somewhere in the docs now

It's hard to trace it without at least file name unless we replace all dollar signs with backticks for math.

> * there seem to be a few warnings for missing docs from extensions?

IIRC they were there for a long time.

> [ ] We could maybe use an emoji upfront links that point to ManifoldsBase? I just did not yet have a good idea which one – or we could use the base symbol if that is not too hard to include in text.

Can it be done automatically? If not, it would be yet another chore to keep it consistent.

mateuszbaran (approved):

kellertuer:
Yes, I looked for the $ a bit but that might really be hard to track down, since Documenter does not (yet) provide the details which file that happens on. 

I also checked whether we can maybe prefix ManifoldsBase.jl links with some symbol / emoji, but had no-good idea, so with the updated news entry to today, I think this should be fine as is and we leave the rest for someone (tm) with too much time and/or creativity :D 

For the missing docs – it would also at some point be nice to document them, but not so urgent I think.

PR #699, opened 2024-01-01
Implement the complex functions (allocation still not working)

kellertuer:
This should (up to specifying the right basis) resolve https://github.com/JuliaManifolds/Manopt.jl/issues/340

For now there is a few things to check though

* it seems the docs for the original `get_coordinates` (real-valued case) are not precise enough, the inner product for $\lambda$ seems wrong and the used $x$ is not defined. Will check. The same holds for the inverse `get_vector`
* We should improve the docs in ManifoldsBase.jl and provide more details on the two approaches used for bases in the complex case
  * real coefficients in a complex-valued basis, which has the advantage that we have as many coefficients as the manifold dimension, and this should be the default
  * complex coefficients in a real-valued basis, which is sometimes, especially for the sphere here a bit strange, since it would mean that in the North Pole the first component is only allowed to be purely imaniary for the spheres
  * currently it still seems that allocation is a bit wrong still, since the real coefficients should – for best of cases, also be real. On the sphere they also seem to be of wrong size still.

Here is a bit of test code that I try to verify the new method with

a) real check for comparison

```julia
M1 = Sphere(2)
p1 = rand(M1)
X1 = rand(M1; vector_at=p1) 
c1 = get_coordinates(M1, p1, X1, DefaultOrthonormalBasis())
# Since this is an ONB, the norm of X1 in the TpM and c1 (usual norm) should be equal
norm(M1, p1, X1) ≈ norm(c1)
```

The same should hold now if we perform that on `M2 = Sphere(2, ℂ)`

```julia
M2 = Sphere(2, ℂ)
p2 = rand(M2)
X2 = rand(M2; vector_at=p2) 
c2= get_coordinates(M2, p2, X2, DefaultOrthonormalBasis()) #probably with `\bbC`?
# Since this is an ONB, the norm of X2 in the TpM and c2 (usual norm) should be equal (still TODO)
norm(M2, p2, X2) ≈ norm(c2)
```

### DIFF
--- a/src/manifolds/Sphere.jl
+++ b/src/manifolds/Sphere.jl
@@ -12,31 +12,31 @@ end
 @doc raw"""
     Sphere{T,𝔽} <: AbstractSphere{𝔽}
 
-The (unit) sphere manifold $𝕊^{n}$ is the set of all unit norm vectors in $𝔽^{n+1}$.
+The (unit) sphere manifold ``𝕊^{n}`` is the set of all unit norm vectors in ``𝔽^{n+1}``.
 The sphere is represented in the embedding, i.e.
 
 ````math
 𝕊^{n} := \bigl\{ p \in 𝔽^{n+1}\ \big|\ \lVert p \rVert = 1 \bigr\}
 ````
 
-where $𝔽\in\{ℝ,ℂ,ℍ\}$. Note that compared to the [`ArraySphere`](@ref), here the
-argument `n` of the manifold is the dimension of the manifold, i.e. $𝕊^{n} ⊂ 𝔽^{n+1}$, $n\in ℕ$.
+where ``𝔽\in\{ℝ,ℂ,ℍ\}``. Note that compared to the [`ArraySphere`](@ref), here the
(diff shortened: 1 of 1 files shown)


mateuszbaran:
> * complex coefficients in a real-valued basis, which is sometimes, especially for the sphere here a bit strange, since it would mean that in the North Pole the first component is only allowed to be purely imaniary for the spheres

I think we should restrict complex-coefficient bases to [complex manifolds](https://en.wikipedia.org/wiki/Complex_manifold), while the complex sphere is actually a [CR manifold](https://en.wikipedia.org/wiki/CR_manifold) but not a complex manifold.

mateuszbaran:
Or, at least, I wouldn't add complex-coefficients bases to non-complex CR manifold without some understanding of the theory behind CR-manifolds.

kellertuer:
That would again be something breaking and different to what we use those currently, but sure that is something to discuss.

kellertuer:
I fixed the one allocation bug I had but now I notice that my idea might not work as intended (doing the same rotation as in the real case and taking the complex value of the first component as the last one). So I fear I have to close this since I have no clue how to compute an ONB or its coordinates here.

PR #700, opened 2024-01-03
Symplectic Grassmann

kellertuer:
this PR is a start to implement the (real) Symplectic Grassmann, which one obtains from Symplectic Stiefel the same way one obtains Grassmann from Stiefel.
It will probably also revise a few symplectic Stiefel functions.

* [x] Introduce `Hamiltonian`, they allow for `H^+` to be nicely defined for the symplectic inverse 
* [x] We should introduce `HamiltonianMatrices`? Maybe not 100% necessary but nice for completeness
* [x] Define `SymplecticGrassmann`
* [x] implement checks
* [x] `inner`  (after (4.10) / Lemma 4.8, BZ21)
* [x] the `test/utils.jl` functions complain since Julia 1.10 to be multiply included. We should at least move those, if not also the constants to the Manufolds.jl/Test.jl extension?
* [x] unify notation concerning J_2n and Q_2n for the symplectic element (and refer to `SymplecticMatrix`).
* [x] unify/simplify notation to refer more often to `symplectic_inverse` when writing `p^+` instead of repeating the formula.
* [x] implement the geodesic
* [x] implement the Cayley retraction and its inverse 
* [x] refactor `rand_hamiltonian` to be `rand(Hamiltonian(2n))`
* [x] implement the gradient conversion (after Lemma 4.8, BZ21)
* [x] (maybe) check for `Symplectic` and `SymplecticStiefel`, so they get `rand!` methods as well. They currently just have `rand` methods
* [x] read docs carefully again
* [x] check test coverage

Both symplectic Stiefel and Grassmann stem from this paper https://arxiv.org/abs/2108.12447

PR #702, opened 2024-01-19
Multinomial randomness

kellertuer:
This PR introduces random for `MultinomialSymmetric` and `MultinomialDoublyStochatstic`, which I saw was missing due to a comment [here](https://stackoverflow.com/questions/76460042/julia-gives-methoderror-rand-when-trying-to-optimize-over-a-certain-manifold).

Checking the corresponding paper  https://ieeexplore.ieee.org/document/8861409
I saw they also introduce MultinoialSPD, which I added here as well. So this short side-track still needs

# ToDo

* [x] `riemannian_gradient` for `Multinomial` and the three doubly stochastic ones
* [x] `riemannian_Hessian` for `MultinomialSymmetric`
* [x] tests
* [x] read docs
 
The paper (its tech report we link in the docs to be precise) does state the Hessian for `MultinomialDoublyStochastic`, but it is quite involved (the symmetric one already is), that I am not sure I can “hack that down” in an efficient and stable way.


kellertuer:
All things I wanted to implement (and that needed fixing) are implemented (and fixed, e.g. the multinomalsymmetric ones have more than just the symmetric check for points now – again).

So this is no draft any longer, but still needs the tests to be checked and extended.

src/manifolds/MultinomialSymmetricPositiveDefinite.jl
```diff
@@ -0,0 +1,52 @@
...
+containing ones and zeros, respectively. More details for this manifold can b found in
+[DouikHassibi:2019](@cite).
+
+# Constructor
+
+    MultinomialSymmetricPositiveDefinite(n)
+
+Generate the manifold of matrices ``\mathbb R^{n×n}`` that are symmetric, positive definite, and doubly stochastic.
+"""
+struct MultinomialSymmetricPositiveDefinite{T} <: AbstractMultinomialDoublyStochastic
```
mateuszbaran, line 29: Could we have a `rand` for these too?

src/manifolds/MultinomialSymmetricPositiveDefinite.jl
```diff
@@ -0,0 +1,52 @@
...
+    p ∈ ℝ^{n×n}\ \big|\ &p_{i,j} > 0 \text{ for all } i=1,…,n, j=1,…,m,\\
+& p^\mathrm{T} = p,\\
+& p\mathbf{1}_n = \mathbf{1}_n\\
+a^\mathrm{T}pa > 0 \text{ for all } a ∈ ℝ^{n}\backslash\{\mathbf{0}_n\}
+\bigr\},
+\end{aligned}
+````
+
+where ``\mathbf{1}_n`` and ``\mathbr{0}_n`` are the vectors of length ``n``
+containing ones and zeros, respectively. More details for this manifold can b found in
```
mateuszbaran, line 20:
```suggestion
containing ones and zeros, respectively. More details for about manifold can be found in
```

src/manifolds/MultinomialSymmetricPositiveDefinite.jl
kellertuer: I would have to check, whether I see them somewhere, for now I just saw that manifold in the paper (where we have the other three from).
If I see a good wax to construct there, sure.

src/manifolds/MultinomialSymmetricPositiveDefinite.jl
kellertuer: `about` sounds better here, but then without the `for` upfront.

mateuszbaran: Yes, I actually meant to replace "for", not "this".

src/manifolds/MultinomialSymmetricPositiveDefinite.jl
mateuszbaran: Probably generating a random totally positive matrix: https://math.stackexchange.com/questions/2773460/how-to-generate-a-totally-positive-matrix-randomly-using-software-like-maple and then projecting to `AbstractMultinomialDoublyStochastic` would work.

kellertuer: That would at least not directly yield an error, but would the result have positive eigenvalues? I am not so sure, but I can check.

kellertuer: I fiddled a bit with the random for MSPD, I fear I am not clever enough for this. Not only does it involve several of our manifolds (symmetric, SPDs, multinomial, Simplex, and all might throw errors somewhere along the way), I can just not get (a) random to work with the positive matrix for points nor the one for the vectors, since I am not sure how the projection looks like.

So in theory – yes. In practice – not.

kellertuer: So if you can do that, feel free to implement `project` for vectors and rand including your idea for the randomness – to which in theory I agree – just that in practice I can not get it to work (but I also did a whole day of writing tests, so my brain is not much more than a pudding by now).

mateuszbaran: I've written `rand` based on that post, seems to work:
```julia
function rand(M::MultinomialSymmetricPositiveDefinite)
    n = get_parameter(M.size)[1]
    L = sort(exp.(randn(n)))
    V = reduce(hcat, map(xi -> [xi^k for k in 0:(n-1)], L))'
    Vlu = lu(V, LinearAlgebra.RowNonZero())
    dm = Diagonal(Vlu.U)
    uutd = dm \ Vlu.U
    random_totally_positive = uutd * dm * Vlu.L
    MMDS = MultinomialDoubleStochastic(n)
    ds = project(MMDS, random_totally_positive)
    sym = (ds .+ ds')./2
    return sym
end
```

mateuszbaran: I will add some documentation later because it's quite involved.

kellertuer: That looks nearly as involved as the Hessian I did. 
And we should do that as a `rand!(rng, M, p; kwargs...)` (maybe we see how to do tangent vectors as well :) )

What – at first glance – I did not have, was the exp

mateuszbaran: `exp` is not strictly necessary, we only need `L` to be positive and sorted. About 2/3 of the work was figuring out how to do the exact variant of LDU decomposition it needs.

kellertuer:
Multinomial is failing on 1.6, but it seems to not be so easy to reproduce, on my machine that works fine.

mateuszbaran:
This is weird. `is_vector` at `/ManifoldsBase/src/decorator_trait.jl:481` assumes that the error returned by `check_vector` has field `val` but `ComponentManifoldError` obviously doesn't have it. It looks like a problem with that `is_vector` perhaps?

kellertuer:
Might also be that, but even more strange is, that there should not happen any error and on Julia 1.10, 1.9 and 1.6 on my machine it also does not happen.

mateuszbaran:
Probably tolerance is too low. It would be more clear if that `is_vector` didn't try to access `val` or had a contingency in case it's not present.

kellertuer:
Yes sure, that would help, but I do not yet directly see, why that is happening yet (also because I can not yet reproduce it here).

kellertuer:
The problem seems to be that this line

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/65780dc51475b255258dfd35a225f09a0cd2810b/src/ManifoldsBase.jl#L769

either there or in the decorator at

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/65780dc51475b255258dfd35a225f09a0cd2810b/src/decorator_trait.jl#L481

implicitly assumes that the error at hand is a DomainError (or one of the other simple errors that has `msg` and `val`). So this is nothing to fix here, but maybe to rethink in ManifoldsBase. I am also not sure when this might occur, since for now I can not reproduce this on my machine. Here I for now tried to raise the tolerance instead.

src/manifolds/MultinomialSymmetricPositiveDefinite.jl
mateuszbaran: I've added it to the PR. I haven't checked tangent vector generation yet but that seems to be less important.

kellertuer: If I see that correctly, Symmetric Multonomial should do the job.

kellertuer: Sure I can check. I wanted to finish the other PR first anyways (the other one has so many strings changed in the docs, so lets better merge it here, i.e. finish that first), so this one is not that urgent.

test/manifolds/multinomial_spd.jl
```diff
@@ -0,0 +1,18 @@
...
+        @test repr(M) == "MultinomialSymmetricPositiveDefinite(3)"
+        @test repr(Mf) == "MultinomialSymmetricPositiveDefinite(3; parameter=:field)"
+        @test get_embedding(M) == MultinomialMatrices(3, 3)
+        @test get_embedding(Mf) == MultinomialMatrices(3, 3; parameter=:field)
+    end
+
+    @testset "Random" begin
+        q = zeros(3, 3)
+        M = MultinomialSymmetricPositiveDefinite(3)
+        @test is_point(M, rand!(Random.Xoshiro(), M, q))
```
kellertuer, line 16: `Random.Koshiro()` seems to not exist on Julia 1.6 – can we use another RNG?

mateuszbaran: Yes, sure, I didn't notice.

mateuszbaran: Hm, Julia 1.6 doesn't have `RowNonZero` either. I will look for a replacement in the evening.

kellertuer:
The one line we do not hit is a false-negative (sadly).

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -60,44 +98,27 @@ end
 Checks whether `p` is a valid point on the [`MultinomialDoubleStochastic`](@ref)`(n)` `M`,
 i.e. is a  matrix with positive entries whose rows and columns sum to one.
 """
-function check_point(
-    M::MultinomialDoubleStochastic,
-    p::T;
-    atol::Real=sqrt(prod(representation_size(M))) * eps(real(float(number_eltype(T)))),
-    kwargs...,
-) where {T}
+function check_point(M::MultinomialDoubleStochastic, p::T; kwargs...) where {T}
```
mateuszbaran, line 101:
```suggestion
function check_point(M::MultinomialDoubleStochastic, p; kwargs...)
```
We don't need that now.

src/manifolds/MultinomialDoublyStochastic.jl
```diff
@@ -60,44 +98,27 @@ end
...
-) where {T}
-    r = sum(X, dims=2) # check for stochastic rows
-    if !isapprox(norm(r), 0.0; atol=atol, kwargs...)
-        return DomainError(
-            r,
-            "The matrix $(X) is not a tangent vector to $(p) on $(M), since its rows do not sum up to zero.",
-        )
-    end
-    return nothing
+function check_vector(M::MultinomialDoubleStochastic, p, X::T; kwargs...) where {T}
```
mateuszbaran, line 116:
```suggestion
function check_vector(M::MultinomialDoubleStochastic, p, X; kwargs...)
```

src/manifolds/MultinomialSymmetric.jl
```diff
@@ -106,23 +112,59 @@ end
...
+
+@doc raw"""
+    rand(::MultinomialSymmetric; vector_at=nothing, σ::Real=1.0, kwargs...)
+
+Generate random points on the [`MultinomialSymmetric`](@ref) manifold
+or tangent vectors at the point `vector_at` if that is not `nothing`.
+
+Let ``n×n`` denote the matrix dimension of the [`MultinomialSymmetric`](@ref).
+
+When `vector_at` is nothing, this is done by generating a random matrix`rand(n,n)`
```
mateuszbaran, line 144:
```suggestion
When `vector_at` is nothing, this is done by generating a random matrix `rand(n, n)`
```

mateuszbaran (approved):
Just a few minor comments.

kellertuer:
This one is now merged with #700, so we just have to wait for the tests.

PR #703, opened 2024-01-21
Fix `var` for `TranslationGroup`

mateuszbaran:
Reported here: https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1811 .


kellertuer (approved):
This looks fine for me, we could, however, also do this in general for groups to pass `var` to the inner manifold?

_edit:_ Please check the one missing code coverage line unless that is due to the tests not yet having finished completely.

mateuszbaran:
Sure, I've generalized it and the line is now covered.

dehann:
Thanks very much!

PR #706, opened 2024-01-27
More random randomness on random (no actually all) Lie groups

kellertuer:
This small fix could even reduce other Lie groups random implementations

* [x] adds `rand` to `UnitaryMatrices`
* [x] adds. a “passthrough” for points on Lie groups (to ask the manifold) and tangent vectors to generate tangent vectors at the identity
* [x] text coverage
* [x] check whether other Lie algebras implicitly do that already and now explicitly could use this generic default instead.

### DIFF
--- a/src/manifolds/ProbabilitySimplexEuclideanMetric.jl
+++ b/src/manifolds/ProbabilitySimplexEuclideanMetric.jl
@@ -21,6 +21,69 @@ function manifold_volume(M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanM
 end
 
 @doc raw"""
+    rand(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}; vector_at=nothing, σ::Real=1.0)
+
+When `vector_at` is `nothing`, return a random (uniform) point `x` on the [`ProbabilitySimplex`](@ref) with the Euclidean metric
+manifold `M` by normalizing independent exponential draws to unit sum, see [Devroye:1986](@cite), Theorems 2.1 and 2.2 on p. 207 and 208, respectively.
+When `vector_at` is not `nothing`, return a (Gaussian) random vector from the tangent space
+``T_{p}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ`
+to have a zero component sum.
+"""
+rand(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}; σ::Real=1.0)
+function Random.rand(
+    M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric};
+    vector_at=nothing,
+    kwargs...,
+)
+    if vector_at === nothing
+        pX = allocate_result(M, rand)
+    else
+        pX = allocate_result(M, rand, vector_at)
+    end
+    rand!(M, pX; vector_at=vector_at, kwargs...)
+    return pX
+end
+function Random.rand(
+    rng::AbstractRNG,
+    M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric};
+    vector_at=nothing,
+    kwargs...,
+)
+    if vector_at === nothing
+        pX = allocate_result(M, rand)
+    else
+        pX = allocate_result(M, rand, vector_at)
+    end
+    rand!(rng, M, pX; vector_at=vector_at, kwargs...)
+    return pX
+end
+
+function Random.rand!(
+    rng::AbstractRNG,
+    ::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric},
+    pX;
+    vector_at=nothing,
+    σ=one(eltype(pX)),
+)
+    if isnothing(vector_at)
+        Random.randexp!(rng, pX)
+        LinearAlgebra.normalize!(pX, 1)
+    else
+        Random.randn!(rng, pX)
+        pX .= (pX .- mean(pX)) .* σ
+    end
+    return pX
+end
+
+function Random.rand!(
+    M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric},
+    pX;
+    kwargs...,
+)
+    return rand!(Random.default_rng(), M, pX; kwargs...)
+end
+
+@doc raw"""
     volume_density(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}, p, X)
 
 Compute the volume density at point `p` on [`ProbabilitySimplex`](@ref) `M` for tangent
--- a/src/manifolds/ProbabilitySimplexEuclideanMetric.jl
+++ b/src/manifolds/ProbabilitySimplexEuclideanMetric.jl
@@ -21,37 +21,6 @@ function manifold_volume(M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanM
 end
 
 @doc raw"""
-    rand(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}; vector_at=nothing, σ::Real=1.0)
-
-
-When `vector_at` is `nothing`, return a random (uniform) point `x` on the [`ProbabilitySimplex`](@ref) with the Euclidean metric
-manifold `M` by normalizing independent exponential draws to unit sum, see [Devroye:1986](@cite), Theorems 2.1 and 2.2 on p. 207 and 208, respectively.
-
-When `vector_at` is not `nothing`, return a (Gaussian) random vector from the tangent space
-``T_{p}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ`
-to have a zero component sum.
-
-"""
-rand(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}; σ::Real=1.0)
-
-function Random.rand!(
-    rng::AbstractRNG,
-    M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric},
-    pX;
-    vector_at=nothing,
-    σ=one(eltype(pX)),
-)
-    if isnothing(vector_at)
-        Random.randexp!(rng, pX)
-        LinearAlgebra.normalize!(pX, 1)
-    else
-        Random.randn!(rng, pX)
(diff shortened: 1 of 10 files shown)


kellertuer:
Hmm, this turns out to be much more involved than I thought.

a) Interaction with power groups is still quite strange (I would have hoped them passing on to PowerManifold would just be fine
b) Introducing a generic fallback has broken a few existing implementations (or fallbacks) as well? This is also very stange. Probably means I have some misunderstanding (again?) on how the group decorator is supposed to work. My hope was that basically no group needs a rand function after this (unless a much more efficient variant is known on said group compared to the manifold one).

Currently I am not able to resolve these two. So I will try again the next few days somewhere. Sorry, this was meant to be a “Saturday-morning-30-minute-PR” but might be a “maybe I understand this in a few weeks” PR then.

kellertuer:
The hint you propose, I do not understand. I can not define rand to work on the identity for _any_ decorated manifold (for tangents) – and also not just for `GroupManifolds` since the whole idea is to also provide it for the general unitary group definition.

mateuszbaran:
The only way to make `rand` work for identity is checking for it in the implementation of `rand` for group (using `if vector_at isa Identity`). You can't do that through dispatch.

kellertuer:
Oh that part is not my problem ;)

I want to map random tangents to those at the identity = Lie algebra (instead of in TpM what the manifold case would be)

kellertuer:
I think my best solution (but I struggle with your macro) is if line 306 above would omit all parameters before the manifold (so none if manifold is first but for example rng or a type) when calling `trait`. That would resolve all my problems basically.

Without that – we can not do this PR, and I could only try to implement more single Lie group `rand!`s

kellertuer:
A (boooooring) alternative is of course to define `active_traits(rng, M, rand!, ...)` to map to `activate_traits(M, rand!,...)`;)

mateuszbaran:
It wouldn't be as general as traits but much easier to do.

mateuszbaran:
I'm simply just tired of fighting with dispatching on traits and don't want to debug any new code that relies on it.

kellertuer:
> I'm not sure why can't you just dispatch on GroupManifold and forget about traits? It's boring but works fine

because some are (just) implicit group manifold by having the trait

> I'm simply just tired of fighting with dispatching on traits and don't want to debug any new code that relies on it.

Yes, me as well, as I said before, when it works it‘s great to reuse code, but a bit hard to debug. Maybe the boring variant should work fine enough, since that can remove the `rng` even for any manifold in the active traits arguments. Will check that tomorrow.

kellertuer:
Hm the “boring” approach really seems to be the better one, it also restricts this “removing a parameter” to just RNGs – and was a 3 line fix – just that it still does not work.

So if we have a `pX = zeros(3,3)` then

```
rand!(Random.default_rng(), Rotations(3), pX)
```

works perfectly fine and rotations is the `base_manifold` of `SpecialOrthogonal(3)`, but

```
rand!(Random.default_rng(), SpecialOrthogonal(3), pX)
```

still errors. As I said in the beginning – I love our traits they make a lot of things really nice _if_ they work, but they are a bit hard to debug. So I have no clue why my new code does not make this last piece of code work. And also no idea left where to look.

mateuszbaran:
I think the boring solution shouldn't cause any issues.

kellertuer:
I think by now the boring one is even the better one, since it does not (further) complicate the macro _and_ localises the “ignore something before the manifold” to really just the one case where we have that (the seed), so we do not do that “by accident” when we should not.

kellertuer:
I noticed that for https://github.com/JuliaManifolds/Manopt.jl/discussions/349 having rand and project would be nice for unitary so I adapted those from Stiefel (which only differs in tangent vector representation).

src/manifolds/ProbabilitySimplexEuclideanMetric.jl
```diff
@@ -20,37 +20,6 @@ function manifold_volume(M::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanM
...
-manifold `M` by normalizing independent exponential draws to unit sum, see [Devroye:1986](@cite), Theorems 2.1 and 2.2 on p. 207 and 208, respectively.
-
-When `vector_at` is not `nothing`, return a (Gaussian) random vector from the tangent space
-``T_{p}\mathrm{\Delta}^n``by shifting a multivariate Gaussian with standard deviation `σ`
-to have a zero component sum.
-
-"""
-rand(::MetricManifold{ℝ,<:ProbabilitySimplex,<:EuclideanMetric}; σ::Real=1.0)
-
-function Random.rand!(
```
mateuszbaran, line 37: Why did you remove this?

kellertuer: it is no longer covered by tests so I checked and it is now passed to the simplex itself. The alternative is to check whether this exact call is still valid, but then we would have to come up with a new test for code lines that we did not actively change.

mateuszbaran: This is bad then. The old calls should use this deleted method. If they don't then we perhaps need to manually tweak dispatch for simplex.

kellertuer: I See. I hope that at least one of my next PRs will not turn into an infinite round of discussions as this one does now as well.

I think this is not bad, since I am even more surprised that rand depends on a metric. But sure. then this PR will stay open for much longer until this method is back and has test coverage. The main problem then is, that I did not write this, so it will take some time to get into all the ideas behind a Euclidean metric here, since I think this is not even a manifold. but well.

kellertuer: For me at least something like this was expected and I consider that code duplication reduction.
The tests still produce valid random points and vectors even for the Euclidean metric, though.

kellertuer: The code that is called now is

https://github.com/JuliaManifolds/Manifolds.jl/blob/e46a864447274646f75fb1e79c5492770fd05a8f/src/manifolds/ProbabilitySimplex.jl#L377-L394

which for the points has an additional abs2 for numerical safety, the vectors have a `change_metric` additionally to the code here, but since the metric here is Euclidean, change_metric is the identity. So that code does the same.

kellertuer: No, that was not correct. But I thinking just had the right idea; we use base_manifold in the new dispatch version; that unwraps "too much" we should just remove one decorator. Changing that should make this version necessary again. Will check tomorrow.

mateuszbaran: The difference is that changing a metric may change the preferred distribution from which we draw random points and tangent vectors. I will see if I can fix that.

kellertuer: Yes that is what I noticed as well then, the one with changing the metric is not called. I am a bit surprised that these 2 are affected by the change, but I can check later today as well how that happens – you are right it should not.

src/manifolds/Unitary.jl
```diff
@@ -111,6 +112,45 @@ project(::UnitaryMatrices{TypeParameter{Tuple{1}},ℍ}, p) = sign(p)
...
+@doc raw"""
+    project(M::UnitaryMatrices,p)
+
+Projects `p` from the embedding onto the [`UnitaryMatrices`](@ref) `M`, i.e. compute `q`
+as the polar decomposition of ``p`` such that ``q^{\mathrm{H}}q`` is the identity,
+where ``⋅^{\mathrm{H}}`` denotes the hermitian, i.e. complex conjugate transposed.
+"""
+project(::UnitaryMatrices, ::Any, ::Any)
+
+function project!(::UnitaryMatrices, q, p)
```
mateuszbaran, line 124: The same `project!` is at Manifolds/src/manifolds/GeneralUnitaryMatrices.jl:968

kellertuer: Interesting, it was not triggered in the rand so I added it, but maybe I was also tired. we could remove it if the general one works fine.

src/manifolds/ProbabilitySimplexEuclideanMetric.jl
mateuszbaran: I think I more or less fixed it but actually I'm not sure if the dispatch was entirely correct before this PR. It looks like it was when `rng` was provided but otherwise not. I don't really like having to define those effectively repeated methods but I don't see a better way.

src/manifolds/Unitary.jl
mateuszbaran: Probably the tangent vector projection was necessary but not the point projection.

kellertuer: Ok, I am still busy today, we could try the tests without the point proj then, or for now leave it a is (and at some point spent a bit of time to reduce actual code duplication throughout the package).

src/manifolds/ProbabilitySimplexEuclideanMetric.jl
mateuszbaran: Yes, it's not ideal but `rand` is kind of an exception to our trait rules. Technically we don't have to block it propagating through `MetricManifold` because we still get a correct point, so it gets propagated by default and stuff happens in wrong order.

mateuszbaran:
It seems like there was an old projection that was working, and your new one was incorrect?

kellertuer:
That is strange, because yesterday locally that worked fine for me? But maybe I confused pX and X as representers somewhere between UnitaryMatrices and Unitary?

kellertuer:
At least I added concrete tests for Unitary random by now.

That is the project I also will return to after this – better testing in Manifolds.jl – that is my next thing to continue here in this package.

mateuszbaran:
Tests on CI stopped failing after I removed that `project!` so I guess it wasn't correct.

> That is the project I also will return to after this – better testing in Manifolds.jl – that is my next thing to continue here in this package.

That's a hard problem but maybe you will figure something out.

kellertuer:
Well I started that in the other pr and have an idea, though I am not yet sure how much the complete rework will be.

But I noticed that we have single groups with currently 10k tests, so reducing that a bit might also be nice.

kellertuer:
Tests seem a bit strange here, as if codecov had a hiccup – bit besides that I would consider this finished ;)

mateuszbaran (approved):
LGTM but you still need to bump version in Project.toml :wink:

PR #709, opened 2024-02-13
Fix several typos in the docs

kellertuer:
This fixes a few math related typos in the docs and continues to use two backticks in Documenter Markdown and Julia Docstrings instead of `$`.

I propose to not make this a new version per se but leave it on master until a feature justifies registering a new 0.9.15 version. This PR still already mentions the two above fixes in the `NEWS.md`

### DIFF
--- a/src/manifolds/SymplecticStiefel.jl
+++ b/src/manifolds/SymplecticStiefel.jl
@@ -23,7 +23,7 @@ The symplectic Stiefel tangent space at ``p`` can be parametrized as [BendokatZi
     &= \{X ∈ ℝ^{2n×2k} ∣ p^{T}J_{2n}X + X^{T}J_{2n}p = 0 \}, \\
     &= \{X = pΩ + p^sB \mid
         Ω ∈ ℝ^{2k×2k}, Ω^+ = -Ω, \\
-        &\qquad & p^s ∈ \mathrm{SpSt}(2n, 2(n- k)), B ∈ ℝ^{2(n-k)×2k}, \},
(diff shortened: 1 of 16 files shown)


src/manifolds/SymplecticStiefel.jl
```diff
@@ -452,9 +452,9 @@ That is, we find the element ``X ∈ T_p\mathrm{SpSt}(2n, 2k)``
 which solves the constrained optimization problem
 
 ````math
-    \operatorname{min}_{X ∈ ℝ^{2n×2k}} \frac{1}{2}||X - A||^2, \quad
+    \displazstyle\operatorname{min}_{X ∈ ℝ^{2n×2k}} \frac{1}{2}||X - A||^2, \quad
```
mateuszbaran, line 455: This doesn't look correct.

mateuszbaran (commented):
Everything except that one issue looks good :+1:

src/manifolds/SymplecticStiefel.jl
kellertuer: Indeed. German and Norwegian keyboards (also German and English) switch y and z so sometimes I still mess that up.

mateuszbaran: Yes, it's annoying when this happens.

mateuszbaran (approved):

src/manifolds/SymplecticStiefel.jl
kellertuer: Fixed it. So it was correct (in muscle movement, just for the wrong layout) ;)

Thanks for spotting this one! I tried checking all formulae I changed and must have missed that one.

PR #710, opened 2024-03-15
Fix some typos in working-in-charts example.

vtjeng:


kellertuer:
Thanks! The CI error is surely not your fault, that seems to be some package we use that did a breaking change without marking it correctly (or us using functions they consider internal).

mateuszbaran:
Thanks for fixes!

The cause of failure is once again BoundaryValueDiffEq.jl, again.

mateuszbaran:
Reported here: https://github.com/SciML/BoundaryValueDiffEq.jl/issues/171

mateuszbaran:
CI will be fixed in #711 by upper-bounding dependencies.

PR #711, opened 2024-03-16
Sectional curvature

mateuszbaran:
Just finishing the work on sectional curvature.


src/utils.jl
```diff
@@ -277,3 +277,83 @@ _eps_safe(::Type{T}) where {T<:Real} = eps(T)
...
+The approximation is derived from the [Bertrand–Diguet–Puiseux theorem](https://en.wikipedia.org/wiki/Bertrand%E2%80%93Diguet%E2%80%93Puiseux_theorem)
+which states that
+````math
+\kappa_p(X, Y) = \lim_{r \to 0^+} 3\frac{2\pi r-C(r)}{\pi r^3},
+````
+where ``C(r)`` is the circumference of the circle of radius ``r`` around `p` in submanifold
+of `M` spanned by `X` and `Y`. The circumference calculation method has a tendency to
+return curvature values larger than the exact ones.
+"""
+function sectional_curvature_bdp(M::AbstractManifold, p, X, Y; r::Real=1e-3, N::Int=10000)
```
kellertuer, line 318: I had to check a while to see why ist is called `bdb`

maybe `estimated_sectional_curvature` (or approximated) would fit better? We could even use an `EstimationMethod` either now or later then (when there are multiple ones).

mateuszbaran: I don't really want to over-engineer this in this PR, so maybe I will just rename it to `estimated_sectional_curvature`. I called it `_bdp` from the theorem name since there are other ways to estimate sectional curvature but we can handle that later.

kellertuer: Yes, I read the doc string quite a few times until I saw where bpd is from.

I also think for now no method is necessary, just calling it `estimates_` should be fine.

mateuszbaran: Updated :slightly_smiling_face:

kellertuer (approved):

kellertuer:
Hm, direct dependency is not so nice, but maybe necessary here to finish this in finite time. Really a bit sad that BVP breaks so often

mateuszbaran:
Yes, I don't like this solution but I hope this is only a temporary measure until they fix the bug. It seems that BVP breaks more often than all our other dependencies combined...

mateuszbaran:
Speaking of bugs in other dependencies...
```
ERROR: LoadError: UndefVarError: `pyisnone` not defined
Stacktrace:
  [1] getproperty
    @ ./Base.jl:31 [inlined]
  [2] _py_bbox(obj::PythonCall.Core.Py)
    @ Plots ~/.julia/packages/Plots/a3u1v/src/backends/pythonplot.jl:219
  [3] _update_min_padding!(sp::Plots.Subplot{Plots.PythonPlotBackend})
    @ Plots ~/.julia/packages/Plots/a3u1v/src/backends/pythonplot.jl:1253
```

mateuszbaran:
See https://github.com/JuliaPlots/Plots.jl/issues/4905 (the name of user opening that issue :smile:)

mateuszbaran:
It looks like with two new constraints, CI is passing. I'd suggest merging #710 and #712 first, and this one last, and then release 0.9.15 @kellertuer .

I've also updated the default vector transport method as discussed there: https://github.com/JuliaManifolds/Manopt.jl/discussions/349#discussioncomment-8811114 .

kellertuer:
Nice, sounds like a good plan to me :)

mateuszbaran:
Great, then I'll go with it.

PR #712, opened 2024-03-19
📚Switch to using DocumenterInterLinks.jl

kellertuer:
Switches ManifoldsBase links to use DocumenterInterLinks


.github/workflows/documenter.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           version: 1.3.361
       - uses: julia-actions/setup-julia@latest
         with:
-          version: 1.9
+          version: 1.10
```
mateuszbaran, line 20:
```suggestion
          version: "1.10"
```
Without `""` YAML considers it to be a floating point number equal to 1.1.

kellertuer: Ah yes debugged for quite a while the same issue on Manopt and was not careful enough yesterday.

kellertuer:
Ah, we seem to have the same problem with BVPs here as well.

mateuszbaran:
I've noticed that it was broken by NonlinearSolve v 3.8 so I've upper-bounded it in my PR. Hopefully CI picks the right version.

kellertuer:
Sue, still a bit annoying that they again broke their package

mateuszbaran:
And CI doesn't respect that upper bound... I've asked on Slack how to upper-bound it.

src/manifolds/VectorBundle.jl
```diff
@@ -20,7 +20,7 @@ fiber_bundle_transport(M::AbstractManifold, ::FiberType) =
     VectorBundle{𝔽,TVS,TM,VTV} = FiberBundle{𝔽,TVS,TM,TVT} where {TVS<:VectorSpaceType}
 
 Alias for [`FiberBundle`](@ref) when fiber type is a `TVS` of type
-[`VectorSpaceType`]https://juliamanifolds.github.io/ManifoldsBase.jl/stable/bases/#ManifoldsBase.VectorSpaceType).
+[`VectorSpaceType`]@extref `ManifoldsBase.VectorSpaceType`).
```
mateuszbaran, line 23:
```suggestion
[`VectorSpaceType`](@extref `ManifoldsBase.VectorSpaceType`).
```

mateuszbaran (approved):

PR #720, opened 2024-04-03
Hyperrectangle

mateuszbaran:
Mostly to make Manopt work with box constraints. The trick is just projecting gradients and projection retraction but it's still better than nothing IMO. Activity analysis needs to be done on Manopt's side.

### DIFF
--- a/src/manifolds/Hyperrectangle.jl
+++ b/src/manifolds/Hyperrectangle.jl
@@ -28,6 +28,13 @@ struct Hyperrectangle{T<:AbstractArray} <: AbstractDecoratorManifold{ℝ}
     end
 end
 
+function active_traits(f, ::Hyperrectangle, args...)
+    return merge_traits(
+        IsDefaultMetric(EuclideanMetric()),
+        IsDefaultConnection(LeviCivitaConnection()),
+    )
+end
+
 function check_point(M::Hyperrectangle, p)
     if !(eltype(p) <: Real)
         return DomainError(
@@ -380,6 +387,12 @@ Return the array dimensions required to represent an element on the
 [`Hyperrectangle`](@ref) `M`, i.e. the vector of all array dimensions.
 """
 representation_size(M::Hyperrectangle) = size(M.lb)
+
+function retract_project!(M::Hyperrectangle, r, q, Y, t::Number)
+    r .= q .+ t .* Y
+    project!(M, r, r)
+    return r
+end
 
 @doc raw"""
     riemann_tensor(M::Hyperrectangle, p, X, Y, Z)
--- a/src/manifolds/Hyperrectangle.jl
+++ b/src/manifolds/Hyperrectangle.jl
@@ -178,24 +178,24 @@ injectivity_radius(::Hyperrectangle) = Inf
     inner(M::Hyperrectangle, p, X, Y)
 
 Compute the inner product on the [`Hyperrectangle`](@ref) `M`, which is just
-the inner product on the real-valued or complex valued vector space
-of arrays (or tensors) of size ``n_1 × n_2  ×  …  × n_i``, i.e.
+the inner product on the real-valued vector space of arrays (or tensors)
+of size ``n_1 × n_2  ×  …  × n_i``, i.e.
 
 ````math
-g_p(X,Y) = \sum_{k ∈ I} \overline{X}_{k} Y_{k},
+g_p(X,Y) = \sum_{k ∈ I} X_{k} Y_{k},
 ````
 
 where ``I`` is the set of vectors ``k ∈ ℕ^i``, such that for all
 
-``i ≤ j ≤ i`` it holds ``1 ≤ k_j ≤ n_j`` and ``\overline{⋅}`` denotes the complex conjugate.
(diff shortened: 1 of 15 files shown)


src/manifolds/Hyperrectangle.jl
```diff
@@ -0,0 +1,526 @@
...
+
+function active_traits(f, ::Hyperrectangle, args...)
+    return merge_traits(
+        IsDefaultMetric(EuclideanMetric()),
+        IsDefaultConnection(LeviCivitaConnection()),
+    )
+end
+
+function check_point(M::Hyperrectangle, p)
+    if !(eltype(p) <: Real)
```
kellertuer, line 39: Would it maybe be easier to check that p is on `Euclidean(length(M.lb))` ? or `size(lb)`?

mateuszbaran: Sure, though re-using that check here still requires a non-trivial amount of code for constructing the manifold and parsing the result, so the complexity isn't reduced here much, if at all.

kellertuer: If you feel its not worth it, we can also duplicate the code here.

mateuszbaran:
Regarding activity analysis: https://web.mit.edu/dimitrib/www/ProjectedNewton.pdf . A small modification to direction update rule would be needed so that update is diagonal in directions orthogonal to boundary.

mateuszbaran:
We have an ambiguity here:
```julia
injectivity_radius(M::AbstractManifold, m::AbstractRetractionMethod) @ ManifoldsBase ~/.julia/packages/ManifoldsBase/Sk7lI/src/ManifoldsBase.jl:528
injectivity_radius(M::Hyperrectangle, p) @ Manifolds ~/work/Manifolds.jl/Manifolds.jl/src/manifolds/Hyperrectangle.jl:173
```
I thought I would leave the global `injectivity_radius` undefined (it's just 0 and not useful) but seeing this I'm not sure. What do you think @kellertuer ?

mateuszbaran:
OK, defining it to be 0 will fix the ambiguity. I can try to explain if you have any specific questions :slightly_smiling_face:

kellertuer:
I think my main problems are twofold

* what is this manifold useful for? A bit of documentation before the actual types would be good
* why does exp potentially lead to points outside the manifold. Is that considered in any way? 

Besides that I think the inner product might not be correctly documented at first sight. For i=2 X^TY is still a matrix not a number – maybe there is a trace or so missing for this case?

mateuszbaran:
> * what is this manifold useful for? A bit of documentation before the actual types would be good

Mostly optimization with box constraints, at least that's my idea.

> * why does exp potentially lead to points outside the manifold. Is that considered in any way?

I think we can't assume exponential map is always well-defined for arbitrary vectors.  Note that the default retraction is `ProjectionRetraction`, and `ProbabilitySimplex` with `EuclideanMetric` behaves the same way.

mateuszbaran:
> Besides that I think the inner product might not be correctly documented at first sight. For i=2 X^TY is still a matrix not a number – maybe there is a trace or so missing for this case?

Yes, it should have a trace. I copied it from `Euclidean` which has the same issue.

kellertuer:
I never considered Probability simplex with euclidean metric a serious candidate for a manifold ;)

mateuszbaran:
> I never considered Probability simplex with euclidean metric a serious candidate for a manifold ;)

It surely is not, it's a manifold with boundary. A different thing :slightly_smiling_face: .

mateuszbaran:
Or, more precisely, those are manifolds with corners: https://arxiv.org/abs/0910.3518

kellertuer:
Ok, while I do not see much how this manifold is useful – I think I also do not have to find it useful.
I am still a bit confused that retractions do map into the manifold and exp does not, but ok.

The only think I would like to see a bit more proimentnly – maybe in the docs before the list of doc strings starts – would be the note that this is a manifold with boundary, and their interface might not yet bet 100% fixed.

mateuszbaran:
> Ok, while I do not see much how this manifold is useful – I think I also do not have to find it useful.

With L-BFGS-B being one of the most widely used optimization algorithm, and working exactly on a hyperrectangle manifold, I think utility should be fairly clear :slightly_smiling_face: . Manopt's qN requires some small tweaks to ensure reasonable convergence for manifolds with boundary or corners but I'm planning on adding that soon-ish.

> The only think I would like to see a bit more proimentnly – maybe in the docs before the list of doc strings starts – would be the note that this is a manifold with boundary, and their interface might not yet bet 100% fixed.

Sure, I will add a note. And the interface would definitely have to be extended a bit but I haven't designed that yet in detail.

> I am still a bit confused that retractions do map into the manifold and exp does not, but ok.

Locally they both map into the manifold. I'm not sure how to explain that idea clearly but for standard Riemannian manifolds you also may have tangent vectors for which `exp` is undefined. Not all manifolds are [geodesically complete](https://en.wikipedia.org/wiki/Geodesic_manifold). It's just that we focus in Manifolds.jl on symmetric manifolds and they are geodesically complete.

kellertuer (approved):
I think all discussed points are clarified. I am looking forward to seeing this “in action”.

PR #731, opened 2024-06-16
Implement and document parallel transport on Grassmann.

kellertuer:
I recently noticed (p. 171 AMS08), that parallel transport on Grassmann (Stiefel representation) has a closed form in the `_direction` variant. So here is an implementation.
Will add tests the next few days.

# 🛣️
* [x] Test coverage
* [x] read documentation carefully that it renders correctly

### DIFF
--- a/src/manifolds/GrassmannStiefel.jl
+++ b/src/manifolds/GrassmannStiefel.jl
@@ -32,9 +32,9 @@ ManifoldsBase.@default_manifold_fallbacks (Stiefel{<:Any,ℝ}) StiefelPoint Stie
 ManifoldsBase.@default_manifold_fallbacks Grassmann StiefelPoint StiefelTVector value value
 
 function default_vector_transport_method(::Grassmann, ::Type{<:AbstractArray})
-    return ProjectionTransport()
+    return ParallelTransport()
 end
-default_vector_transport_method(::Grassmann, ::Type{<:StiefelPoint}) = ProjectionTransport()
+default_vector_transport_method(::Grassmann, ::Type{<:StiefelPoint}) = ParallelTransport()
 
 @doc raw"""
     distance(M::Grassmann, p, q)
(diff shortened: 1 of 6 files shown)


kellertuer:
hm, first just my activation of tests fail, now I am confused.

If I do
```Julia
using Manifolds
M = Grassmann(3,2)
p1 = [1.0 0.0; 0.0 1.0; 0.0 0.0]
X = [0.0 0.0; 0.0 0.0; 0.0 1.0]
p2 = exp(M, p1, X)
Z = zero_vector(M, p2)
parallel_transport_direction!(M, Z, p1, X, X)
```

it does not hit the new implementation but some path into nested-ness – though I thought these would just be on the decorator and hence the more specific definition here should work?
(Sure I have to implement the allocating variant myself uptherwise it would go to the `to` method)

Could you take a look @mateuszbaran ? I seem to miss something why It dives down the decorator though I implemented the specific version of the `!` variant. I am a bit lost why it does since it dives into some method defined by a macro and I am not able to debug that.

src/manifolds/GrassmannStiefel.jl
```diff
@@ -182,6 +183,54 @@ function log!(M::Grassmann, X, p, q)
...
+geodesic starting in direction ``\dot γ (0) = Y``.
+
+ Let ``Y = USV`` denote the SVD decomposition of ``Y``.
+Then the parallel transport is given by the formula according to Equation (8.5) (p. 171) [AbsilMahonySepulchre:2008](@cite) as
+
+```math
+\mathcal P_{p,Y} X = -pV \sin(S)U^{\mathrm{T}}X + U\cos(S)U^{\mathrm{T}}X + (I-UU^{\mathrm{T}})X
+```
+
+where the since and cosine applied to the diagonal matrix ``S`` are meant to be elementwise
```
mateuszbaran, line 199:
```suggestion
where the sine and cosine applied to the diagonal matrix ``S`` are meant to be elementwise
```

mateuszbaran (approved):

Issue #701, opened 2024-01-03
Rename Symplectic to SymplecticMatrices?


kellertuer:
Currently we usually suffix `Matrices` for quite a few manifolds where the name itself is not specific enough, e.g. `SymmetricMatrices` since we use `Symmetric` for the single matrix.

This is a bit inconsistent to one other case we have, which I therefore would like to rename

* [x] `Symplectic` should be called `SymplecticMatrices`
* [x] `SynplecticMatrix` should be called `SymplecticElement`
* [ ] `Symplectic` could be introduced as a type indicating a symplectic matrix (similar to `Symmetric`)

I think this would even only be mildly breaking, only the constructors need to be deprecated but could still keep working for a while. Breaking would be if other use these types, since we could not const them when reusing them. But it would be nice to have this consistently with LinearAlgebra in the naming

What are opinions on this? I am reworking a bit of this in #700, so I could even include it in there is the breaking part is considered fine enough.

kellertuer:
Cool. I will carefully add the functionality I need on #700, and will maybe already add a few notes what might change in the future as comments

kellertuer:
If we just deprecate the constructors, the only breaking part would be if someone dispatches on the old `Symplectic` type, the others could be kept as deprecated const definitions even.
Would that still be too bad to be considered breaking? I could otherwise include this in #700.

mateuszbaran:
I think we can do the first two points in #700 with deprecation and aliases but the third one IMO needs to wait for a breaking release.

kellertuer:
That sounds fair enough. I will do those two then next in that PR.

PR #732, opened 2024-08-03
Manifolds.jl v0.10

mateuszbaran:
List of changes:

- [x] Make Distributions.jl, HybridArrays.jl and RecursiveArrayTools.jl weak dependencies
- [x] Merge PR #683 into here and finish it.
- [x] #701 
- [x] Remove deprecated things.
- [x] Break other things that need breaking.
- [x] Investigate ambiguities.
- [x] Make CI happy
- [x] set version to v0.10

With these weak dependencies, load time of Manifolds.jl is reduced to 0.72 s on my PC. To go lower we'd have to turn both Graphs.jl and StaticArrays.jl into weak dependencies but that seems to be too much work for too little gain, and basically everything else either has marginal impact on load time or is indispensable.


kellertuer:
So this supersedes #693 ? I would close it then once this is finished.

mateuszbaran:
Yes, this PR supersedes #693 .

ext/ManifoldsDistributionsExt/distributions.jl
```diff
@@ -0,0 +1,128 @@
+"""
+    FVectorvariate
+
+Structure that subtypes `VariateForm`, indicating that a single sample
+is a vector from a fiber of a vector bundle.
+"""
+struct FVectorvariate <: VariateForm end
```
kellertuer, line 7: Is that a good idea to introduce new types in extensions? I would prefer to only add new methods to functions in extensions.

mateuszbaran: It's not a great design but it's an intermediate thing until we have "manifold distributions done right" in a separate package. Putting these types there makes them private. If anyone relies on that they can still use it but I would discourage any new uses.

kellertuer: Hm, ok, I get your point. I am not a fan of this, but I do see, that we would need someone to do distributions right. Let‘s find someone ;)

mateuszbaran: I fully support your idea of finding someone to do distributions right :wink:

kellertuer: We could try to ask in the forum. Besides that we could try to convince one of my students next year to work a bit on that, but that is probably more like for us to get into that then.

mateuszbaran: I think it's not a good student project because it's in part an API development problem. There is also lots of confusing papers with misleading claims on this topic. Anyway, I'm not going to be looking for anyone myself but if you find someone, I could guide them.

kellertuer: That‘s why I said they probably can bring me only to know a bit more about it, since it will be quite some API design. 

Just out of interest I might ask in the forum if someone would be interested to join. but its also not my main focus, I am ok with our current `rand(M)` stuff, that is what I mainly need.

kellertuer:
In relation to #701, would it make sense to introduce the `Symplectic` type for the matrices like we have for skew symmetric and from linear algebra for symmetric matrices?

ext/ManifoldsDistributionsExt/distributions.jl
kellertuer: I just felt like opening something here https://discourse.julialang.org/t/call-for-contributors-distributions-on-manifolds/117820 – let‘s see, whether someone would be interested in something like a ManifoldDistributions.

mateuszbaran:
> In relation to #701, would it make sense to introduce the `Symplectic` type for the matrices like we have for skew symmetric and from linear algebra for symmetric matrices?

Maybe but it can be added later in a non-breaking release so I don't want to work on that interface in this PR.

kellertuer:
That is correct, the breaking change would happen here to “free” the name. Adding the matrix type could hen later be a nonbreaking change.

NEWS.md
```diff
@@ -11,9 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `Distributions.jl`, `RecursiveArrayTools.jl` and `HybridArrays.jl` were moved to weak dependencies to reduce load time and improve extensibility.
 * `translate_diff`, `inv_diff` and thus `apply_diff_group`, are available for all the groups with invariant tangent vector storage
-* `adjoint_action` takes a direction argument
-* `adjoint_action!` is the necessary method to implement in any given group for the above to work properly
-* On groups with invariant vector storage, `exp` and `log` are based on `exp_lie` and `log_lie` and thus group invariant
+* `adjoint_action` takes a direction argument; by default it is `LeftAction`.
+* `adjoint_action!` is the necessary method to implement in groups with left-invariant tangent vector storage.
```
kellertuer, line 15:
```suggestion
* `adjoint_action!` is the necessary method to implement in groups with left-invariant tangent vector representation.
```

Am I right here? I would make the distinction as follows:
* __representation__ is the question/indication how the vector is represented mathematically (in the Lie algebra, in a tangent space, ....)
* __storage__ is the type of data stored, a single array or a sparse array

The abstract type below also is called representation.

mateuszbaran: Yes, representation is more suitable here.

kellertuer: I will try to find time – maybe on the weekend – to check that these two terms are consistently used throughout. I think there was a mixup on the other PR as well somewhen.

mateuszbaran:
@olivierverdier I've adapted SpecialEuclidean here so that it has two possible tangent vector representations: the old one, called `TangentVectorRepresentation` (which is the default) and `LeftInvariantRepresentation`, which is the representation from your PR #683 . Choosing it just requires calling `SpecialEuclidean(3; gvr=LeftInvariantRepresentation())` instead of `SpecialEuclidean(3)`. Is that fine for you?

The tests are now passing up to one rare random issue with SpecialEuclidean and some missing coverage.

NEWS.md
mateuszbaran: Yes, that's a good idea, thanks! The PR is now more or less complete, I'm only polishing it and consider if I should maybe look for anything else to break.

kellertuer:
What does `gvr` stand for? Can we maybe use a keyword argument that is a bit longer/descriptive? `tangent_vector_representation` would be one coming to my mind.

mateuszbaran:
It was supposed to stand for "group vector representation". We can change it but I'd prefer something shorter than `tangent_vector_representation`.

mateuszbaran:
OK, I see, then I will make `LeftInvariantRepresentation` the default for `SpecialEuclidean` and name the old one `HybridTangentRepresentation`.

kellertuer:
Concerning 3, @olivierverdier , if you feel that that group can also live well within Manifolds.jl, we can surely include it here as well. If you want to keep it in your own package, that is of course also just fine :)

kellertuer:
> It was supposed to stand for "group vector representation". We can change it but I'd prefer something shorter than `tangent_vector_representation`.

that was just my first idea. Rust `representation=`is maybe a bit ambiguous? Ah the types carry that as well, maybe `vectors=`? that reads nicely in `vectors=LeftInvariantRepresentation()`.

mateuszbaran:
`vectors` sounds fine :+1:

mateuszbaran:
I've copied a part of the rotations tutorial from #355 to have a more introductory section in the group how-to.

kellertuer:
At first glance looks good to me.

NEWS.md
```diff
@@ -5,6 +5,36 @@ All notable changes to this project will be documented in this file.
...
+* `Distributions.jl`, `RecursiveArrayTools.jl` and `HybridArrays.jl` were moved to weak dependencies to reduce load time and improve extensibility.
+* `translate_diff`, `inv_diff` and thus `apply_diff_group`, are available for all the groups with invariant tangent vector storage.
+* `SpecialEuclidean` group now has a different default tangent vector representation, the left-invariant one; to get the old representation pass `vectors=HybridTangentRepresentation()` to the constructor of `SpecialEuclidean`.
+* `adjoint_action` takes a direction argument; by default it is `LeftAction`.
+* `adjoint_action!` is the necessary method to implement in groups with left-invariant tangent vector representation.
+* Fixed a few typos in the doc string of the SPD fixed determinant description.
+
+### Added
+
+* Introduced `exp_inv` and `log_inv` based on `exp_lie` and `log_lie`. They are invariant to group operation.
```
kellertuer, line 21:
```suggestion
* Introduced `exp_inv` and `log_inv` based on `exp_lie` and `log_lie`. They are invariant to the group operation.
```

tutorials/groups.qmd
```diff
@@ -0,0 +1,244 @@
...
+| `ProductGroup` | Product of connections in each submanifold | Yes if both connections are invariant separately, otherwise no |
+| `SemidirectProductGroup` | Same as underlying product | ❌ |
+| `TranslationGroup` | `CartanSchoutenZero` | ✅ |
+| `CircleGroup` | `CartanSchoutenZero` |✅|
+| `GeneralLinearGroup` | Metric connection from the left invariant metric induced from the standard basis on the Lie algebra | ❌ |
+| `GeneralUnitaryMultiplicationGroup` | `CartanSchoutenZero` (explicitly)  | ✅ |
+| `HeisenbergGroup` | `CartanSchoutenZero` | ✅ |
+| `SpecialLinearGroup` | Same as `GeneralLinear` | ❌ |
+
+: Invariance of `exp` and `log` for various groups {#tbl-exp-log-inv}
```
mateuszbaran, line 234: @kellertuer do you know why this just renders as text "[TABLE]" in docs instead of the actual table? Do I need some  `{=commonmark}` thing around the table?

kellertuer: I can check tomorrow (after my exams), I would have thought since this is in markdown that it stays as is, but maybe your Markdown is slightly off from the common mark quarto assumes? (To me it looks fine, but my idea of a markdown table is maybe also quite intuitive. Does it already `[TABLE]` in the markdown produced or only in Documenter? 

And sure, `{=commonmark}` would fix it probably but I would prefer fixing the actual error and not the symptom.

mateuszbaran: "[TABLE]" is already in the .md file. I tried fixing it but nothing seems to work except `{=commonmark}`.

kellertuer: I hope there is just a missing space somewhere, I can check later after my oral exams (probably around 5).

kellertuer: Hm our setup is maybe broken on that even just an empty file with an example table from the quarto page renders into `[TABLE]` so I have to check what is happening there.

kellertuer: 🥳(after another oral exam) found it! Your lines were too long, but I could add a `variant:` to quarto, after reading about the corresponding Pandoc extension and now tables are rendered over as is if they are too long.

I also changed the column alignment to (left, left, center) instead of all of them being right aligned.

So this should now work, though the last column does not look that nice, maybe one could do the first with an orange circle or such and explain that below the table?

mateuszbaran: Nice! But it looks like we still need `{=commonmark}`, right?

I will replace the first one with a circle.

kellertuer: It works without, sorry I had that tested in an extra file to have that created faster and missed to adapt the code block. So I just committed that.
Makes typesetting tables easier then now :)

mateuszbaran: Cool, thanks!

mateuszbaran:
I think this is more or less ready for review now. I don't see any more breaking changes worth making.

NEWS.md
```diff
@@ -5,6 +5,41 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.0] – unreleased
+
+### Changed
+
+* `Distributions.jl`, `RecursiveArrayTools.jl` and `HybridArrays.jl` were moved to weak dependencies to reduce load time and improve extensibility.
```
kellertuer, line 12: We could maybe mention what this implies in practice? Everyone who until now uses the `ArrayPartition` (probably everyone using product manifold until now) has to load `RecursiveArrayTools` manually now. 

Where does this change things for the other two?

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+
+This is a short overview of group support in [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/) and how to get started working with them.
+
+Groups currently available in `Manifolds.jl` are listed in ``[group section](@ref GroupManifoldSection)``{=commonmark}.
+
+You can read more about the theory of Lie groups for example in [Chirikjian:2012](@cite).
+An example application of Lie groups in robotics is described in [SolaDerayAtchuthan:2021](@cite).
+
+First, let's load libraries we will use.
+`RecursiveArrayTools` is necessary because `ArrayPartition` defined in it is used as one of the possible representations of elements of product and semidirect product groups.
```
kellertuer, line 25: We could for both packages maybe write `.jl` behind and link to the packages docs or repository?
I would also leave out the “defined in it”; one could just write “its” upfront ArrayPartition. We could even add a link to that type but maybe not so important.

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+Let's say we want to define a manifold point `p_i` some rotation θ from the [`identity_element`](@ref) reference rotation `p0` (another point on the manifold that we will use as reference)
+
+```{julia}
+# + radians rotation from x-axis on plane to point i
+xθi = π/6
+```
+
+### From Coordinates
+
+To get our first Lie algebra element we can use the textbook [`hat`](@ref), or equivaliently a more generalized [`get_vector`](@ref), function:
```
kellertuer, line 70: I do not understand the phrase “the textbook” here. What does this refer to?

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+In this case, the same would work given the base manifold [`Rotations(2)`](@ref):
+```{julia}
+_X_ = hat(M, p0, xθi)             # Lie groups definition
+_X = get_vector(M, p0, xθi, B)   # generalized definition
+@assert _X_ == xXi; @assert _X == xXi
+```
+One more caveat here is that for the Rotation matrices, the tangent vectors are always stored as elements from the Lie algebra.
+:::
+
+Now, let's place this algebra element on the manifold using the exponential map [`exp`](@ref):
```
kellertuer, line 91: “place this” also sounds a bit strange and informal. What does this actually refer to?

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+xRi = exp(G, p0, xXi)
+# similarly for known underlying manifold
+xRi_ = exp(M, p0, xXi)
+
+@assert isapprox(xRi, xRi_)
+```
+
+### To Coordinates
+
+The logarithm transform from the group back to algebra (or coordinates) is:
```
kellertuer, line 102:
```suggestion
The logarithm transforms from the group back to algebra (or coordinates) is:
```

kellertuer: But also “transforms” is not clear to me what this means here.

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+```{julia}
+xXi_ = log(G, p0, xRi)
+xXi__ = log(M, p0, xRi)
+@assert xXi ≈ xXi__
+```
+
+Similarly, the coordinate value can be extracted from the algebra using [`vee`](@ref), or directly from the group using the more generalized [`get_coordinates`](@ref):
```
kellertuer, line 109: “the coordinate value” is not correct I think? especially also because the function has a plural ;) but I think it is a bit vague, and one might explain, that the basis is just one basis vector here?

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+@assert isapprox(xθi, xθi_); @assert isapprox(xθi, _xθi_)
+@assert isapprox(xθi, xθi__); @assert isapprox(xθi, _xθi__)
+```
+
+::: {.callout-note}
+The disadvantage might be that the representation of `X` is not nice, i.e. it uses too much space or doing vector-calculations is not so easy.  E.g. fixed rank matrices are overloaded for all vector operations, but maybe that is “not enough” for a general user application that really wants vectors. But: Given a basis `B` one can look at the coefficients of the tangent vector `X` with respect to basis `B`.  From the Sphere example note above the basis vectors would be `Y1=[0.0, 1.0, 0.0]` and `Y2=[0.0, 0.0, 1.0]`, so to get the coordinates would be `c = get_coordinates(Sphere(2), p, X, B)`.  Vice versa, if you have a coordinate vector with respect to a basis `B` of the tangent space at `p` and want the vector back, then you do `X2 = get_vector(M, p, c, B)` (and you should have `X2==X`).  The coordinate vector `c` might also have the advantage of saving memory. For example SPD matrix tangent vectors take $n^2$ entries to save, i.e. storing the full matrix, but the coordinate vectors only take $\frac{n(n+1)}{2}$.
```
kellertuer, line 126: This is a bit long and I do no understand what this is for. Sphere example above? also a short reference to fixed rank is unclear to me why this is here.

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+:::
+
+
+### Actions and Operations
+
+With the basics in hand on how to move between the coordinate, algebra, and group representations, let's briefly look at composition and application of points on the manifold.  For example, a `Rotations(n)` manifold is the mathematical representation, but the points have an application purpose in retaining information regarding a specific rotation.
```
kellertuer, line 132: rotations could be a `@ref` maybe?

tutorials/groups.qmd
```diff
@@ -0,0 +1,242 @@
...
+for aV in [aV1; aV2; aV3]
+    bV = apply(A_left, bRa, aV)
+    # test we are getting the rotated vectors in Euclidean(2) as expected
+    @assert isapprox(bV[1], norm(aV) * cos(bθa))
+    @assert isapprox(bV[2], norm(aV) * sin(bθa))
+end
+```
+
+::: {.callout-note}
+In general, actions are usually non-commutative and the user must therefore be weary of [`LeftAction`](@ref) or [`RightAction`](@ref) needs. As in this case, the default `LeftAction()` is used.
```
kellertuer, line 158: what does “weary” mean here? Also from thereon the grammar of the sentence seems to be broken. I do not understand this sentence.

kellertuer (commented):
Here are a few comments, mainly on the tutorial. Am I right that this is based on another PR this would close as well then?

mateuszbaran:
> Am I right that this is based on another PR this would close as well then?

The first part of the tutorial is copied from #355 with minor editing so we can close that PR after this is merged.

NEWS.md
mateuszbaran: Practical implications are already discussed in the "Removed" section, do you think it's insufficient?

tutorials/groups.qmd
mateuszbaran: I will rephrase it.

tutorials/groups.qmd
mateuszbaran: Good point, I didn't catch it when reading this for the first time.

tutorials/groups.qmd
mateuszbaran: I wasn't sure about this note either so maybe let's remove it?

kellertuer: Yes, I think until we understand more closely what disadvantage is meant, we should remove this.

NEWS.md
kellertuer: Removed just says which ones are no longer removed – either there or here we should then mention “where to get them from instead”

mateuszbaran: I've added some more details. For `HybridArrays.jl` there are no practical implications, and for `Distributions.jl` its mostly "if you have to use it, it's still there but private; we would like to have a proper exported replacement in the future but other projects take priority"

tutorials/groups.qmd
mateuszbaran: I've rephrased it.

tutorials/groups.qmd
mateuszbaran: I have rephrased it.

tutorials/groups.qmd
mateuszbaran: I have rephrased this.

PR #734, opened 2024-08-07
RFC / WIP Start submersions

kellertuer:
This is a start to get submersions and their differentials implemented for the manifolds where that is reasonable. 

Before continuing a bit of feedback about the approach and naming of the functions would be great.

### DIFF
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -385,7 +385,7 @@ include("groups/group.jl")
 include("distributions.jl")
 include("projected_distribution.jl")
 include("statistics.jl")
-
+include("submersion.jl")
 # Meta Manifolds II: Products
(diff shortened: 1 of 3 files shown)


mateuszbaran:
As far as I can tell, Boumal and Absil's books you refer to in the comment refer to significantly different things. You probably want the $$h$$ function from Boumal's section 7.7 ("Manifolds defined by h(x) = 0)? I wouldn't call that a submersion, maybe something related to level sets?

kellertuer:
We also indeed might not need these, we just noticed, so it would anyways be a point whether we do those.

It is a super special case of a submersion, but I would also be fine with level set

kellertuer:
Since for now we do not directly need these, I will close this PR and open an issue for that first to discuss the details before restarting an implementation.

PR #746, opened 2024-09-13
`translate_diff` documentation clarification

olivierverdier:
The order in description and equations was confusing


kellertuer:
Thanks, now I also understand all 4 cases – I was a bit lost in the old description and hence even did not follow up on the corresponding PR then any longer – now it makes more sense for the as well – thanks for fixing that!

PR #747, opened 2024-09-14
GroupManifold accepts two arguments

olivierverdier:
The default representation for GroupManifold is `LeftInvariantRepresentation`, this ensures backward compatibility with previous versions of `Manifolds.jl`

### DIFF
--- a/src/groups/GroupManifold.jl
+++ b/src/groups/GroupManifold.jl
@@ -28,11 +28,8 @@ struct GroupManifold{
     vectors::VR
 end
 
-function GroupManifold(
-    M::AbstractManifold{𝔽},
-    op::AbstractGroupOperation,
-    rep::AbstractGroupVectorRepresentation=LeftInvariantRepresentation(),
-) where {𝔽}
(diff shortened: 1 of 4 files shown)


kellertuer:
Looks reasonable to me. Thanks :) Could you wait a bit for #748, then add this as an entry to the changelog (including setting the date); and bump the version in the project toml? Then wo could register this as 0.10.2

src/groups/GroupManifold.jl
```diff
@@ -24,6 +24,14 @@ struct GroupManifold{
     vectors::VR
 end
 
+function GroupManifold(
+    M::AbstractManifold{𝔽},
+    op,
+    rep=LeftInvariantRepresentation(),
```
mateuszbaran, line 30:
```suggestion
    op::AbstractGroupOperation,
    rep::AbstractGroupVectorRepresentation=LeftInvariantRepresentation(),
```

mateuszbaran (commented):
I think this is fine, I'd only adding this default value to the docstring of `GroupManifold` and changelog.

kellertuer (approved):
lgtm.

PR #751, opened 2024-09-25
Fix typo in atlases documentation

GregVernon:
Typo can be seen on https://juliamanifolds.github.io/Manifolds.jl/stable/features/atlases.html

image

Looks like a missing `\` so I added it, though I haven't previewed if that does, in fact, fix it.


kellertuer:
Thanks for spotting that, indeed there is a `\` missing. I am confident this does not need a preview. It will be visible in the dev-version after merge and be part of the next release then.

Thanks for the PR.

PR #754, opened 2024-10-01
Basis number system swap

mateuszbaran:
- [x] Adapt `Euclidean`
- [x] Adapt `ProjectiveSpace` (I'm not sure the old code is even correct for real basis of complex projective space?)
- [x] Adapt `SymmetricMatrices`
- [x] Adapt quaternionic unitary matrices?


mateuszbaran:
I've made the tests for ProjectiveSpace pass but I haven't verified if the code is indeed correct (it's just the same logic as on master). Notably, sphere only has the real coefficient basis.

kellertuer:
I am also not sure the old Euclidean code was correct. But then we should check that. Just that today I am busy (a bit behind with my lecture for tomorrow, and a bit of grading). So I'll probably only tomorrow have time to take a look.

mateuszbaran:
I haven't found any issues with `Euclidean` -- do you have a specific concern? Anyway, this can wait until tomorrow :).

kellertuer:
The original example

```
using Revise
using Manifolds, ManifoldDiff, RecursiveArrayTools, FiniteDifferences

r_backend = ManifoldDiff.TangentDiffBackend(
    ManifoldDiff.FiniteDifferencesBackend()
)

f(p) = 1
grad_f(M,p) = ManifoldDiff.gradient(M, f, p, r_backend)
M = ℂ^3 × PowerManifold(Rotations(6), NestedPowerRepresentation(),3)
p = rand(M)
# causes the same error
grad_f(M,p)
```

also threw errors in Euclidean, but I have to check whether your fix in ManifoldsBase already fixed that then.

kellertuer (approved):
This looks good to me, if code coverage agrees of course.

PR #755, opened 2024-10-02
add segre and warped segre manifold

sjacobsson:


kellertuer:
Thanks for getting this started!
If you grant me access to your repository/fork, I can also push a few small things along the way.

A very short first look

* Signatures of functions in the doc string can also just be one line
* we usually added doc-strings to the non-inlace variant (e.g `exp`) while implementing the mutating one (e.g. `exp!`).
* doc strings would be great with a formula for most of the functions (inner, exp, log, norm,...)
* you already added the reference, you can refer to that in the doc string as well.
* I can help setting up the following as well
  * a separate page in the documentation for the new manifold
  * a bit for file structures see the other manifolds different files for different metrics
  * a first idea for testing

sjacobsson:
> Thanks for getting this started! If you grant me access to your repository/fork, I can also push a few small things along the way.
> 
> A very short first look
> 
>     * Signatures of functions in the doc string can also just be one line
> 
>     * we usually added doc-strings to the non-inlace variant (e.g `exp`) while implementing the mutating one (e.g. `exp!`).
> 
>     * doc strings would be great with a formula for most of the functions (inner, exp, log, norm,...)
> 
>     * you already added the reference, you can refer to that in the doc string as well.
> 
>     * I can help setting up the following as well
>       
>       * a separate page in the documentation for the new manifold
>       * a bit for file structures see the other manifolds different files for different metrics
>       * a first idea for testing

Thanks, you should now be added as a collaborator to the fork.

I'll take a look at the doc stuff and try to conform to the rest of the repo. For the testing, I never figured out how to run the tests locally so I just commited my best guess. But once tests run, they should at least guarantee that the methods that Segre implements are consistent with each other.

kellertuer:
No worries! Some parts of such a PR are a bit involved. I will _both_ set up the tests for you and help you running them locally :) 
On the other hand – tests are quite important to make sure the code continues to work as we expect, so we should definetly add them – also because we have a few checks here, that require to have a certain fraction (quite large part actually)of your code covered by tests.
But again, no worries, I am confident we will manage that together!

sjacobsson:
Hi again,
Sorry, is there some way to view the documentation that I'm writing? I tried to run docs/make.jl but it ran into some errors.

kellertuer:
HI,
an easy, but slow way, is to make sure Manifolds.jl compiles at least and then you can (now with the label added) render the docs here If you want.

Locally on your machine you have to set the package into development mode and then call `docs/make.jl` (either from shell as it is a shell script or including it because it is also a Julia script).

The development mode is maybe the tricky one (see https://pkgdocs.julialang.org/v1/managing-packages/#developing), running from shell should be a bit easier. We wrote the docs in its own environment so that should work fine.
It might be that running the make fiel for the docs itself errors, that happens whenever there is documented functions/types that are not yet added to the documentation. But as I said, when I find time I can create the page for your manifold and the metric.

If you can not get it to run you could post the error message here or we can check (after I found time) in a video call.

sjacobsson:
Thanks, I'll try those things. No stress!

mateuszbaran:
Thanks for the contribution! I see there is an error with an unrelated test, I will fix it soon in a separate PR.

kellertuer:
I allowed myself to fix a few small things here

* merged master so the last fix from Mateusz also works here (not related to your code, but made CI fail)
* created a page in the docs.
* made sure docs work
* made Manifolds.jl compile – you had a few undeclared parameters. Not using them, one can even remove them
  * usually our field 𝔽 is the first parameter, I switched the order in `Segre` to make it first there as well since it confused me _and_ makes a few function signatures nicer in the metric file.
  * I simplified said signatures – V could be removed in most places
  * removes all `@assert` within a function. We take the approach “input is correct/valid“ as far as possible – there is the `ValidatioNManifold` to activate those tests on a global level for a manifold. The advantage their is you can even choose the level of check – do you want to see an error for example or just an info
  * I removed the `exp` for now since that default we have implemented as well – if so we might want to write the corresponding `allocate_result` (https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/7b52f311f3cc5bcd4671edb5479f798250c02876/src/exp_log_geo.jl#L16) – then much more functions also only need their `!`-variants.
* it would maybe be nicer to document `is_point` instead of `check_vector´ – it would also be great to have a bit more of a desciption there.
* checked that tests are included
  * they currently still fail, since you initialise some `AlphaWarpedMetric`? Was that renamed?
* ran JuliaFormatter (the format check here, run it with `using JuliaFormatter; format(".")` in the base folder of `Manifolds.jl`)
* started an entry in the news.
* Commented out tests for now, since then we get a first code coverage report.
* made them “standalone” they can just be run with `include("test/manifolds/segre.jl")` instead of running all tests – load other packages you need also there (and check they are in the extras of the project toml)

What would be great
* continue with the documentation, the formulae deyail in the beginning looks great (will read them carefully somewhen)
* check performance
* check test coverage
* add you to the contributors also in the zenodo info file for example
* I did not yet read any of the docs or code more closely but will definetly do that to learn about the new manifold as well

Besides all these comments – super cool that you got this started!

sjacobsson:
Thanks for fixing those things!

The problem is indeed a rename of AlphaWarpedMetric. I still don't know if the current name, WarpedMetric, is a good choice. The problem is that the euclidean metric is already a warped product metric, but then you can have a whole family of metrics based on your warping parameter and most of those are not euclidean...

Sorry, I still can't find where to view the documentation.

sjacobsson:
I shall meditate on the name for today and tomorrow, otherwise let's just go with WarpedMetric.

I also tried running make.jl but I get the following errors:
```
simonj@beltrami:~/Workspace/Manifolds.jl$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.3 (2023-08-24)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using Manifolds

help?> Segre
search: Segre NestedReplacingPowerRepresentation base_group isinteger HeisenbergGroup SymplecticGrassmann

      \mathcal{S} = \operatorname{Seg}(𝔽^{n_1} \times \dots \times 𝔽^{n_d})

  is the space of rank-one tensors in 𝔽^{n_1} \otimes \dots \otimes 𝔽^{n_d}.

  When 𝔽 = ℝ, the Segre manifold is represented as

      \mathcal{S} \sim ℝ^{+} \times S^{n_1 - 1} \times \dots \times S^{n_d - 1}.

  This is a local diffeomorphism, and the manifold is a locally a warped product
  (https://en.wikipedia.org/wiki/Warped_product) of ℝ^{+} and S^{n_1 - 1} \times \dots \times S^{n_d -
  1}. The tuple (n_1, \dots, n_d) is called the valence of the manifold.

  The geometry is summarized in JacobssonSwijsenVandervekenVannieuwenhoven:2024 (@cite).

  Constructor
  ≡≡≡≡≡≡≡≡≡≡≡≡≡

  Segre(valence::NTuple{V, Int}; field::AbstractNumbers=ℝ)

  Generate a valence V Segre manifold.

julia> include("docs/make.jl")
  Activating project at `~/Workspace/Manifolds.jl/docs`
   Resolving package versions...
    Updating `~/Workspace/Manifolds.jl/docs/Project.toml`
  [1cead3c2] ~ Manifolds v0.10.4 `..` ⇒ v0.10.4 `~/Workspace/Manifolds.jl`
    Updating `~/Workspace/Manifolds.jl/docs/Manifest.toml`
  [1cead3c2] ~ Manifolds v0.10.4 `..` ⇒ v0.10.4 `~/Workspace/Manifolds.jl`
  No Changes to `~/Workspace/Manifolds.jl/docs/Project.toml`
  No Changes to `~/Workspace/Manifolds.jl/docs/Manifest.toml`
[ Info: Precompiling ManifoldsRecipesBaseExt [37da849e-34ab-54fd-a5a4-b22599bd6cb0]
    CondaPkg Found dependencies: /home/simonj/Workspace/Manifolds.jl/docs/CondaPkg.toml
    CondaPkg Found dependencies: /home/simonj/.julia/packages/PythonCall/Nr75f/CondaPkg.toml
    CondaPkg Found dependencies: /home/simonj/.julia/packages/PythonPlot/469aA/CondaPkg.toml
    CondaPkg Dependencies already up to date
[ Info: Precompiling ManifoldsRecursiveArrayToolsExt [e29539f8-2f71-57d7-85ce-957fb5efd4f1]
[ Info: Precompiling ManifoldsOrdinaryDiffEqExt [95431769-3d64-508c-9576-79148c2c82de]
[ Info: Precompiling ManifoldsBoundaryValueDiffEqExt [eb713886-0b96-5c41-a09d-5e8967e02f7d]
[ Info: Precompiling ManifoldsNLsolveExt [b22937d9-c54a-59ec-962e-7448ad3eacb1]
[ Info: Precompiling ManifoldsOrdinaryDiffEqDiffEqCallbacksExt [f484757d-caa2-58a4-a95c-7868e5350983]
[ Info: SetupBuildDirectory: setting up build directory.
ERROR: LoadError: 'tutorials/getstarted.md' is not an existing page!
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] walk_navpages(visible::Bool, title::String, src::String, children::Vector{Any}, parent::Documenter.NavNode, doc::Documenter.Document)
    @ Documenter ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:185
  [3] walk_navpages
    @ ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:200 [inlined]
  [4] walk_navpages
    @ ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:202 [inlined]
  [5] #81
    @ ./none:0 [inlined]
  [6] iterate
    @ ./generator.jl:47 [inlined]
  [7] collect(itr::Base.Generator{Vector{Pair{String, String}}, Documenter.var"#81#82"{Documenter.NavNode, Documenter.Document}})
    @ Base ./array.jl:782
  [8] walk_navpages
    @ ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:203 [inlined]
  [9] walk_navpages
    @ ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:190 [inlined]
 [10] walk_navpages(title::String, children::Vector{Pair{String, String}}, parent::Nothing, doc::Documenter.Document)
    @ Documenter ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:199
 [11] walk_navpages(p::Pair{String, Any}, parent::Nothing, doc::Documenter.Document)
    @ Documenter ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:202
 [12] (::Documenter.var"#81#82"{Nothing, Documenter.Document})(p::Pair{String, Any})
    @ Documenter ./none:0
 [13] iterate
    @ ./generator.jl:47 [inlined]
 [14] collect_to!(dest::Vector{Documenter.NavNode}, itr::Base.Generator{Vector{Any}, Documenter.var"#81#82"{Nothing, Documenter.Document}}, offs::Int64, st::Int64)
    @ Base ./array.jl:840
 [15] collect_to_with_first!
    @ ./array.jl:818 [inlined]
 [16] collect(itr::Base.Generator{Vector{Any}, Documenter.var"#81#82"{Nothing, Documenter.Document}})
    @ Base ./array.jl:792
 [17] walk_navpages
    @ ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:203 [inlined]
 [18] runner(#unused#::Type{Documenter.Builder.SetupBuildDirectory}, doc::Documenter.Document)
    @ Documenter ~/.julia/packages/Documenter/C1XEF/src/builder_pipeline.jl:133
 [19] dispatch(#unused#::Type{Documenter.Builder.DocumentPipeline}, x::Documenter.Document)
    @ Documenter.Selectors ~/.julia/packages/Documenter/C1XEF/src/utilities/Selectors.jl:170
 [20] #89
    @ ~/.julia/packages/Documenter/C1XEF/src/makedocs.jl:272 [inlined]
 [21] withenv(::Documenter.var"#89#91"{Documenter.Document}, ::Pair{String, Nothing}, ::Vararg{Pair{String, Nothing}})
    @ Base ./env.jl:197
 [22] #88
    @ ~/.julia/packages/Documenter/C1XEF/src/makedocs.jl:271 [inlined]
 [23] cd(f::Documenter.var"#88#90"{Documenter.Document}, dir::String)
    @ Base.Filesystem ./file.jl:112
 [24] #makedocs#87
    @ ~/.julia/packages/Documenter/C1XEF/src/makedocs.jl:271 [inlined]
 [25] top-level scope
    @ ~/Workspace/Manifolds.jl/docs/make.jl:95
 [26] include(fname::String)
    @ Base.MainInclude ./client.jl:478
 [27] top-level scope
    @ REPL[3]:1
in expression starting at /home/simonj/Workspace/Manifolds.jl/docs/make.jl:95

julia> 
```

kellertuer:
Oh, sorry! I forgot about the (bit involved) Quarto examples. You either have to install Quarto (and add `--quarto`) to render them or pass `--exclude-tutorials` to the arguments, and yeah sorry that means for now it does not just work by include, since that both are command line args.

Sorry forgot about those nitty gritty details. And sure, for best of cases we should come up with something that “just works”, but for now rendering Quarto with pure Julia is still not yet there (and we would depend on that).
So for now you would have to call it from terminal then, sorry.

sjacobsson:
No worries! Sorry for bothering you when you said you were busy, I just wanted to be able to see the docs so that I could work on them in the meantime.

I get the same error again with `docs/make.jl --exclude-tutorials` :(
When I look in `make.jl` there is an `if "--quarto" ∈ ARGS`, but I don't see any lines with `--exclude-tutorials`.

kellertuer:
Ah, hm, yes. That comes from “I am busy but I can answer this in 2 minutes so let's just do it”-answers (mine, not yours).

What I wrote does hold for Manopt.jl, but not yet for Manifolds.jl (but then we should introduce this soon!) – so then, sorry, you would have to run first with `--quarto` to generate the files Documenter complains about and for that you have to install quarto,...
so maybe – you were right from the start – you can currently only render the docs with a bit of effort locally.
But since Manopt.jl has a solution for that I will transfer it here as soon as time permits.

sjacobsson:
I got it to work now by commenting out lines 109-114:
```
            "🚀 Get Started with `Manifolds.jl`" => "tutorials/getstarted.md",
            "work in charts" => "tutorials/working-in-charts.md",
            "perform Hand gesture analysis" => "tutorials/hand-gestures.md",
            "integrate on manifolds and handle probability densities" => "tutorials/integration.md",
            "explore curvature without coordinates" => "tutorials/exploring-curvature.md",
            "work with groups" => "tutorials/groups.md",
```
Thanks for the help! Now I'll try to write a nice documentation page like the ones for the Stiefel etc :)

kellertuer:
Ah, yeah, the ARG `--exclude-tutorials` in principle does the same thing ;) Great that you got it to work, just remember not to commit the commenting out.

I also got you started already on such a page, see `docs/src/manifolds/segre.md`, which automatically includes all docs from both files you write – but sure a bit more structure like in the Stiefel page or the SPD case are also nice to have, especially with a default metric and a second (the warped) one.

sjacobsson:
I see the segre.md that you added, very nice! I also see some texing errors that I made along the way. Thanks!

kellertuer:
HI,
it was a bit busy the last month on my side. How is your implementation going, anything I can help with?

sjacobsson:
Hi Ronny,
For me it has been busy as well. Sorry for not getting back to you until now. I think there are lots of small things that I need help with, but right now the biggest thing is getting the tests to run. I have a bunch of tests that I wrote as a script, and if you have the time maybe you could help me get them into the test suite.

kellertuer:
Sure, maybe not the next few days - busy finishing a semester, two talks and some travels - but turning the script into tests should not be too complicated.

You can also get inspired by what the other manifold have in tests. Otherwise I'll find some time next week

sjacobsson:
Sounds good. I'll start looking at this again on monday.

sjacobsson:
Hi, I have a quick question about determining when points are connected by a geodesic. The Segre manifold is not geodesically complete, so before I test the logarithmic map I would like to make sure that the points can actually be connected by a geodesic. I wrote a method compatible() that checks this but I was wondering if there is any standard way to do this in Manifold.jl?
If not, may I make this method visible?

kellertuer:
Hi,
for now there is no standard wyay, since most manifold we have _are_ complete.
My idea would be that log is not defined for two points that are not connected by a geodesic;
that also happens for example on the sphere for opposing points – that is not defined, so it is also not clear what the function would return or whether it would error.

So that is what I would maybe follow here as well.

What would be your signature of the new function, `compatible(M, p, q)`? And where is the name from, for me the name sounds a bit too generic? what about `connected(M, p, q)`?
I would maybe not call that in every log and leave it really undefined – but one could document that one can check this upfront with that function.

sjacobsson:
I agree connected() is better.
connected(M, p, q) would return a bool

sjacobsson:
The reason for using it in the tests I guess is that I would like to check both that log . exp = "identity on T_p M" and exp . log = "identity on M", and for the second check I want to generate connected points.

kellertuer:
Having the check is totally fine, I would just do that “external” so outside of the log, for performance reasons. but sure for testing it would be great to check that, and deterministally choosing points that are close enough.

sjacobsson:
right now it's literally just
```
while !connected(M, p, q)
    p = rand(M);
    q = rand(M);
end
```
I want connected() rather than injectivity_radius() since the injectivity radius condition is satisfied by exponentially fewer points...

kellertuer:
For the tests, especially maintaining them later(!), please choose two deterministic points, because this randomness might fail every now and then differently, which is horrible if it hen only fails every now and then.

sjacobsson:
Ah I see. But then I don't actually need to make connected() visible :).

kellertuer:
It might still be nice to have for others to use as well :)

We can maybe also still come up with a better name (but naming is hard).

sjacobsson:
how about: `connected_by_geodesic()`?

kellertuer:
That sounds good :)

kellertuer:
To get the format check pass, just locally on the folder run `using JuliaFormatter; format(".")`.

I canhelpt a bit with the tests next week, but still have oral exams until Monday.

sjacobsson:
Thanks, I'll run that.
Right now I just commited the tests I had already made, but I still get errors when I run for example `test_manifold(Segre(...`. I don't know how much work it is to fix that, but either way it's not for this week :)

kellertuer:
I will take a look, maybe some tests also have to be deactivated because you do not provide implementations for that.

Did you allow for pushes to this PR? If not I'll try to note a few things here (otherwise I can push them).

sjacobsson:
You should be able to push to this PR.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+Generate a valence `V` Segre manifold.
+"""
+struct Segre{𝔽,V} <: AbstractManifold{𝔽} end
+
+function Segre(valence::NTuple{V,Int}; field::AbstractNumbers=ℝ) where {V}
+    return Segre{field,valence}()
+end
+
+valence(::Segre{𝔽,V}) where {𝔽,V} = V
+ndims(::Segre{𝔽,V}) where {𝔽,V} = length(V)
```
kellertuer, line 30: These two functions no where tested or used. Are they really necessary? Then they should be covered by tests. Since they are also not exported, they would for now only be internal functions anyways.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+    check_size(M::Segre{𝔽, V}, p)
+
+Check whether `p` has the right size for `Segre` manifold `M`.
+"""
+function check_size(M::Segre{𝔽,V}, p;) where {𝔽,V}
+    p_size = only.(size.(p))
+    M_size = [1, V...]
+
+    if p_size != M_size
+        return DomainError(
```
kellertuer, line 43: Here and for the other following domain errors, it would be good to have a (wrong/non-)point to hit these in a test of the form

```
@test_throws DomainError is_point(M, p, true)
```

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+    if p[1][1] <= 0.0
+        return DomainError(p[1][1], "$(p) has non-positive modulus.")
+    end
+
+    for (x, n) in zip(p[2:end], V)
+        e = check_point(Sphere(n - 1)::AbstractSphere, x; rtol=1e-10, atol=1e-10, kwargs...)
+        if !isnothing(e)
+            return e
+        end
+    end
```
kellertuer, line 94: (besides writing a test for this as well...)

These checks currently have a small disadvantage. if the components of `p` that are checked here have several mistakes, the user would always only see the first one.
After fixing that he would see the second,...

In the product manifold, we for example “collect” these and have created a special error message type to report all such errors, cf. https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/76d6a5fb05a1593325dd342ae780dd89f55afbcd/src/ProductManifold.jl#L169-L175 and the `ComponentManifoldError` that contains a tuple of errors. Would that maybe make sense here as well?

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+"""
+    function check_vector(M::Segre{ℝ, V}, p, v, kwargs...)
+
+Check whether `v` is a tangent vector to `p` on `M`, i.e. after `check_point(M, p)`, `v` has to be of same dimension as `p` and orthogonal to `p`. The tolerance can be set using the `kwargs...`.
+"""
+function check_vector(M::Segre{ℝ,V}, p, v, kwargs...) where {V}
+    e = check_point(M, p, kwargs...)
+    if !isnothing(e)
+        return e
+    end
```
kellertuer, line 108:
```suggestion
```
This should not be done here; `is_vector` has an option to enable or disable the check and does perform this. for the `check_vector` function the idea is hence that either the user did not want to check it or in other words we may assume this is correct anyways.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+            return e
+        end
+    end
+
+    return nothing
+end
+
+"""
+    function get_coordinates(M::Segre{𝔽, V}, p, v; kwargs...)
+"""
```
kellertuer, line 130: Could we extend this doc string a bit?

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+            get_coordinates(Sphere(n - 1), x, xdot, DefaultOrthonormalBasis(); kwargs...) for (n, x, xdot) in zip(V, p[2:end], v[2:end])
+        ]...,
+    ]
+
+    return vcat(coords...)
+end
+
+"""
+    function get_vector( M::Segre{𝔽, V}, p, X; kwargs...)
+"""
```
kellertuer, line 147: Could we extend this doc string a bit?

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+end
+
+@doc raw"""
+    function norm(M::Segre{𝔽, V}, p, v)
+
+Norm of tangent vector ``v`` at ``p``.
+"""
+function norm(M::Segre{𝔽,V}, p, v) where {𝔽,V}
+    return sqrt(inner(M, p, v, v))
+end
```
kellertuer, line 192:
```suggestion
```
This is the default anyways, so it should not be necessary, cf https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/76d6a5fb05a1593325dd342ae780dd89f55afbcd/src/ManifoldsBase.jl#L942-L948

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+
+"""
+    function rand(M::Segre{ℝ, V}; vector_at=nothing)
+"""
```
kellertuer, line 196: Please extend also this doc string and provide more details.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+
+@doc raw"""
+    function embed(M::Segre{𝔽, V}, v)
+
+Embed ``p \doteq (\lambda, x_1, \dots, x_d)`` in ``𝔽^{n_1 \times \dots \times n_d}`` using the Krönecker product:
+````math
+    (\lambda, x_1, \dots, x_d) \mapsto \lambda x_1 \otimes \dots \otimes x_d.
+````
+"""
+function embed(M::Segre{𝔽,V}, p) where {𝔽,V}
```
kellertuer, line 219: It would be better to implement the in-place variant instead (`embed!(M, q, p)`) – as well as `get_embedding` to state what the embedding is. Maybe this also fixed that `q` is allocated automatically and there is no need for `embed(M,p)` at all.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+    function m(M::Segre{ℝ, V}, p, q)
+
+Let ``p \doteq (\lambda, x_1, \dots, x_d)``, ``q \doteq (\mu, y_1, \dots, y_d) \in \mathcal{S}``.
+Then
+````math
+    m(p, q) = \sqrt{\sphericalangle(x_1, y_1)^2 + \dots + \sphericalangle(x_d, y_d)^2}
+````
+is the distance between ``(x_1, \dots, x_d)`` and ``(y_1, \dots, y_d)`` on ``S^{n_1 - 1} \times \dots \times S^{n_d - 1}``.
+"""
+function m(M::Segre{ℝ,V}, p, q) where {V}
```
kellertuer, line 250: Should this function maybe be called `distance(M, p, q)`? Or am I misreading the doc string? What is the function used for? We should otherwise maybe find a better name.

kellertuer: If it is only used in the function below, then it should maybe also be only part of that function and be defined and explained in the doc string there.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+    return sqrt(
+        sum([distance(Sphere(n - 1), x, y)^2 for (n, x, y) in zip(V, p[2:end], q[2:end])]),
+    )
+end
+
+"""
+    function connected_by_geodesic(M::Segre{ℝ, V}, p, q)
+
+Check if two points, `p` and `q`, can be connected by a geodesic.
+"""
```
kellertuer, line 260: Please provide more details here in the doc string. Also “can be” is a bit vague. I think it would be better to write “are”

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+function connected_by_geodesic(M::Segre{ℝ,V}, p, q) where {V}
+    q_ = closest_representation(M, p, q)
+
+    return m(M, p, q_) < pi
+end
+
+"""
+    function closest_representation(M::Segre{ℝ, V}, p, q)
+
+Find the representation of `q` that is closest to `p`.
```
kellertuer, line 270: Please explain this, I am not sure what `p` and `q` are – not what the closest representation here refers to.
Is q not on M?

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+For a proof, see theorem 4.4 in [JacobssonSwijsenVandervekenVannieuwenhoven:2024](@cite).
+"""
+function log(M::Segre{ℝ,V}, p, q) where {V}
+    q_ = closest_representation(M, p, q)
+    if connected_by_geodesic(M, p, q_)
+        v = zeros.(size.(p)) # Initialize
+        log!(M, v, p, q_)
+        return v
+    else
+        return Nothing
```
kellertuer, line 392: This seems a bit random and not very type stable.
Also please implement.`log!` (the same way you nicely implemented `exp!`) instead.

We could fill the result with `NaN` if it does not exist or so instead?

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+    \langle (\nu, u_1, \dots, u_d), (\xi, v_1, \dots, v_d) \rangle_{(\lambda, x_1, \dots, x_d)} = \nu \xi + (A \lambda)^2 (\langle u_1, v_1 \rangle + \dots + \langle u_d, v_d \rangle).
+````
+``A`` is called the _warping factor_ and ``A = 1`` corresponds to the usual Segre manifold.
+
+The geometry is summarized in [JacobssonSwijsenVandervekenVannieuwenhoven:2024](@cite).
+"""
+struct WarpedMetric{A} <: AbstractMetric end
+
+valence(::MetricManifold{𝔽,Segre{𝔽,V},WarpedMetric{A}}) where {V,A,𝔽} = V
+ndims(::MetricManifold{𝔽,Segre{𝔽,V},WarpedMetric{A}}) where {V,A,𝔽} = length(V)
```
kellertuer, line 19: See comment in the Segre manifold – are these needed – should they be documented/exported?

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+
+"""
+    function get_coordinates(M::MetricManifold{𝔽, Segre{𝔽, V}, WarpedMetric{A}}, p, v; kwargs...)
+"""
+function get_coordinates(
```
kellertuer, line 24: Things like coordinates should pass “pass through” usually Also – is  this missing the statement of the basis used? Cf. https://juliamanifolds.github.io/ManifoldsBase.jl/stable/bases/#ManifoldsBase.get_coordinates

kellertuer: And same as above, it is better to only implement `get_coordinates!`.

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+    v;
+    kwargs...,
+) where {V,A,𝔽}
+    return get_coordinates(M.manifold, p, v; kwargs...)
+end
+
+"""
+    function get_vector(M::MetricManifold{𝔽, Segre{𝔽, V}, WarpedMetric{A}}, p, X; kwargs...)
+"""
+function get_vector(
```
kellertuer, line 36: Here, the same two comments as for `get_coordinates` hold.

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+end
+
+@doc raw"""
+    function norm(M::MetricManifold{ℝ, Segre{ℝ, V}, WarpedMetric{A}}, p, v)
+
+Norm of tangent vector ``v`` at ``p``.
+"""
+function norm(M::MetricManifold{𝔽,Segre{𝔽,V},WarpedMetric{A}}, p, v) where {V,A,𝔽}
+    return sqrt(inner(M, p, v, v))
+end
```
kellertuer, line 65:
```suggestion
```
Same as above, since this is the default, it should not be necessary to be implemented.

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+@doc raw"""
+    function m(M::MetricManifold{ℝ, Segre{ℝ,V}, WarpedMetric{A}}, p, q)
+
+When ``p``, ``q \in ℝ^{+} \times S^{n_1 - 1} \times \dots \times S^{n_d - 1}``, this is the distance between the ``S^{n_1 - 1} \times \dots \times S^{n_d - 1}`` parts of ``p`` and ``q``.
+"""
+function m(M::MetricManifold{ℝ,Segre{ℝ,V},WarpedMetric{A}}, p, q) where {V,A}
+    return sqrt(
+        sum([distance(Sphere(n - 1), x, y)^2 for (n, x, y) in zip(V, p[2:end], q[2:end])]),
+    )
+end
```
kellertuer, line 76: Same with the naming – is this the same as above? Then this could be “passed on” to.`M.manifold`?

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+
+"""
+    function connected_by_geodesic(M::MetricManifold{ℝ, Segre{ℝ,V}, WarpedMetric{A}}, p, q)
+
+Check if two points, `p` and `q`, can be connected by a geodesic.
+"""
```
kellertuer, line 82: Maybe it is not necessary to document this again?

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+For a proof, see theorem 4.4 in [JacobssonSwijsenVandervekenVannieuwenhoven:2024](@cite).
+"""
+function log(M::MetricManifold{ℝ,Segre{ℝ,V},WarpedMetric{A}}, p, q) where {V,A}
+    q_ = closest_representation(M, p, q)
+    if connected_by_geodesic(M, p, q)
+        v = zeros.(size.(p)) # Initialize
+        log!(M, v, p, q_)
+        return v
+    else
+        return Nothing
```
kellertuer, line 193: Here the same as for Segre counts – returning nothing is maybe too type instable.

src/manifolds/SegreWarped.jl
```diff
@@ -0,0 +1,284 @@
...
+``R_{\mathcal{S}_A}`` is zero in the remaining (orthogonal) directions.
+"""
+function riemann_tensor(
+    M::MetricManifold{ℝ,Segre{ℝ,V},WarpedMetric{A}},
+    p,
+    u,
+    v,
+    w,
+) where {V,A}
+    # Can we avoid the deep-copies here? That looks a bit inefficient
```
kellertuer, line 248: I think they should not be necessary, none of the functions below does modify them.

Can you maybe commute the tenor just on the components 2,... and add zero anyways upfront? The first term does not seem to use the `[1]` component, the last term (l. 261) can probably be adapted accordingly to just use a different inner and add the first summand accordingly?

test/manifolds/segre.jl
```diff
@@ -0,0 +1,4149 @@
...
+            @test(isapprox(v, get_vector(M, p, get_coordinates(M, p, v))))
+        end
+
+        @testset "get_vector()" begin
+            @test(isapprox(X, get_coordinates(M, p, get_vector(M, p, X))))
+        end
+    end
+end
+
+# TODO: Test that distance and inner are compatible
```
kellertuer, line 4149: `sectional_curvature` seems to be the main thing missing (besides embedding and a few points/vectors that fail the checks on purpose).

kellertuer: ... for both the warped and non-warped metric, so doing that once in the for loop should be fine :)

kellertuer (changes requested):
Sorry that this is quite a few comments. We can surely also work together towards fixing them, but for a few I also need a bit of clarifying responses.

Overall this already looks quite good, this is all just a few small things to do I hope.

kellertuer:
I also already merged the current main branch, so this is up to date – I can also do the news update and the about page update in the end for sure :)

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,473 @@
...
+The geometry is summarized in [JacobssonSwijsenVandervekenVannieuwenhoven:2024](@cite).
+
+# Constructor
+    Segre(valence::NTuple{V, Int}; field::AbstractNumbers=ℝ)
+
+Generate a valence `V` Segre manifold.
+"""
+struct Segre{𝔽,V} <: AbstractManifold{𝔽} end
+
+function Segre(valence::NTuple{V,Int}; field::AbstractNumbers=ℝ) where {V}
```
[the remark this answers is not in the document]
sjacobsson: Hi, thanks for going through this!

I agree that it would be clearer to use Segre((10,)) etc in the tests.

I don't think we should define a specific constructor for one integer. I considered not implementing it at all for one integer since Segre((10,)) ~ Euclidean(10).

test/manifolds/segre.jl
```diff
@@ -0,0 +1,4149 @@
+using Manifolds, Test
+
+# Approximate derivative of f at x
+# https://en.wikipedia.org/wiki/Finite_difference_coefficient
+function finite_difference(f::Function, x::Float64, h::Float64; order=1::Int64) #={{{=#
```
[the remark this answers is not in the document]
sjacobsson: Yes FiniteDiff would work. They implement first and second derivatives, which is all we need :)

src/manifolds/Segre.jl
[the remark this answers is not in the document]
kellertuer: Your last sentence to me sounds a lot like `Segre(10)` totally makes sense. one could mention in the doc string what this is equivalent to (I for example was not aware of that).

test/manifolds/segre.jl
```diff
@@ -0,0 +1,4149 @@
...
+    MetricManifold(Segre{ℝ,(2, 2)}(), WarpedMetric{1.1978398577942357}()),
+    MetricManifold(Segre{ℝ,(9, 6, 10)}(), WarpedMetric{1.4545138169484464}()),
+    MetricManifold(Segre{ℝ,(9, 9, 2)}(), WarpedMetric{0.886048645869852}()),
+    MetricManifold(Segre{ℝ,(10, 10, 7)}(), WarpedMetric{1.2810851738110494}()),
+    MetricManifold(Segre{ℝ,(9, 3, 8, 10)}(), WarpedMetric{1.396673190458706}()),
+    MetricManifold(Segre{ℝ,(10, 9, 7, 6)}(), WarpedMetric{1.294280190668267}()),
+    MetricManifold(Segre{ℝ,(7, 2, 8, 8)}(), WarpedMetric{1.2374529454448573}()),
+]
+
+# ps[i] is a point on Ms[i]
```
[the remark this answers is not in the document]
sjacobsson: I agree that it's not very readable. These points are generated from [rand(M) for M in Ms]. I don't know if you would be okay with using rand() and Random.seed!(1) or something like that.

test/manifolds/segre.jl
```diff
@@ -0,0 +1,4149 @@
...
+            (13 / 8) * f(x - 1 * h) +
+            (-13 / 8) * f(x + 1 * h) +
+            (1) * f(x + 2 * h) +
+            (-1 / 8) * f(x + 3 * h)
+        ) / h^3
+    end
+end #=}}}=#
+
+# Manifolds to test
+Ms = [
```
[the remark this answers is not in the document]
sjacobsson: For sure, we could test only one Segre with each "valence". Then we could for example test
```
    Segre{ℝ,(10,)}(),
    Segre{ℝ,(7, 2)}(),
    Segre{ℝ,(7, 9, 9)}(),
    Segre{ℝ,(9, 3, 6, 6)}(),
    MetricManifold(Segre{ℝ,(10,)}(), WarpedMetric{1.2025837056880606}()),
    MetricManifold(Segre{ℝ,(2, 9)}(), WarpedMetric{1.1302422072971439}()),
    MetricManifold(Segre{ℝ,(9, 6, 10)}(), WarpedMetric{1.4545138169484464}()),
    MetricManifold(Segre{ℝ,(9, 3, 8, 10)}(), WarpedMetric{1.396673190458706}()),
```

src/manifolds/Segre.jl
sjacobsson: I'll remove them!

src/manifolds/Segre.jl
sjacobsson: `m()` is an unhelpfully named helper function that is used in `distance()`, `log()`, `connected_by_geodesic()`, and `closest_representation()`. I can come up with a better name for it maybe, but it's really just a shorthand for a specific expression that pops up alot in the formulas.

sjacobsson:
Thanks for going through the PR. I'll make these updates!

sjacobsson:
The only question I have right now I think is about using `rand()` with `Random.seed!()`

kellertuer:
I would prefer if directly before the use of `rand` you would set the seed – for reproducibility.
And use `rand` only (as you do for now I think) to test that correct points/tangent vectors are generated.

All other tests should be deterministic with concrete points (which I think is the case). Since otherwise they might “randomly fail” if there is a bug and that is usually super hard to narrow down / maintain in code.

sjacobsson:
I should mention about `closest_representation()` as well. Topologically, the Segre is 2^d-sheeted cover of R x S^n1 x ... x S^nd, similarly to how real projective space is a double cover of the sphere. But then when we compute the geodesics between two points [p] and [q] on the Segre, we need to use the representatives p and q that are closest on  R x S^n1 x ... x S^nd.

src/manifolds/Segre.jl
sjacobsson: I think will call it `spherical_angle_sum()` and explain in its docstring that it's a helper function.

kellertuer: You do not have to explicitly state that it is a helper function.
The name sounds good, add the formula as well, then you could `@ref` to the function in all 4 mentioned functions above :)
And sure if it is used 4 times it is worth its own functions.

src/manifolds/Segre.jl
kellertuer: Neat idea indeed! You have to switch it, first M then p, but the rest looks good :)

sjacobsson:
Hi again,
Okay I think I managed to fix most things you mentioned!

I didn't manage to get `ComponentManifoldError` to work. Something about the type of points on the product manifold was complicated... I tried for a while but in the end I just reverted to how it was.

I implemented `get_coordinates_orthonormal!` and `get_vector_orthonormal!`. I don't know exactly how all the passthroughs work but it seems to pass the tests. However, when I implemented `get_vector_orthonormal!`, it only worked if I also returned the vector I was modifying, is this intentional?

I tried to explain a bit better in the docs how the Segre manifold is a Riemannian covering, but these things are always more difficult to explain the more you've worked with the manifold...

Sorry also for not using your `test_manifold` method, I hope it didn't create too much extra work in the end.

kellertuer:
Hi, I can take a look at the changes the next few days.

A few answers:

* yes a function modifying something should usually return the thing it modifies. Both `get_coordinates_orthonormal!` and `get_vector_orthonormal!` should hence return the coordinates and vector they modify, respectively. The reason is, that this allows to “chain” such modifications. The rest, i.e. to exactly implement these, sounds very good.
* yeah, I also noticed that my componentError idea is a bit more advanced than I thought, we should probably first do that a bit more general than just for a product manifold
* no worries about not using the `test_manifold` function. The first main thing is that you have enough test coverage. We wrote that more for us, since it made tests a bit easier. It is also not optimal and will be reworked (for a modular idea see the current tests in `LieGroups.jl`. Sometimes when reworking, I also include other, new tests in the new test suite ideas – sometimes I am too lazy for that. Again, the main point is to have the test coverage and maybe not over-test (for time on CI reasons).

So one thing you could check – once code-cov is done, where you might miss lines, but I think I commented on most missing ones.

sjacobsson:
Hi,
Thanks. I'll see what I can do to make codecov happy.

src/manifolds/Segre.jl
sjacobsson: Hi, I noticed now that even if `log!` fills `v` with nans, it still gets interpreted as floats in `log`. So it outputs for example something like `[[4.0e-323], [6.9057512179161e-310, 6.90574352987124e-310], [6.90574432576356e-310, 6.90570212344215e-310], [-0.0245041404219386, 0.002821512911371685], [-0.01594663570864798, -0.01842389578578475], [-0.06178003383876796, -0.0713773692384812]]`. So I don't know what the best way to deal with this is. Maybe to put the check in `log` instead?

kellertuer: Well, `NaN`s are floats. So of course that is the case. That is intentional – the `Nothing` this comment is based on would be type-unstable. And I would not overwrite `log`

What would you wish for to return? 

As an example – `log`/`log!` on the sphere does not to any checks whether from `M, p, q` the `q` is on the sphere (unit norm). If it is not, the result is simply not defined and you might get out _anything_. The same for opposing points.
I would do the same here. If it is not well-defined, just don't care what you return.
Actually also on the sphere, your new function `connected_by_geodesic` might also be helpful – maybe with a keyword argument to ask whether the shortest geodesic is unique (because only then then log would be well-defined for sure)

sjacobsson: Ah, I didn't realize nans are floats as well. Then it must be something else then... I replaced the `return Nothing` above with `return v = NaN * p`, and I was confused by this behaviour:
```
julia> M = Segre(2, 2, 2, 2, 2);

julia> p = [[1.0], [1.0, 0.0], [1.0, 0.0], [1.0, 0.0], [1.0, 0.0], [1.0, 0.0]];

julia> q = [[1.0], [0.0, 0.1], [0.0, 1.0], [0.0, 1.0], [0.0, 1.0], [0.0, 1.0]];

julia> v = deepcopy(p);

julia> log!(M, v, p, q)
6-element Vector{Vector{Float64}}:
 [NaN]
 [NaN, NaN]
 [NaN, NaN]
 [NaN, NaN]
 [NaN, NaN]
 [NaN, NaN]

julia> log(M, p, q)
6-element Vector{Vector{Float64}}:
 [5.0e-324]
 [6.365987374e-314, 1.0e-323]
 [6.93009044731883e-310, 6.93009779249477e-310]
 [6.9300904323103e-310, 6.9300947100619e-310]
 [6.93008946239225e-310, 6.9300977939335e-310]
 [6.93009783360024e-310, 6.9300978225541e-310]

julia> 
```
In my experience, the e-310's usually indicate that I tried to interpret something that isnt a float as a float, which is why I assumed that's what had happened. But if you don't usually do this check on the input, maybe I'll just remove it :).

kellertuer: Maybe two points

1. is it really necessary to fill it with nans? Does the other code (if the geodesic exists) error if called with wrong input? Or does it just fill the result with nonsense? If the second is the case, we can also leave it at that nonsense (instead of nans)
2. The current form allocates quite a bit, I think `v .= NaN .* p` should be better, maybe there is even a better `fill!.(v, NaN)` possible (both ideas, not tested).

the `e-310` just mean that you allocated memory for `v` but never filled it with any values also not with `NaN`s. The thing I like less here, is that `log` and `log!` return different results, that should never be the case.

sjacobsson: yea just returning nonsense is easier, let's do that.

I have no idea why `log` and `log!` were returning different results, it's possible it was because `log!` was not also returning `v`, which I fixed now, but I don't know

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,465 @@
+@doc raw"""
+    Segre{𝔽,V} <: AbstractManifold{𝔽}
+
+The Segre manifold
```
kellertuer, line 4: I have a final question on the name:
I am fine with the the `Segre` – but there is one place, the menu entry, where you write “Segre-Veronese” (as in the paper title) and otherwise you always write Segre in the text. Should we maybe use a unified form in text?
The Type can stay as is for sure.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,465 @@
...
+    return Segre{field,(n...,)}()
+end
+
+manifold_dimension(::Segre{𝔽,V}) where {𝔽,V} = (1 + sum(V .- 1))
+
+"""
+    check_size(M::Segre{𝔽, V}, p)
+
+Check whether `p` has the right size for `Segre` manifold `M`.
+"""
```
kellertuer, line 35: Thi doc string (and the next four doc strings) does not provide any information.

We could either document properly what it means for `p` (or `X` for the next) to be of right size – or leave these two doc strings out. Overall this could even better be documented with `is_point` or `is_vector` doc strings, since these are the “API” functions a user would use. The `check_`s are merely the internal ones one should implement).

Yes I am writing a lot of comments – if you feel this gets too much work, just mention it, this one for example I know how I would do that as well.

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,465 @@
...
+is the set of rank-one tensors in ``𝔽^{n_1} \otimes \dots \otimes 𝔽^{n_d}``.
+
+When ``𝔽 = ℝ``, the Segre manifold is a normal Riemannian covering of
+````math
+    \mathcal{P} = ℝ^{+} \times S^{n_1 - 1} \times \dots \times S^{n_d - 1}
+````
+with the [warped product metric](https://en.wikipedia.org/wiki/Warped_product) [`inner`](@ref inner(::Segre, ::Any)). The tuple ``(n_1, \dots, n_d)`` is called the _valence_ of the manifold.
+
+The geometry is summarized in [JacobssonSwijsenVandervekenVannieuwenhoven:2024](@cite).
+
```
kellertuer, line 17: Side track: I usually love nerdy details. Could we finish this doc strip with a sentence why this is called the Segre(-Veronese) manifold? Maybe with a reference to a paper and a wikipedia link to the one/two mathematicians(?) that came up with this?

src/manifolds/Segre.jl
```diff
@@ -0,0 +1,465 @@
...
+````math
+    (\lambda, x_1, \dots, x_d) \mapsto \lambda x_1 \otimes \dots \otimes x_d.
+````
+"""
+function embed!(M::Segre{𝔽,V}, q, p) where {𝔽,V}
+    return q = kron(p...)
+end
+
+# ManifoldsBase doesn't export embed_vector! right now
+# @doc raw"""
```
kellertuer, line 221: Why do you think `ManifoldsBase` should export that? It does not even exist. The function to embed tangent vectors is `embed(M, p, X)` (mutating: `embed!(M, Y, p, X)`. So we can indeed introduce this in exactly that signature.

src/manifolds/Segre.jl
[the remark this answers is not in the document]
kellertuer: Ah the new constructor now even does not take a Typle any longer but really just `(n...)` that is an even better solution!

test/manifolds/segre.jl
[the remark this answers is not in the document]
kellertuer: If necessary yes, the disadvantage would still be that the tests change when Julia changes its default random number generator (ok, they only ever did that once). For now this is still several hundred lines of magic numbers, then maybe really better a `rand(M)` for each of the manifolds, maybe even in the for-loop?

test/manifolds/segre.jl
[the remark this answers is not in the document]
kellertuer: That already looks better, I would still change them to `Segre(10)`, `Segre(7,2)`, and so on, since that is much more readable.
Since these are all real – is it planned (or covered even in theory) to do complex Segre ones?

kellertuer (commented):
Thanks for already thoroughly working through so many comments – here is four more.

I also noticed the Domain errors are not yet covered, should I give that a try? The idea is to call `is_point` (or `is_vector`) with wrong points/vectors that exactly trigger each of the Domain errors once.

_edit:_ I also resolved all points in the above list that are fixed, I think there are mainly the domain errors left, which one can also see in codecov.

src/manifolds/SegreWarpedMetric.jl
```diff
@@ -0,0 +1,301 @@
+@doc raw"""
+    WarpedMetric{A} <: AbstractMetric
```
kellertuer, line 2: Can we maybe
* provide a bit of information, what `A` is? Always just a number?
* provide a nicer constructor (and a corresponding show method) `WarpedMetric(a::Real)` or so?

test/manifolds/segre.jl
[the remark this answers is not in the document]
kellertuer: I had an idea and at least reduced the amount of lines to generate `ps`, `qs`, `us`, `vs`.  It's still a line per point, but at least not at least dimension+7 many. Same for the coordinates – where I sometimes shortened the nonzero entries a bit to also squeeze the high dimensional ones in one line

I have not yet understood after what scheme and how many `dy`s are necessary, could you maybe adapt these as well to such a scheme that is not fully random but takes merely nice-to-produce numbers?

src/manifolds/Segre.jl
sjacobsson: It should be Segre everywhere. The Segre-Veronese is when you are symmetric in some indices. It's not that hard to describe geometrically in terms of a larger Segre manifold, it's just a totally geodesic submanifold, but I didn't implement any of this.

It's on my todo list to create the Segre-Veronese as well, but I have to think through how I want to do the `connected_by_geodesic` since there is no nice closed-form expression like for the Segre.

sjacobsson: It should be Segre everywhere. The Segre-Veronese is when you are symmetric in some indices. It's not that hard to describe geometrically in terms of a larger Segre manifold, it's just a totally geodesic submanifold, but I didn't implement any of this.

It's on my todo list to create the Segre-Veronese as well, but I have to think through how I want to do the `connected_by_geodesic` since there is no nice closed-form expression like for the Segre.

kellertuer: Nice, I can change that. And now that you have gone through a manifold once (the first is the one that takes longest), the next one will be easier. I will adapt the places to Sagre then :)

docs/src/misc/about.md
```diff
@@ -17,6 +17,7 @@
 - [Nick Dewaele](https://github.com/Nikdwal) contributed the [Tucker manifold](../manifolds/tucker.md)
 - [Renée Dornig](https://github.com/r-dornig) contributed the [centered  matrices](../manifolds/centeredmatrices.md) and the [essential manifold](../manifolds/essentialmanifold.md)
 - [David Hong](https://github.com/dahong67) contributed uniform distributions on the Stiefel and Grassmann manifolds.
+- [Simon Jacobsson](https://github.com/sjacobsson) contributed the [Segre manifold](../manifolds/segre.md) including its [warped metric](@ref segre-warped-metric-sec) thereon.
```
kellertuer, line 20: We started (actually after you started this PR) to list all contributions beyond fixing typos and that are not done by the main developers on this page.

I added you with your GitHub page as link. Feel free to change this to any link you would like to link this to.
I can also add you to the .zenodo file if you like – we do not write “and contributors” there but list everyone who is listed here as well in the metadata of the zenodo DOI entry

src/manifolds/Segre.jl
sjacobsson: I'll just leave that docstring out then I think. But I think the docstring
```
    check_point(M::Segre{ℝ, V}, p; kwargs...)

Check whether `p` is a valid point on `M`, i.e. `p[1]` is a singleton containing a positive number and `p[i + 1]` is a point on `Sphere(V[i])`. The tolerance can be set using the `kwargs...`.
```
actually describes `is_point` rather than `check_point`. I'll put the docstring there as well.

I don't feel it gets too much. I use this library in my research and I have more manifolds I want to contribute so I don't mind :).

src/manifolds/Segre.jl
kellertuer: I fixed this and did as described here.

src/manifolds/Segre.jl
kellertuer: Perfect! Exactly what I had in mind. One could also in general add a similar doc string to `is_vector`, that would be great


And: Great, that you do not mind and super cool that you use the library. I am very happy both about that and that you contribute. I think this PR is indeed nearly done.

kellertuer:
Oh I noticed that also the docs are not yet fully correct. I was too used to docs failing when doc strings are not included. But here we have switched that to just warning (instead of errors). Will work through that and fix the docs somewhen the next days.
You can give it a try as well – with the current version on master (finally!) running `docs/make.jl` should “just work” also if you do not have (or intend to run) quarto.

src/manifolds/SegreWarpedMetric.jl
sjacobsson: Here is a suggestion for information on `A`:
```
``A = 1`` corresponds to the usual Segre manifold.
The Segre manifold is a cone in the sense that if ``p \in \mathcal{S}``, then ``a p \in \mathcal{S}`` for all ``r \neq 0``. The tangent subspace at ``p`` defined ``\mathrm{d} (r p) / \mathrm{d} r`` is called the _radial_ direction. ``A < 1`` puts less weight on the directions orthogonal to the radial direction compared to ``\mathcal{S}``, while ``A > 1`` puts more weight on those directions.
```

sjacobsson: >     * provide a nicer constructor (and a corresponding show method) `WarpedMetric(a::Real)` or so?

For sure!

docs/src/misc/about.md
sjacobsson: Thanks!

Sure, you can add me to the zenodo file if you want. I have an orcid that I try to keep up to date
https://orcid.org/0000-0002-1181-972X

src/manifolds/SegreWarpedMetric.jl
kellertuer: The doc string should mention the constructor, the rest looks very good.

sjacobsson:
Hi, I have a couple of thoughts:
- The `dys` could maybe be code golfed a bit. It was generated as `[rand(4 * manifold_dimension(M)) for M in Ms]` but really it just needs to be a list of vectors of the correct length that are generic enough, so not `ones(4 * manifold_dimension(M))` but anything that is more "random" than that would be fine.
- Equality of points on the Segre isn't `p == q` but rather `embed(M, p) == embed(M, q)`, but embedding is super expensive, so that's why I wrote the tests to use points that are close enough that `p == q` also holds. It's maybe not the nicest solution but that's how it's implemented right now.
- I didn't manage to compile the docs, but it's a me problem. I get some error message like "LoadError: Unsatisfiable requirements detected for package ManifoldDiff"... I'll try to fix this so that I can compile the docs locally.

kellertuer:
I carefully read the rendered docs and did a bit of formatting – and e.g. ordered the function definitions alphabetically in every file and such unifying things

Final remarks:

1. now both the sphere and the Segre manifold are denoted \mathcal S in their docs – the sphere in Segre is just S.
2. usually tangent vectors are denoted X and Y in variables and docs, here it is u and v
3. simiarily coordinates are denoted c (sometimes d) 

Would it maybe be ok to unify / adapt this a bit? That would be nice.

4. if we could do a 10-line variant of `dys` as well that would be great. Could I do the same with them as with the `us`? There I shortened all numbers to one digit after the . and if it were too many, the remaining ones are zeros. Would that be valid for the `dy`s as well?

sjacobsson:
1. The Sphere is denoted \mathbb{S}? Sure, this could be updated in the Segre docs. If you still want to free up \mathcal{S}, we can use \mathrm{Seg} for the Segre.

2. and 3. sounds good, I can change it tomorrow :)

4. maybe we can do something like `[0.1, 0.3, -0.1, zeros(5)..., -0.5, 0.7, zeros(3)...]`? The problem with only right-padding with zeros is that that corresponds to some of the control points never moving. But the actual vector `dy` can in principle be anything (I don't know why it only has positive entries atm...), the digits of pi, pi^2, ... or something like that but then at that point maybe we're just reimplementing the wheel.

5. I'll go through the warnings tomorrow :)

kellertuer:
1. you are right, the sphere is $\mathbb S^d$, where $d$ is the dimension of the sphere. If you could unify that, that would be great! No, we do not have to free up $\mathcal S$ then, I misread my own latex code, so we can keep that for Segre – or any other (not yet used) Symbol you like.
2. and 3. – thanks :)
4. sure, anything that we can fit in a line is fine with me. I just want to avoid several hundred lines of numbers.
5. nice!

sjacobsson:
I managed to compile the docs, I was just trying to do it with the wrong julia version earlier (:

Now I just have to complete the code coverage and then I think I'm happy.

sjacobsson:
Sorry, did we lose the codecov check?

kellertuer:
No, it is only triggered when I push, not when you do. The simple reason is, that external contributors might otherwise “inject” code. I will do a small commit soon :)

kellertuer:
Interesting, now your last commit has coverage, mine does not ;) Maybe it is bit delayed.

sjacobsson:
I'll make another commit. For good measure :).

kellertuer (approved):
Nice! Even with full code coverage. This is done. I can merge it and register a new version tomorrow.

sjacobsson:
Okay perfect! Thanks for helping me getting it merge-ready. The next manifold will be a breeze :)

sjacobsson:
And happy holidays!

kellertuer:
Same to you; god jul / frohe Weihnachten! 🎄

sjacobsson:
I had not seen your paper with Eggleston before, thanks for linking it!

Sounds good. I'll let you know if there are any updates.

kellertuer:
The only thing that could brought you here is, that the name might appear in the literature we cite ;)

I am also not aware of anyone else planning to implement this manifold, but I can help on technical questions regarding Manifolds (or the ManifoldsBase interface) for sure :)

PR #756, opened 2024-10-14
More general columnwise multiplication action

olivierverdier:
This makes `ColumnwiseMultiplicationAction` more general, as it works now with any subtype of `AbstractMatrixGroup`, a new abstract type above `GeneralLinear` and `GeneralUnitary`.

Feel free to change the new name if you don't like it!

Maybe other groups could be subgroups of this new abstract type?

I also think more actions could be defined with respect to that abstract type instead of `GeneralUnitary`.

What do you think?

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install JuliaFormatter and format
         run: |
           using Pkg
-          Pkg.add(PackageSpec(name="JuliaFormatter"))
+          Pkg.add(PackageSpec(name="JuliaFormatter", version="1"))
           using JuliaFormatter
           format("."; verbose=true)
         shell: julia --color=yes {0}
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,18 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.10.4] - 2024-10-20
 
 ### Added
 
 - Union type `MatrixGroup`
 - Columnwise group action with arbitrary matrix groups
-
-## [0.10.4] - 2024-10-18
-
-### Added
-
-* `uniform_distribution` now has an error hint explaining what has to be done to make it work.
+- `uniform_distribution` now has an error hint explaining what has to be done to make it work.
 
 ## [0.10.3] - 2024-10-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,8 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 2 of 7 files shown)


olivierverdier:
Yes, that's a good point, I agree with that.

Maybe something else is required for that: a *diagonal action*, that is, when you have an action of $`G`$ on a space $`V`$, you should be able to construct the *diagonal action* of $`G`$ on $`V^n`$ for some integer $`n`$.

Then the columnwise action on $`d\times n`$ matrices is essentially a diagonal action.

That would solve the issue in the most general way, I believe?

So a rough sketch of how it would look like is:
```julia
struct ColumnwiseMultiplicationAction{
    TA<:AbstractGroupAction{TAD}
} <: AbstractGroupAction{TAD}
    action::TA
end

function apply(A::ColumnwiseMultiplicationAction, a, p)
    return apply(DiagonalAction(A.action, size(p,2)), a, p) # but DiagonalAction doesn't exist yet (and not sure if `p` or `eachcol(p)`)
end
```

kellertuer:
Just a first super small remark. In `LieGroups.jl` there is only one (concrete) type `LieGroup{Field, Manifold, GroupOperation}` would a MatrixLieGroup there make sense as well? Then we need `AbstractLieGroup` somewhen there (maybe anyways) – or is it more useful to do an `AbstractMatrixMultiplicationGroupOperation` since we have a `MultiplicationGroupOperation` already anyways?

mateuszbaran:
> Maybe something else is required for that: a _diagonal action_, that is, when you have an action of 

Good idea, I will think about it.

> Just a first super small remark. In `LieGroups.jl` there is only one (concrete) type `LieGroup{Field, Manifold, GroupOperation}` would a MatrixLieGroup there make sense as well? Then we need `AbstractLieGroup` somewhen there (maybe anyways) – or is it more useful to do an `AbstractMatrixMultiplicationGroupOperation` since we have a `MultiplicationGroupOperation` already anyways?

At the moment it's not entirely clear what that abstract type would exactly mean. I think we can implement the key idea here without that type anyway, so maybe there is no need to think about implementing it in LieGroups.jl.

olivierverdier:
Yes, I suppose it could make sense for `LieGroups.jl` as well?

But I'm not 100% percent sure that this is the correct way to handle the issue in this case, since, as @mateuszbaran observed, multicolumn action should also be available for direct and semidirect product, and the current solution doesn't address these cases. I honestly don't know what the best solution would be.

kellertuer:
Sure, taking this as a bit of a playground and experiment sounds good. I feel it might be nicer to introduce an `AbstractMultiplicationOp` and a `MatrixOp` or so on the operations side and subtype there. But I have not yet fully thought through that and all ist implications either.

mateuszbaran:
I've switched `AbstractMatrixGroup` to a union type `MatrixGroup` for now. This should now be a simple extension that won't need too much consideration and we can think about something more general for LieGroups.jl later. Is that OK?

olivierverdier:
Yes, I think that will do, thank you.

mateuszbaran (approved):

PR #757, opened 2024-10-18
Maybe don't split distributions extension

mateuszbaran:


mateuszbaran:
JuliaFormatter v2 has some serious issues reported so maybe let's keep using v1 for now.

kellertuer:
v2 was even yanked I think, so while I look forward to it, since it does use JuliaSyntax – yes we should wait with that for sure.

kellertuer (approved):

mateuszbaran:
I have no idea why the Manopt integration test suddenly fails... should we maybe try a different Julia version on CI?

kellertuer:
I also have no clue why it claims Manifolds is not in the dependencies there. That seems to be unrelated to the PR here?

mateuszbaran:
It is unrelated, so I will merge this anyway.

kellertuer:
I do not even et all what the error message wants to tell me. And it worked before. So I have no clue what is happening there nor howI should debug a dependency error from within an Extension that is based on the package it reports to be missing?!

mateuszbaran:
Hopefully it will just resolve itself.

PR #759, opened 2024-10-19
Introduce `has_components` for Euclidean

kellertuer:
Since we now have available that

```julia
distance(Euclidean(3), p, q, 1)
```

for the -norm, this introduces it also for Euclidean. I do not directly see other manifolds that would be available for that.


kellertuer:
Currently `Has_components` does not “pass transparently” through `MetricManifold`. Should we add that to ManifoldsBase? 

To be precise, we currently have that

```
M = Manifolds.MetricManifold(E, Manifolds.EuclideanMetric())
```

yields `has_components(M)` to be `false`. I think if we fix that we have to use traits – and this would have to be done in `ManifoldsBase.jl` I think.

mateuszbaran:
Nice. Could you also add tests that `distance` and `norm` work with `r`?

> Currently `Has_components` does not “pass transparently” through `MetricManifold`. Should we add that to ManifoldsBase?

Right, yes, `has_components` should be a trait function with transparent passing.

kellertuer:
sure I can add those lines.

For the transparency, we should probably do that in `ManifoldsBase`?

mateuszbaran:
Yes, transparency should be in ManifoldsBase.jl.

kellertuer:
Sure, will do.

kellertuer:
Added the test for distance and norm. Also started the `ManifoldsTestSuite`. Similar to Lie groups I will try to now lower-level add functions therein and slowly transfer to using the test suite then. Yes that means I tweak that into other PRs a bit every now and then, but that is easier for me, I think, than the current open PR for that.

mateuszbaran:
Sure, that's fine.

mateuszbaran:
BTW, I haven't tagged 0.10.4 yet so this could be a part of it together with Oliver's PR.

kellertuer:
Just saw that 0.10.4 is not yet registered, then this can also be part of that. But I will do the transparency first in ManifoldsBase maybe.

mateuszbaran (approved):

kellertuer:
Already merged master here, but since we just merged a branch in 3 packages, all CIs might take a time ;) afterward I will merge this as well. We could register those two as 0.10.4 for sure.

PR #760, opened 2024-10-20
Rewrite contributing.md

### LINKED ISSUE #750: (Update / Improve) A Developer guide
in @Affie said in https://github.com/JuliaManifolds/LieGroups.jl/issues/5

> Perhaps it will help to develop a developer guide over time with steps that has to be checked off with details such as: run the formatter, put tests here, add docstrings like this, add these traits.

I will check what needs to be done, but to first document the state “as is”:
 We have 

* a page about the design ideas https://juliamanifolds.github.io/ManifoldsBase.jl/stable/design/ especially explaining the usual order of variables, 3 levels we work on (to avoid ambiguities) and the mutating/allocating ideas.
* a Paige about the decorator idea and the current trait system https://juliamanifolds.github.io/ManifoldsBase.jl/stable/decorator/ – that I was a large fan of to get started with, but it turns out to be not so user friendly, neither to the developer nor to someone using it because the error messages are sometimes bit weird. I would love to have a better system here, but maybe a very technical, complicated wish (due to performance)
* We have the https://juliamanifolds.github.io/Manifolds.jl/stable/misc/CONTRIBUTING.html (a rendered version of https://github.com/JuliaManifolds/Manifolds.jl/blob/master/CONTRIBUTING.md mentioning for example code style.

So at first glance I can spot two things we missed to update, probably in the last file

* any change should be documented in the https://github.com/JuliaManifolds/Manifolds.jl/blob/master/NEWS.md – and since we ourselves forgot about that often (and with we I also specifically mean me) we have a test to fail when this file is not changed on a PR
* doc strings, and what they should contain

Let me know what else we a re missing and I will do a PR working on / improving these here; I will check afterwards wherelse this should then be adapted (ManifoldsBase/Manopt/LieGroups/...)

kellertuer:
Resolves #750.

@Affie can you take a look whether this is clearer now? I especially added an overview of files to consider when contributing.
Another Idea I can do, is to give a super short summary of what to do and which files to consider in the beginning of https://juliamanifolds.github.io/ManifoldsBase.jl/stable/tutorials/implement-a-manifold/, but that is something for another PR in that repository.

This also initialises vale.sh as a linter for the text to improve style and easier find bugs, since vale also highlights hints and errors in VS code when using their plugin. Vale is not perfect but maybe a good guide.


Affie:
Hi, yes the overview with the links will help a lot. I'm guessing the group "how to" will go into "LieGroups.jl". Perhaps when that is done, I should try to contribute the `SpecialGalileanGroup` we have here (if you want it): https://github.com/JuliaRobotics/RoME.jl/blob/master/src/factors/Inertial/IMUDeltaFactor.jl
I can follow the guide and let you know where I get stuck.

kellertuer:
Nice!

Yes I plan to write a “How to Lie Group” once we have at least group operations fixed in `LieGroups.jl`, which should not take so long I hope – the first Lie group is nearly done. I want to do a second to be safe, and use that to write a tutorial.

But sure, new Lie groups are then super welcome over there!

PR #762, opened 2024-10-22
Group retraction from invariant exp/log

### LINKED ISSUE #682: Wrong parameter for `GroupExponentialRetraction`
Currently, `GroupExponentialRetraction` depends on a type `ActionDirectionAndSide`, but, unless I am mistaken, it should only depend on `GroupActionSide`. 

Put differently, the `GroupExponentialRetraction{LeftBackwardAction}` and `GroupExponentialRetraction{RightForwardAction}` *make no sense*.

Indeed: in the code to `retract`, there is a call to `inverse_translate_diff(G, p, p, X, conv)`, which is supposed to send the tangent vector `X` to the Lie algebra. But this only holds if `conv` is either `LeftForwardAction` or `RightBackwardAction`.

---
**Suggested solution**:
Implement the generally useful `translate_to_id` which translates a tangent vector to the identity (to the Lie algebra) like so
```julia
_conv_from_side(::LeftSide) = LeftForwardAction()
_conv_from_side(::RightSide) = RightBackwardAction()

translate_to_id(G, p, X, side::GroupActionSide) = translate_diff(G, p, p, X, switch_side(_inv_conv_from_side(side)))
```
then change the implementation of `GroupExponentialRetraction` as
```julia
struct GroupExponentialRetraction{D<:GroupActionSide} <: AbstractRetractionMethod end

function GroupExponentialRetraction(side::GroupActionSide=LeftSide())
    return GroupExponentialRetraction{typeof(side)}()
end

group_action_side(::GroupExponentialRetraction{D}) where {D} = D()
```
then essentially keep the current implementation of `retract` as
```julia
function retract(
    ::TraitList{<:IsGroupManifold},
    G::AbstractDecoratorManifold,
    p,
    X,
    method::GroupExponentialRetraction,
)
    side = group_side(method)
    Xₑ = translate_to_id(G, p, X, side)
    pinvq = exp_lie(G, Xₑ)
    q = translate(G, p, pinvq, _conv_from_side(side))
    return q
end
```
---
**Note 1** there is probably exactly the same problem (solved in the same manner) for `GroupLogarithmicInverseRetraction`.

**Note 2** `inverse_translate_diff` may be superfluous in general, or at least it should be implemented as `inverse_translate_diff(G, p, q, X, conv) = inverse_translate_diff(G, p, q, X, switch_direction(conv))`.

Let me know if the problem and/or the solution make any sense to you.

olivierverdier:
This PR will
- Reduce code duplication
- Fix #682

It deprecates using `GroupExponentialRetraction` with a parameter. This parameter is now unused and should be scheduled for removal it in a breaking release.

The functionality of `GroupExponentialRetraction` is unchanged.

- [x] Replace `GroupExponentialRetraction` with invariant exp/log
- [x] Fix `exp_inv`/`log_inv`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * rewrote the `CONTRIBUTING.md` and adapt it to todays links and references.
+* Deprecate direction and side parameter in group exponential/logarithm retractions
 
 ## [0.10.4] - 2024-10-20
 
--- a/src/groups/GroupManifold.jl
+++ b/src/groups/GroupManifold.jl
@@ -81,10 +81,7 @@ function inverse_retract(
     q,
     method::GroupLogarithmicInverseRetraction,
 )
-    conv = direction_and_side(method)
-    pinvq = inverse_translate(G, p, q, conv)
-    Xₑ = log_lie(G, pinvq)
-    return translate_diff(G, p, Identity(G), Xₑ, conv)
+    return log_inv(G, p, q)
 end
 
 function inverse_retract!(
@@ -95,10 +92,7 @@ function inverse_retract!(
     q,
     method::GroupLogarithmicInverseRetraction,
 )
-    conv = direction_and_side(method)
-    pinvq = inverse_translate(G, p, q, conv)
-    Xₑ = log_lie(G, pinvq)
-    return translate_diff!(G, X, p, Identity(G), Xₑ, conv)
+    return log_inv!(G, X, p, q)
 end
 
 function is_point(
--- a/src/groups/group.jl
+++ b/src/groups/group.jl
@@ -1244,9 +1244,11 @@ For more details, see
 """
 struct GroupExponentialRetraction{D<:ActionDirectionAndSide} <: AbstractRetractionMethod end
 
-function GroupExponentialRetraction(conv::ActionDirectionAndSide=LeftForwardAction())
-    return GroupExponentialRetraction{typeof(conv)}()
+function GroupExponentialRetraction()
+    return GroupExponentialRetraction{LeftForwardAction}()
 end
+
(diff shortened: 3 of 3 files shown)


olivierverdier:
Interesting, I must have made a mistake: although all the tests pass, something is wrong with this PR.

olivierverdier:
I see, the problem is that there is a mistake in the implementations of `exp_inv` and/or `log_inv`. Those have to be fixed first.

olivierverdier:
The tests fails because `SpecialEuclidean` with `HybridTangentRepresentation` is broken: it doesn't implement `translate_diff`. I can't fix it on my computer, it takes too much time, even with just one test file. 🐌

Honestly, this `HybridTangentRepresentation` should be removed, what is it good for? I will delay this PR until that is fixed.

kellertuer:
Just out of curiosity: What do you actually rund and why is it slow?
Sure, we do not have a nice test suite here – tried starting it once, but only got a nice scheme now that I started on in Lie groups.

for the hybrid representation – I think we will not fix that here, but maybe introduce different representations in Lie groups then. For now LieGroups.jl only supports representation of tangents in the Lie algebra as we discussed yesterday; extending that is only a next step when a Lie group comes along that requires that. On the other hand “converting to” Manifold tangent vectors might be a nice additional function to consider/introduce there as well.

kellertuer:
> I'm only running the test in test/group/special_euclidean.jl and it takes forever, I don't know why.

Ah, yeah maybe we overcast there a bit too much.

> One day the HybridTangentRepresentation will be removed: it is incomplete 

One day (tm) the GroupManifold will be removed here anyways, since that will be superseded by LieGroups.jl, what we introduce there in representations is then a discussion for that package.
But I already have on my plans to add deprecations here, that will be released as soon as the corresponding functionality is released on LieGroups.jl.

kellertuer:
the hybrid representation does no longer exist over on `LieGroups.jl` – we can maybe reintroduce it if necessary, for now there all things are left-invariant-Lie-Algebra-Tangents – so for now I will close this here, since we already deprecate these functions, cf. #776.

PR #764, opened 2024-10-23
introduce invertible matrices (merely for completeness)

kellertuer:
This is merely for completeness and it makes a few steps in LieGroups.jl a bit easier/generic.

### DIFF
--- a/src/manifolds/InvertibleMatrices.jl
+++ b/src/manifolds/InvertibleMatrices.jl
@@ -1,0 +1,116 @@
+@doc raw"""
+    InvertibleMatrices{𝔽,T} <: AbstractDecoratorManifold{𝔽}
+
+The [`AbstractManifold`](@extref `ManifoldsBase.AbstractManifold`)  ``\operatorname{Sym}(n)``
+consisting of the real- or complex-valued invertible matrices, that is the set
+
+```math
+\bigl\{p  ∈ 𝔽^{n×n}\ \big|\ \det(p) \neq 0 \bigr\},
+```
+where the field ``𝔽 ∈ \{ ℝ, ℂ\}``.
+
+# Constructor
+
+    InvertibleMatrices(n::Int, field::AbstractNumbers=ℝ)
+
+Generate the manifold of ``n×n`` invertible matrices.
+"""
+struct InvertibleMatrices{𝔽,T} <: AbstractDecoratorManifold{𝔽}
+    size::T
+end
+
+function InvertibleMatrices(n::Int, field::AbstractNumbers=ℝ; parameter::Symbol=:type)
+    size = wrap_type_parameter(parameter, (n,))
+    return InvertibleMatrices{field,typeof(size)}(size)
+end
+
+function active_traits(f, ::InvertibleMatrices, args...)
+    return merge_traits(IsEmbeddedSubmanifold())
+end
+
+function allocation_promotion_function(
+    ::InvertibleMatrices{ℂ,<:Any},
+    ::typeof(get_vector),
+    args::Tuple,
+)
+    return complex
+end
+
+@doc raw"""
+    check_point(M::InvertibleMatrices{n,𝔽}, p; kwargs...)
+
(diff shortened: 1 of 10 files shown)


src/manifolds/InvertibleMatrices.jl
```diff
@@ -0,0 +1,116 @@
+@doc raw"""
+    InvertibleMatrices{𝔽,T} <: AbstractDecoratorManifold{𝔽}
+
+The [`AbstractManifold`](@extref `ManifoldsBase.AbstractManifold`)  ``\operatorname{Sym}(n)``
```
mateuszbaran, line 4: Maybe we should find a different symbol for this manifold?

src/manifolds/InvertibleMatrices.jl
```diff
@@ -0,0 +1,116 @@
...
+    size::T
+end
+
+function InvertibleMatrices(n::Int, field::AbstractNumbers=ℝ; parameter::Symbol=:type)
+    size = wrap_type_parameter(parameter, (n,))
+    return InvertibleMatrices{field,typeof(size)}(size)
+end
+
+function active_traits(f, ::InvertibleMatrices, args...)
+    return merge_traits(IsEmbeddedSubmanifold())
```
mateuszbaran, line 28: Since in LieGroups.jl GL(n) is going to be a connection manifold, it might be a better idea to have the left-GL-invariant, right-O-invariant metric here?

src/manifolds/InvertibleMatrices.jl
kellertuer: Oh just missed to remove this, since I did not find a good symbol here, I for now went with no symbol.

src/manifolds/InvertibleMatrices.jl
kellertuer: My plan was to keep this one here lightweight and do the “heavy” metric/exp/log/... stuff in LieGroups?
Otherwise, the idea to make this one a quick PR and then work on GL in LieGroups is a bit turned upside down and this here will take quite a bit of time, because next I first have to read up on the metric you mention (O is the set of orthogonal matrices for example?). And what is a connection manifold?
Since that metric Is also heavily tailored to the Lie group, it is maybe better to document it there?

mateuszbaran: I think it would be easier to discuss this in-person.

mateuszbaran: There is nothing that needs changing here since the metric I was talking about seems to result in an invariant exponential map.

src/manifolds/InvertibleMatrices.jl
```diff
@@ -0,0 +1,106 @@
...
+        )
+    end
+    return nothing
+end
+
+"""
+    check_vector(M::InvertibleMatrices{n,𝔽}, p, X; kwargs... )
+
+Check whether `X` is a tangent vector to manifold point `p` on the
+[`InvertibleMatrices`](@ref) `M`, which are all matrces of size ``n×n```
```
mateuszbaran, line 52:
```suggestion
[`InvertibleMatrices`](@ref) `M`, which are all matrices of size ``n×n``
```

src/manifolds/InvertibleMatrices.jl
```diff
@@ -0,0 +1,106 @@
...
+
+Return true. [`InvertibleMatrices`](@ref) is a flat manifold.
+"""
+is_flat(M::InvertibleMatrices) = true
+
+@doc raw"""
+    manifold_dimension(M::InvertibleMatrices{n,𝔽})
+
+Return the dimension of the [`InvertibleMatrices`](@ref) matrix `M` over the number system
+`𝔽`, which is the same dimension as its embedding, the [`Euclidean`](@ref)`(n,n)`.
```
mateuszbaran, line 81:
```suggestion
`𝔽`, which is the same dimension as its embedding, the [`Euclidean`](@ref)`(n, n; field=𝔽)`.
```

mateuszbaran (approved):
LGTM

Issue #529, opened 2022-09-08
Implement the rotations manifold also for quarternions


kellertuer:
For `Rotations(3)` we could also implement the main functions using quaternions., since this is used in a lot of applications.

sethaxen:
There's (at least) Quaternions.jl (the most used) and Quaternionic.jl. Quaternions is working to drop its only dependency (https://github.com/JuliaGeometry/Quaternions.jl/issues/79#issuecomment-1235627348).

 Then Rotations.jl has a 3x3 matrix whose underlying storage is a unit quaternion. This type could already be used with Manifolds.jl, but I think everything would end up being promoted to a 3x3 static array instead of working with the unit quaternions directly much.

PR #765, opened 2024-10-24
New rotation actions

mateuszbaran:
Two actions useful in robotics and one new function.


src/groups/rotation_action.jl
```diff
@@ -304,3 +304,108 @@ function optimal_alignment(
...
+
+"""
+    quaternionic_rotation_matrix(g::Quaternions.Quaternion)
+
+Compute rotation matrix for [`RotationAction`](@ref) corresponding to
+[`QuaternionRotation`](@ref) by `g`.
+
+See https://www.songho.ca/opengl/gl_quaternion.html for details.
+"""
+function quaternion_rotation_matrix(g::Quaternions.Quaternion)
```
kellertuer, line 395: And `Quaternions.jl` does not yet have such a function? I would have expected that.

src/groups/rotation_action.jl
```diff
@@ -304,3 +304,108 @@ function optimal_alignment(
...
+struct QuaternionRotation <: AbstractGroupAction{LeftAction} end
+
+base_group(::QuaternionRotation) = Unitary(1, ℍ)
+
+group_manifold(::QuaternionRotation) = Euclidean(3)
+
+@doc raw"""
+    apply(A::QuaternionRotation, g::Quaternion, p)
+
+Rotate point `p` from [`Euclidean(3)`](@ref) manifold through conjugation by the group
```
kellertuer, line 359: $\mathbb R^3$, $\mathbb C^3$, or both?

Also for the actions you wrote the math term of the space they act on, here you write the manifold type/constructor. Let's maybe unify that?

src/groups/rotation_action.jl
```diff
@@ -304,3 +304,108 @@ function optimal_alignment(
...
+    return copyto!(q, apply(A, g, p))
+end
+function apply!(::ComplexPlanarRotation, q, ::Identity{MultiplicationOperation}, p)
+    return copyto!(q, p)
+end
+
+@doc raw"""
+    QuaternionRotation
+
+Action of the unit quaternion group [`UnitaryMatrices`](@ref)`(1, ℍ)` on ``ℝ^3``.
```
kellertuer, line 348: Does that mean, that we missed that unitary matrices already make #259 a bit obsolete? Because I just gave that task to students to get a bit into the code base here

src/groups/rotation_action.jl
mateuszbaran: `Quaternions.jl` has it in documentation but not in code. Maybe it's not worth adding here. I can delete it if you prefer.

src/groups/rotation_action.jl
mateuszbaran: Only $\mathbb{R}^3$, or at least I haven't verified if it makes sense in the complex case. I will ref the type here.

src/groups/rotation_action.jl
kellertuer: The point would be, if they have code, then we should use their code. But I am surprised that they have docs but no code? Then its maybe worth contributing this function there?

src/groups/rotation_action.jl
mateuszbaran: How is #259 related to this?

kellertuer: To exactly zero percent. I copied https://github.com/JuliaManifolds/Manifolds.jl/issues/529 – and have no clue how that got mixed up.

src/groups/rotation_action.jl
mateuszbaran: They have it here: https://juliageometry.github.io/Quaternions.jl/stable/examples/rotations/#Convert-a-quaternion-to-a-rotation-matrix so there isn't much to contribute actually. I didn't realize they even have it until you asked and I checked. There is also https://github.com/JuliaGeometry/Rotations.jl which has a lot more rotation-related stuff, much of which we already replicate anyway in some form.

kellertuer: hm, ok, then let's keep this here.

src/groups/rotation_action.jl
mateuszbaran: Ah, no, #529 is a superset of what I've done here.

kellertuer: Ok, perfect. We might get that somewhen from two students in Heidelberg, if all works well.

kellertuer (approved):

kellertuer:
While we now have the rule in LieGroups, that one approval is necessary (unless overwritten on squash and merge), should we add that here as well?

mateuszbaran:
> While we now have the rule in LieGroups, that one approval is necessary (unless overwritten on squash and merge), should we add that here as well?

OK, we can do that.

PR #770, opened 2024-11-24
Add some type annotations in `GeneralUnitaryMatrices` `get_vector`

mateuszbaran:
I had a nasty infinite recursion due to lack of an upper bound for that type. This didn't exactly solve the recursion but at least it was a less bad one with a useful stack trace.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.8] – unreleased
+
+### Changed
+
+Some methods related to `get_vector` for `GeneralUnitaryMatrices` now have `AbstractVector` upper bound for coefficients.
+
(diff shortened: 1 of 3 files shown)


kellertuer (approved):
While I am not sure there might be someone coming along with more general “vectory-things” for the coordinates, this looks good to me in general.

mateuszbaran:
In a few places we allow using plain numbers for coordinates (if there is only one coordinate). Other than this, `<:AbstractVector` shouldn't be an issue.

mateuszbaran:
I will merge this when CI passes and tag 0.10.8.

kellertuer:
You also have to update the `_quarto.yml` if you bump the docs CI Julia version

mateuszbaran:
I've reverted to building documentation on Julia 1.10 because the config file for quarto (used for some documentation pages) has the version of Julia in it and I don't know if we can have it automatically inferred.

kellertuer:
No, you have to switch the version in both, I do not think this can be inferred.

mateuszbaran:
Then maybe let's keep both at 1.10 for now. #772 changed it from "1.10" to "1" which won't work, so here I'm reverting this change.

PR #771, opened 2024-11-24
Upgrade to ManifoldsDiff v0.4

gdalle:
> [!TIP]
> Please review and merge #772 first.

- [x] Bump ManifoldsDiff compat to v0.4.0
- [x] Remove import of `AbstractDiffBackend`
- [ ] Undo Pkg tricks

> [!WARNING]
> Before merging, revert Pkg tricks to install custom ManifoldsDiff (v0.4.0 should be released by then)


NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.8] – unreleased
+
+### Changed
+
+* Bump compat for ManifoldsDiff to v0.4.0
```
kellertuer, line 12: When we drop support for Jluia 1.6, we should maybe mention that here as well.

kellertuer:
One thing that could be done here as well then, is to remove all the requires and `using ..X` tricks in the extension, since from 1.10 onwards they are not needed. But sure we can also do that on a next PR.

gdalle:
Good point, I'll do the Julia 1.10 upgrade first and then rebase this PR on it.

gdalle:
@kellertuer this should be good to go

mateuszbaran (approved):

PR #772, opened 2024-11-24
Bump Julia compat to 1.10

gdalle:
- Bump Julia compat in `Project.toml` from 1.6 to 1.10
- Bump Manifolds version to v0.10.8
- Update CI to run on `lts`, `1` (stable) and `pre` (pre-release)
- Remove Requires and its extension shenanigans

PR #773, opened 2024-11-24
Adjust permissions of format action

gdalle:
This is a test to see if I can overcome the formatting permissions issue of #772 

Related:

- https://github.com/reviewdog/action-eslint/issues/148

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -5,6 +5,12 @@ on:
     tags: [v*]
   pull_request:
 
+permissions:
+  contents: read
+  checks: write
+  issues: write
+  pull-requests: write
+
 jobs:
   format:
     name: "Format Check"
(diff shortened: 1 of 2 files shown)


gdalle:
This fails because PRs from forks cannot give a write permission to the `GITHUB_TOKEN`

> Finally, if the workflow was triggered by a pull request from a forked repository, and the Send write tokens to workflows from pull requests setting is not selected, the permissions are adjusted to change any write permissions to read only.

<https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job>

kellertuer:
But format also does not commit here, it would just _report_ the errors as code watch comments. Sure, these comments you might not get, but running `using JluiaFormatter; format(".");` is all we require here. So that can also just be considered as a reminder to do that similar to the Changeling CI, that we added after we (maybe mainly I) missed to write an entry there a few times in a row.

gdalle:
Fair enough, I see that it succeeds now because for a well-formatted code it doesn't try to write the comments. Closing.

PR #775, opened 2024-12-16
Heisenberg matrices

mateuszbaran:
So that we can add `HeisenbergGroup` to LieGroups.jl.


docs/src/manifolds/heisenbergmatrices.md
```diff
@@ -0,0 +1,7 @@
+# Heisenberg matrices
```
kellertuer, line 1: we are not fully unified here, but I think calling the file just `Heisenberg.md` would also be fine.

src/manifolds/HeisenbergMatrices.jl
```diff
@@ -0,0 +1,140 @@
...
+Heisenberg matrices `HeisenbergMatrices(n)` is the manifold of ``(n+2)×(n+2)`` matrices [BinzPods:2008](@cite)
+
+```math
+\begin{bmatrix} 1 & \mathbf{a} & c \\
+\mathbf{0} & I_n & \mathbf{b} \\
+0 & \mathbf{0} & 1 \end{bmatrix}
+```
+
+where ``I_n`` is the ``n×n`` unit matrix, ``\mathbf{a}`` is a row vector of length ``n``,
+``\mathbf{b}`` is a column vector of length ``n`` and ``c`` is a real number.
```
kellertuer, line 13:
````suggestion
```math
\begin{bmatrix} 1 & \mathbf{a} & c \\
\mathbf{0}_N & I_n & \mathbf{b} \\
0 & \mathbf{0}_n & 1 \end{bmatrix}
```

where ``I_n`` is the ``n×n`` unit matrix, ``\mathbf{a}`` is a row vector of length ``n``,
``\mathbf{b}`` is a column vector of length ``n``, ``\mathbf{0}_n`` is the zero-vector of length ``n``,  and ``c`` is a real number.
```
Maybe this is consistent with `I_n` ? It could also be shortly mentioned int the `where`-lines

File: src/manifolds/HeisenbergMatrices.jl [hunk 3]
```diff
@@ -0,0 +1,140 @@
...
+            "The matrix $(p) does not lie on $(M), since p[n + 2, 1:(n+1)] is not equal to 0.",
+        )
+    end
+    if !isapprox(I, p[2:(n + 1), 2:(n + 1)])
+        return DomainError(
+            det(p[2:(n + 1), 2:(n + 1)] - I),
+            "The matrix $(p) does not lie on $(M), since p[2:(n+1), 2:(n+1)] is not an identity matrix.",
+        )
+    end
+
```
kellertuer, line 69: Similar to the product manifold – would it maybe make sense to collect all errors into a type of group error and return that? might be nicer for the user.
If you feel that is too much work, we can also not do that of course.

kellertuer (commented):
Just a few small comments

src/manifolds/HeisenbergMatrices.jl
mateuszbaran: I don't think your edit is entirely right. There is one column 0 vector and one row 0 vector. Would you prefer to use transposition to differentiate between them? Or maybe give both dimensions in subscript?

kellertuer: ah and I have a N/n problem. I would prefer it to be a column vector I think, then – yes – the line 9 one should get a transpose.

docs/src/manifolds/heisenbergmatrices.md
mateuszbaran: OK, done.

src/manifolds/HeisenbergMatrices.jl
mateuszbaran: I think I've applied it correctly.

src/manifolds/HeisenbergMatrices.jl
mateuszbaran: Do you mean `CompositeManifoldError`? I think it would be good to first factor out the code that constructs it in `ProductManifold` so that it can be re-used here. It's not much code but still, I'd prefer to not copy it here.

Maybe let's do it in a new PR.

kellertuer: Ok, let's do it in a new PR.

kellertuer (approved):

PR #781, opened 2025-01-25
Adapt to ManifoldsBase 1.0

kellertuer:
This depends on https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/221 being merged.

Overall I went through all occurrences of `exp` and `exp!` and split/renamed them to `expt` and `expt!`, respectively, where necessary.
There are two tricky areas left

* interaction with the `MetricManifold`, which should carefully be checked
* interaction with `solve_exp_ode` where I can not yet see why they fail because the error messages basically kill my terminal in length.

Formerly this is breaking, because we drop support for ManifoldsBase 0.15?

Furthermore all occurrences of `retract`, `retract_` and all its follow up functions need to be split for each and every retraction method on every manifold.

* [x] add deprecations for the renamed types from `TVector` to `Tangentvector`


mateuszbaran:
`retract` rework is more or less done. I have some issues with OrdinaryDiffEq.jl failing to even precompile right now so that part will have to wait a bit. Next part is renaming and `injectivity_radius`.

kellertuer:
My rework from the weekend makes all tests pass, so after releasing 1.0, we only have to check for test coverage by now.

NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.13] - unreleased
+
+## Changed
```
mateuszbaran, line 10: We probably need to note the rename of `PoincareBallTVector` to `PoincareBallTangentVector` ? And also for other such types? Do we keep the old names as deprecated aliases?

kellertuer: Yes, right, I renamed all in a global search frenzy.

Since this is breaking I do not really want to keep them as deprecated.

mateuszbaran: Ah, so this is going to be Manifolds 0.11?

kellertuer: No!

You asked whether we shall keep the old names as deprecated – I wrote, _no_ I do not want that. We should just remove them since this is breaking.

mateuszbaran: Why shouldn't it be 0.11 if we make breaking changes? I'm confused.

kellertuer: Sure, we can hop back and forth for a while between 1.0 and 0.11 a few times now. For now all 3 PRs I am working on are prepared for 1.0. But if you want to go back to 0.11, sure. Then we go back. It was your idea to make this 1.0 in the first place,

mateuszbaran: This is Manifolds.jl, not ManifoldsBase.jl. I'm talking about Manifolds.jl here.

kellertuer: Sure, then we should add deprecations here as well. I also think, collecting deprecations is the better idea.

kellertuer: I added the deprecations as a point to resolve here still.

kellertuer:
This should be finished ± checking for code coverage after all the changes to exp/retr.

And formerly this is breaking since we stop compatibility for 0.15?

kellertuer:
Docs are already fixed here locally, checking tutorials already. But will only push after I see how the tests are doing and maybe working on code cov there already.

kellertuer:
Raised number of ambiguities allowed to 35 – noticed that it is also 35 on 1.11; but we exclude 1.11 from testing ambiguities; so it did not error locally for me. Do we want to keep that exclusion? Do we want to keep the ambiguity test or is the one from Aqua enough? (I just do not remember the arguments maybe).

mateuszbaran:
Maybe let's keep the exclusion; it prevents spurious failures on CI when a new Julia version is released. The bump in allowed ambiguities should be OK. The Aqua test makes wider exclusions than ours so I'd prefer to keep both.

kellertuer:
Ok, just means that me running on 1.11 would not see this happening and only CI would fail. But I checked the new ambiguities and they look harmless.

kellertuer:
We have code coverage – I even fixed a few previously uncovered lines, to slowly increase code cov here.

If it is ok, I would like to merge this and register a new version.

src/Manifolds.jl
```diff
@@ -702,20 +699,22 @@ export HyperboloidPoint,
...
-    ProjectorTVector,
-    StiefelTVector
+export HyperboloidTangentVector,
+    OrthogonalTangentVector,
+    PoincareBallTangentVector,
+    PoincareHalfSpaceTangentVector,
+    TuckerTangentVector,
+    UMVTangentVector,
+    ProjectorTangentVector,
+    StiefelTangentVector
```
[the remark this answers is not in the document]
kellertuer: They should be exported, in the tests, I do not have to write `Manifold.X` so they are indeed still exported (by default of the deprecate macro).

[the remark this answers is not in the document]
mateuszbaran: I see, I forgot about that :+1:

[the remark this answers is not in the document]
kellertuer: No worries, I usually forget these technical details as well, but when writing the tests I noticed that they are still there :)

mateuszbaran (approved):
LGTM

PR #785, opened 2025-02-18
Improved implementation of PT on Rotations(3)

mateuszbaran:
This could be further optimized but at least we won't call generic matrix `exp` on rotations in 3D.

### DIFF
--- a/src/manifolds/Rotations.jl
+++ b/src/manifolds/Rotations.jl
@@ -344,6 +344,33 @@ function Random.rand!(
     return pX
 end
 
+"""
+    _get_expdhalf(::Rotations, d)
+
+Calculate `exp(d / 2)`.
+"""
+_get_expdhalf(::Rotations, d) = exp(d / 2)
+"""
+    _get_expdhalf(::Rotations{TypeParameter{Tuple{3}}}, d)
+
+Calculate `exp(d / 2)` for the manifold of `Rotations(3)` based on the Rodrigues' rotation
+formula.
+"""
+function _get_expdhalf(::Rotations{TypeParameter{Tuple{3}}}, d)
+    θ = norm(d) / (2 * sqrt(2))
+    if θ ≈ 0
+        a = 1 - θ^2 / 6
+        b = θ / 2
+    else
+        a = sin(θ) / θ
+        b = (1 - cos(θ)) / θ^2
+    end
+    a /= 2
+    b /= 4
+
+    return I + a .* d .+ b .* (d^2)
+end
+
 @doc raw"""
     parallel_transport_direction(M::Rotations, p, X, d)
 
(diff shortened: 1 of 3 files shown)


src/manifolds/Rotations.jl
```diff
@@ -344,6 +344,33 @@ function Random.rand!(
...
+Calculate `exp(d / 2)`.
+"""
+_get_expdhalf(::Rotations, d) = exp(d / 2)
+"""
+    _get_expdhalf(::Rotations{TypeParameter{Tuple{3}}}, d)
+
+Calculate `exp(d / 2)` for the manifold of `Rotations(3)` based on the Rodrigues' rotation
+formula.
+"""
+function _get_expdhalf(::Rotations{TypeParameter{Tuple{3}}}, d)
```
kellertuer, line 359: is the `d` in the name necessary? And the `_get`?
I would prefer `_exp_half`? or – if I am correct, this is actually expsqrt?

src/manifolds/Rotations.jl
```diff
@@ -344,6 +344,33 @@ function Random.rand!(
...
+    θ = norm(d) / (2 * sqrt(2))
+    if θ ≈ 0
+        a = 1 - θ^2 / 6
+        b = θ / 2
+    else
```
kellertuer, line 364: Also already in earlier code, I did not understand these lines too much. Here `θ` is nearly zero, maybe up to a small type dependent tolerance.

But then up to the same tolerance `a` is just 1 (actually with a much smaller tolerance) and `b` is rust 0, or in short this could (together with the `a/2` later just return `I + 0.5 .* d`?

src/manifolds/Rotations.jl
```diff
@@ -344,6 +344,33 @@ function Random.rand!(
...
+    _get_expdhalf(::Rotations, d)
+
+Calculate `exp(d / 2)`.
+"""
+_get_expdhalf(::Rotations, d) = exp(d / 2)
+"""
+    _get_expdhalf(::Rotations{TypeParameter{Tuple{3}}}, d)
+
+Calculate `exp(d / 2)` for the manifold of `Rotations(3)` based on the Rodrigues' rotation
+formula.
```
kellertuer, line 357: If it is not too much effort, maybe documenting the formula that is used would be cool?

kellertuer (commented):
Looks nice! I just have a few small comments on naming, docs, and one on the code complexity.

src/manifolds/Rotations.jl
mateuszbaran: I will rename it to `_exp_half`.

src/manifolds/Rotations.jl
mateuszbaran: It's just temporary measure until someone has time to derive efficient formula for the entire PT. It's very standard code so nothing is gained by elaborating on it.

src/manifolds/Rotations.jl
kellertuer: Hm, the mention of AD I do not understand, for me these lines look useless ;)

src/manifolds/Rotations.jl
kellertuer: Hm, well I would gain that I maybe understand it, but I do see that there is no black/white border on what is “easy enough”. I had colleagues that claimed that Matlab is high-level enough that one never needs documentation...

But ok, then maybe document the entire PT variant more detailed once it is available.
I know that these formulae here are also used somewhere else and they come from a time before we documented formulae and their origin, I would like to slowly get rid of undocumented formulae.

src/manifolds/Rotations.jl
kellertuer: Sure, but that is even no where mentioned here, so I can not guess that it is necessary for that

mateuszbaran: You can always ask me if you need to know.

src/manifolds/Rotations.jl
mateuszbaran: This is not a time for handling that issue. This is not a new complicated formula.

mateuszbaran:
I'm sorry, but I don't have time to properly explain everything for every minor change.

kellertuer (approved):

mateuszbaran:
Thanks, I hope to get back to it and implement a better version with docs in the future.

PR #788, opened 2025-03-21
Add matrices of determinant one

kellertuer:
@mateuszbaran 
I had an idea how to maybe (slowly) restructure and replace the traits from ManifoldsBase.

I was a bit bored to yet again write a wrapper to check for det being one, so here is a new idea how to maybe do that.
One could also put the tuple type as a first argument in a THHT trick sense.

What do you think? Or is this not efficient?
One difference is that this is not a nested but a tuple approach.

For now I wanted to first introduce that only for matrices sub typing `MatrixManifold` to slowly introduce that for others as well – maybe in the long run replacing the decorated manifold.


mateuszbaran:
The main issue with this approach to traits is that it tries to appear modular but in practice it's not modular at all. The traits must be in a specific order for dispatch and any partial dispatches on properties are going to introduce massive ambiguity issues. Quite likely we could simplify `check_point`/`check_vector` call maze a bit with a generic `ArrayManifold` type but then I'd prefer to have `Properties` as a single `struct` instead of a tuple, roughly the same way `GeneralUnitaryMatrices` currently works. Maybe even we could start here by making `GeneralUnitaryMatrices` a special case of something like `GeneralMatrices` with `AbstractMatrixType` that is a subtype of `AbstractUnitaryMatrixType`, or something. Such hierarchy would necessarily look a bit arbitrary but is much less prone to ambiguities. Alternatively, we could offer `unitary`/`nonunitary` as a separate property type argument -- two simple type arguments for dispatch should be manageable.

kellertuer:
Hm, then I have to think about this. Because I could implement this with not too many lines of code by basically copying everything from invertible matrices, but I felt a bit bad copying code again. Maybe I do that here then because it is faster than spending a year on redesigning traits.

I do not mind that this is not flexible – if we do not drown in ambiguities and it is easier to use then the nested traits. And the idea was here to maybe transition to something nicer.
But maybe I am also just not able to program that.

So I will probably just copy Invertivle Matrices then and change 2-3 places where det1 is checked.

mateuszbaran:
I think we shouldn't introduce a new solution for matrix manifolds here. I'm fine with either copying invertible matrices or extending general unitary matrices to non-unitary cases. IMO copying is still better than introducing an unnecessary abstraction in case it might be useful in the future.

kellertuer:
Well, my hope was this idea could serve as a replacement for treats, but it seems that is not the case.
It was not meant as another additional layer but a replacement idea. But it seems it is no good.

So I will copy yet again.

kellertuer:
Hm, it seems in my internal renaming of `DeterminantOneMatrices` to `DeterminantOneMatiricesType` I might not have been careful enough. Will check tomorrow.

src/manifolds/DeterminantOneMatrices.jl
```diff
@@ -0,0 +1,136 @@
...
+)
+    rand!(rng, get_embedding(M), pX; kwargs...)
+    if vector_at == nothing # for points ensure invertibility
+        n = size(pX)[1]
+        e = eigen(Symmetric(pX' * pX)) # sym to get real EV
+        ev = real.(e.values)
+        (prod(ev) < 0) && (ev[1] *= -1)
+        # determinant is  already positive, rescale to one
+        ev .*= 1 / (prod(ev)^(1 / n))
+        pX .= e.vectors' * diagm(ev) * e.vectors
```
mateuszbaran, line 123:
```suggestion
while det(pX) < sqrt(eps(eltype(pX)))
            rand!(rng, get_embedding(M), pX; kwargs...)
        end
        if det(pX) < 0
            pX[1, :] .*= -1
        end
        pX ./= det(pX)^(1/n)
```
I think something like this would be much easier? And it seems a bit less prone to numerical errors.

kellertuer: I got lost for about an hour in this, do not ask me what is easier here.

src/Manifolds.jl
```diff
@@ -636,12 +637,14 @@ export AbstractCotangentVector,
     AbstractManifold, AbstractManifoldPoint, AbstractTangentVector
 # Manifolds
 export AbstractSphere, AbstractProjectiveSpace
+export MatrixManifold
```
mateuszbaran, line 640:
```suggestion
```

kellertuer: oops, missed that. Surprised Aqua did not complain that we export a non defined thingy.

mateuszbaran (approved):
LGTM, let's just make sure coverage is fine.

kellertuer:
So CI is back, but I am not yet sure how to cover the lines for the cases where we want to be sure the matrix is invertible, when this has such a low probability to occur

mateuszbaran:
I've found a way to test that.

kellertuer:
So this is hopefully only waiting for the docs to render again once the new versions in SciML are all up to date, but Chris said yesterday he is looking into this.

kellertuer:
It finally works, merging this and updating the other two PRs should bring us up to date and to the next version 🥳

mateuszbaran:
Great!

mateuszbaran (approved):

PR #789, opened 2025-03-24
Fix `rand!` on complex circle

kellertuer:


kellertuer:
I do not understand how rand on the complex circle (added) should make Kendalls shape space fail. That does make no sense.

mateuszbaran:
Maybe this has a bug: https://github.com/andreasvarga/MatrixEquations.jl/commit/69516a51b7a17074609c0dc2e0532565f1883d67 ?

kellertuer:
Uff, let's see whether we can narrow that down to an MWE then and report it, but yes that updates exactly the function we use – and locally. Will check and compare.

mateuszbaran (approved):

PR #803, opened 2025-03-29
fix a misleading error message.

kellertuer:
Due to copy paste one error message was wrong and even quite misleading. This fix can still wait for the nest release.


mateuszbaran (approved):
LGTM up to that one minor issue.

PR #805, opened 2025-04-07
`get_vector` on `Circle` no longer returns `SArray` when no static ar…

mateuszbaran:
…rays are passed as arguments

Otherwise LieGroups.jl doesn't like it.


kellertuer (approved):
Let me extend to the original post – I also did not like that because that hit me quite a few times already ;) 

Or positively phrased – I feel it is great when this is type-consistent. Also because the new test suite might test that as generic as Lie groups already does.

PR #806, opened 2025-05-03
fix supertype of a few tangent vectors that were `<: AbstractManifoldPoint`

kellertuer:
Just saw that a super type was wrong – this can wait for the next larger thing to be fixed I think, but awe could maybe merge it already to master.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   test:
-    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
+    name: Julia ${{ matrix.julia-version }} - ${{ matrix.group }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,18 +1,15 @@
-name: CI
+name: nightly
 on:
-  push:
-    branches: [master]
-    tags: [v*]
   pull_request:
 
 jobs:
   test:
-    name: Julia ${{ matrix.julia-version }} - ${{ matrix.group }} - ${{ matrix.os }}
+    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ["lts", "1"]
-        os: [ubuntu-latest, macOS-latest]
+        julia-version: ["pre"]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
         group:
           - 'test_manifolds'
           - 'test_lie_groups'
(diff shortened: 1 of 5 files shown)


kellertuer:
@mateuszbaran I have a question about `StiefelFactorization` (maybe also @sethaxen since you might have programmed this):

The type is a `<: AbstractManifoldPoint` but used as a point and a tangent vector. Does this make sense – or should we maybe either

* use it in both but not make it a `<: AbstractManifoldPoint` but more like a generic representation of things just used for both
* introduce two types, one for tangent vectors one for points?

I think I would prefer the first.

kellertuer:
I moved `pre` also to its own workflow here, it is maybe ok to mark that one as “do not fail CI if this fails”?

mateuszbaran:
`PoincareBallTangentVector` is still a point on the manifold of tangent vectors at a point so it's not entirely wrong. I generally consider `AbstractManifoldPoint` and `AbstractTangentVector` generally irrelevant anyway. You should be able to safely remove the supertype if use're not sure which one to use.

mateuszbaran:
> it is maybe ok to mark that one as “do not fail CI if this fails”?

Yes, that would be OK.

kellertuer:
> `PoincareBallTangentVector` is still a point on the manifold of tangent vectors at a point so it's not entirely wrong.

Well that would make indeed both types 100% obsolete.

I am not so sure why we introduced them, and they are, as you said, not necessary in general, but as a meta-information still nice.

> You should be able to safely remove the supertype if use're not sure which one to use.

I do not want to remove them, I want to make them semantically right. That is all “manifold point type” structs should be `<: AbstractManifoldPoint` and all (tangent/vector fibre?) vectors should be `<: AbstractTangentVector`. I also mainly noticed this bug only by looking at all subtypes of these two.

kellertuer:
For the failing `pre` CI I also checked on slack with sciml - and opened https://github.com/JuliaLang/julia/issues/58312

mateuszbaran:
> I do not want to remove them, I want to make them semantically right. That is all “manifold point type” structs should be `<: AbstractManifoldPoint` and all (tangent/vector fibre?) vectors should be `<: AbstractTangentVector`. I also mainly noticed this bug only by looking at all subtypes of these two.

That's fine, I just don't care about those supertypes. We can have them but they are not useful for me.

> For the failing `pre` CI I also checked on slack with sciml - and opened [JuliaLang/julia#58312](https://github.com/JuliaLang/julia/issues/58312)

Cool, another issue involving `BoundaryValueDiffEq` :laughing:

kellertuer:
That is fine, I think they are mainly useful on a meta level. They might be handy in the `Manopt/JuMP` extension actually, where we need to handle those “non-array” points/vectors separately.

No, the issue is more general and this time not due to SciML/BoundaryDiffEq but more general/generic to Julia itself.

kellertuer:
Can we merge this? It does not need its own version since it is not urgent and can just be part of a next release somewhen

PR #810, opened 2025-05-29
Fix `riemannian_gradient` on Fixed Rank,

### LINKED ISSUE #809: Add riemannian_gradient for FixedRankMatrices
We already have `riemannian_Hessian` somehow but not the gradient.

kellertuer:
this PR fixes #809, since the `EuclideanMetric` was not yet declared as the default metric for fixed rank, and there was a small bug in one of the fallbacks.

We now have again

```
julia> using Manifolds

julia> p = rand(M);

julia> Manifolds.riemannian_gradient(M, p, zeros(4,3)) # Assume the last argument is the Euclidean gradient, this now converts
UMVTangentVector{Matrix{Float64}, Matrix{Float64}, Matrix{Float64}}
U factor:
 4×2 Matrix{Float64}:
  0.0  0.0
  0.0  0.0
  0.0  0.0
  0.0  0.0
M factor:
 2×2 Matrix{Float64}:
  0.0  0.0
  0.0  0.0
Vt factor:
 2×3 Matrix{Float64}:
  0.0  0.0  0.0
  0.0  0.0  0.0
```

Should we release this as a patch version then? Then I would bump version and such.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,12 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.10.18] unreleasdd
+## [0.10.18] unreleasd
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


kellertuer:
We could ignore pre probably, it‘s SciML again.

Issue #811, opened 2025-06-03
Generalizing SPD manifolds to Hermitian matrices?


sethaxen:
Can we generalize `SymmetricPositiveDefinite` and `SPDFixedDeterminant` to handle also Hermitian matrices? Mostly I'm wondering if key results like exponential and logarithmic maps with our current metrics are known for the Hermitian generalizations of these manifolds.

This came up in JuliaManifolds/ManifoldDistributions.jl#8, where we need optimization on `SPDFixedDeterminant` but supporting also Hermitian matrices.

PR #813, opened 2025-06-04
[WIP] HermitianPositiveDefinite

kellertuer:
I started the HPDs as proposed in #811, but here at home I noticed that programming on a German layout is not something I am used to.

The idea is to 

# 📋
* [ ] introduce as many functions as possible by generalising them from the SPD case.
* [ ] if reasonable maybe even make the SPD case a `const` of the HPD-HermitianPositiveDefinite{ℝ} case

Check a bit for literature and document carefully. Feel free to check functions as well otherwise I will slowly work through the affine invariant metric.


mateuszbaran:
Yes, I think the SPD should just be an alias for real-coefficient HPD.

kellertuer:
Well, if we manage to finish that, for now I struggle a bit to find time to continue, but I will try to continue this step by step though maybe I am a bit slow since for anything beyond checks of points/vectors I first have to read up on the formulae as well.

PR #814, opened 2025-06-09
Fix project and embedd of tangent vectors for General Linear

kellertuer:
Though already deprecated, this is a fix of GL(n) concerning embed/project of tangent vectors due to their representation in the Lie algebra, cf. https://github.com/JuliaManifolds/Manopt.jl/issues/473#issuecomment-2955075922

# 📋
* [x] fix project/embed
* [x] extend documentation
* [x] `News.md` updated
* [x] bump version number
* [ ] check code coverage

### DIFF
--- a/ext/ManifoldsTestExt/tests_general.jl
+++ b/ext/ManifoldsTestExt/tests_general.jl
@@ -603,7 +603,7 @@ function test_manifold(
                 end
                 # check projection idempotency
                 for i in 1:N
-                    Test.@test isapprox(M, p, project(M, p, bvectors[i]), bvectors[i])
+                    Test.@test isapprox(M, p, project(M, p, bvectors[i]), p \ bvectors[i])
                 end
             end
             if !isa(btype, ProjectedOrthonormalBasis) && (
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.19] unreleased
+
+### Fixed
+
+* Fix the projection and embedding of tangent vectors on `GeneralLinear`.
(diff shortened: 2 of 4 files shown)


ext/ManifoldsTestExt/tests_general.jl
```diff
@@ -603,7 +603,7 @@ function test_manifold(
                 end
                 # check projection idempotency
                 for i in 1:N
-                    Test.@test isapprox(M, p, project(M, p, bvectors[i]), bvectors[i])
+                    Test.@test isapprox(M, p, project(M, p, bvectors[i]), p \ bvectors[i])
```
mateuszbaran, line 606: This is a generic test, so you can't modify it like that. It's also used for manifolds that don't do this translation to identity. Easiest fix would be disabling or special-casing this test for GL(N).

kellertuer: Oh, yes, I was too fast there. My usual problem – thinking I can fix things fast. Then this has to wait until I find time again.

kellertuer: I at least did an undo on this. But then someone (tm) still has to adapt the tests to make this pass again. I added a todo in the PR.

kellertuer: That is a bit hacked, but given that this is mainly on deprecated code probably fine.

mateuszbaran: It will be removed in the next breaking version anyway, so I don't care that much.

mateuszbaran:
Yes, this appears to be correct. I've made a small code for verifying the gradient and it seems fine:
```julia
using Manifolds, Manopt, Plots, LinearAlgebra

N = 4

M = GeneralLinear(N)

q = rand(M)

f(M, p) = norm(p - I)^2
grad_eucl(M, p) = 2*(p - I)

function grad_riem(M, p)
    return Manifolds.riemannian_gradient(M, p, grad_eucl(M, p))
end

Manopt.check_gradient(M, f, grad_riem; plot=true)
```

PR #816, opened 2025-06-16
Fix in-place exp! for GeneralLinear

mateuszbaran:
Resolves the issue spotted in https://github.com/JuliaManifolds/Manopt.jl/issues/473 .


kellertuer (approved):
LGTM, can you check the last code cov report? Ah that might also just due to Lie groups not having finished yet at time of writing.

PR #826, opened 2025-07-26
An attempt to remove traits

mateuszbaran:
The goal is to make the internal structure easier to understand.


mateuszbaran:
It's (hopefully) small from the point of view of end users. Feel free to remove it if you think it doesn't fit.

kellertuer:
From that perspective yes, from ours, no ;) Internally it is quite a rework, but I like that the `default_` function approach gets more unified now as well.

kellertuer:
I fixed all manifolds generally,
in `tests/metric.jl` there is still a lot of failings, since the interaction with the `IsMetricManifold` has to be reworked. Some of that is the default to solve an ODE for metric manifolds. 

I have to think a bit how to rewrite that logic so that the function `exp!` in line 52 `ManifoldsOrdinaryDiffEqExt.jl` is accordingly “activated” once one loads that package. Currently still the new version without traits that checks for default metric is called even after extension is loaded.
I would expect that the interaction / removing `IsMetricManifold` would solve most those remaining errors in metric.jl tests.

kellertuer:
I am running into a small problem with the “dispatch logic” for `exp` on metric manifolds without traits.

Formerly:
* the metric that is now the `default_metric` would pass on to the underlying manifold (with implicit metric)
* all others would error usually,
* some “magically” called the ODE solver when `IsMetricManifold` was set, though that is a retraction actually

The last two can not really be distinguished by now. I would propose to leave out the last one as a “magic default” since if no exp is implemented, the user has to specify a retraction anyways – but could use the ODEsolver one?

kellertuer:
I continued removing the Connection based traits `IsDefaultConnection` and `IsConnectionManifold`. I noticed that there is a `connection(M)` function, so I am now considering calling `default_metric(M)` also just `metric(M)` since that seems not only more consistent but also more in line with: For every manifold we assume a metric and only one, so there is no need for a `metric_` case.

I also fixed the default of the ODE solver to use the `ODEExponentialRetraction` instead of overwriting `exp_fused!` and also defined that for both retract and its fused version – I think we might have missed to define the first one when we split those two functions.

There is still a few errors in the tests, but it is getting better and I think also overall more consistent.

kellertuer:
> I'm fine with removing groups as soon as https://github.com/JuliaManifolds/LieGroups.jl/issues/52 is resolved, but not earlier.

Hm, ok. That issue might still take a while, because it is a lot of types that I never felt had any use actually. But sure we should also check that for each case.

But who knows how long this rework will take, maybe I also find time until then to check at least a few functions. Or we do two breaking changes here in the end.

mateuszbaran:
I think I'd prefer to align removal of traits with splitting out Lie groups. BTW, not all things on that list are actions. There are also Cartan-Schouten connections or some Jacobians.

kellertuer:
Sure. I also specifically only referred to actions. For the connections I have to check, those I probably really simply forgot.

src/groups/heisenberg.jl
```diff
@@ -26,10 +26,11 @@ function HeisenbergGroup(n::Int; parameter::Symbol=:type)
     return HeisenbergGroup{typeof(size)}(size)
 end
 
+is_embedded_manifold(::HeisenbergGroup) = true
```
kellertuer, line 29: Just as a general remark, this function might need a case where it depends on the point type? Some representations might be embedded, others not. 

Though that might not affect this specific case, we at least should have that in mind and handle it similar to other cases like rand.

kellertuer:
> To be honest, I think that ODE solver for the geodesic is really bad. The one from `ManifoldsOrdinaryDiffEqDiffEqCallbacksExt` is the right approach. I'd prefer to just remove the old one. I'm thus perfectly fine with no longer using it automatically.

Hm I am by now very confused. But maybe that will just become my default status also on these PRs.
We do have a `solve_exp_ode!` in `ManifoldsOrdinaryDiffEqExt.jl` – with which I just fought for a while, bringing it into a retraction, since the `ODEExponentialRetraction` always requires a retraction internally to work, which does not fit to these. Besides that this looks totally fine to me? Why is it really bad? We even wrote that ODE in the Kalman paper. What is bad with that ODE?
I am tempted to introduce a retraction without an internal one for this ODE, but you said we can even remove it? I am confused, why we should remove the default ODE one solves for geodesics.

On the other hand `ManifoldsOrdinaryDiffEqDiffEqCallbacksExt.jl` works in charts? That I would consider the strange one then? Why is that the right approach?

Sorry, got lost on that and hence we have another day without much progress from my side. I struggle a bit coming from the code back to any theory, simply because there is none documented.

src/groups/heisenberg.jl
mateuszbaran: That's a tricky point. It might indeed have to depend on the point type. I will think about it.

mateuszbaran:
> Besides that this looks totally fine to me? Why is it really bad? We even wrote that ODE in the Kalman paper. What is bad with that ODE?

The main issue is that it doesn't let the user select a chart to work in. It doesn't do proper chart switching. It basically assumes working in `Euclidean` with a custom metric. It uses `christoffel_symbols_second` which is an inefficient approach and only has an implicitly selected chart, in a way that has not been sufficiently verified to work well. And there is no point in verifying it because a better approach is already implemented.

The Kalman paper ODE is what's implemented in `ManifoldsOrdinaryDiffEqDiffEqCallbacksExt.jl`.

> I am tempted to introduce a retraction without an internal one for this ODE, but you said we can even remove it? I am confused, why we should remove the default ODE one solves for geodesics.

Because it's a poor default.

> On the other hand `ManifoldsOrdinaryDiffEqDiffEqCallbacksExt.jl` works in charts? That I would consider the strange one then? Why is that the right approach?

Yes, the geodesic equations are basically always expressed in a chart. What's so strange about it?

mateuszbaran:
I think I've figured out an even easier fix. I will push a commit to ManifoldsBase in a moment.

kellertuer:
Short status update: I think I fixed a few more manifolds (Segre, Stiefel), but now to a bit suck on the interplay between metric manifold and the embedding, or to be precise:

I think on the metric manifold, so using a nondefault metric, the non_simple embedding does not work correctly – or I missed to specify for this combination the unwrapping of SPDPoints (towards the embedding?) wrongly.

kellertuer:
Short status update:

* `UnitaryMatrices` has a part on SpecialUmitary failing, (`is_flat`2x, `manifold_dimension`, and `)
* statistics has still failing parts that I also struggle to hunt down
* metric has still failing parts with the same problem as the previous two for me.

overall not many left (4 failed 5 errored)

I am unsure what to do about the GroupManifolds?

kellertuer:
For Grassmann it is implicit enough that it can stay here.

Sure we can also move Kendall if you think we can not do it Grassmann-styleish enough.

kellertuer:
I would prefer keeping a Grassmann here, moving that would be super breaking, because it is used by quite a few people.

Sure if we can “enhance” Grassmann within LieGroups, that's cool, but I would do that as an enhancement over there.

And sure, if you feel such a concrete Kendall here does not make so much sense, we can move it; we are working towards a breaking release here anyways.

kellertuer:
My problem with moving Grassmann is that it really would break a lot a lot of code, including my current PR to DFTK.

That is why I proposed to have a simple or the current version just here and a boosted/enhanced version in Liegroups.

I mean the current Grassmann version here works quite okay and fine – just without LieGroups. I do not see a reason to remove it, just because in principle it is a quotient. Same for the sphere. And maybe even for Kendall. But Kendall seems to actually need LieGroups tools, the other two do not directly need that.

kellertuer:
Merged runic, brought it here.

For me locally (with the other ManifoldsBase branch checked out) some small things on Grassmann and Kendall still fail (one of them due to the orbit-function that was deleted, Grassmann with exp?) but maybe that was there before. Can look at it closer tomorrow.

kellertuer:
With ManifoldsBase 2.0 this PR only errors on the distance of Kendalls shape space, since we do not longer have the optimal alignment available here. Can we maybe hardcode this specific case in that distance to make it work again?

All other tests run locally for me (but I also have to switch parameters).

mateuszbaran:
IIRC `log` also won't work without `optimal_alignment`. I considered doing it locally here but I don't like it because `optimal_alignment` is the natural place to define that functionality and we'd duplicate some code or do weird dispatches. I don't like other solutions either so I'm not strongly opposed to this one. Maybe we could have `LieGroupsBase.jl` that defines `optimal_alignment` and is a dependency of `Manifolds.jl`? That's more work but otherwise it might be the cleanest  solution.

kellertuer:
Hm. We are Tendenz towards more and more Detroits. Can we finish 11 where we delete these and do that as a next step?

mateuszbaran:
That's fine too, sure.

kellertuer:
I finished the refactor of all manifolds to have the F upfront. I noticed that I either had to switch or could simplify some `ManifoldName{<:Any, F}` occurrences.
All those changes are documented.

So now locally only a few statistics methods fail. that I do not fully understand why they do.

I think we can think about releasing ManifoldsBase 2.0, announcing that and looking here for code coverage (and bumping the remaining packages).

ext/ManifoldsOrdinaryDiffEqExt.jl
```diff
@@ -45,13 +44,25 @@ function solve_exp_ode(
     q = sol.u[end][(n + 1):(2 * n)]
     return q
 end
-# also define exp! for metric manifold anew in this case
-function exp_fused!(
-        ::TraitList{IsMetricManifold}, M::AbstractDecoratorManifold, q, p, X, t::Number;
-        kwargs...
+
+function ManifoldsBase.retract_exp_ode!(
```
mateuszbaran, line 48: Since we're making a breaking release here, I think we could finally remove this solver? `solve_chart_exp_ode` should be the replacement.

docs/make.jl
```diff
@@ -110,6 +101,22 @@ using Test, FiniteDifferences
 ENV["GKSwstype"] = "100"
 
 # (e) add CONTRIBUTING.md and NEWS.md to docs
+
+function add_links(line::String, url::String = "https://github.com/JuliaManifolds/Manopt.jl")
```
mateuszbaran, line 105: Shouldn't it link to Manifolds.jl instead of Manopt.jl?

test/groups/general_unitary_groups.jl
```diff
@@ -1,261 +0,0 @@
...
-        e = Identity(MultiplicationOperation())
-        @test_throws DomainError is_vector(SU2, e, Xe, true; error = :error) # Xe not skew hermitian
-
-        @test manifold_volume(SpecialUnitary(1)) ≈ 1
-        @test manifold_volume(SpecialUnitary(2)) ≈ 2 * π^2
-        @test manifold_volume(SpecialUnitary(3)) ≈ sqrt(3) * π^5
-        @test manifold_volume(SpecialUnitary(4)) ≈ sqrt(2) * 4 * π^9 / 12
-    end
-
-    @testset "SO(4) and O(4) exp/log edge cases" begin
```
mateuszbaran, line 195: Should we keep these tests in Manifolds.jl as tests for `exp`/`log` or move them to LieGroups.jl?

docs/src/manifolds/shapespace.md
```diff
@@ -21,12 +21,6 @@ q = [
 # let's plot them as triples of points on a plane
 fig = scatter(p[1,:], p[2,:], label="p", aspect_ratio=:equal)
 scatter!(fig, q[1,:], q[2,:], label="q")
-
-# aligning q to p
-A = get_orbit_action(M)
-a = optimal_alignment(A, p, q)
-rot_q = apply(A, a, q)
-scatter!(fig, rot_q[1,:], rot_q[2,:], label="q aligned to p")
```
mateuszbaran, line 29: This is a relatively important part of shape spaces, so maybe it would be better to move the entire tutorial to LieGroups.jl instead of keeping an incomplete one here.

kellertuer: Ah yes, sorry, I did not have that in mind, that it might make the example too minimalistic.

I think it would be great example for ManifoldExamples – maybe even defining the optimal alignment just in the example might work for now.

IN general I think, having more like “how to do this” tutorials here and nice examples over on the examples packages might be a good idea to follow.

test/groups/general_unitary_groups.jl
kellertuer: I think they are already done as edge cases on Liegroups.jl? See e.g. https://github.com/JuliaManifolds/LieGroups.jl/blob/main/test/LieGroupsTestSuite.jl/so4_edge_cases.jl where they are defined even within the test suite.

kellertuer:
The last 23 lines are a bit hard to spot since code cov is a bit confused and there is a few more lines than that not covered. Do you want to give it a try? I want at least through easy cases I could check.

mateuszbaran:
I've added a few tests. It's easier to see what's missing by looking at the entire file on codecov like here: https://app.codecov.io/gh/JuliaManifolds/Manifolds.jl/pull/826/blob/src/manifolds/GeneralUnitaryMatrices.jl instead of the default overview.

test/groups/general_unitary_groups.jl
kellertuer: Maybe you are right and added them back already for the matrices. So we can close this.

docs/src/manifolds/shapespace.md
kellertuer: I understand and sure then lets move it over there.

kellertuer (approved):

docs/src/manifolds/shapespace.md
kellertuer: Ok, so then let's just remove code of the example here.

PR #833, opened 2025-10-09
Be more precise when extending functions that were not imported.

kellertuer:
On Julia 1.12, there is now a warning when extending a function that was loaded with `using` but not imported. 
I prefer to slowly move over to not import but sully specify, so the two fixes here follow that idea as well.


mateuszbaran (approved):
That's a good idea I think.

PR #834, opened 2025-10-17
[WIP] A test suite

kellertuer:
Currently with a small problem around

https://github.com/JuliaLang/Pkg.jl/issues/1585

that causes the tests on 1.10 to fail in this setup, maybe?


kellertuer:
I have no ideas left what to do on 1.10 actually for a proper test suite. So before I waste another weekend, maybe it is time to recognise, that a new Test suite will even just due to technical difficulties before it was even started, just not happen.

Maybe at some point later when I am again anonoyed by not having integration tests, I might give it a another try, but it seems my current idea is simply not doable on Julia 1.10, since it either errors with the package of local `[source]` (in later versions) being “not registered” or with an error that projects could not be merged.

PR #835, opened 2025-10-23
Fix passthroughs for retractions, inverse retractions, and vector transport on the `MetricManifold` decorator

### LINKED ISSUE #823: retract_project! does not get passed through to the manifold of a MetricManifold
Hi all,
I've encountered a problem with the way `retract_project!` works for a manifold of type `MetricManifold`. For `M::MetricManifold` It should in my opinion use `retract_project!(M, q, p, X) = retract_project!(M.manifold, q, p, X)`. It can be reproduced with the following code-
```julia
using Manifolds
using Manopt
using LinearAlgebra

struct DummyMetric <: RiemannianMetric end

function Manifolds.inner(M::MetricManifold{ℝ,<:Sphere{<:Any,ℝ},<:DummyMetric}, p, X, Y)
    return (dot(X,Y))
end

# Construct sphere with dummy metric
n = 4
dummy_metric = DummyMetric()
M = MetricManifold(Sphere(n-1), dummy_metric)

# Construct model problem: Eigenvalue minimization of spd matrix
A = diagm(1:n)

obj = ManifoldFirstOrderObjective(;
    cost = (M,p) -> dot(p, A*p), 
    gradient = (M,p) -> (A*p - dot(A*p,p) * p)
)

# solve using RCG
retraction_method = ProjectionRetraction()
vector_transport_method = ProjectionTransport()

p = rand(M)
conjugate_gradient_descent!(M, obj, p; retraction_method, vector_transport_method)
```
This throws the following error.
```
ERROR: MethodError: inner(::AbstractManifold, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64}) is ambiguous.
```
It can be fixed by introducing
```julia
# fix
function Manifolds.retract_project!(M::MetricManifold{ℝ,<:Sphere{<:Any,ℝ},<:DummyMetric}, q, p, X)
    return Manifolds.retract_project!(M.manifold, q, p, X)
end
```
Notably, this problem does not occur with the vector transport method. It seems, here the nested manifold of the the `MetricManifold` gets passed through correctly. 

On another note, the convergence here is extremely slow, since RCG randomly stagnates, probably because the stepsize gets extremely small. I will create another issue on the `Manopt` page for this.

kellertuer:
This fixes #823 in a general sense:

Since every retraction is a retraction with respect to every metric, i.e. a first order approximation of the exponential map associated to that metric, the metric manifold will “pass through” all retractions, **unless** it is the exponential retraction.

Similarly the inverse retraction (and log) as well as both vector tranports (to & direction and the corresponding parallel transport) are fixed.

This just needs a bit of testing to keep code coverage happy.

The example from @jonas-pueschel now does run (when adapting that the sphere now has its field parameter first):

```
using Manifolds
using Manopt
using LinearAlgebra

struct DummyMetric <: RiemannianMetric end

function Manifolds.inner(M::MetricManifold{ℝ,<:Sphere{ℝ},<:DummyMetric}, p, X, Y)
    return (dot(X,Y))
end

# Construct sphere with dummy metric
n = 4
dummy_metric = DummyMetric()
M = MetricManifold(Sphere(n-1), dummy_metric)

# Construct model problem: Eigenvalue minimization of spd matrix
A = diagm(1:n)

obj = ManifoldFirstOrderObjective(;
    cost = (M,p) -> dot(p, A*p), 
    gradient = (M,p) -> (A*p - dot(A*p,p) * p)
)

# solve using RCG
retraction_method = ProjectionRetraction()
vector_transport_method = ProjectionTransport()

p = rand(M)
conjugate_gradient_descent!(M, obj, p; retraction_method, vector_transport_method)
```

nicely returns

```
julia> conjugate_gradient_descent!(M, obj, p; retraction_method, vector_transport_method)
4-element Vector{Float64}:
  0.8945753502239056
 -0.3643281895014556
 -0.24708139400126913
  0.07714076642578808
```


kellertuer:
Hm, there are a bit unexpected side-effects, that exp/log for the submersion metric on Stiefel no longer works as expected, but I did not change any of the functions, as far as I am able to follow this. But the points are random points, so maybe some randomness changed? Hard to track down (spent the last 2 hours on that, besides a short dinner break).

mateuszbaran (approved):

PR #839, opened 2025-11-05
Towards a Test submodule (refactoring tests, Chapter MCLXVII)

kellertuer:
I know there have now been a few approaches towards test functions. But based on the Lie groups test suite that I really start to like, this is an approach that is maybe a good way to go.

* a submodule `Manifolds.Test`
* a test function per function from the API
* a test suite running through all functions (specifying what to test and what to expect in tolerances and such).

Especially the submodule with a weak test dependency will allow others to test their own implementations as well.


kellertuer:
The old tests are running again, I managed to do rear/inverse retr today. A few functions might need some extra tricks, like injectviity radius. Will think about how to best design that to check their values.

kellertuer:
I started writing a test for `flat` but struggle a bit, since there is

https://github.com/JuliaManifolds/Manifolds.jl/blob/bf70d69242d5d051e59504b442ed9724aa96d720/src/cotangent_space.jl#L7

That `flat` happily returns, but there is no copy or anything supported, so how can I best generically test `flat!` then? @mateuszbaran ?
For now I´ll set `test_mutating` to false, but maybe we can get cotangents to be better first-class citizens? Also we could just support arrays here as well maybe? (implicitly assuming they are on the right manifold and right cotangent space)?

test/manifolds/test_sphere.jl
```diff
@@ -0,0 +1,356 @@
...
+        #= Add the following to the test suite as functions
+            @testset "Weingarten" begin
+                M = Sphere(2)
+                p = [1.0, 0.0, 0.0]
+                X = [0.0, 2.0, 0.0]
+                V = [0.3, 0.0, 0.0]
+                Y = -X * (p'V)
+                @test Weingarten(M, p, X, V) == Y
+            end
+        =#
```
kellertuer: This should be covered, did I forget to remove this? Weingarten is tested I hope.

test/manifolds/test_sphere.jl
```diff
@@ -0,0 +1,356 @@
...
+    V = [2.0, 0.0, 0.0]  # normal vector at p
+    ξ = [0.0, π / 4, 0.0]  # covector at p
+
+    @testset "Basics" begin
+        @test is_default_metric(M, EuclideanMetric())
+        @test !is_default_metric(M, AffineInvariantMetric())
+        @test base_manifold(M) === M
+        @test representation_size(M) == (3,)
+        @test !is_flat(M)
+        @test is_flat(Sphere(1))
```
kellertuer: We should discuss which of these do make sense to be moved to the test suite.

mateuszbaran: `representation_size` and `is_flat` definitely make sense to be in the suite, I'm not sure about the other ones.

kellertuer:
@mateuszbaran  can you take a look at the gazillion cases of Circle?
I especially struggle with what I thought would be the allocating cases, i.e. when we have `Ref(1.)` things. The code seems to implement a lot for that, however, even without mutation I can not even get exp and log to work on Refs?

kellertuer:
One alternative would also be, since we have test coverage here, to maybe add one test somewhere for flat/sharp – then merge and register this, so it can be used by others and continue with the rework on a new PR. What do you think?

mateuszbaran:
I think we have to stick to either plain numbers and 1-element array (`fill(number)`) representation.

kellertuer:
I am also fine to switch to 0-size arrays instead of refs. We did that in LieGroups as well I think.

And sure for the PR and how to continue lets check for that tomorrow then

mateuszbaran (approved):

PR #849, opened 2025-12-12
Port more metric functions to the new style

mateuszbaran:
With this only a few functions remain to be ported (flat, sharp, change_metric, change_representer, inner).

To discuss: we should probably replace Eigen.jl with a more modern alternative; Tullio.jl seems to be the best choice but I'm open to suggestions.


kellertuer (commented):
Nice, and sure the switch in dependence seems reasonable.

PR #854, opened 2026-01-24
Make Hyperrectangle a component manifold with r-norm and r-distance

mateuszbaran:
I need r-norms for a stopping criterion in L-BFGS-B.


kellertuer (approved):
Ok.

PR #857, opened 2026-02-23
Improve GPU compatibility of UnitaryMatrices rand! and QR retraction …

mateuszbaran:
…methods

A more simple solution of some issues raised in #856 . This should also be slightly faster.


kellertuer (approved):
lgtm.

PR #863, opened 2026-03-11
Fix embed! on Segre

mateuszbaran:
See issue #862


kellertuer (approved):
Uh, a subtle small bug!
Thanks for also adding a test here :)

PR #867, opened 2026-03-21
Tucker manifold fixes

mateuszbaran:
Noticed while investigating https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/256 .


kellertuer (approved):
I would not have seen the bottom eltype thing,

and great that you also added defaults :)

PR #872, opened 2026-03-30
Fix bug in get_vectors on Stiefel

mateuszbaran:
Fixes #871


kellertuer (approved):
lgtm, thanks for the quick fix.

PR #878, opened 2026-04-03
Introduce Workspace for the Project.toml

kellertuer:
This closes  #874, #875, #876, and #877.
It moves to using dependabot, since that groups these PRs in the future (and works well on Manopt.jl) and to a workspace based Project.toml ecosystem.

Requires
* https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/260


mateuszbaran (approved):
LGTM

kellertuer:
Something else (besides ManifoldBase) seems to hold RecursiveArrayTools back, but  I do not see which.

kellertuer:
But we can maybe also fix that if it does then actually break somewhen.

PR #885, opened 2026-04-24
Improve performance of embed for FixedRankMatrices (fixes #884)

### LINKED ISSUE #884: Improve performance of `embed` for `FixedRankMatrices`
The current implementation of `embed(::FixedRankMatrices, p::SVDMPoint, X::UMVTangentVector)` for the `FixedRankMatrices` manifold is quite slow, compared to the inplace variant `embed!` (even factoring in the allocation of the result).
The reason is that the computation in `embed` does a lot of allocations, compared to the more optimized computation in `embed!`.
See here
https://github.com/JuliaManifolds/Manifolds.jl/blob/956a872ebc47906b2b4a8900301d9d06c5aa638f/src/manifolds/FixedRankMatrices.jl#L436

I suggest to simply allocate a result matrix and call the quicker `embed!` instead of manually computing the embedding. 

Below are some quick benchmarks.

```julia
using Manifolds
using Random
using BenchmarkTools

key = Xoshiro(42)
m, n, k = 1000, 1000, 40

M = FixedRankMatrices(m, n, k)

p = rand(key, M)
X = rand(key, M, vector_at=p)

@benchmark Y = embed(M, p, X)
@benchmark begin
    Y = Matrix{Float64}(undef, m, n)
    embed!(M, Y, p, X)
end
```
Results:
```
BenchmarkTools.Trial: 210 samples with 1 evaluation per sample.
 Range (min … max):   2.723 ms … 265.048 ms  ┊ GC (min … max):  0.00% … 98.65%
 Time  (median):      3.159 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   23.819 ms ±  58.019 ms  ┊ GC (mean ± σ):  81.50% ± 33.57%

  █▂                                                            
  ██▆▁▁▁▄▇█▅▄▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▅▄▅▆▆▅ ▅
  2.72 ms       Histogram: log(frequency) by time       223 ms <

 Memory estimate: 23.50 MiB, allocs estimate: 15.

BenchmarkTools.Trial: 1545 samples with 1 evaluation per sample.
 Range (min … max):  1.975 ms … 295.690 ms  ┊ GC (min … max):  0.00% … 98.88%
 Time  (median):     2.528 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.233 ms ±  10.459 ms  ┊ GC (mean ± σ):  20.59% ± 12.35%

  ▂▅▇▇▇▆█▅▂▂▂▃▃▄▅▅▃                                           ▁
  ██████████████████▇▆▅▅▄▅▅▁▄▁▄▅▅▁▅▄▄▅▆▅▆▆█▅▅▄▅▄▅▁▆▆▅▅▆▄▅▅▄▅▆ █
  1.97 ms      Histogram: log(frequency) by time      7.18 ms <

 Memory estimate: 7.93 MiB, allocs estimate: 6
```

NKlug:
See #884 for a description of the problem.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,13 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.11.21] 2025-04-03
+## [0.11.22] 2026-04-24
+
+### Changed
+
+* Improved performance of `embed` for embedding tangent vectors of the `FixedRankMaticesManifold` (#884).
+
+## [0.11.21] 2026-04-03
 
 ### Changed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* Improved performance of `embed` for embedding tangent vectors of the `FixedRankMaticesManifold` (#884).
+* Improved performance of `embed` for embedding tangent vectors of the `FixedRankMaticesManifold` (#885).
 
 ## [0.11.21] 2026-04-03
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manifolds"
 uuid = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.11.21"
+version = "0.11.22"
 
 [workspace]
 projects = ["test", "docs", "tutorials"]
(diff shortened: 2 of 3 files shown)


kellertuer:
Overall this looks good to me. Thanks for checking that!
I think that line of code is more than 7 years old, so revisiting it is a very good idea.

For allocating `Y` maybe @mateuszbaran can check whether that follows our general scheme, I think for now we never allocated with `undef`.

There is currently one test failing – we are carefully keeping a changelog.
So add a new version there and add a small sentence what was added – maybe even refer to this PR, since `(#885)` would be linked to this PR in there rendered version of the Changelog (that is put into the docs).

You can also already bump the version number accordingly in the Project.toml if you like :)

_edit:_ oh the changelog has different names here and there. I am referring to the `NEWS.md` file for this repository.

mateuszbaran (approved):
Thanks! This is completely fine, allocating with `undef` is the right way to go here.

kellertuer (approved):
LGTM as well, thanks for directly fixing our PRs (we should maybe set up a PR template somewhen).
And thanks for the quick fix after seeing the issue yourself. 

Dankeschön!

PR #887, opened 2026-04-30
Fix ManifoldsBoundaryValueDiffEqExt resolver / precompile failure

ChrisRackauckas-Claude:
## Summary

Fixes the precompile failure of `ManifoldsBoundaryValueDiffEqExt` reported in [SciML/DifferentialEquations.jl#1136](https://github.com/SciML/DifferentialEquations.jl/issues/1136).

When a user adds `OptimizationOptimJL` (or `SymbolicAnalysis`, etc.) to an env containing `Manifolds`, an older `OptimizationBase` is selected. This transitively forces `BoundaryValueDiffEqCore` down to `2.1.0`, and the resolver also picks `BoundaryValueDiffEqMIRK` / `BoundaryValueDiffEqFIRK` `1.13.0` (whose Project.toml declares only `BoundaryValueDiffEqCore = \"2\"`). Their `@compile_workload` calls `__initial_guess_on_mesh(...; tune_parameters)` — a kwarg added only in `Core` `2.2` — so precompilation errors out, and the Manifolds extension never loads:

```
✗ BoundaryValueDiffEqMIRK
✗ BoundaryValueDiffEqFIRK
✗ BoundaryValueDiffEq
✗ Manifolds → ManifoldsBoundaryValueDiffEqExt
```

## Fix

The extension only uses `MIRK4` and `BVProblem`, both available in `BoundaryValueDiffEqMIRK`. This PR:

- Switches the extension trigger weakdep from `BoundaryValueDiffEq` (the meta-package, which forces the FIRK / MIRK / Shooting / etc. precompile workloads) to `BoundaryValueDiffEqMIRK`.
- Pins compat to `BoundaryValueDiffEqMIRK = \"1.14\"`, the version where the `BoundaryValueDiffEqCore = \"2.2 - 2\"` constraint was tightened. This transitively forces `Core >= 2.2` for any environment where Manifolds is loaded, which is enough to make `MIRK`/`FIRK` `1.13`'s `tune_parameters` calls dispatch correctly even if the resolver ends up keeping those older subpackage versions.
- Updates `using BoundaryValueDiffEq` → `using BoundaryValueDiffEqMIRK` in the extension. `MIRK4` and `BVProblem` (re-exported via `SciMLBase`) are still available.
- `using BoundaryValueDiffEq` from user code still triggers the extension, because BVDE pulls `MIRK` in transitively.
- Bumps version to 0.11.23 and adds a NEWS entry.

## Test plan

- [x] Reproduced the precompile failure on master with the exact MWE from #1136 (Julia 1.12.6, `DifferentialEquations 7.17.0`, `Optimization 5.4.0`, `OptimizationOptimJL 0.4.8`, `Manifolds 0.11.22`).
- [x] With this PR, an env (`BoundaryValueDiffEq`, `Manifolds`, `OptimizationOptimJL`) resolves to `Core 2.3.1` / `MIRK 1.14.0` / `FIRK 1.14.0` / `OptimizationOptimJL 0.4.13` / `OptimizationBase 5.1.1` and precompiles cleanly. `Base.get_extension(Manifolds, :ManifoldsBoundaryValueDiffEqExt)` returns the loaded extension.
- [x] `solve_chart_log_bvp` and `estimate_distance_from_bvp` on `EmbeddedTorus` still produce sensible answers.
- [x] `test/manifolds-old/embedded_torus.jl` passes (44/44).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kellertuer:
One small problem that I have here, is that Optim and Manifolds in general do not play well together, since Optim shipps their own idea of how to do manifolds?
But that is maybe also more of a philosophical question.

For the restriction here / switch to another package to be extended I can not say much.

kellertuer:
Would your last commit make this PR be a replacement for #880 then? That PR was waiting for all the changes in your ecosystem to be finished.

ChrisRackauckas:
> Would your last commit make this PR be a replacement for https://github.com/JuliaManifolds/Manifolds.jl/pull/880 then? That PR was waiting for all the changes in your ecosystem to be finished.

Yes, this will do the ecosystem bump as well. It looks like it's possible to make it stay backwards compatible without much effort. 

> One small problem that I have here, is that Optim and Manifolds in general do not play well together, since Optim shipps their own idea of how to do manifolds?

This at least doesn't add any more Optim than there already is.

kellertuer:
Concerning Optim – yeah that is merely a philosophical question, we had an open issue about that but never found anyone who wanted to move Optim to use Manifolds.jl instead – so they are basically (still) incompatible.

The last commit is probably good to check that everything is working per se, but I would prefer that the bug is fixed as opposed to us testing less. This PR should hence wait for the bug to be fixed.

mateuszbaran (approved):
LGTM, thanks for the fixes 👍

ChrisRackauckas:
> The last commit is probably good to check that everything is working per se, but I would prefer that the bug is fixed as opposed to us testing less. This PR should hence wait for the bug to be fixed.

The AI went a bit rogue disabling adaptivity on that test 😅 . It should probably be enabled. I am still fixing upstream though, so I'll fix that then come back here to re-enable it.

mateuszbaran:
Regarding Optim.jl, there is no inherent incompatibility between it and JuliaManifolds -- both can be used at the same time. The only issue is that Optim.jl has its own concept of a manifold, represented by different types than we use in JuliaManifolds. I think there is not much benefit in working on better interoperability because Manopt.jl already nearly matches what Optim.jl can do, and even surpasses it in a few areas. Even more features will come soon in Manopt.jl. Useful tricks in nonlinear optimization that don't generalize to manifolds seem to be very infrequent.

> The AI went a bit rogue disabling adaptivity on that test 😅 . It should probably be enabled. I am still fixing upstream though, so I'll fix that then come back here to re-enable it.

OK, then I'll wait with registering a new version :+1:

PR #893, opened 2026-05-05
Add projection transport for Segre

sChoiKr:
This PR implements `vector_transport_to_project!` for `Segre{ℝ,V}`.

This is my first code contribution to Manifolds.jl, so feedback on style, documentation, and test placement is very welcome.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.11.24] 2026-05-07
 ### Added
-`vector_transport_to(::Segre, p, X, q, ::ProjectionTransport)` for Segre Manifold
+`vector_transport_to(::Segre, p, X, q, ::ProjectionTransport)` for Segre Manifold (#893)
 
 ## [0.11.23] 2026-04-29
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,10 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.24] 2026-05-07
+### Added
+`vector_transport_to(::Segre, p, X, q, ::ProjectionTransport)` for Segre Manifold
+
 ## [0.11.23] 2026-04-29
 
 ### Changed
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,8 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.11.24] 2026-05-07
+## [0.11.24] 2026-05-19
+
 ### Added
 `vector_transport_to(::Segre, p, X, q, ::ProjectionTransport)` for Segre Manifold (#893)
 
--- a/src/manifolds/Segre.jl
+++ b/src/manifolds/Segre.jl
@@ -170,16 +170,12 @@ function ManifoldsBase.vector_transport_to_project!(M::Segre{ℝ, V}, Y, p, X, q
     d = length(V)
     λ = q[1][1]
 
-    # Initialize output tangent vector Y at q.
     for Yi in Y
         fill!(Yi, zero(eltype(Yi)))
     end
 
-    # Reusable storage for dot products <z_k, x_k>.
     dots = Vector{typeof(dot(p[2], q[2]))}(undef, d)
 
-    # term = 0 corresponds to ν y₁ ⊗ ⋯ ⊗ y_d.
-    # term = k corresponds to λ_p y₁ ⊗ ⋯ ⊗ u_k ⊗ ⋯ ⊗ y_d.
     for term in 0:d
         c = term == 0 ? X[1][1] : p[1][1]
 
--- a/src/manifolds/Segre.jl
+++ b/src/manifolds/Segre.jl
@@ -168,64 +168,53 @@ the embedded tangent vector ``D\Phi_p[X]`` onto the tangent space of `M` at `q`.
 
 Let
 ````math
-p = (\mu, y_1,\ldots,y_d), 
-\qquad 
-X = (\nu,u_1,\ldots,u_d) \in T_pM,
+p = (μ, y_1,…,y_d),
+\qquad
+X = (ν,u_1,…,u_d) ∈ T_pM,
 ````
-and let ``q = (\lambda,x_1,\ldots,x_d)``.
+and let ``q = (λ,x_1,…,x_d)``.
 
 Under the Segre parametrization
 ````math
-\Phi(\lambda,x_1,\ldots,x_d) 
-= 
-\lambda\, x_1\otimes\cdots\otimes x_d,
+Φ(λ,x_1,…,x_d)
+=
+λ\, x_1 ⊗⋯⊗ x_d,
 ````
 
 the embedded tangent vector at ``p`` is
 ````math
-D\Phi_p[X]
+DΦ_p[X]
 =
-\nu\, y_1\otimes\cdots\otimes y_d
+ν\, y_1 ⊗⋯⊗ y_d
 +
-\mu
-\sum_{k=1}^d
-y_1\otimes\cdots\otimes u_k\otimes\cdots\otimes y_d .
+μ \sum_{k=1}^d y_1 ⊗⋯⊗ u_k ⊗⋯⊗ y_d .
 ````
 
 Projection vector transport from ``T_pM`` to ``T_qM`` is given by
 ````math
-Y = \Pi_{T_qM}\bigl(D\Phi_p[X]\bigr),
+Y = Π_{T_qM}\bigl(DΦ_p[X]\bigr),
 ````
 
 where
 ````math
-Y = (\dot{\lambda},v_1,\ldots,v_d)\in T_qM.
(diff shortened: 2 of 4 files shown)


src/manifolds/Segre.jl
```diff
@@ -159,6 +159,56 @@ function embed!(::Segre{𝔽, V}, u, p, X) where {𝔽, V}
...
 end
 
+
+@doc raw"""
+    vector_transport_to_project!(M::Segre{ℝ, V}, Y, p, X, q)
+
+Compute projection vector transport on the [`Segre`](@ref) manifold by projecting
+the embedded tangent vector ``D\Phi_p[X]`` onto the tangent space of `M` at `q`.
+"""
+function ManifoldsBase.vector_transport_to_project!(M::Segre{ℝ, V}, Y, p, X, q) where {V}
```
kellertuer:
```suggestion
@doc raw"""
    vector_transport_to(M::Segre{ℝ, V}, Y, p, X, q, ::ProjectionTransport)

Compute projection vector transport on the [`Segre`](@ref) manifold by projecting
the embedded tangent vector ``D\Phi_p[X]`` onto the tangent space of `M` at `q`.
"""
vector_transport_to(M::Segre{ℝ, V}, Y, p, X, q, ::ProjectionTransport) where {V}

function vector_transport_to_project!(M::Segre{ℝ, V}, Y, p, X, q) where {V}
```
* You do not have to prefix `ManifoldsBase` here
* the function `vector_transport_to_project!` is internal, so documenting it is not so useful, better to do the one I propose in the change above (the first block “attaches” the docs to the right function without implementing that one)
* it would be nice to provide a formula here from the derivations you and Simon took ,that way a user of the function can also directly read about what is computed in the docs. We usually try to document every function also with its mathematical formula. Use

````
```math
\frac{1}{2}
```
````

in your doc string. Since it is already `raw` you can directly write Markdown/LaTeX

**Comment by kellertuer** [comment]
Thanks! This is a very good start. Here is a few comments:

1. We always write a small entry in our Changelog for every PR (that is the first Test failing). That is the file https://github.com/JuliaManifolds/Manifolds.jl/blob/master/NEWS.md . You want to start a new “section” for a version 0.11.24 with an `Added` subsection, where you describe in one line what you added. We usually also refer to the PR where this is added (just att `(#893)` for your case at the end of the line ✅
2. We format our code with Runic, see https://github.com/fredrikekre/Runic.jl. Surprisingly, you directly hit that style in your commit here – great! You could reduce the code a bit by removing some of the empty lines inside the new function ✅
3. Write tests. Those are CI 2&3 failing (0% of your added lines are covered, for the whole package coverage goes down). But we already discussed on the issue where to add those tests. ✅
4. Once the documentation is complete I have to check here locally that all looks good. This PR does not “preview” docs, since that is only done when one of us developers opens PRs (that is a security measure). So let me know once the docs are finished so I can take a look and maybe fix a few typos. ✅
5. Since you write something for a new release, we can also agree to make this the new version overall. For that, change the version number in the `Project.toml` (https://github.com/JuliaManifolds/Manifolds.jl/blob/master/Project.toml line 4) in the main folder as well. ✅

File: src/manifolds/Segre.jl (hunk 1 above)
Comment (sChoiKr) [reply src/manifolds/Segre.jl]: Thank you for your detailed guidelines. I got the first two! 

> it would be nice to provide a formula here from the derivations you and Simon took ,that way a user of the function can also directly read about what is computed in the docs. We usually try to document every function also with its mathematical formula.
I will write the general formula (X_i, where i = 1 ... d) with its proper form. :) Thanks.

**Comment by sChoiKr** [comment]
OK, @kellertuer I think I passed all the checkpoints now? Could you please double-check, make sure if everything I made a change is correct? :)

**Comment by kellertuer** [comment]
I changed all reasonable places to UTF8 (for example \lambda to λ) in the docs, so that the docs get more readable also on terminal

The one thing I found a bit strange is the `stackrel{+}=` which gets just rendered to `+=`. This is common in coding but not in math. Can we change that to be just a `=` somehow?

**Comment by sChoiKr** [comment]
> I changed all reasonable places to UTF8 (for example \lambda to λ) in the docs, so that the docs get more readable also on terminal
> 
> The one thing I found a bit strange is the `stackrel{+}=` which gets just rendered to `+=`. This is common in coding but not in math. Can we change that to be just a `=` somehow?

Hi, thank you for your guidance. I still need to get used to the math format, tried to fix them but need reviews from you.

**Comment by kellertuer** [comment]
Hm, but now I am more lost on the math and the notation you use.

Formerly there was a `\dot{\lambda}` – maybe a tangent vector with the dot? I have no idea where you have the notation from hence, me asking. Now it has a further Delta upfront? What does that mean? And why is now the + just gone?

Or in short: Mathematically I did not understand what the += meant, if your `a += b` meant `a = a+b` then mathematically that is still not nice to write.

From the new notation I have sadly even less idea what you are trying to write there :(

**Comment by sChoiKr** [comment]
> Hm, but now I am more lost on the math and the notation you use.
> 
> Formerly there was a `\dot{\lambda}` – maybe a tangent vector with the dot? I have no idea where you have the notation from hence, me asking. Now it has a further Delta upfront? What does that mean? And why is now the + just gone?
> 
> Or in short: Mathematically I did not understand what the += meant, if your `a += b` meant `a = a+b` then mathematically that is still not nice to write.

Oh maybe I created the confusion. a <- a + b : meaning "add this contribution to the current value of a" It is not mathematical, it was borrowed from programming notation. So mathematically, the clean replacement is not \dot{\lambda}.

**Comment by sChoiKr** [comment]
IMG_3240
This was what the function was coded.

**Comment by sChoiKr** [comment]
Thus, I should use the summand. I will modify it.

**Comment by kellertuer** [comment]
Hm, if you would TeX exactly that it would be fine with me. 

Note that your handwritten part has _neither_ some `+=` (which confused me first) nor any Delta (which I currently do not understand what they refer to). So just that note should be fine, yes.

**Review by kellertuer — APPROVED** [review]
It looks good to me now.

File: NEWS.md [hunk 2]
```diff
@@ -5,6 +5,10 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.24] 2026-05-07
+### Added
```
mateuszbaran:
```suggestion
## [0.11.24] 2026-05-19

### Added
```

sChoiKr: Thank you for reviewing!

mateuszbaran: Also thank you for a nice contribution!

src/manifolds/Segre.jl
```diff
@@ -159,6 +159,110 @@ function embed!(::Segre{𝔽, V}, u, p, X) where {𝔽, V}
...
+λ\, x_1 ⊗⋯⊗ x_d,
+````
+
+the embedded tangent vector at ``p`` is
+````math
+DΦ_p[X]
+=
+ν\, y_1 ⊗⋯⊗ y_d
++
+μ \sum_{k=1}^d y_1 ⊗⋯⊗ u_k ⊗⋯⊗ y_d .
```
mateuszbaran, line 190:
```suggestion
μ \sum_{k=1}^d y_1 ⊗⋯⊗ y_k ⊗⋯⊗ y_d .
```
I guess that was supposed to be $y_k$? If not, what's $u_k$?

mateuszbaran: Ah, sorry, I see now -- it's fine.

sChoiKr: No, that should be u_k since we fix all other modes and a tangent vector k-th component, differentiate only the k_th mode (u_k). See my handwritten post above (the first line)?

mateuszbaran: Now I see, thanks!

src/manifolds/Segre.jl
```diff
@@ -159,6 +159,110 @@ function embed!(::Segre{𝔽, V}, u, p, X) where {𝔽, V}
...
+= \sum_{r=0}^{d} c_r \prod_{j=1}^{d} ⟨ z_{r,j},\, x_j ⟩
+````
+
+and, for each ``k=1,…,d``,
+````math
+v_k
+= 
+\sum_{r=0}^{d} \frac{c_r}{\lambda}
+\left( ∏_{j≠k} ⟨ z_j, x_j ⟩ \right)
+\left( z_k - ⟨ z_k, x_k⟩ x_k \right).
```
mateuszbaran:
```suggestion
\left( ∏_{j≠k} ⟨ z_{r,j}, x_j ⟩ \right)
\left( z_{r,k} - ⟨ z_{r,k}, x_k⟩ x_k \right).
```
Maybe let's keep `r` for clarity here?

mateuszbaran (approved):
Everything else looks fine, thanks again!

src/manifolds/Segre.jl
sChoiKr: Yes, I found that missing index just now too. Thank you!

kellertuer:
@sChoiKr we discussed and committed Mateusz suggestions to the code (see his commit).

The usual workflow as described [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) would be that the PR creator would do that – after a discussion and agreeing what to do of course.

One can also discuss and decide to do something different, do a commit, and close the discussion without “commit suggestions”.

And no worry, this is just for your information, and that we did commit that is hopefully fine for the to changes, we only did it to put you in “Easy mode” on PRs, since this is your first one.
I know in the beginning all the workflow with PRs and such might be a bit overwhelming.

sChoiKr:
I thank all of you for the support. It was a great experience for me to be a part of it. Next time, I will follow the workflow as guided. Hopefully, I will be faster next time. :)

kellertuer:
You're welcome and thanks for the PR;

and speed is never the quality measure here, (code) quality is :)

sjacobsson:
Nice!

PR #895, opened 2026-05-19
Allocating Levi-Civita, Christoffel symbol bugfix

hajg-ijk:
This PR adds an allocating variant of `levi_civita_affine_connection!`, i.e. `levi_civita_affine_connection`.
Furthermore, an argument name for `christoffel_symbols_second!` was changed from `i` to `index`. 
The previous `i` overlapped with a for loop index, also named `i`, later in the function, thus resulting in incorrect behavior.

On this note, I'm not entirely sure the `levi_civita_affine_connection!` does exactly what it's supposed to, in that this function seems to only return the "sum" of
```math 
\Gamma^l_{jk} = \tfrac{1}{2} g^{lr} \left( \partial _k g_{rj} + \partial _j g_{rk} - \partial _r g_{jk} \right)
```
evaluated on the coordinates of the input tangent vectors `Xc` and `Yc`, as opposed to
```math
 (\nabla_j Y)^l = \partial_jY^l + \Gamma^l_{jk} Y^k.
 ```
Maybe there is something I'm missing out.


mateuszbaran:
Thanks for the bugfix!

> On this note, I'm not entirely sure the levi_civita_affine_connection! does exactly what it's supposed to, in that this function seems to only return the "sum" of

Maybe the naming is slightly unfortunate but this is intended, as we have no generic access to  $\partial _j Y^l$ part -- that needs to be handled by specific algorithms that use this function.

hajg-ijk:
> Maybe the naming is slightly unfortunate but this is intended, as we have no generic access to ∂ j Y l part -- that needs to be handled by specific algorithms that use this function.

I see. I implemented a similar function that does what I mentioned, and that works with some AD for the partial derivatives of Y.
The current signature is `M::AbstractManifold, A::AbstractAtlas, chart_index, p, X, Y; backend=AutoFiniteDifferences(central_fdm(5, 1))`, where `X, Y` are vector fields, so functions that take `(M, p)` and return a tangent vector to `M` at `p`.
Do you think this would fit somewhere here?

mateuszbaran:
> Do you think this would fit somewhere here?

Good question. We generally don't have an API for vector fields right now and if we were to add that, I'd like to have them with a larger batch of features, such as their linear algebra or Lie derivatives and a tutorial. If you want to work on that then I think it would fit but just having the connection wouldn't allow us to properly settle the design and demonstrate use cases.

mateuszbaran:
By the way, could you add a test for the new function in `embedded_torus.jl`?

hajg-ijk:
> Good question. We generally don't have an API for vector fields right now and if we were to add that, I'd like to have them with a larger batch of features, such as their linear algebra or Lie derivatives and a tutorial. If you want to work on that then I think it would fit but just having the connection wouldn't allow us to properly settle the design and demonstrate use cases.

Fair enough. I might look into doing that eventually, just not something I currently have time for.

kellertuer:
I think a vector field API and a few related functions would really be great to have (also using different symbols than single vectors X,Y for them).

I got that question now a few times (over several years though). So I would support work on that and help when I can :)

mateuszbaran (approved):

mateuszbaran:
Yes, a few people have already expressed interest in vector field support, and also more general symbolic computation capabilities as they are more mainstream in differential geometry than AD. I've looked into it a bit but unfortunately Julia seems to have very limited symbolic capabilities, so we'd have to rely on SymPy which already has some differential geometry tools but worse Julia interoperability. So I haven't pursued it any further.

I have some plans for adding ODE-based Jacobi field computation this summer, a few more atlases and probably also generalizing the embedded torus example a little.

PR #897, opened 2026-05-22
Update the function of vector_transport_to for Segre

sChoiKr:
The implementation avoids explicitly forming the ambient tensor and precomputing the inner products to reduce the computational cost.


sChoiKr:
I accidently created PR with the previous branch. I will close it and redo a new PR.

PR #899, opened 2026-05-27
Fix two small typos.

kellertuer:
Fix 2 small typos I noticed in the docs.


kellertuer:
I also added a small update to ignore patch updates from typos, we still get minor and major versions of course.
Having Typos is great but they seem to put out a lot of patch releases.

Is this ok @mateuszbaran ? I already did the same on Manopt.

mateuszbaran:
OK, it's really a lot of these patches lately so it makes sense.

mateuszbaran (approved):

PR #901, opened 2026-05-28
Move PositiveNumbers tests to new suite

sdraeger:
## Summary

- Move the PositiveNumbers tests from `test/manifolds-old` to `test/manifolds`.
- Replace the old `Manifolds.test_manifold` calls with `Manifolds.Test.test_manifold` coverage for `PositiveNumbers()` and `PositiveVectors(2)`.
- Keep the explicit constructor, representation, metric conversion, Hessian, volume, random, and field-parameter checks from the old file.

Part of #870.

## Validation

- `julia --project=test test/manifolds/test_positive_numbers.jl`
- `MANIFOLDS_TEST_SET=manifolds julia --project=test test/runtests.jl` was attempted twice and timed out after 20 minutes in unrelated old `symplectic` tests; the run reached and passed `manifolds/test_positive_numbers.jl` before timing out.


kellertuer:
Thats a nice start. Thanks!

Several of the functions in the first block are also available in the new test suite and do no longer have to be tested separately.
The validation sounds a bit AI-ish (ran for 20 minutes? The whole Manifolds.jl tests on my machine runs 25 or so minutes) How much AI was involved here? Just curious and like transparency, especially on AI things.
Besides that let's wait for the test results & coverage.

Your last name sounds German, I am originally from Dräger-Town (or Marzipanien or just good old Lübeck)!

sdraeger:
Thank you! Yes, this was AI-assisted, and I reviewed the resulting diff and ran the targeted PositiveNumbers test locally. I agree the validation wording was awkward and will change it accordingly.

I’ll also trim the first block to keep only the constructor/extra behavior checks that are not already covered by `Manifolds.Test.test_manifold`.

kellertuer:
Thanks for the transparency.
I was mainly curious, because by now I often spent time fixing other peoples AI slop (both in my daily job and here in open source software. That does not mean, your code is bad, as I wrote it is a very good start.

Let me know when you have rechecked it, I can help looking through the rest as well. But for example `repr` can be tested in the suite, see https://juliamanifolds.github.io/Manifolds.jl/stable/misc/testsuite/#Manifolds.Test.test_repr, so line 7 can be moved; provide the expectation (the second dict) and it also checks that the string is right, cf https://github.com/JuliaManifolds/Manifolds.jl/blob/5bb7aa22befc376cacfdd7d719d2201c4f2203fe/test/manifolds/test_sphere.jl#L57

The reason to really only check edge cases outside that function is, that in a next step, “filtering” tests gets easier – one idea is to have a short integration test for example.

Also for the functions, we are using runic for code formatting, so feel free to group functions reasonably together on lines.
Good that you kept the structure with the `using` line upfront, that way a single manifold test can be run individually.

Thanks for working on this, I know it is a bit dull (and hence maybe even suited for a bit of AI), but a good thing to slowly revise for a bit faster and flexible CI.

PR #904, opened 2026-06-05
An ONB for Tangent spaces on Grassmann in Stiefel Representation.

kellertuer:
This PR introduces an ONB for Grassmann in the Stiefel representation and makes the doc string of the manifold more precise. The old statement of the tangent space was imprecise/wrong.

### DIFF
--- a/src/manifolds/GrassmannStiefel.jl
+++ b/src/manifolds/GrassmannStiefel.jl
@@ -100,7 +100,7 @@ end
 
 Given a point `p` on the [`Grassmann`](@ref) manifold `M` in Stiefel representation,
 i.e. ``p ∈ ℝ^{n×k}`` compute the coordinates ``c ∈ ℝ^{k(n-k)}`` representing the
-tangent vector `X`.
+tangent vector `X` with respect to the [`DefaultOrthonormalBasis`](@extref `ManifoldsBase.DefaultOrthonormalBasis`).
 
 The tangent space is characterized by ``p^{\mathrm{T}}X = 0_{k×k}``,
 where ``0_{k×k}`` is the ``k×k`` matrix containing zeros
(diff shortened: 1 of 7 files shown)


kellertuer:
Oh, Ambiguities was not what I expected to fail. We will have to check what is happening there, maybe with `StiefelPoint` or so? I would have thought they just “pass to array”.

Even more interestingly: On my machine, ambiguities are fine...

_edit:_ Found it. It only is more ambiguities on LTS, and these will not be hit. The reason is as follows:
We define the forwards for StiefelPoint/StiefelTangentVector for `Grassmann` and the new ONB is defined for `Grassmann{ℝ}`, so this is not even a problem on 1.11 or 1.12; but on 1.10 this raises 4 ambiguities.
In practice these will not be hit, since even _with_ the types points/vectors the field will always be given.

So I raised that ambiguity bound slightly for 1.10.

src/manifolds/GrassmannStiefel.jl
```diff
@@ -94,6 +94,50 @@ function exp!(M::Grassmann, q, p, X)
...
+get_coordinates(::Grassmann{ℝ}, p, c, ::DefaultOrthonormalBasis)
+
+function get_coordinates_orthonormal!(M::Grassmann{ℝ}, c, p, X, ::RealNumbers)
+    n, k = get_parameter(M.size)
+    V = nullspace(p') # from SVD, so we have (n-k) ON columns in R^n
+    for i in 1:k
+        # The sum above in the docs is a MV product in blocks
+        # c[1:(n-k)], c[(n-k)+1:2*(n-k)],...with V
+        c[((i - 1) * (n - k) + 1):(i * (n - k))] .= V \ X[:, i]
+    end
```
mateuszbaran:
```suggestion
    c .= vec(V \ X)
```
This should work the same up to ordering of components. It solves all the linear systems in one call.

src/manifolds/GrassmannStiefel.jl
```diff
@@ -113,6 +157,44 @@ function ManifoldsBase.get_forwarding_type(::Stiefel, f, ::Type{<:StiefelPoint})
...
+"""
+get_vector(::Grassmann{ℝ}, p, c, ::DefaultOrthonormalBasis)
+
+function get_vector_orthonormal!(M::Grassmann{ℝ}, X, p, c, ::RealNumbers)
+    n, k = get_parameter(M.size)
+    V = nullspace(p') # from SVD, so we have (n-k) ON columns in R^n
+    for i in 1:k
+        # The sum above in the docs is a MV product in blocks c[1:(n-k)], c[(n-k)+1:2*(n-k)],...with V
+        X[:, i] .= V * c[((i - 1) * (n - k) + 1):(i * (n - k))]
+    end
```
mateuszbaran:
```suggestion
    mul!(X, V, reshape(c, n - k, k))
```
This matches the edit for `get_coordinates`.

mateuszbaran (commented):
Thanks, this looks nearly fine. I just have one small suggestion that should make this faster.

src/manifolds/GrassmannStiefel.jl
kellertuer: Oh that look neat!

mateuszbaran (approved):

PR #906, opened 2026-06-14
optimize unitary exponential map

araujoms:
The exponential map of the unitary manifold was not exploiting the fact that the gradient is anti-Hermitian, which makes it slower for larger dimensions. Funnily enough, my benchmarks showed that the transition point where the overhead is worth it is around dimension 42.

I also tested it for the real case, and my benchmarks showed that it is never worth it, so I only touched the complex one.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,9 +5,15 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.28] unreleased
+
+### Changed
+
+* the exponential map for complex `UnitaryMatrices` to use an optimized algorithm for large matrices (#906)
+
 ## [0.11.27] 2026-06-06
 
-## Added
+### Added
 
 * a `DefaultOrthonormalBasis` for the `Grassmann` manifold in Stiefel representation. (#904)
 * improved documentation of the tangent space.
--- a/src/manifolds/GeneralUnitaryMatrices.jl
+++ b/src/manifolds/GeneralUnitaryMatrices.jl
@@ -213,18 +213,18 @@ Compute the exponential map, that is, since ``X`` is represented in the Lie alge
 \exp_p(X) = p\mathrm{e}^X
 ```
 
-For different sizes, like ``n=2,3,4``, there are specialized implementations.
+In the real case there are specialized implementations for ``n=2,3,4``. The algorithm used is a more numerically stable form of those proposed in [GallierXu:2002](@cite) and [AndricaRohan:2013](@cite).
 
-The algorithm used is a more numerically stable form of those proposed in
-[GallierXu:2002](@cite) and [AndricaRohan:2013](@cite).
+In the complex case an eigendecomposition of `im * X` is used for `n ≥ 42`, as benchmarks indicate that it is advantageous for these dimensions.
+
 """
 exp(::GeneralUnitaryMatrices, p, X)
 
 function exp!(M::GeneralUnitaryMatrices, q, p, X)
     return copyto!(M, q, p * exp(X))
 end
-function exp!(M::GeneralUnitaryMatrices{ℂ}, q, p, X)
-    if size(X, 1) < 42
+function exp!(M::GeneralUnitaryMatrices{ℂ, TypeParameter{Tuple{n}}}, q, p, X) where {n}
+    if n < 42
         return copyto!(M, q, p * exp(X))
     else
         return copyto!(M, q, p * _anti_exp(X))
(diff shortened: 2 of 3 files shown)


kellertuer:
Thanks, this is a quite interesting observation. Here are a few comments.

* we have a static way to dispatch on the dimension of the manifold (or equivalently input size of the matrices) see https://juliamanifolds.github.io/ManifoldsBase.jl/stable/types/#type-parameter
* until now we did not yet use that as a barrier like yours, but we could try that for sure
* the new implementation does not yet have test coverage 
* The barrier of 42 should somewhere be documented in the doc string, maybe as an admonition / technical detail – or that we even write two doc strings for the two dispatch cases implemented 

For me the last point is an important one – ok I am also the one annoying everyone about the docs and testing ;)  – but if I would look at the code in about a year or two, I would wonder a lot why it was 42 and what this distinction does.

araujoms:
I did see that, but I couldn't figure out how to dispatch on a size range, only how to dispatch on specific sizes. Also, would there really be any advantage in static dispatch here?

Of course you're right about the tests and documentation, I'll handle it shortly.

araujoms:
Nevermind, figured it out now.

kellertuer:
Test coverage looks fine, but also here we need a changelog entry, just that here it is the NEWS.md file (just to confuse everyone we have different names for that file in different repositories...joke aside, they have mainly different names for historical reasons).

araujoms:
No problem, I've dealt with worse legacy code ;)

mateuszbaran:
Could you share your benchmarking script? I get consistently worse performance with this new variant. For n=50 I get 161.848 μs for `_anti_exp` vs 128.794 μs for plain `exp`, while for n=100 it's respectively 1.196 ms and 739.986 μs.

I just did a simple script like this:
```julia
using BenchmarkTools, Manifolds, LinearAlgebra
M = UnitaryMatrices(100)
p = rand(M)
X = rand(M; vector_at=p)
@btime exp($X)
@btime _anti_exp($X)
```

It's not that surprising to me since `exp` is just doing basic arithmetic + LU decomposition, which is cheaper than eigendecomposition.

araujoms:
That's what I used:
```julia
julia> using BenchmarkTools, LinearAlgebra

julia> function _anti_exp(X)
           λ, U = eigen(Hermitian(im * X))
           D = Diagonal(exp.(-im .* λ))
           return U * D * U'
       end
_anti_exp (generic function with 1 method)

julia> @btime exp(A) setup = (n = 50; B = randn(ComplexF64, n, n); A = B - B');
  967.826 μs (24 allocations: 236.11 KiB)

julia> @btime _anti_exp(A) setup = (n = 50; B = randn(ComplexF64, n, n); A = B - B');
  328.303 μs (33 allocations: 237.47 KiB)

julia> @btime exp(A) setup = (n = 100; B = randn(ComplexF64, n, n); A = B - B');
  2.442 ms (24 allocations: 940.11 KiB)

julia> @btime _anti_exp(A) setup = (n = 100; B = randn(ComplexF64, n, n); A = B - B');
  2.253 ms (33 allocations: 864.28 KiB)
```
With a bit more testing the advantage seems to be rather noisy, though. This is visible if we use `@elapsed` instead of `@btime`:
```julia
julia> function texp(n)
           B = randn(ComplexF64, n, n)
           A = B - B'
           return @elapsed exp(A)
       end
texp (generic function with 1 method)

julia> function taexp(n)
           B = randn(ComplexF64, n, n)
           A = B - B'
           return @elapsed _anti_exp(A)
       end
taexp (generic function with 1 method)

julia> n = 2:100
2:100

julia> ta = taexp.(n);

julia> ts = texp.(n);

julia> using Plots

julia> plot(n,[ta, ts])
```
The resulting graph is 
exptime

Of course, if the speedup is not robust the added complexity is not worth it.

mateuszbaran:
These are my results with your script:
```
julia> @btime exp(A) setup = (n = 50; B = randn(ComplexF64, n, n); A = B - B');

  173.872 μs (24 allocations: 236.11 KiB)

julia> @btime _anti_exp(A) setup = (n = 50; B = randn(ComplexF64, n, n); A = B - B');
  175.587 μs (33 allocations: 237.47 KiB)

julia> @btime exp(A) setup = (n = 100; B = randn(ComplexF64, n, n); A = B - B');
  1.005 ms (24 allocations: 940.11 KiB)

julia> @btime _anti_exp(A) setup = (n = 100; B = randn(ComplexF64, n, n); A = B - B');
  1.293 ms (33 allocations: 864.28 KiB)
```
I think it would be more productive to port for example this algorithm to Julia: https://www.gicas.uji.es/Research/ExpSkewHermitian.html to have faster `exp`.

araujoms:
Unfortunately it's not worth it then. Nevertheless I'd be rather curious to see how the timing graph looks in your machine.

mateuszbaran:
Sure, that's my graph:
image

kellertuer:
Thanks still for the idea and the discussion! It is always good to reconsider some ideas and check :)

PR #907, opened 2026-06-15
Add orthonormal coordinates on UnitaryMatrices and extend change_metric/change_representer to all general unitary matrices

mateuszbaran:
Addresses missing features mentioned in https://github.com/JuliaManifolds/ManifoldDiff.jl/issues/88 (cc @araujoms)

### DIFF
--- a/.github/workflows/documenter.yml
+++ b/.github/workflows/documenter.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: quarto-dev/quarto-actions/setup@v2
         with:
-          version: "1.7.29"
+          version: "1.9.38"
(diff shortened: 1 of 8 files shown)


kellertuer:
Ih, I would have thought, we either have an “isometric embedding” trait that handles this automatically or a check against the default metric (which here would be Euclidean) to handle this so a specific implementation of those two methods would not have been necessary. But maybe I also remember that wrongly (or from an earlier state where we had more involved traits ;))

mateuszbaran:
I think it's currently not covered by the traits and I generally don't mind a bit of boilerplate code like here. We used to overuse traits.

kellertuer (approved):

PR #909, opened 2026-06-22
Add Tucker projection transport

NKlug:
During one of my projects I implemented an efficient projection transport for the Tucker manifold and thought I share the code in this PR.

For the implementation I needed a way to efficiently compute contractions for tensors of arbitrary dimensions, such as for example for `D = 3`
```
result[k, r2, r3] = core[r, r2, r3] * F[k, r]
result[r1, k, r3] = core[r1, r, r3] * F[k, r]
result[r1, r2, k] = core[r1, r2, r] * F[k, r]
```
For this I used Tullio. As far as I know, Tullio does not provide a way to implement such contractions for arbitrary dimensions. Following the maintainer's advice in https://github.com/mcabbott/Tullio.jl/issues/190#issuecomment-3134605597 I sort of hard-coded these functions for dimensions 2, ..., 16, see the code in Tucker.jl lines 1049ff.
This is, of course, not particularly nice, however, it works very efficiently. Also, computations with Tucker tensors of order higher than 16 are maybe unrealistic anyways (?). The number 16 could of course also be increased by changing a single line of code. What do you think about this?

Maybe you also have another idea how to perform these tensor contractions efficiently without so much hard-coding?
I would appreciate your input and also other suggestions for improvements :)

I also did some very basic unit testing against `project!(..., embed(...))`, but not within the test framework of this package. Maybe it would be nice to add some unit tests here as well, but I am not quite sure how and where.


kellertuer:
Thanks for the PR – and for using your package!

Here are a few remarks

* the test failing for the plots recipes is not your fault, either rerun the test (because that error was not from us maybe it was fixed) or I take a look at that the next few days
* please also document the formula of the function you compute, for best of cases with a reference from literature
* ` vector_transport_to_project!` is an internal function, please add the documentation to the “public” function. A very detailed example is the recent one in Segre https://github.com/JuliaManifolds/Manifolds.jl/blob/87bb18eced9452a72ccbfdf183b8181c6883413c/src/manifolds/Segre.jl#L163-L223
* I do not yet understand the reason for the hard-coded functions, but make sure you have code coverage at least (you can for example use LocalCoverage to check that locally, or as soon as the tests pass here, you get it here)
* we use `Runic.jl` to have a unified code format. Make sure to run it locally before you commit/push, cf. https://github.com/fredrikekre/Runic.jl

NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.29] 
+
+### Added
+
+* `vector_transport_to_project!` for `Tucker` manifold.
```
kellertuer:
```suggestion
* `vector_transport_to_project!` for `Tucker` manifold. (#909)
```

Now that the PR exists (we have its number) mentioning it here renders it as a link both in HTML and when looking at the Changelog here on GitHub.

kellertuer:
We just fixed the CI on master, so if you merge that here, the CI should properly report e.g. coverage again.

NKlug:
Thanks for your feedback! I'll add proper documentation and some tests hopefully in the next days :)

Regarding the (somewhat) hard-coded functions, let me try to explain further:
Assume $D$ is fixed but unknown (at compile time) and $1 \leq k \leq D$. Let $C$ be an order- $D$ tensor and $F$ a matrix of compatible size.
Lets say I want to compute $C \times_k F$, i.e. the mode- $k$ contraction of C with F.
As an Einsum, this would look something like
```
result[r_1, ..., r_{k-1}, R, r_{k+1}, ..., r_D] = C[r_1, ..., r_{k-1}, K, r_{k+1}, ..., r_D] * F[K, R]
```
This needs to be done very efficiently. I currently use Tullio for this. However, Tullio unfortunately does not allow expressions like above (with "..."), but needs indices for each mode at compile time.
So the current best way that I found is to generate ("hard-code") all expressions needed as individual functions up to some fixed maximum $D$. E.g. for $D = 3$, this means 3 functions are generated.
Of course, there might be another, better way to perform such contractions that I am unaware of.

Btw, the contraction above is just an example. Other types of contractions involving order- $D$ tensors also need to be performed, such as the contraction of two order- $D$ tensors along all but one mode.

kellertuer:
Thanks for the explanations and great to see that we were able to fix the CI, so we now see which code lines are missing coverage – and great that you plan on documenting soon :)

For the benchmarks. I am not a good “performance prgrammer” that is more something for @mateuszbaran.
And sure, Tullio is not a problem, it is a depenency of this package already

I think there might even be the following tradeoff: I do not expect contributors to “hang around” and help debugging / extending their code, especially not for PhD students. That is absolutely fine to leave it be. But that also means that there are the main developers here left (Matreusz and me for this package) that have to take care of potential bugs, necessary maintenance. 
So if your code is the most clever one (and reasonably performant) it might be too clever for me. Parts of Tucker are actually, I was not careful enough there.

That leaves a tradeoff between “performant” (which you on your own can totally do!) and “usable/maintainable” (on purpose both terms in quotation marks.
So please have that in mind. We could even add proper documentation strings to some of the new functions and open an “internal functions” section on the Tucker docs page.

One example, the line
https://github.com/NKlug/Manifolds.jl/blob/8b2b0789e31ee6a9778179160cbd67ad75e25b1f/src/manifolds/Tucker.jl#L986
does help me to zero extend to understand what the function is about. Sure the next two are a bit better ;)

mateuszbaran:
I think the Tullio code is almost clear enough and perfectly maintainable for me. I've pushed a minor patch -- I personally prefer a style of macros where more things are put in quotes instead of building code piece by piece. I would only ask you to unify the indices between the code and the comment to the same notation (`n1` vs `n`, `r1` vs `j` etc).

kellertuer:
Ok, then we can keep the current code; a bit more documentation –for the internal one the comment style Nick started would also be ok– and testing is missing only then.

NKlug:
I now added further documentation and also some rudimentary unit tests against `project(..., embed(...))`. Not quite sure if the unit tests are in the right place though. Feel free to improve both if you see fit :)

src/manifolds/Tucker.jl
```diff
@@ -918,3 +1015,125 @@ function ManifoldsBase.allocate_result(M::Tucker, ::typeof(embed), p, args...)
...
+``
+q = (C_q, U_q^1, …, U_q^D)
+``
+Computes 
+``
+[core ×_{j ≠ i} factors[j] ×_i Ui (I - U_q^i (U_q^i)^T)]_{(i)} * (C_q)_{(i)}^T * Σ_inv
+``
+and adds to or overwrites `result`, depending on whether `add_to_result == true`.
+"""
+function _compute_projection_summand!(
```
kellertuer: In general this looks good, but if we document them with a doc string (which I think is a good idea for detail) we should
* not let then start with a `_`
* repeat the signature in the first line of the doc string as docs usually do
* add the in a separate block on the tucker docs page at https://github.com/JuliaManifolds/Manifolds.jl/blame/master/docs/src/manifolds/tucker.md

For the last, it is probably best to copy the autodocs block add `Public=true` and `Private=false` to the first and the other way around for the second (see an example at) and add a headline (maybe `## Helper functions`) before the second one with the private functions, see https://github.com/JuliaManifolds/ManifoldsBase.jl/blame/db44779765cc69a7c7114b16f135dbef842077f6/docs/src/vector_transports.md#L22-L28 for such an autodocs example

kellertuer: Or for the function naming to be more precise `_fname` are for me “superinternal” ones, maybe with a short comment as docs, but a user should for 100% ignore them
`fname` documented but not exported are internal helpers, where the documentation would still allow a user to maybe use it or improve it (e.g. on a GPU or so), but this is still considered more an internal helper.
`fname` that is exported should be documented super throughly and is considered the “official” API of Manfiolds.jl

and your new functions are on the second level I think. Documenting them is a good idea.

test/manifolds-old/tucker.jl
```diff
@@ -197,6 +197,50 @@ include("../header.jl")
...
+            if d == 2
+                ranks = (3, 3)
+            else
+                ranks = Tuple(2:(2 + d - 1))
+            end
+            dims = Tuple(10:-1:(10 - d + 1))
+            M = Tucker(dims, ranks)
+
+            # test transport for out-of-place update
+            p1 = rand(key, M)
```
kellertuer: Even if you _would_ set `Rand.seed!(42)` before calling `rand` (which makes the tests somewhat reproductibe), I am not a fan of testing on random points. Could you use fully deterministic tests please?

With random points, debugging in case of errors is not really fun.

test/manifolds-old/tucker.jl
```diff
@@ -197,6 +197,50 @@ include("../header.jl")
             "Tucker((4, 5, 6), (2, 3, 4), ℝ; parameter=:field)"
     end
 
+    @testset "projection transport" begin
+        key = Xoshiro(42)
+        atol = 5eps(Float64)
+        rtol = sqrt(eps(Float64))
+        approx(A, B) = isapprox(A, B, atol = atol, rtol = rtol)
```
kellertuer: This function is called 4 times. Can we maybe omit this definition and do the explicit keywords in all 4 places? That does “hide” this a bit less.

Also, we do have specific point (`isapprox(M, p, q)`) and tangent vector (`isapprox(M, p, X, Y)`) functions available. Would these maybe make more sense to be used? They of course take the same kwargs.

test/manifolds-old/tucker.jl
kellertuer, line 168: Just as a side note: This is an old test suite we slowly want to get rid of (the new one is in its own submodule https://juliamanifolds.github.io/Manifolds.jl/stable/misc/testsuite/#Manifolds.Test.test_manifold),
so doing your tests separate is a reasonable idea, unless you feel adventurous to rewrite this call to the new suite as well ;) – but you do not have to of course.

test/manifolds-old/tucker.jl
NKlug: I can change that to deterministic points. Would you prefer testing against hard-coded results then, too?
Might make sense I guess, in case some error gets introduced to `project(..., embed(...))`?

kellertuer: Yeah maybe, though it might tend to “magic numbers” than that are on the level of “the following is the result, trust me!”
Find a good balance, that is deterministic and not too many “magic numbers”. :)

NKlug: I left the test against `project(..., embed(...))` for now and used deterministic points for 4 different tensor orders (2, 3, 4, 5).
I leave it up to you to decide if this is too much, as hard-coding the points makes the source code quite bloated.

test/manifolds-old/tucker.jl
NKlug: Yeah I realized that. Unfortunately I don't have the time at the moment to dive deeper into the (new) test suite.

kellertuer: WEll, as you can see in not all manifolds “having transitioned yet” – neither do I ;)

kellertuer:
Could you either here in the web interface or locally on your machine Merge the master branch here? It is mainly a conflict in the news (on the version 11.29, where the last PR added something and your addition is already mentioned on this PR as well? 
Let me know if you feel lost on that, I can resolve it as well (on the web interface) then you just have to pull once before continuing on your machine..

test/manifolds-old/tucker.jl
```diff
@@ -197,6 +197,76 @@ include("../header.jl")
...
...19588331669149 0.4147887836081399 0.15768481896945105; 0.1835714487029855 0.10787934125539969 -0.21868978264543462; -0.07983969935320319 -0.04181921596405814 -0.5940967077660939; -0.23012151486233037 -0.19212012334444772 0.04697357478445809; -0.20930465020180836 -0.2286895352811398 0.5337840389337893; -0.5160824474389014 -0.6851196169878814 -0.14001671507653302; -0.16259241484679066 0.2114293022118128 0.03603985432665287], [0.044078516760720665 0.2259586648613519 0.28670972022369023; 0.5028027867637208 0.3801494871877342 -0.06099937895726573; 0.21408115765652025 -0.13648385234837618 -0.5424939109789019; -0.2161518078477106 0.4315763872688958 -0.049259585620541814; -0.2035024903209966 -0.3245061236706567 -0.025715954127858497; -0.5553360934958359 0.24714641544326688 -0.20914392235987955; 0.5135604694548546 0.20651774786619606 0.08001036271549553; 0.1694350896927468 -0.36289414079211696 -0.543637222238994; -0.10212761708509162 0.5087393405211225 -0.5205812941020956]), [-2.5122995873056295 -0.0 0.0; -0.0 -1.5158183833563994 0.0; -0.0 -0.0 0.7930676278052453], ([2.5122995873056304, 1.5158183833563998, 0.7930676278052451], [2.5122995873056295, 1.5158183833563994, 0.7930676278052453]))),
```
kellertuer: Hm. Here you just copied in the rand? That is basically the same as rand. Do we maybe have something a bit smaller we can test on?

Tests are not meant to demonstrate how large problems can get but to have a way to verify things work. For example I would always test on the 2-sphere, not the 1k-sphere.

_edit:_ Maybe there is best to keep in mind the following two scenarios where these tests are really needed:
I do not expect you to “stick around” (for example after your PhD) to do maintenance.
(a) But there might be someone coming along reporting either a bug in your code or propose a “speedup”. Then the tests shall make sure “everything still works as expected”
(b) we do a restructure somewhere, because we add features. Then we still want to make sure, these changes are not breaking other peoples code. To some extend, tests can check for that.

But that does mean Test have  to be a bit minimalistic in size, maybe have in mind that they can be “checked by hand on paper” for the very best of cases – or are not too far away from that.

test/manifolds-old/tucker.jl
```diff
@@ -197,6 +197,76 @@ include("../header.jl")
...
+        ]
+        Ranks = [
+            (3, 3),
+            (2, 3, 4),
+            (2, 3, 4, 5),
+            (2, 3, 4, 5, 6)
+        ]
+        Dims = [
```
kellertuer: Are these all different implementations?
Otherwise we are over-testing here a bit. In other words: Sure you can check 100 times that a retraction works correctly, but if all use the same implementation probably 2-3 checks are enough. 
The main problems with over-testing is (a) in code that tests looks more complex than they are (harder to verify if something goes wrong) and (b) at some point run time of the test suite.

kellertuer:
If you feel this is all too much work and too complicated – sorry for that, but at least I try to properly test and document everything in this package, even if it takes more effort.
We could then do a short meeting online and discuss how to best finish this PR; maybe I can help a bit along the way.

NKlug:
I agree that it's better to do things thoroughly :)
I don't have terribly much time at the moment, so realistically I can only work on this PR every couple of days or so. So if you want to move on quicker, maybe it is a good idea for you to take over parts of this. Otherwise it might just take longer (but there is no time pressure anyways I guess).

kellertuer:
No, there is absolutely no time pressure – I am happy about everyone who contributes!
There is also no one depending on or waiting for this feature, so then take your time.

My main worry might be that you feel all this takes too long :) So no rush, take your time. My offer for a short consultation online of course is still available.

test/manifolds-old/tucker.jl
NKlug: I now used deterministic and "simple" points and tangent vectors. However, the result of the projection still looks quite arbitrary, but there is no way around this without vast oversimplification. Still better than fully random points though.

NKlug:
I tried to improve the docu and simplified the tests as discussed. If you don't have any further comments this is ready for merging from my side.

kellertuer (approved):
This looks good to me. Do you want to take a look (again) as well? @mateuszbaran

kellertuer:
Do you want to have this registered as a new version? Then you have to set the version in the `Project.toml` and we should set the date in the changelog accordingly.

kellertuer:
Oh, my post could be misunderstood. What I meant is having it registered directly. We also do not have so urgent next PRs, so waiting is maybe also too long then. So maybe set the date in the Changelog then and update the Versioning the Project toml as a last commit here :)

NKlug:
For some reason the version in `Project.toml` is already 0.11.29. But I updated `NEWS.md`.

kellertuer:
Then it's also fine – I see a previous PR set that already. We do not have that completely specified when it should be increased – but since this is the last PR before we register the version good that you then checked that it is on 0.10.29 :)

PR #913, opened 2026-06-29
Remove PythonPlot from docs

mateuszbaran:
This should fix the recent errors.


kellertuer:
During first coffee I had an idea what to do, after I deep dived yesterday into recipes and GR, and then even found the old note I had written about the inconsistencies in the Plots recipes, which means that not all backends can do surfaces or wireframes.

My latest commit therefore removes the plotting example, refers do ManifoldMakie and only mentions the legacy sphere plot recipe.
In a breaking change we could maybe even remove either just the sphere recipe or even the whole extension – plotting Hyperbolc seems fine though.

mateuszbaran:
I think this is a perfectly fine solution.

kellertuer (approved):

Issue #900, opened 2026-05-27
Meaning of injectivity radii on the Sphere


sblelong:
Hi all,

The `Sphere` class includes an extension of `injectivity_radius` to the `ProjectionRetraction`, with value $\pi/2$:

https://github.com/JuliaManifolds/Manifolds.jl/blob/89289d53850da3efafca69bd7de431348575a09a/src/manifolds/Sphere.jl#L321

I initially did not understand why this value was chosen, since the projection from $T_p\mathcal{M}$ to $\mathcal{M}$ is globally injective. A kind explanation from @kellertuer made me realize that this value is to be understood as the largest geodesic distance between $p$ and $q\in\mathcal{M}$ for which `inverse_retract(M, p, q, ::ProjectionInverseRetraction)` is defined.

Since in one case it points to a radius within $T_p\mathcal{M}$ and in the other case, to a radius for the geodesic distance, I think it would be beneficial to disambiguate the use of `injectivity_radius` with inverse retractions.

After suggestion from @kellertuer, I will try to open a PR on `ManifoldsBase.jl` to define a `injectivity_radius(M, p, m::AbstractInverseRetractionMethod)`. This issue is mainly opened to be referenced in this PR, since it should resolve it.

mateuszbaran:
Hi! Thanks for looking into this. I think we weren't careful enough to properly define and use this function, though I'm not sure if we need the injectivity radius of an inverse retraction for something? `injectivity_radius` is currently mostly used to limit stepsize in different algorithms.

kellertuer:
Hi! I think it is not so much about carefulness as it is, that this silly has not been considered before and we just generalised the `injectivity_radius` to include a retraction because we can and we wanted to distinguish that.

Actually for the case mentioned we exactly hit a discussion point that might be breaking when trying to limit stepsiyes. the projection retraction on the sphere is injective on the whole tangent space and even invertible,
but of course it just has a range of the upper half sphere when p is the North Pole,
so “stepsize” as norm of the gradient might be very different from “actual step taken” (in distance on the sphere).
And for the inverse q can not be further away from p than pi/2 (again measured on the sphere) for the inverse to be well-defined.

I am also not 100% sure we need that, it might still be nice to be clear and transparent in distinguishing both;
but for step size limiting one maybe should restrict it actually to the minimum of both radii?
At least if both are defined – and it gets trivial again for exp/log where both are equal.

mateuszbaran:
[...]

PR #915, opened 2026-07-06
Adjust injectivity radii for (inverse) retractions on the Sphere (#900)

sblelong:
Hi!

Here's a first attempt to address the issue raised in #900 about injectivity radii for retractions and their inverse.

I tried to follow the unit testing pattern on the sphere but I'm requesting approval before proceeding with tests for the `bijectivity_radius`.


src/manifolds/Sphere.jl
```diff
@@ -286,7 +286,7 @@ function get_vector_orthonormal!(M::AbstractSphere{ℝ}, Y, p, c, ::RealNumbers)
 end
 
 _doc_injectivity_radius_sphere = raw"""
-    injectivity_radius(M::AbstractSphere[, p, ::ExponentialRetraction])
```
kellertuer, line 289: No. We can not do this.
If you __replace__ ExponentialRetraction here with the Stabilized one, the function is now undefined for ExcponentialRetraction. Anyone relying on that because they use it (it is the standard even!) has their code breaking.
One thing we should always avoid is breaking other peoples code. Always under all circumstances. In 6 years (since we started 2019) we had 10 reasons to do so and we did also those only very carefully. So please do not delete the old variant.

You can for sure add a copy that does this for the stabilised one but I would even argue the stablized one just has a numerical check around so it could even just call infectivity radius on its inner retraction.

src/manifolds/Sphere.jl
```diff
@@ -301,12 +301,11 @@ _doc_injectivity_radius_sphere_projection = raw"""
     injectivity_radius(M::Sphere, p, ::ProjectionRetraction)
 
 Return the injectivity radius for the [`ProjectionRetraction`](@extref `ManifoldsBase.ProjectionRetraction`) on the
-[`AbstractSphere`](@ref), which is globally ``\frac{π}{2}``.
+[`AbstractSphere`](@ref), which is globally ``\infty``.
```
kellertuer, line 304: If we are here we should document this much more precise, what is injective here, we had a long discussion so lets also be precise here.

src/manifolds/Sphere.jl
```diff
@@ -317,8 +316,8 @@ end
 function injectivity_radius(M::AbstractSphere, p, m::AbstractRetractionMethod)
     return _injectivity_radius(M, p, m)
 end
-_injectivity_radius(::AbstractSphere, ::ExponentialRetraction) = π
```
kellertuer, line 320: Removing this definition here is what is actually breaking. Let's not do that, but as mentioned define the stabilised one to return the one of the retraction it stabilises .

src/manifolds/Sphere.jl
```diff
@@ -337,6 +336,62 @@ function inverse_retract_project!(::AbstractSphere, X, p, q)
...
+
+function injectivity_radius(M::AbstractSphere, m::AbstractInverseRetractionMethod)
+    return _injectivity_radius(M, m)
+end
+function injectivity_radius(M::AbstractSphere, p, m::AbstractInverseRetractionMethod)
+    return _injectivity_radius(M, p, m)
+end
+
+function _injectivity_radius(M::AbstractSphere, m::LogarithmicInverseRetraction)
+    return injectivity_radius(M, StabilizedRetraction())
```
kellertuer, line 359: No!
(a) Log is the mathematically most important thing, so that one should have the number it returns defined specifically
(b) the stabilised one is just a numerical thing
(c) why are you passing from an inverse retraction to a retraction? That looks at least misleading.

src/manifolds/Sphere.jl
```diff
@@ -337,6 +336,62 @@ function inverse_retract_project!(::AbstractSphere, X, p, q)
...
+Return the bijectivity radius for the abstract sphere `M` endowed with (inverse) retraction method `m`, which is the smallest radius applicable to both the tangent spaces and the Riemannian metric for which the (inverse) retraction `m` has a well-defined inverse. This radius is computed as the minimum between the injectivity radius of `M` endowed with (inverse) retraction method `m` (at `p`), and the injectivity radius of the inverse of `m` (at `p`).
+"""
+
+@doc "$(_doc_bijectivity_radius_sphere)"
+bijectivity_radius(::AbstractSphere, ::Union{AbstractRetractionMethod, AbstractInverseRetractionMethod})
+@doc"$(_doc_bijectivity_radius_sphere)"
+bijectivity_radius(::AbstractSphere, p, ::Union{AbstractRetractionMethod, AbstractInverseRetractionMethod})
+
+function bijectivity_radius(M::AbstractSphere, m::Union{AbstractRetractionMethod, AbstractInverseRetractionMethod})
+    if m ∈ (StabilizedRetraction(), StabilizedInverseRetraction())
```
kellertuer, line 381: This looks very strange and will often not yield what you want 
(a) you compare instances. Even if someone comes along with an instance of the type, the instance might be a different one
(b) if you want to go for `if` you should check types, so that `typeof(m) === StabilizedRetraction`. The usually best way is – as all examples in the code before – to do that with dispatch.
(c) also here anything stabilised (since it is a computer/implementation/stability thing) should “pass” to the inner retraction (or inverse retraction) it stores.

src/Manifolds.jl
```diff
@@ -308,6 +308,7 @@ using ManifoldsBase:
     SoftmaxRetraction,
     SoftmaxInverseRetraction,
     StabilizedRetraction,
+    StabilizedInverseRetraction,
```
kellertuer, line 311: This is not defined in ManifoldsBase? How can you import it from there then?

test/manifolds/test_sphere.jl
```diff
@@ -33,7 +33,7 @@ using ManifoldDiff
             :InvalidVectors => [p],
             :NormalVectors => [V],
             :Covectors => [ξ],
-            :RetractionMethods => [ExponentialRetraction(), ProjectionRetraction()],
```
kellertuer, line 36: Sane as before – no. The sphere should absolutely be tested on the ExponentialRetraction. Stabilisation is something extra – and exp is still the default. If you want to test it, add it at the end.

NEWS.md
```diff
@@ -5,6 +5,15 @@ All notable changes to ´Manifolds.jl´ will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.30] Unreleased
+
+### Added
+
+* `injectivity_radius(M, p, m)` and `injectivity_radius(M, m)` where `m` is an `AbstractInverseRetractionMethod` and `M` is a `Sphere` are introduced.
+* `injectivity_radius(M, m)` where `M` is a `Sphere` and `m` is a `LogarithmicInverseRetraction` or a `ProjectionInverseRetraction`.
+* `bijectivity_radius(M, p, m)` and `bijectivity_radius(M, m)` are defined for `Sphere`s.
```
kellertuer, line 14: You could also add tests for this?

kellertuer (changes requested):
Thanks for the start. Here are some first preliminary comments.

The test failure on CI does seem more related too SciML destroying something, so for comments on that we have to wait for that to work again (I wrote on their channel on slack)

src/manifolds/Sphere.jl
sblelong: (c) I thought this was the right way to implement your suggestion in [this message](https://github.com/JuliaManifolds/Manifolds.jl/issues/900#issuecomment-4563744085): the injectivity radius for log calls the one for exp. But I agree this is misleading, I will change that.

sblelong:
Thank you for this detailed feedback, I'll address all changes!

kellertuer:
As a concrete example – I could convince the tests a bit to run and this

https://github.com/JuliaManifolds/Manifolds.jl/actions/runs/28779489714/job/85391973067?pr=915#step:6:544

should never happen. You did not touch Stiefel at all, but something you changed made that fail now.

You did change expected numbers in the tests (and whether that is breaking is something we should discuss/conclude in the end) – but nothing else should change – otherwise we have a breaking change.

You can also test this locally. Do you know about development mode and `] test Manifolds` to run the tests locally?
If you feel a bit lost, we can check to have a short meeting, but not this week, since officially I am on vacation.

sblelong:
I was indeed negligent with running tests locally: I ran the `The Spheres` testset only. I do know about the `] test Manifolds` command, but wrongly decided not to test thoroughly this time. My mistake.

The error encountered in the CI brings an interesting question though: as soon as an inverse retraction is defined for a manifold, should we expect the injectivity radius for this inverse retraction to be implemented?

src/manifolds/Sphere.jl
sblelong: This does make much more sense. I will keep it `Manifolds.jl` so we can run tests before moving it to `ManifoldsBase.jl` where stabilized retractions are defined.

kellertuer:
No worries, errors are also always a good occasion to learn.

> The `Sphere`s testset only.

Yes, we made the possible on purpose, since the overall tests run for a while, one can focus on a single one of on does work _only_ on these. You however also introduce a new signature globally!

> as soon as an inverse retraction is defined for a manifold, should we expect the injectivity radius for this inverse retraction to be implemented?

That is an interesting point that affects the test suite (so not whether something is breaking) – For consistency, I think it would be good if it _were_ implemented for all manifolds we have that now run the new tests, and I think a reasonable start is to set it to the same value as the retraction case – but explicitly to that number and add a doctoring to that implementation as well.
That is a path that takes a bit more time to implement, but does not require so much “thinking” in the sense that you have to prove something or look something up (mainly copy a function and its docs and adapt it)
Would you be available to do that?

The alternative is to make the new tests inactive by default and introduce a boolean keyword – though that might also lead to us “forgetting” to add these functions. But it is less work for sure.

src/manifolds/Sphere.jl
kellertuer: That is our usual path. We introduce new features here in Manifolds.jl and once we notice they are good – well-designed and tested “in the real world”, we move them to ManifoldsBase

sblelong:
> Yes, we made the possible on purpose, since the overall tests run for a while, one can focus on a single one of on does work _only_ on these. You however also introduce a new signature globally!

And that would have been a good reason for me to be more patient and run the tests entirely before opening the PR! One lesson learned. :)

> That is an interesting point that affects the test suite (so not whether something is breaking) – For consistency, I think it would be good if it _were_ implemented for all manifolds we have that now run the new tests, and I think a reasonable start is to set it to the same value as the retraction case – but explicitly to that number and add a doctoring to that implementation as well. That is a path that takes a bit more time to implement, but does not require so much “thinking” in the sense that you have to prove something or look something up (mainly copy a function and its docs and adapt it) Would you be available to do that?

That sounds feasible. I am available and will work on that in the upcoming days, while addressing the other issues raised in this PR.

kellertuer:
> That sounds feasible. I am available and will work on that in the upcoming days, while addressing the other issues raised in this PR.

Nice, thanks! And no worries, a PR is nearly never a sprint, I always prefer thoroughness and completeness (and less bugs) over speed.

PR #916, opened 2026-07-08
Use direct SciML owner imports in extensions

ChrisRackauckas:
## Summary

- import SciML interface objects directly from `SciMLBase` in the BVP and ODE extension code
- import `Rodas5P` and `AutoVern9` from their owner packages instead of through `OrdinaryDiffEq`
- add the corresponding weakdeps/extension triggers for `SciMLBase`, `OrdinaryDiffEqRosenbrock`, and `OrdinaryDiffEqVerner`
- remove stale unused `ADTypes` imports from the main module

## Motivation

This avoids relying on transitive SciML exports such as `BoundaryValueDiffEqMIRK.BVProblem`, `BoundaryValueDiffEqMIRK.solve`, and `OrdinaryDiffEq.SciMLBase`.

`CommonSolve` is intentionally not added as a direct weakdep; `solve` is imported from `SciMLBase`, which provides the public SciML interface used here.

## Verification

- Fresh extension load check for `ManifoldsBoundaryValueDiffEqExt`
- Fresh extension load check for `ManifoldsOrdinaryDiffEqDiffEqCallbacksExt`
- ExplicitImports rescan of SciML-related findings:
  - `sciml_nonpublic=0`
  - previous BVP/OrdinaryDiffEq transitive SciML findings removed
- `Pkg.test("Manifolds")`
  - `20891` pass
  - `1` broken
  - `20892` total
  - `29m49.6s`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.11.29] Unreleased
 
+### Changed
+
+* SciML-related extensions now import interfaces and solvers from their owner packages and explicitly trigger on the required SciML weak dependencies.
+
 ### Fixed
 
 * PythonPlot was replaced with GR to fix documentation building on CI.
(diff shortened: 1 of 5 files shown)


kellertuer:
Formally I think this even changes when the extension is loaded? Then this would even be a breaking change.
Though maybe in practice it does not?

mateuszbaran:
LGTM. The changes in the loading order either shouldn't matter or we would have to make them anyway at some point. I think it would be enough to mention them in the NEWS.md file.

kellertuer:
Yes, a News.md entry is missing and maybe bumping the patch version would be nice in order to merge this.

mateuszbaran (approved):

PR #917, opened 2026-07-13
🌾

mateuszbaran:
This time it's Jacobi fields in coordinates and a few additional things.

- [x] `differential_exp_argument`
- [x] `differential_exp_basepoint`
- [x] `differential_log_argument`
- [x] `differential_log_basepoint`
- [x] `volume_density`
- [x] examples in docs/tutorial
- [x] complete tests on sphere
- [x] maybe atlases on Grassmann or Stiefel
- [x] in-place problems instead of out-of-place ones
- [x] parametric surfaces
- [x] power/product atlases for power and product manifolds


kellertuer:
We are back to single-Emoji-PRs? 
I am not sure I understand this emoji though. Wheat? Because it is Jacobi but not Wheat fields ;) ?

So I do not yet fully see where this leads, but you have a tutorial on your list, so I am looking forward to reading that :)

mateuszbaran:
It's a rice emoji, it's supposed to represent a bunch of geodesics emanating from a single point, a bit like blades of grass bending apart.

I've opened the PR a bit early to have CI, SciML is throwing odd warnings locally for me.

mateuszbaran:
OK, this should be ready for a review now. I've added the Runic md/qmd formatting here. If you want to format it locally from REPL you now need to run `using Runic; Runic.main(["-i", "-v", "-d", "--extensions=jl,md,qmd", "."])`.

kellertuer:
How much of the 3k+ lines do you want to have a focus on in the review?

mateuszbaran:
I think roughly in order of importance:
1. If the new stuff in the "Working in charts" tutorial looks fine.
2. Parametric surface and new atlases for Grassmann and Stiefel. Changes in the main Manifolds.jl file.
3. All other changes in the `/src` folder.
4. Everything else.

So I think it would be nice if you go at least through 1 and partway through 2 :smile: .

docs/src/manifolds/parametric.md
```diff
@@ -0,0 +1,75 @@
...
+
+```@example mobius-band
+using Manifolds, Plots, RecursiveArrayTools, ForwardDiff
+
+half_width = 0.35
+M_param = ProductManifold(Hyperrectangle([-half_width], [half_width]), Circle())
+M_embed = Euclidean(3)
+
+function mobius!(q, p)
+    u, θ = p.x[1][1], p.x[2][]
```
kellertuer, line 25: p is not typed but implicitly has a `p.x` so it is probably a point on M and an ArrayPartition?

Originally p,q are always points on _one_ manifold, here it seems one of them is in “parameter space”? We could consider using a new name (or two new) for that ?
At least for 2D I often see `u,v`? The general could be `u1,...,un` – if we do not use `u` yet of course.

docs/src/tutorials/working-in-charts.md
```diff
@@ -0,0 +1,534 @@
+# Working in charts
```
kellertuer, line 1: This was rendered on CI before and is now checked in?

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
+@doc raw"""
+    ParametricSurface{TMP<:AbstractManifold{ℝ},TE<:Euclidean,TF,TIF,TDF} <: AbstractDecoratorManifold{ℝ}
+
+Surface in ℝⁿ described by a parametric function `f` defined on a parameter space `M_param`.
```
kellertuer, line 4: You write this with Rn (maybe better in math?) but the structuct called it even `M_embed` (though restricted in type to Euclidean. Is there a plan to allow even more general embeddings?

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+    inverse_f!::TIF
+    jacobian_f!::TDF
+end
+
+"""
+    decorated_manifold(M::ParametricSurface)
+
+Return the parameter-space manifold of `M`.
+"""
+decorated_manifold(M::ParametricSurface) = M.M_param
```
kellertuer: I am not sure this is a decorated manifold, can you elaborate why?

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+
+Check whether the embedded point `q` lies on `M` by applying the inverse parametrization and
+reconstructing it with `f!`.
+"""
+function check_point(M::ParametricSurface, q; kwargs...)
+    p = ManifoldsBase.allocate_on(M.M_param)
+    M.inverse_f!(p, q)
+    q_reconstructed = similar(q)
+    M.f!(q_reconstructed, p)
+    if !isapprox(q, q_reconstructed; kwargs...)
```
kellertuer, line 94: hm, I do not fully understand why M -> parameters -> M is the right test here. Is it maybe enough to check whether is_point(M.M_param, p)` holds?

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+    check_vector(M::ParametricSurface, p, X; atol::Real=sqrt(eps(float(number_eltype(p)))), kwargs...)
+
+Check whether `X` is tangent to the parametric surface `M` at embedded point `p`.
+"""
+function check_vector(
+        M::ParametricSurface,
+        p,
+        X;
+        atol::Real = sqrt(eps(float(number_eltype(p)))),
+        kwargs...,
```
kellertuer:
```suggestion
        M::ParametricSurface, p, X;
        atol::Real = sqrt(eps(float(number_eltype(p)))), kwargs...,
```

More like a remark, I feel this is more readable and (as opposed to JuliaFormatter) allowed with runic. One could even check whether it still fits one line.

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+    )
+    parameters = ManifoldsBase.allocate_on(M.M_param)
+    M.inverse_f!(parameters, p)
+    J = Matrix{promote_type(eltype(p), eltype(X))}(
```
kellertuer: can we maybe document a bit what this does? I would heave expected that ...ah we do not have Df_inv? because Df_inv(X) should be in T_uM_Param (u parameters)

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+"""
+    project!(M::ParametricSurface, q, p)
+
+Project `p` onto `M` by applying the inverse parametrization followed by `f!` and store the
+result in `q`.
+"""
+function project!(M::ParametricSurface, q, p)
+    parameters = ManifoldsBase.allocate_on(M.M_param)
+    M.inverse_f!(parameters, p)
+    M.f!(q, parameters)
```
kellertuer: Does this maybe also need a project on M_param?

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+function project!(M::ParametricSurface, Y, p, X)
+    parameters = ManifoldsBase.allocate_on(M.M_param)
+    M.inverse_f!(parameters, p)
+    J = Matrix{promote_type(eltype(p), eltype(X))}(
+        undef,
+        manifold_dimension(M.M_embed),
+        manifold_dimension(M.M_param),
+    )
+    M.jacobian_f!(J, parameters)
+    Y .= J * (J \ X)
```
kellertuer, line 163: this could be used within check:vector?

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,313 @@
...
+
+Store in `cX` the coordinates of embedded tangent vector `X` at `p` in the basis induced by
+forwarded atlas `B`.
+"""
+function get_coordinates_induced_basis!(
+        M::ParametricSurface,
+        cX,
+        p,
+        X,
+        B::InducedBasis{ℝ, TangentSpaceType, <:ParametricSurfaceAtlas},
```
kellertuer:
```suggestion
        M::ParametricSurface, cX, p, X,
        B::InducedBasis{ℝ, TangentSpaceType, <:ParametricSurfaceAtlas},
```

Since we can also reformat this, one could do 2-lines here at least without loosing readability?

kellertuer (commented):
I checked roughly the first half of files and here are just a few small comments.

Maybe a few more math formulae in general would be nice?

src/manifolds/ParametricSurface.jl
kellertuer: Actually this refactor of code formatting was one of the few things the AI could do quite well, but for other functions than the one on this PR we could do that in a follow up

docs/src/manifolds/parametric.md
mateuszbaran: I will add `ArrayPartition`. Yes, here we have a slightly more complicated scenario because we have two manifolds (`M_param` and `M_embed`. I think I can do `u`, `v` for `M_param`.

docs/src/tutorials/working-in-charts.md
mateuszbaran: Yes, that's correct.

src/manifolds/ParametricSurface.jl
mateuszbaran: Currently it's even limited to one-index `Euclidean` and the only flexibility is in using type-based or value-based parameters. I might consider extending to multi-index `Euclidean` in the future but I think it's worth the trouble of implementing it. I currently have no plans to allow non-Eucildean embeddings.

src/manifolds/ParametricSurface.jl
mateuszbaran: I wasn't sure about doing it and it's probably better to leave it out. I will delete this.

src/manifolds/ParametricSurface.jl
mateuszbaran: The manifold consists of points `p` that can be expressed as `p=f(u)` for some `u`. I will consistently rename to `u`/`v` to make it clear why we need this construction.

src/manifolds/ParametricSurface.jl
mateuszbaran: I will add some docs. No, we don't have Df_inv but I think Df is enough. Matrix dimensions agree (and we have them different in tests between parameter space and the embedding) so I don't think I've mixed that.

src/manifolds/ParametricSurface.jl
mateuszbaran: `inverse_f!`is supposed to handle this, I will add an explanation in docs.

src/manifolds/ParametricSurface.jl
mateuszbaran: Yes, that's right. I will use it there.

src/manifolds/ParametricSurface.jl
kellertuer: So its curves on manifolds? Ah no its Rn but not R^(nxm) ?

mateuszbaran: Yes, curves/surfaces in Rn only, not even R^(nxm) yet. We can do generalization later when we find a good motivation.

kellertuer: Ok, no that is totally fine and a very good start; we can surely do generalisations to parameters on Rmn or even other manifolds (🤯) later. I just wanted to understand the scope here, no judgement.

src/manifolds/ParametricSurface.jl
kellertuer: Sure we do not need dfinv for now, was again just for clarification.

src/manifolds/ParametricSurface.jl
mateuszbaran: Cool. Note that the *parameters* can be on any manifold, it's the *embedding* that is restricted to R^n. I think this is already quite generic.

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,347 @@
...
+
+    ParametricSurface(M_param, M_embed, f!, inverse_f!, jacobian_f!)
+"""
+struct ParametricSurface{
+        TMP <: AbstractManifold{ℝ},
+        TE <: Euclidean,
+        TF,
+        TIF,
+        TDF,
+    } <: AbstractDecoratorManifold{ℝ}
```
kellertuer, line 40:
```suggestion
struct ParametricSurface{
        TMP <: AbstractManifold{ℝ}, TE <: Euclidean, TF, TIF, TDF,
    } <: AbstractDecoratorManifold{ℝ}
```

with runic this is possible and I will try to unify this somewhen to this format, which is a bit easier to read I think

src/manifolds/ParametricSurface.jl
```diff
@@ -0,0 +1,347 @@
...
+    return p
+end
+
+"""
+    get_coordinates_induced_basis!(
+        M::ParametricSurface,
+        cX,
+        p,
+        X,
+        B::InducedBasis{ℝ, TangentSpaceType, <:ParametricSurfaceAtlas}
```
kellertuer:
```suggestion
        M::ParametricSurface, cX, p, X,
        B::InducedBasis{ℝ, TangentSpaceType, <:ParametricSurfaceAtlas}
```

Manopt uses now this form (unified) when the last one is something long (basis, retraction...)

kellertuer (commented):
Review part 2 (or probably 3)

* the manifold tests are usually called e.g. `test_sphere.jl` to not confuse the file with the one of the manifold – this seems not unified yet, but could we at least name new ones this way?
* for the new manifolds further: Can we do them directly with the test suite? Sure I plan to transfer manifold by manifold somewhen – but if all new ones are created without the suite, that extends my work later

The part 3 – when I find time – is just to read the updated/new tutorials once again, the rest is already fine with me.

mateuszbaran:
OK, I've used the new test suite for what looked reasonable and renamed the file.

kellertuer:
Sure that is every now and then a balance, whether one wants to “squeeze” everything into the suite or does a few small tests separately as well

kellertuer (approved):
The tutorials are fine with me.

mateuszbaran:
Thanks!

PR #921, opened 2026-08-21
Bump DocCitations, add CodeBlocks and LandingPage.

kellertuer:
Adds DocumenterCodeBlocks and a landing page.

# ToDo
* [x] codeblock warns on a few doc strings either missing or having a too long first line (<200 chars required)
* [x] the tagline should be improved
* [x] the 6 tiles should be checked and maybe improved.
* [x] add a changelog entry.


mateuszbaran:
Nice. I think the landing page looks good, I don't have any suggestions for improvements.

kellertuer:
Perfect. Then I will just work through the CodeBlocks warnings – they all seem to be “too long first paragraphs” if I see correctly.

kellertuer:
We have a few warnings from Codeblocks, but
* all but one will be resolved once we add the doc strings in ManifoldsBase (when moving there to CodeBlocks)
* the last one is from `LinearAlgebra.norm` where I am unsure which Interlinks inventory to load.

That mainly means that some functions currently do not have a tool tip.

The rest is fine, we could merge this and wait for a next actual feature to be included before we register a new version?

kellertuer:
I tried a few things and now left the Julia Inventory in the interlinks loading, that does still not remove the one warning that is not due to ManifoldsBase – `LinearAlgebra.norm` with just one argument currently warns that it has no tooltip.

I have no clue where that is in the docs nor how to fix that, since when it does not even work when loading the whole inventory of Julia (which includes the LinearAlgebra docs) – then no clue what else to do. I fear we have to just leave that warning in then.

The rest is fixed and could be merged.

mateuszbaran (approved):
LGTM.

Issue #902, opened 2026-05-30
Implement the Veronese manifold


sChoiKr:
Hi from Italy,

Since our next item on the TensorKitchen.jl to-do list is to add symmetric CP decomposition, we will need a Veronese manifold.

`Veronese{𝔽,N,D} <: AbstractManifold{𝔽}`
`struct Veronese{𝔽,N,D} <: AbstractManifold{𝔽} end`

As far as I understand, the Veronese manifold can be viewed as the symmetric, diagonal analogue of the Segre manifold: it corresponds to the rank-one tensor case where all tensor factors are tied together,

```math
x_1 = x_2 = \cdots = x_d = x.
```

While working on vector transport for the Segre manifold, I started to understand better how the basic manifold structure and its necessary functions are organized. Thanks to your tutorial on how to create custom manifolds, I would like to suggest implementing a `Veronese` manifold in Manifolds.jl, although I am still not fully confident about the design.

This would also be useful for TensorKitchen.jl, since symmetric CP decomposition naturally uses rank-one symmetric tensors of the form

```math
\left\{
\lambda\, x^{\otimes d}
\;\middle|\;
\lambda \in \mathbb{R}_{>0},\;
x \in \mathbb{S}^{n-1}
\right\}
\subset
(\mathbb{R}^n)^{\otimes d}.
```

Of course, I would work on this with my supervisor’s guidance. I am not completely sure whether adding a general Veronese manifold to Manifolds.jl makes sense from the ecosystem perspective, so I would be very happy to hear your thoughts or suggestions. If you think this should instead be handled differently, I would also completely understand.

kellertuer:
Manifolds.jl aims to collect all manifolds where a reasonable numerical “things” can be done. I think originally, even Simon had the manifold in mind, but never found the tie to create it.

So – sure, that manifold is very very welcome!

yes we have the [tutorial on how to write a manifold](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/tutorials/implement-a-manifold/), you already know all the git and GitHub workflow.

So – sure – go for it and implement it. I can guide you along the way with technical help, how to use a few tricks to not implement things more than once, how to setup and do the documentation and the testing.

I have never worked with nor do I know much about the Veronese manifold, so it would be great if the documentation of the manifold also contains links to literature, see also the `references.bib` in the docs folder.

kellertuer:
[...]

PR #924, opened 2026-09-14
Add the Veronese manifold

sChoiKr:
This PR starts adding the Veronese manifold of nonzero symmetric rank-one tensors.

It includes:

* the manifold implementation with Euclidean induced metric,
* point and tangent embeddings,
* random point and tangent vector generation,
* tests for the manifold interface, embedding, metric, and so on,
* documentation and a `NEWS.md` log entry.

The implementation follows the structure from Segre manifold and the geometry of Segre–Veronese manifolds described in Jacobsson et al. (2024).

Remark: compared with the initial proposal in #902, the radial parameter is allowed to be any nonzero real number, (not only positive numbers). This ensures that the manifold contains all nonzero symmetric rank-one tensors; in particular, for even $D$, restricting $\lambda>0$ would exclude negative multiples of $x^{\otimes D}$, since $(-x)^{\otimes D}=x^{\otimes D}$.


sChoiKr:
I hope you've had a wonderful summer.

It's been a busy one for me that I finally now have time to catch up on implementing Veronese manifold as we discussed before summer. :/

1. One design question: I currently store `N` and `D` directly as type parameters, following the `Segre` implementation. I noticed that newer updates and refactoring like`Tucker` support now `parameter=:type` / `:field` via `TypeParameter`.
Would you prefer Veronese manifold to follow that newer parameter-storage pattern as well, or is the simpler `Veronese{N,D}` representation just like Segre appropriate here?

2. I noticed there is a recent published reference, would you prefer addressing that paper rather than the previous one?

As much as I took enough time for checking through the scripts line by line, followed by your guideline from CONTRIBUTING.md this time, there might be mistakes. I'd appreciate your reviews and guidance. 

Thank you.

kellertuer:
HI and welcome back!

> I hope you've had a wonderful summer.

Mine could also have been calmer. By now I am even again fully submerged in teaching (we start mid/end of August)
But no worries, nothing here is urgent – and I prefer thorough code than rushing something.

## TypeParameter
The two variants based on the `parameter = :field` or `:type` cases. It is not necessary but it would be cool to have both.
The `field` case is a bit easier to implement (see Segre), but the `type` one is cool, because it allows to _dispatch_ on specific values of the type parameter.
This is for example used on the Rotation matrices to provide efficient implementations for the low dimensional cases.

If you expect to have efficient implementations based on the values of the manifold parameters, we should definetly already do both. If you feel that this is too techical/complicated – just do the current approach and we add the other one when it is actually needed.

## References

Take the one that best reflects where you got your sources from or where you think the user finds more details best. If all your implementation is based on Simons paper, then use that.

## Tests

A good thing of this PR is that we can already see where we are missing tests – currently 77 lines are uncovered...oh and while I wrote the two above it reduced to 9 lines :)
Carefully check the missing lines and reflect about how to best test that those lines work as intended.

If possible, follow the new method of using the `Manifolds.Test` test suite. I see you are currently following the `manifolds-old` scheme. The new one is hopefully easier to use (see the examples in the `manifolds` folder).
And again, if that seems to complicated, I can do that somewhen later, though my plan to rework all tests is very slowly moving forwards recently.


And sure – I try to find some time soon to take a closer look at the code. No worries!

sChoiKr:
> By now I am even again fully submerged in teaching (we start mid/end of August)

Hope you enjoy teaching, and I guess then the semester will end earlier than ours? ;)


Here are some investigation and fixes:

1. The Julia 1.13 CI failures appear unrelated to this PR. Both Ubuntu and Windows fail at `test/test_utils.jl:62`: 
         `@test Manifolds.isnormal(skew(randn(ComplexF64, 3, 3)))\`

2. Regarding the reference: I think the newer reference I mentioned above is actually the published SIAM version of Simon's previous preprint. :) The change from λ > 0 to λ ≠ 0 also comes from this newer version. For now, I only updated the existing entry in references.bib from 2024 to 2026, since the same citation key is also used by Segre.jl and SegreWarpedMetric.jl. Would it make sense to replace that entry entirely with the full 2026 journal reference instead?

3. TypeParameter: I updated Veronese to support both `parameter=:type` and `parameter=:field`, with `:type` as the default following Tucker.jl 

4. I also migrated the Veronese tests to the newer `Manifolds.Test/test_manifold` setup.

Due to the first error, this attempt will also likely fail. :/  

As always, thank you for your encouragement! :)

kellertuer:
Yes I like to teach – through every now and then that yields a bit of a high workload. We finish the lectures end of November and have the exams in December – so yes that is a bit earlier than the German scheme.

I saw the error on utils already over on #923 and I am working on that :)

For the reference, ah – then it makes sense to replace the old preprint by the new Journal version, sure! Even more, if the new version has more detail _and_ you use that detail. 

3 and 4 sound good and look fine at first glance – I will try to take a closer look soon.

One additional thing you can already check is
* add an entry for yourself on the docs/src/misc/about.md, link your name to a site of yours and refer to the new docs page
* add yourself also in the Zenodo metadata (`.zenodo.yml` in the main folder) – then you are listed on the metadata pages as well. Maybe run both cases just follow the example of Simon :)

_edit:_ The error in the `test_utils` is fixed on master. Merging the master branch here should fix it here as well.

kellertuer:
Yay! We have test coverage.

Could you move yourself in the zenodo a bit up (basically to line 20 or so) so the people in there are ordered alphabetically?

sChoiKr:
> Could you move yourself in the zenodo a bit up (basically to line 20 or so) so the people in there are ordered alphabetically?

Sure, I was copying Simon's lines right there. ;) Will do!

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,497 @@
...
+    T = number_eltype(p)
+    return [zeros(T, 1), zeros(T, n)]
+end
+
+function ManifoldsBase.allocate_result_embedding(
+        ::Veronese,
+        ::typeof(project),
+        A,
+        p,
+    )
```
kellertuer, line 150:
```suggestion
function ManifoldsBase.allocate_result_embedding(
        ::Veronese, ::typeof(project), A, p,
    )
```

One thing we can have since we switched from JuliaFormatter to Runic is that in a list of parameters or a vector, the arguments can share a line,
they can even be split to cover 2 lines if that fits reasonably (for example one for the args one for the kwargs)

You do not have to adapt your code, but if you would, that would be nice. Otherwise I plan to “brush over” this somewhen in the future with a bit of a Regex or an AI :)

test/manifolds/test_veronese.jl
```diff
@@ -0,0 +1,231 @@
...
+        Y = [[-0.3], v]
+        Z = [[0.1], w]
+
+        invalid_points = Any[
+            [[1.0, 2.0], x],
+            [[0.0], x],
+            [[Inf], x],
+            [[1.0], x[1:3]],
+            [[1.0], 2 .* x],
+        ]
```
kellertuer:
```suggestion
        invalid_points = [
            [[1.0, 2.0], x], [[0.0], x], [[Inf], x], [[1.0], x[1:3]], [[1.0], 2 .* x],
        ]
```

this is for example also readable (and though edited on the web interface hopefully runic-ed). I think the Any can also be just left out.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+    n > 0 || throw(ArgumentError("The vector-space dimension n must be positive."))
+    d > 0 || throw(ArgumentError("The tensor order d must be positive."))
+    size = wrap_type_parameter(parameter, (n, d))
+    return Veronese{typeof(size)}(size)
+end
+
+function ManifoldsBase.allocate_result(M::Veronese, ::typeof(rand))
+    n, _ = get_parameter(M.size)
+    return [zeros(1), zeros(n)]
+end
```
mateuszbaran, line 129: If all points always have exactly two components, I'd suggest using tuples to represent them instead of lists for type stability and performance.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+
+function copyto!(::Veronese, Y, p, X)
+    copyto!(Y[1], X[1])
+    copyto!(Y[2], X[2])
+    return Y
+end
+
+function _isapprox(M::Veronese, p, q; kwargs...)
+    if is_point(M, p) && is_point(M, q)
+        return isapprox(distance(M, p, q), 0; kwargs...)
```
mateuszbaran: Comparing by `isapprox` with 0 is a bit odd by default:
```julia
julia> isapprox(1e-100, 0)
false
```
I generally prefer a test like `abs(distance(M, p, q)) < atol` instead where `atol::Real` is a keyword argument that defaults to something positive (like `eps()`)

mateuszbaran (commented):
Thanks for another contribution!

It looks mostly fine to me, I just have two relatively minor comments.

src/manifolds/Veronese.jl
sChoiKr: Hi Mateusz,

Thank you for checking. I simply followed vector formed component representation from Segre.jl. But, true, you are right that Veronese will always have two factors unlike Segre. So, we don't need variable arity. I will switch all the outer point rep to tuples and update the allocation methods,

Hmmm, but then I think I should replace `check_size` to tuples as well:

```julia
function check_size(M::Veronese, p)
    p_size = only.(size.(p))
    n, _ = get_parameter(M.size)
    M_size = (1, n)

    if p_size != M_size
        return DomainError(...)
    end
    return nothing
end

```
?

I will review through the entire script to double-check.

mateuszbaran: I see, I think a slight deviation from Veronese is justified in this case.

I think this `check_size` code should also work fine for tuples.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+````math
+\mathcal V_{N,D}
+=
+\left\{
+    \lambda x^{\otimes D}
+    \;\middle|\;
+    \lambda\in\mathbb R\setminus\{0\},\;
+    x\in\mathbb S^{N-1}
+\right\}.
+````
```
kellertuer, line 16: Can you add a short sentence what the $\otimes D$ means? and D is an integer right?

That is probably also written in the paper, but if possible it would be good if the docs are self-contained.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+    return check_vector(
+        Sphere(n - 1; parameter = get_parameter_type(M)), p[2], X[2]; kwargs...,
+    )
+end
+
+@doc raw"""
+    closest_representative!(M::Veronese, q, p)
+
+Replace `q` by the representative of the same embedded tensor that is closest
+to `p` in the signed parameterization of [`Veronese`](@ref).
```
kellertuer: what is “the signed representation”? If there is different representations (for me that reads as point/tangent vector “types”) that could best be added in the manifold description.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+
+The representative change is parity-aware:
+
+````math
+(\lambda, x) \sim ((-1)^D\lambda, -x).
+````
+
```
kellertuer, line 274: To me it is not fully clear how \lambda and x are related to either p or q … could that maybe be added?

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+    x\in\mathbb S^{N-1}
+\right\}.
+````
+
+Points are represented by `p = [[λ], x]` and tangent vectors by `X = [[ν], u]`, where ``u`` is
+orthogonal to ``x``. The parametrization ``\Phi(\lambda,x)=\lambda x^{\otimes D}`` is two-to-one,
+with ``(\lambda,x)\sim\bigl((-1)^D\lambda,-x\bigr).``
+
+The metric is induced by the Euclidean metric on the full tensor space ``(ℝ^N)^{\otimes D}``.
+Its geometry is a special case of the Segre--Veronese geometry described in
```
kellertuer: Is Segre--Veronese the manifold that is called Segre here? Than we could extend this a bit and link to Segre as well here.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+@doc raw"""
+    connected_by_geodesic(M::Veronese, p, q)
+
+Return whether `p` and `q` are connected by a geodesic that does not pass
+through the excluded zero tensor. For closest representatives, this requires
+matching signs of the scale and
+
+````math
+\sqrt{D}\,d_{\mathbb S}(x,y) < \pi.
+````
```
kellertuer, line 327: Since I do not fully understand \otimes D – d_S is “just” the distance on a single sphere? and basically we have p consists of \lambda, x and q of \mu and y?

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+    n, d = get_parameter(M.size)
+    sphere = Sphere(n - 1; parameter = get_parameter_type(M))
+    return sqrt(d) * distance(sphere, p[2], q_closest[2]) < pi
+end
+
+@doc raw"""
+    distance(M::Veronese, p, q)
+
+Compute the Riemannian distance between `p` and `q`. For closest
+representatives with scale magnitudes ``r`` and ``s``, let
```
kellertuer: closest representatives could link to the function above?

kellertuer:
```suggestion
representatives with scale magnitudes ``r`` and ``s``, respectively, let
```

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+
+````math
+m = \min\bigl(\sqrt{D}\,d_{\mathbb S}(x,y),\pi\bigr).
+````
```
kellertuer, line 376: Same remark as before, but here even more important, is p now \lambda x or r? is q \mu y or s?

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+function inner(M::Veronese, p, X, Y)
+    _, d = get_parameter(M.size)
+    return X[1][1] * Y[1][1] + d * p[1][1]^2 * dot(X[2], Y[2])
+end
+
+@doc raw"""
+    get_coordinates(M::Veronese, p, X, ::DefaultOrthonormalBasis; kwargs...)
+
+Return the coordinates of `X = [[ν], u]` in a default orthonormal basis of
+``T_pM``. If ``p = [[λ], x]`` and ``c_{\mathbb S}(u)`` are the default
```
kellertuer: We usually use `\mathcal M` for the manifold, but here you probably even mean \mathcal V (with parameters)?

kellertuer: “the default orthogonal parameters” should probably link to what our default here is ;) That is to the function on Sphere describing that.

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,480 @@
...
+end
+
+get_parameter_type(::Veronese{<:TypeParameter}) = :type
+get_parameter_type(::Veronese{Tuple{Int, Int}}) = :field
+
+@doc raw"""
+    inner(M::Veronese, p, X, Y)
+
+Compute the Riemannian inner product induced by the Euclidean metric of the
+full tensor embedding. For `p = [[λ], x]`, `X = [[ν], u]`, and
```
kellertuer: Ah here you write `p = [[λ], x]` :) Feel free to also write that in math instead, that is maybe easier to read in the docs.

kellertuer (commented):
I read the docs. Since the following comments are very German-style – not sure whether you learned about that: No feedback in German means, everything is good – feedback is always the negative things.

Good points:
* all formulae render fine
* the text can be read well

One general thing:
It would be nice if the functions in the file could be ordered alphabetically, that makes them easier to find later on :)

src/manifolds/Veronese.jl
```diff
@@ -0,0 +1,686 @@
...
+````math
+g_{(\lambda,x)}\bigl((\nu,u),(\xi,v)\bigr)
+=
+\nu\xi+D\lambda^2\langle u,v\rangle.
+````
+
+Thus the spherical directions are scaled by ``\sqrt{D}|\lambda|`` relative to
+the radial direction.
+
+The Veronese manifold is a special case of the Segre manifold: the Veronese case
```
kellertuer:
```suggestion
The Veronese manifold is a special case of the [`Segre`](@ref) manifold: the Veronese case 
```

then it even links to The Segre manifold type :)

sChoiKr:
Wow, thank you very much for such detailed reviews. I really appreciate that I am learning a lot from both of you. :)

I am still working through all the details, most of them make sense to me. Just you know, the correction process from your comments is not yet complete. For you not waste your busy time, I will let both of you know once it's done. 

One thing I am considering is changing the way I represent the symmetric tensor to a computationally more efficient alternative. I currently embed it in the full tensor space, which was simpler to implement because the Kronecker product gives the entries directly in the expected order (just the way for Segre). However, this stores repeated symmetric entries, so the ambient representation has size (N^D), and the redundancy becomes increasingly expensive as (N) and (D) grow.

I am wondering how not to repeat the sym entires but still fit the index order of all the computations. I have not seen this implementation detail discussed explicitly in any papers, so I am still thinking through what representation would fit Manifolds.jl best.

kellertuer:
You can also mark points you consider solved as resolved – I just open new ones if/when I would notice something.

src/manifolds/Veronese.jl
sChoiKr: @doc raw"""
    get_coordinates(M::Veronese, p, X, ::DefaultOrthonormalBasis; kwargs...)

Return the coordinates of ``X\simeq(\nu,u)`` in the [`DefaultOrthonormalBasis`](@ref) of ``T_p\mathcal V_{N,D}``. Write ``p\simeq(\lambda,x)`` and let ``c_{\mathbb S}(u)`` denote the default orthonormal coordinates of ``u`` in ``T_x\mathbb S^{N-1}``, as returned by [`get_coordinates`](@ref) on [`Sphere`](@ref). 
"""

Would @ref 'get_coordinates' on @ref 'Sphere' here be sufficient?

kellertuer: yes for example :)

sChoiKr:
I try to rewrite docstrings to make math self-contained explanation (which is a quite of self-learning for myself to know what is obvious and what's not.) Please review them and let me know whether your clarification questions on previous comments are answered enough. 

I also added a vector_transport_to function since our TensorKitchen.jl def needs it for RCG. ;) Check please.

exp/log will be added next.

kellertuer:
Great! Thanks for the effort. I really prefer to have the math details in the docs here.

And no worries about your speed – you do that at your own pace and let us know when you are done so we check the new functions and docs :)

Schönes Wochenende! (not sure you are learning German)

sChoiKr:
Guten Morgen,

Yes, I was able to speak quite ok German in every day life before doing a math phd. Now, I am forgetting it quickly since I rarely have occasions mingling with local 'normal' falks, speaking German. :/

Here, I updated and checked through all docstrings and finalized the last pieces of geodesic functions, which I am still in the process of understanding myself.

So far, the way I picture them geometrically is that:
split the tangent vector X into its radial and angular velocity components, unfold the warped cone into a two-dimensional plane, move along a straight line there, and then map the polar angle back to a geodesic on the sphere(Segre). 

Remark: Unlike other functions, which I've been familiarized from Segre while implementing TensorKitchen.jl before, geodesic maps were never used so far. In order to make sure, the description of geodesic(exp! and log!) is first generated by codex and I've compared to my limited understanding from Simon's paper and Segre.jl implementation. Since I am still in the process of understanding fully myself the entire derivations, I am letting you know that. I might refine them later if needed.

I think most functions I need for TensorKitchen.jl are implemented. Please let me know if there is anything else to be added.

Danke für die Unterstützung! :)

kellertuer:
Oh I know that language problem – I do learn Norwegian, but I speak too often English or even German here.

Thanks for the transparency on the LLM usage for the doc strings – then I have to take a bit of time too check them carefully (I personally do not trust these models!) – which might only happen next week, since besides teaching I also have a guest this week. 

For additions – if this for now covers all functions you need and the tests pass and we have coverage – that is surely enough. We can add further functions once people (or even yourself) need them.

sChoiKr:
Enjoy time hosting your guest then. 

Until next week, hopefully, I might be able to derive it myself, understand better, and then able to check whether everything is correct.:)

So far, in terms of computational efficiency, my Veronese implementation still operates like Segre in the full ambient tensor representation, so it does not yet take any real computational advantage of symmetry. But, for now, I will take some time to digest the differential geometry part of Veronese and work on the symmetric tensor decomposition implementation in TensorKitchen.jl.

Later, I defintely want to come back to the compressed symmetric coordinate representation, since that should be useful for TensorKitchenjl in terms of storage and computation.

Also, since we eventually need partially symmetric tensor decomposition, I think it would be good to implement a Segre-Veronese manifold as well, to complete the family from Simon's paper. What do you think? I am not sure how much of its implementation would differ from Segre + Veronese rather than being one independent manifold, but I will look into that later.

kellertuer:
That sounds good. Also the improvements with symmetry can be next PRs we do not have to do all those here.

The Segre-Veronese is also very welcome as a future PR for sure :)

PR #927, opened 2026-09-16
lower bound Julia in all Project.toml s of the workspace.

kellertuer:


mateuszbaran (approved):

kellertuer:
Oh I forgot the comment. Without the bound, dependabot adds the < 0.01 to the compat which is annoying. Was a discussion on Slack or so yesterday.

mateuszbaran:
That's an odd behavior. Anyway, we can just add proper bounds of course.

kellertuer:
It is a bit odd, but I think Michael gave a longer explanation why it is correct. However; lower bounding the sub environments is nevertheless a good idea.

